Issue #4, opened 2016-02-02
Initial command to build (zip up) a web extension


kumar303:
This should zip a directory of source into a web extension package

kumar303:
I'd like to use zip-dir for this once https://github.com/jsantell/node-zip-dir/pull/8 lands

kumar303:
This is documented https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build

Issue #33, opened 2016-02-08
Add a logger


kumar303:
We need a logger for all web-ext commands, one that can be configured for verbose logging.

muffinresearch:
Having been using bunyan on addons-linter I quite like it and would recommend it. It's very easy to use and has configurable levels. Also has a binary for nicer formatting of the messages which is useful sometimes.

kumar303:
Note to self: it would be nice if `web-ext` dumped a verbose error log in the case of an unexpected exception.

kumar303:
> it would be nice if web-ext dumped a verbose error log in the case of an unexpected exception.

We could use `bunyan.RingBuffer` for this

PR #35, opened 2016-02-10
Add a command to build the web extension

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/4


kumar303:
The console calls are all just placeholders until https://github.com/mozilla/web-ext/issues/33

src/build.js
```diff
@@ -0,0 +1,43 @@
...
+          console.log(`Your web extension is ready: ${xpiPath}`);
+          return {
+            xpiPath: xpiPath,
+          };
+        });
+    });
+}
+
+
+function prepareBuildDir(buildDir) {
```
kumar303: I forgot to add a test for this. Will do so...

src/build.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
+import path from 'path';
+import {createWriteStream} from 'fs';
+import streamToPromise from 'stream-to-promise';
+
+import * as fs from './util/promised-fs';
+import {zipDir} from './util/zip-dir';
+
+
+export default function build({sourceDir, buildDir}: Object) {
```
muffinresearch: I'd be interested to know more about the use of flow here. Do you really need it? To me being unfamiliar with it, it's making the code much less readable.

tests/cli/test.cli.js
```diff
@@ -0,0 +1,41 @@
...
+
+describe('cli.Program', () => {
+
+  it('executes a command callback', () => {
+    let thing = spy(() => new Promise((resolve) => resolve()));
+    let program = new Program(['thing'])
+      .command('thing', 'does a thing', thing);
+    return program.run()
+      .then(() => {
+        expect(thing.called).to.be.equal(true);
```
muffinresearch: Was there a rationale to use expect style here over the TDD style used elsewhere e.g. addons-linter or is this what was used on JPM?

The thing I always like about the standard assert style is it's super direct and obvious.

kumar303: I don't have strong feelings about it. I like the readability of expect but it is slightly harder to type than `assert.equal()` so meh.

src/build.js
kumar303: I thought I'd try Flow and see how annoying/helpful it is. I want to solve these problems:

* Unit tests are the fastest tests but they don't cover the implicit contracts between compositions of units
* Refactoring is hard with a suite of only unit tests
* Changing the interface of any function is hard in a dynamic language

Flow can solve all of these problems. It also has a nice escape hatch where you can annotate a type as `any`. I would like to focus on solving the aforementioned problems only so hopefully we can ignore typing when not necessary.

In this `build()` signature example, the `Object` is just a placeholder. I want to make it into a more meaningful type with: https://github.com/mozilla/web-ext/issues/39

As a counter example, we had a bug in `jpm` once where a refactoring changed the signature of all command handlers but since the unit tests only ran each command handler directly, it was uncaught. The alternative approach to this is to run every test as a shell command so that you invoke the entire process but this adds a magnitude of slowness (at scale) that is unacceptable to me.

By declaring *intent* of usage with type annotations you can catch many incorrect assumptions made by callers when they misunderstand the intent. Flow actually can do a lot of this without the need for explicit type annotations.

muffinresearch: I think it would be good if it didn't require additional syntax. ES6 is one thing, but it's a future standard. Adding non-standard syntax to me seems a step too far, just my 2 cents. It may also hinder contribution.

I wonder if there's any alternative way to get the benefits without compilation?

I suppose there's also an argument here having some integration testing even if the core tests in the tree are just unit tests for the reasons you describe. Might make the flow stuff a bit less necessary.

kumar303: > having some integration testing even if the core tests in the tree are just unit tests

That's exactly what I'd like to avoid though. If we start to rely on integration tests to solve the problems I described then we will make the suite slow. A slow test suite is a slow death. I've seen it on every project I've worked on. When is the last time you were excited about running all of addons-server's tests?

To be clear, I think we should eventually add integration tests but only a very few -- i.e. to act more as "smoke tests" rather than tests that address the integration of components.

> I think it would be good if it didn't require additional syntax

It will definitely require additional syntax because some types can't be inferred (it's a limitation of JavaScript). I did some searching and I see that comment-based syntax is supported but not really documented: http://flowtype.org/blog/2015/02/20/Flow-Comments.html At first it seemed like a good idea but adding a transpilation step is not a problem for us. I fear that relying on comments will just make the code harder to read.

muffinresearch: > To be clear, I think we should eventually add integration tests but only a very few -- i.e. to act more as "smoke tests" rather than tests that address the integration of components.

Yeah that's what I was thinking i.e. not running integration tests at all as part of the normal test suite. E.g. you could run unit tests day to day but only run the integration tests against PRs or merges.

> It will definitely require additional syntax because some types can't be inferred (it's a limitation of JavaScript). I did some searching and I see that comment-based syntax is supported but not really documented: http://flowtype.org/blog/2015/02/20/Flow-Comments.html At first it seemed like a good idea but adding a transpilation step is not a problem for us. I fear that relying on comments will just make the code harder to read.

The biggest issue I have is that the flow type annotations are not easy to parse to the point of making reading and reviewing annotated code much harder, plus it sort of looks like it could be some kind of es-next syntax which is worse imho. All in all that's going to confuse anyone looking at this code until they realise what it is. 

I wonder if a type checking lib might be easier to grok? - e.g: https://www.npmjs.com/package/type-check to just pick one example out of thin air. Although it's a bit crude it's seems much more obvious.

It's up to you but I'm beginning to feel that compiled non-standard extensions to JS are not a good idea - this is on the way to being typescript. Don't get me wrong; I see the problem - I'm just not sure this is a good solution.

kumar303: It's early but there is an [ES8 proposal](https://github.com/sirisian/ecmascript-types) (which flow seems to follow) for type annotations. FWIW.

> https://www.npmjs.com/package/type-check

This is a runtime check which I'd like to avoid because it has performance implications. Flow is completely static. The first babel transform just strips out the annotations.

I definitely think typescript is a bad way to go because it forces you to type everything. I like how Flow is opt-in and how it supports duck-typed interfaces (like Go). I personally don't find the annotations hard to read. I think that for anyone who has even rudimentary knowledge of another static language (Go, Scala, Java, Objective C, ...), it would be intuitive to figure out Flow's annotations. Flow's docs are pretty good although I find their layout confusing. 

I hear what you're saying about hindering contributions though. It will slow someone down the first time if they've never seen Flow. Ultimately, I think it could make for higher quality contributions -- we could rely on interfaces to throw usage errors and it should be easier for a novice to write unit tests because they won't have to "know" too much about the program structure. That is, they will be yelled at by the compiler which should help them figure out the structure

> Don't get me wrong; I see the problem - I'm just not sure this is a good solution.

I'm open to other ideas for solutions. I don't think it would be hard to strip out Flow annotations if we had to. It only requires you to annotate exported functions (inputs and outputs) so I've been trying to do no more than that.

kumar303:
I switched everything to `assert`.

kumar303:
@muffinresearch thanks for all the feedback. I hope to assess Flow's usefulness more fully when I try to address specific problems like https://github.com/mozilla/web-ext/issues/39

Issue #39, opened 2016-02-11
Protect command functions via static typing


kumar303:
Since we have Flow, I think there is a way to protect all command handlers from signature breakage by annotating them with a custom. This would allow us to unit test them directly without worry since either the test invocation or program invocation would fail when the type changes.

PR #58, opened 2016-02-20
Protect handlers from using incorrect options

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/39

Example of an error you get when you use an unknown option:
````
src/build.js:11
 11: export default function build({sourceDir, buildDir}: ProgramOptions) {
                                               ^^^^^^^^ property `buildDir`. Property not found in
 11: export default function build({sourceDir, buildDir}: ProgramOptions) {
                                                          ^^^^^^^^^^^^^^ ProgramOptions
````

This protects all command handlers so that they don't use an unknown option and thus they can be unit tested directly without relying on the construction of a complete program object.

It does not protect against defining a new option on a yargs object and *forgetting* to annotate it in ProgramOptions. I can't figure out how to solve that. Even without that protection, the benefit of using one annotation object, ProgramOptions, for all command handlers is that it's shared. If you add a new handler that relies on a new option then all other handlers will be verified against the new spec.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -8,6 +8,9 @@ import {Program} from './program';
 
 export function main() {
   let program = new Program();
+  // yargs uses magic camel case expansion to expose options on the
+  // final argv object. For example, the 'build-dir' option is available
+  // as argv.buildDir.
   program.yargs
     .usage('Usage: $0 [options]')
     .help('help')
--- a/src/build.js
+++ b/src/build.js
@@ -5,9 +5,10 @@ import streamToPromise from 'stream-to-promise';
 
 import * as fs from './util/promised-fs';
 import {zipDir} from './util/zip-dir';
+import {ProgramOptions} from './program';
 
 
-export default function build({sourceDir, buildDir}: Object) {
+export default function build({sourceDir, buildDir}: ProgramOptions) {
   console.log(`Building web extension from ${sourceDir}`);
   return prepareBuildDir(buildDir)
     .then(() => zipDir(sourceDir))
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by kumar303** [comment]
@muffinresearch it would be cool to get a review from you on this if you have time.

**Comment by tofumatt** [comment]
This makes sense to me, fwiw. Seems like extra boilerplate but if it makes testing better I'm r+

File: src/cli.js [hunk 1]
```diff
@@ -8,6 +8,9 @@ import {Program} from './program';
 
 export function main() {
   let program = new Program();
+  // yargs uses magic camel case expansion to expose options on the
+  // final argv object. For example, the 'build-dir' option is available
+  // as argv.buildDir.
```
muffinresearch: I see this as more that it always provides a camelCased alias. iirc if you have a foo-bar option it is still available as `foo-bar` as well as `fooBar`.

muffinresearch: In fact isn't that where this starts to fall apart. You type stuff is only dealing with one of those two.

kumar303: This approach can't catch a typo in an option configuration or a removal of an option that some handler relies on. I can address that in a follow-up patch if I can figure out how to do it :) 

This patch is still nice because *all* handler's adhere to a contract for what options they receive. If we had ten handlers, for example, each one of their function signatures would be validated.

kumar303:
Managing my other branches is starting to get hairy without this. Thanks for taking a look and feel free to follow up if you want to.

Issue #37, opened 2016-02-10
Name built xpi according to manifest


kumar303:
The build command should name the destination XPI accordingly. It might be easy to do https://github.com/mozilla/web-ext/issues/34 first.

Issue #36, opened 2016-02-10
Exit non-zero upon exception


kumar303:
There is just a placeholder right now where a fatal error just gets logged to the console. We should:
* Exit non-zero
* Show some nice output
* Preserve the traceback

Issue #34, opened 2016-02-09
Do basic manifest validation


kumar303:
This is to check for some very basic manifest validation errors before building a package. This is not to add full integration with [addons-linter](https://github.com/mozilla/addons-linter).

See: https://github.com/mozilla/web-extension-manifest-schema/
Also: https://github.com/mozilla/addons-linter/pull/494

PR #66, opened 2016-02-24
Name a zipped extension file based on manifest

kumar303:
* Fixes https://github.com/mozilla/web-ext/issues/37 (manifest naming)
* Fixes https://github.com/mozilla/web-ext/issues/34 (basic manifest validation)
* Fixes https://github.com/mozilla/web-ext/issues/36 (program exit)


kumar303:
@rpl would you have time to review this? I'll send an email with some details.

tests/test-cmd/test-build/test.build.js
```diff
@@ -0,0 +1,103 @@
...
+describe('build', () => {
+
+  describe('build', () => {
+
+    it('zips a package', () => {
+      let zipFile = new ZipFile();
+
+      return withTempDir(
+        (tmpDir) =>
+          adapter.buildMinimalExt(tmpDir)
```
rpl: Proposal: How about introducing [co](https://github.com/tj/co) (at least in the test cases, because, in my own opinion, currently the implementation is clear enough even with the use of plain Promises) to get nicer async code?

as an example this test case could be turned into something like:
```js
it('zips a package', () => withTempDir(
  (tmpDir) =>
     co(function*() {
       let zipFile = new ZipFile();
 
       let {extensionPath} = yield adapter.buildMinimalExt(tmpDir);
 
       assert.match(extensionPath, /minimal_extension-1\.0\.xpi$/);
 
       yield zipFile.open(extensionPath);
 
       let fileNames = [];
 
       yield zipFile.readEach((entry) => fileNames.push(entry.fileName));
 
       assert.deepEqual(fileNames, ['manifest.json']);
     })
));
```

and as a plus: this mostly matches to our current coding style for async tests in the Firefox source code (e.g. what is achieved in the test suites using [add_task](https://dxr.mozilla.org/mozilla-central/source/browser/components/extensions/test/browser/browser_ext_browserAction_popup.js#129) and [Task.async](https://dxr.mozilla.org/mozilla-central/source/browser/components/extensions/test/browser/browser_ext_browserAction_popup.js#110)

tests/test-util/test.manifest.js
```diff
@@ -0,0 +1,105 @@
...
+          assert.match(error.message, /Could not read manifest\.json/);
+          // Make sure the filename is included in the exception message.
+          // This is actually done by default in file system error messages.
+          assert.include(error.message, nonExistantFile);
+        }));
+    });
+
+    it('reports an error for invalid manifest JSON', () => withTempDir(
+      (tmpDir) => {
+        let badManifest = 'NOT ... VALID ... JSON';
```
rpl: nit, how about using a more realistic invalid manifest.json? something like:
```js
let badManifest = `{
  "name": "I'm an invalid JSON Manifest
  "version": "0.0.0"
}`;
```

it is not going to make a big difference for the current test case, but if we decide to enhance the json validation in a follow-up (e.g. by validating the JSON using [jsonlint](https://github.com/zaach/jsonlint) npm package as a library) it can be an useful example about how to test realistic manifest errors and be sure we catch and report them as planned.

src/cmd/build.js
```diff
@@ -0,0 +1,64 @@
+/* @flow */
+import path from 'path';
+import {createWriteStream} from 'fs';
+import streamToPromise from 'stream-to-promise';
+
+import * as fs from '../util/promised-fs';
+import {zipDir} from '../util/zip-dir';
+import {ProgramOptions} from '../program';
+import getValidatedManifest from '../util/manifest';
+
```
rpl: Nit + Proposal: how about introducing a new eslint rules to raise errors (or warnings) for "multiple empty lines", e.g.

```
    "no-multiple-empty-lines": [2, {"max": 1}]
``` 

or even setting "max" to 2 if we want to use two empty lines for readability from time to time.

rpl:
@kumar303 Thanks a lot for asking me to review this, it was an helpful way to start to look into the internals of the newborn tool :-)

I found the code very clean and readable, then I haven't too much to add (just a couple of proposal and nits, that I hope can be useful)

 :+1:

tests/test-util/test.manifest.js
```diff
@@ -0,0 +1,105 @@
...
+            }));
+        })
+    ));
+
+  });
+
+});
+
+
+function writeManifest(destDir, manifestData) {
```
rpl: Nit, how about moving this helper in ```tests/helpers.js``` file?

tests/test-cmd/test-build/test.build.js
kumar303: omgyes! This is way more concise. I think we should convert the entire code base to this so I filed an issue for it: https://github.com/mozilla/web-ext/issues/72

tests/test-util/test.manifest.js
kumar303: Good idea. I like realistic examples in tests.

src/cmd/build.js
kumar303: Yeah, good idea to control this. I will set the max to two 2 because I am used to reading code in the [Python pep8 style](https://www.python.org/dev/peps/pep-0008/#blank-lines) of two blank lines between functions. I don't see an eslint way to enforce that rule exactly but it's ok :)

The original [setuptools](https://pypi.python.org/pypi/setuptools) code was the most infamous Python offender where there were entire pages of blank lines. It was refused entry into the standard library because of this.

tests/test-util/test.manifest.js
kumar303: The reason I didn't was because I'm not sure any other test will need this helper function. All other tests should be able to work directly with manifest objects.

Issue #76, opened 2016-02-27
Remove all the Flow regex-ignore hacks


kumar303:
Flow's slow startup time has been fixed in 0.22 (see https://github.com/facebook/flow/issues/1461 for context). We can now remove the auto-generated ignore rules. Things to remove:

* config fixing tasks
* config check from tests and travis CI
* all mention of it in the docs

We will still need to ignore some modules because of Flow bugs (e.g. https://github.com/facebook/flow/issues/1420)

kumar303:
Funny enough, the regular expressions were also slowing Flow down (well, originally the ignore rules provided a 2x speedup). After removing all regex hacks, startup time is 2.6 seconds. With Flow 0.21, startup time was about 6 seconds.

PR #89, opened 2016-03-01
Add support for code coverage reports

### LINKED ISSUE #73: Add code coverage
Before we get too deep into web-ext, let's set up code coverage

rpl:
Fixes #73

This PR adds optional code coverage to the project.

The coverage reports can be produced by setting the COVERAGE environment variable:
```
COVERAGE=y npm test
```

In the .travis.yml there are changes to run the test suite with and without the code coverage instrumentation.


rpl:
changed the environment var check to test for "COVERAGE=enabled" which looks more explicit.

package.json
```diff
@@ -44,6 +44,8 @@
   "devDependencies": {
     "babel-core": "6.6.0",
     "babel-eslint": "5.0.0",
+    "babel-istanbul": "^0.6.0",
```
kumar303: I've been trying to use exact requirements (``"babel-istanbul": "0.6.0"``) so that there are no surprises when another contributor installs a fresh copy of the repo. Theoretically the caret bug fix releases won't introduce breaking changes but you have to trust that all other developers will be good about that.

kumar303: Also, with exact requirements it lets us use [greenkeeper](http://greenkeeper.io/) to test with new releases as they happen.

tests/runner.js
```diff
@@ -3,5 +3,24 @@
...
+    let reporter = new istanbul.Reporter();
+    let sync = false;
+
+    collector.add(global.__coverage__);
+
+    reporter.add('text');
+    reporter.addAll([ 'lcov', 'clover' ]);
+    reporter.write(collector, sync, function() {
+      console.log('All reports generated');
+      done();
```
kumar303: Can we cause the test suite to fail if there is a low amount of coverage, maybe 90% or something? That could be done in a follow up patch too.

kumar303:
Looks like there is a conflict that needs resolving but I'm not sure what it is. Maybe the flowconfig? That will go away in https://github.com/mozilla/web-ext/issues/76

kumar303:
This looks great so far. It also needs documentation for how to run coverage, maybe a new heading under the [development section](https://github.com/mozilla/web-ext#development-of-web-ext).

kumar303:
Maybe @muffinresearch could also take a quick look

package.json
muffinresearch: This patch should fix that in future: https://github.com/mozilla/web-ext/pull/92

tests/runner.js
muffinresearch: If you set up coveralls you can get it to fail when coverage drops below a certain overall percentage threshold or by a defined percentage. It will comment on PRS too.

.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: node_js
+env:
+- COVERAGE=
```
muffinresearch: Looks like a dupe.

muffinresearch:
Does babel-istanbul provide coverage that points to the original es6 src or the transpiled src?

rpl:
@muffinresearch yep, the coverage should points to the original es6 src

.travis.yml
kumar303: How so? It looks to me like it will do one run of the tests without coverage and one with coverage.

However, I wonder if it matters. Could we just do one run of the tests with coverage enabled?

rpl: @muffinresearch the goal was to run the tests with and without the code coverage instrumentation applied.

I'm not sure if we want to keep it, the main reason can be to spot any failure which doesn't happen if the code coverage instrumentation has not been applied (it should not happen, but it is not completely impossible)

tests/runner.js
rpl: That is interesting, I didn't know about the coveralls service. 
I'm looking into it, it seems that it would be possible to integrate it by just sent to coveralls the coverage results in the lcov format, e.g.:

```
npm install coveralls && cat ./coverage/lcov.info | coveralls
```

package.json
rpl: @kumar303 @muffinresearch sorry, I installed the deps using "npm install --save-dev" and I didn't notice the exact requirements of all the other deps.

.travis.yml
kumar303: ok, I'd rather do one run of the test suite with coverage enabled. If, at some point, we get suspicious of the results then we can start running the suite twice. The reason I'd like only one run of the test suite is to keep the tests very fast. I know they are fast now but they will only get slower over time. Also, the Flow startup is still a bit slow (3 seconds).

muffinresearch: Ah I get it now. I didn't understand the intention.

.travis.yml
```diff
@@ -5,7 +5,8 @@ node_js:
 - '4'
 before_install:
 - npm install -g npm
-script: npm test
+script: COVERAGE=y npm test
+after_script: "npm install coveralls@2.11.8 && cat ./coverage/lcov.info | coveralls"
```
kumar303: can't we just put `coveralls@2.11.8` into dev requirements?

rpl: @kumar303 yep, we can, but as far I read from the coveralls docs (which I'm reading right now for the first time and so it's perfectly possible that I'm wrong) it needs a separate config file with the developer account keys to work locally, so I thought to reduce the amount of dependencies installed locally but installing it directly on travis
(but I'm not against to add it into the devDependencies)

kumar303: I think it would be better to add it to dev dependencies just to keep the configuration of the build all in one place. I see that it's only needed for Travis but things could get confusing if we only install dependencies on Travis and nowhere else.

rpl:
@kumar303 @muffinresearch here are the changes applied:

- coverage enabled by checking ```process.env.COVERAGE === 'y'``` (I know... it is the third time I'm changing it, I was in the search for a good compromise)
- build with coverage by default on travis
- added coveralls to the dev dependencies
- submit coverage info to coveralls on travis builds
- sync=true when writing the coverage report (because it happened that the coverage has [turned the travis build into a failure in my fork for timeout reason in the afterAll hook](https://travis-ci.org/rpl/web-ext/jobs/113188032#L1294), but I can't be sure if this change helps until it eventually fails again)

rpl:
@kumar303 @muffinresearch and probably the repo has to be explicitly enabled on coveralls.

muffinresearch:
Yeah I've flipped that on but you'll need the secret. Give me a sec and I'll give you the encrypted travis details.

muffinresearch:
This patch should sort it:

```diff
diff --git a/.travis.yml b/.travis.yml
index 7d59cd9..d8cf4db 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,6 @@ notifications:
     - irc.mozilla.org#amo-bots
     on_success: change
     on_failure: always
+env:
+  global:
+    secure: datNSXsx0c1bq2sNrBxLsL8eACAwiHmRB8GR/Eb01dAT4mZ46JmcaC+COFUXgLrFIm9gic+U+GOrV0Zc8HOB6/PwocnhbatRI0ztWSzDlZU6TyeUlw6LLq1KIINbNAeUc2L44HUJhdYlaAoqXQ4wv21zJHf5H/SqfwneYjvhfJtPyHA3MwyE43y5i6ri1bv0FAqVCnHBKRnSVVqGwbFN+mqSZEzcBU7nTg9Snffe4cNTJHFD/H0yUbhHSHv4Kwefy/KzZ2xrzVo3OMl7y4DyiQ6TGGr+adKUZIcPlGrGr5WCBSWfrZEEz517aW7244v5SAvQXWHL7kWqkhHD1hPKB0vhZrvvAphd3g2C8IBnuSt7otNLaky+Jd6kG2/mEKE4DAyxSoB/SGspJkjlqhiQVfWFIENOHvM2MHJuOypshVmvlNQvKJFHvwZFHM51YZwDuFf+GZI39Y5nNhwxdjL49KuKWeO3GoYkTef4v934X4h0hhMBdSHbFY1hafrBvhlzYbKGC99MCmq0NjdF+qJFAlV0Y++dA2NhQYEuOZgUPdaKTlLA9z3015ZVj8pnnSrqrKdsyLO8aBmATeLaV2yT3N7IWvF6eS+u818MKPeHHeijjVQ83DN/HokpwzO200vRbnkixjkXYH/vrNNWd/couheBdL502fr9r2dxK7yX8kk=
```

kumar303:
I would also like to see `COVERAGE=y` documented in the [development section](https://github.com/mozilla/web-ext#development-of-web-ext).

README.md
```diff
@@ -87,6 +87,20 @@ to regenerate the ignore rules. Do so like this:
 
 After that, commit and push the new `.flowconfig`.
 
+### Code Coverage
+
+You can generate Code Coverage reports every time you run the test suite run
+by settings `$CODE_COVERAGE` in the environment. Here is an example of running
```
kumar303: should be: '...every time you run the test suite by setting ...'

README.md
```diff
@@ -87,6 +87,20 @@ to regenerate the ignore rules. Do so like this:
...
+the test suite on the instrumented source code:
+
+    CODE_COVERAGE=y npm test
+
+or even when the test suite re-runs automatically as you edit files:
+
+    CODE_COVERAGE=y npm develop
```
kumar303: this is actually `npm run develop`

README.md
```diff
@@ -76,6 +76,20 @@ the test suite without lint checks:
 This project relies on [flow](http://flowtype.org/) to ensure functions and
 classes are used correctly. Run all flow checks with `npm run flow-check`.
 
+### Code Coverage
+
+You can generate Code Coverage reports every time you run the test suite
+by settings `$CODE_COVERAGE` in the environment. Here is an example of running
```
kumar303: should be '...by setting' (instead of settings).

Issue #87, opened 2016-03-01
Remove all Flow adapters that are used for testing


kumar303:
Dev dependencies are ignored by Flow because of slow startup time. This meant that it was hard to opt-in test files to Flow. After https://github.com/mozilla/web-ext/issues/76 I think we will be able to opt-in all test files. This also means we can delete all the `tests/.../adapter.js` files which were a hacky way to get some Flow coverage.

Issue #71, opened 2016-02-26
Run extension in a custom Firefox binary


kumar303:
https://github.com/mozilla/web-ext/issues/6 installs/runs the extension in the default Firefox. Add an option to specify a path to a custom Firefox.

kumar303:
This is documented https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run

PR #90, opened 2016-03-01
Adds run command -> install and launch extension

kumar303:


kumar303:
This pattern of adding `adapter.js` to the tests is something that will go away in https://github.com/mozilla/web-ext/issues/87 . It was just a hack.

rpl:
@kumar303 I think that we need to add to the cli a new "--firefox-bin" option (as a non mandatory option), because the firefox binary auto-discovery sometimes fails (e.g. on my ubuntu it decides that the binary should be /usr/lib64/firefox doesn't even exist on my system)

src/firefox/index.js
```diff
@@ -0,0 +1,133 @@
...
+    })
+    .catch(onlyErrorsWithCode('ENOENT', () => {
+      console.log(`Creating extensions directory: ${profile.extensionsDir}`);
+      return fs.mkdir(profile.extensionsDir);
+    }))
+    .then(() => {
+      let readStream = nodeFs.createReadStream(extensionPath);
+      let id = manifestData.applications.gecko.id;
+
+      // TODO: also support copying a direcotry of code to this
```
rpl: nit, typo: direcotry -> directory

kumar303:
> I think that we need to add to the cli a new "--firefox-bin" option

Yep, adding that in https://github.com/mozilla/web-ext/issues/71

package.json
```diff
@@ -37,6 +37,8 @@
     "babel-polyfill": "6.6.1",
     "es6-error": "2.0.2",
     "es6-promisify": "3.0.0",
+    "firefox-profile": "^0.3.11",
```
rpl: @kumar303 This should be turned into an exact version match

kumar303: d'oh! Good catch.

src/firefox/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+        firefox.on('error', (error) => {
+          // TODO: show a nice error when it can't find Firefox.
+          // if (/No such file/.test(err) || err.code === 'ENOENT') {
+          console.log(`Firefox error: ${error}`);
+          throw error;
+        });
+
+        firefox.stderr.on('data', (data) => {
+          console.log(`stdout: ${data.toString().trim()}`);
```
rpl: @kumar303 shouldn't this be stderr? (e.g. ```console.log(`stderr: ...`);```)

why not using console.error here to route these messages to the stderr?

src/firefox/index.js
```diff
@@ -0,0 +1,133 @@
...
+        });
+
+        firefox.stdout.on('data', function(data) {
+          console.log(`stderr: ${data.toString().trim()}`);
```
rpl: @kumar303 if I wasn't wrong in the above comment, then this should be console.log(`stdout: ...`);

src/firefox/index.js
kumar303: oh yeah, heh, good catch. I also plan to fix up all the logging in https://github.com/mozilla/web-ext/issues/33

src/firefox/index.js
kumar303: ha! I don't eve know how that happened.

rpl:
@kumar303 great! this PR looks ready.

I'm so happy that I can now run a webextension addon without leaving the command line :-)

kumar303:
Thanks for the review! Yes, definitely let me know how it works for you. There are, of course, a lot of missing features (a lot of issues filed).

PR #92, opened 2016-03-02
Add save-prefix setting via .npmrc

muffinresearch:
This will default all npm deps to be added without `^`.


kumar303:
gold-star

Amazing! I always wondered how to fix this. Thank you.

Issue #99, opened 2016-03-02
web-ext --version results in ENOENT: no such file or directory, open 'package.json'


kumar303:
The code that checks for a package.json file is broken. We rely on `__dirname` which webpack apparently sets to a relative directory which is unhelpful. We need a better way to detect the path and find the package.json file. Maybe there is a way to reach into node internals? For some reason `require.main.filename` is also undefined. 

ideas:
* Look at `$PATH` and see if we can find the `web-ext` executable
* Look at some system module and try to figure out where web-ext is

rpl:
@kumar303 I've just pushed the PR #102 which contains a proposed fix for this issue.

This issues is probably related to webpack, the simpler fix that I found is to convert the relative `__dirname` path into an absolute path to the `./bin` dir (by using `path.resolve`) and 
then to pass the resulting path to the main function as a parameter. 
(which works correctly, because it happens before the `require` is going to be redefined by `webpack`).

kumar303:
Fixed in https://github.com/mozilla/web-ext/pull/102

PR #102, opened 2016-03-03
Fix ENOENT on running the --version cli option

rpl:
This PR contains a proposed fix for #99


kumar303:
Good idea! Thanks. I just tried the branch and it worked for me.

PR #105, opened 2016-03-03
Run extension with a custom Firefox binary

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/71

Some other things:
* Moved `cli.js` into `program.js` because all this logic is related
* Added missing tests for `temp-dir.js`
* Improved coverage of the `main()` program code
* Fixed up option handling since this feature required command sub-options for the first time
* Regressed static coverage of handler options, to be addressed in https://github.com/mozilla/web-ext/issues/106
* Made `--build-dir` specific to `build` rather than a global option


kumar303:
@rpl could you review?

src/program.js
```diff
@@ -1,45 +1,54 @@
...
 
 
 /*
- * Pseudo-class for all global program options.
- *
- * This is used to validate the definition of command handlers.
- * Each class instance variable is a camel case expanded program
- * option. Each option is defined in program.yargs.option(...).
- */
-export class ProgramOptions {
```
kumar303: I removed this because it wasn't working (too many errors I could not resolve). I filed https://github.com/mozilla/web-ext/issues/106 to look for a solution.

src/cmd/index.js
```diff
@@ -0,0 +1,4 @@
+/* @flow */
+import build from './build';
+import run from './run';
+export {build, run};
```
rpl: This is probably a good candidate for an `export default {build, run}'.

if we do this kind of change, we need to change the corresponding imports [1] to something like:

```
import defaultCommands from  './cmd';
```

[1]: in https://github.com/mozilla/web-ext/pull/105/files#diff-9589e3a0cc72e26a4814bc9ef05aa2d4R6 
  and 
in https://github.com/mozilla/web-ext/pull/105/files#diff-2d9bd4ae7220a18dfa5b6b6707f45213R7

kumar303: ah, good idea. I didn't know that was possible :)

rpl:
@kumar303 looks great

+1

Issue #69, opened 2016-02-26
Run an extension in a custom profile


kumar303:
https://github.com/mozilla/web-ext/issues/6 will create a temporary profile every time. Add an option to copy an existing profile.

kumar303:
This is documented https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#--firefox-profile_-p

PR #124, opened 2016-03-08
Run extension in a custom Firefox profile

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/69

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -48,7 +48,7 @@ export function onlyInstancesOf(
 
 
 /*
- * Sugar-y way to catch only errors with a certain code.
+ * Sugar-y way to catch only errors having certain code(s).
  *
  * Usage:
  *
@@ -57,15 +57,31 @@ export function onlyInstancesOf(
  *      // error.code is guaranteed to be ENOENT
  *    }))
  *
+ *  or:
+ *
+ *  Promise.resolve()
+ *    .catch(onlyErrorsWithCode(['ENOENT', 'ENOTDIR'], (error) => {
+ *      // ...
+ *    }))
+ *
  * All other errors will be re-thrown.
  *
  */
 export function onlyErrorsWithCode(
-    codeWanted: string, errorHandler: Function): Function {
+    codeWanted: string | Array<string>, errorHandler: Function): Function {
   return (error) => {
-    if (error.code !== codeWanted) {
+    let throwError = true;
+
+    if (Array.isArray(codeWanted) && codeWanted.indexOf(error.code) !== -1) {
+      throwError = false;
+    } else if (error.code === codeWanted) {
+      throwError = false;
+    }
+
+    if (throwError) {
       throw error;
     }
+
     return errorHandler(error);
   };
 }
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -2,10 +2,13 @@
 import nodeFs from 'fs';
 import path from 'path';
 import defaultFxRunner from 'fx-runner/lib/run';
-import FirefoxProfile from 'firefox-profile';
+import FirefoxProfile, {copyFromUserProfile as defaultUserProfileCopier}
+  from 'firefox-profile';
 import streamToPromise from 'stream-to-promise';
-
 import fs from 'mz/fs';
+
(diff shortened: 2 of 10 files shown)


kumar303:
@rpl this next one is ready for review

I'm still a little confused about what's happening with the missing coverage. Maybe the flow adapter removal patch did introduce a regression somehow.

kumar303:
Weird. This new PR based on master appears to be submitting coverage (it's not finished yet but I see the progress bar) https://github.com/mozilla/web-ext/pull/109

kumar303:
> This new PR based on master appears to be submitting coverage

It never finished

````
¯\_(ツ)_/¯
````

File: src/errors.js [hunk 1]
```diff
@@ -57,15 +57,31 @@ export function onlyInstancesOf(
...
  *
  */
 export function onlyErrorsWithCode(
-    codeWanted: string, errorHandler: Function): Function {
+    codeWanted: string | Array<string>, errorHandler: Function): Function {
   return (error) => {
-    if (error.code !== codeWanted) {
+    let throwError = true;
+
+    if (codeWanted instanceof Array && codeWanted.indexOf(error.code) !== -1) {
```
rpl: This can be ```Array.isArray(codeWanted)``` (but it is not very important because this code runs in node, and so it will not make a great difference)

src/program.js
```diff
@@ -67,10 +67,14 @@ export class Program {
         if (cmd) {
           prefix = `${cmd}: `;
         }
-        console.error(prefix + error.toString());
+
+        console.error();
```
rpl: These empty ```console.error()``` method calls  look strange, I suppose that they are used to create empty lines, 
how about prefixing the interpolated strings with a ```\n```? 
e.g. ```console.error(`\n${prefix}${error.stack}`);```

src/firefox/index.js
```diff
@@ -89,12 +99,54 @@ export function createProfile(
...
+ * It resolves with the new profile object.
+ *
+ * The temporary profile will be deleted when the system process exits.
+ *
+ * The existing profile can be specified as a directory path or a name of
+ * one that exists in the current user's Firefox directory.
+ */
+export function copyProfile(
+    profileDirectory: string,
+    {copyFromUserProfile=FirefoxProfile.copyFromUserProfile}: Object = {})
```
rpl: I suppose that this `copyFromUserProfile` option and its default value are mainly used to be able to inject a fake ```copyFromUserProfile``` method to simplify the testing of this helper.

how about using the same convention used in the other methods like this, by defining it in the module as a `const defaultCopyFromUserProfile`?

```
const defaultCopyFromUserProfile = FirefoxProfile.copyFromUserProfile;
...
export function copyProfile(
    profileDirectory: string,
    {copyFromUserProfile=defaultCopyFromUserProfile}: Object = {})
...
```

src/firefox/index.js
```diff
@@ -89,12 +99,54 @@ export function createProfile(
...
+    : Promise {
+
+  let copy = promisify(FirefoxProfile.copy);
+  let copyByName = promisify(copyFromUserProfile);
+
+  return isDirectory(profileDirectory)
+    .then((dirExists) => {
+      if (dirExists) {
```
rpl: Currently this `copyProfile` helper doesn't set the preferences on the profile after copying it, as it is done in its `createProfile` counterpart.

We have to set the preferences even when we are cloning an existent profile (either by setting the preferences in this method or by refactoring out the "set the app preferences" part from the `createProfile` method and share it with both the methods that are used to create a profile) or the installed addon will probably not work correctly in the copied profile (e.g. the addon is disabled because is not signed and the copied preferences doesn't disable the signatures checks)

src/program.js
kumar303: ah, yes, good idea. Hopefully this will all look nicer once [logging is in place](https://github.com/mozilla/web-ext/issues/33).

src/firefox/index.js
kumar303: ah, good catch.

kumar303: Fixed

Issue #85, opened 2016-03-01
Ignore common artifacts when zipping a directory


kumar303:
The build command (https://github.com/mozilla/web-ext/issues/4) currently zips up the entire directory. It should ignore things like `.git` and other artifact. This is how jpm does it: https://github.com/mozilla-jetpack/jpm/blob/master/lib/ignore.js#L23

kumar303:
It should also not include copies of itself -- this currently happens because the local build directory gets included.

kumar303:
This is documented https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build

PR #130, opened 2016-03-09
Ignore commonly unwanted files when building XPI

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/85

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -1,5 +1,6 @@
 /* @flow */
 import path from 'path';
+import minimatch from 'minimatch';
 import {createWriteStream} from 'fs';
 import streamToPromise from 'stream-to-promise';
 
@@ -11,7 +12,7 @@ import getValidatedManifest from '../util/manifest';
 
 export default function build(
     {sourceDir, buildDir}: Object,
-    {manifestData}: Object = {}): Promise {
+    {manifestData, fileFilter}: Object = {}): Promise {
 
   console.log(`Building web extension from ${sourceDir}`);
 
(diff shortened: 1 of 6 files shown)


src/cmd/build.js
```diff
@@ -62,3 +69,35 @@ export function prepareBuildDir(buildDir: string): Promise {
...
+
+
+/*
+ * Allows or ignores files when creating a ZIP archive.
+ */
+export class FileFilter {
+  filesToIgnore: Array<string>;
+
+  constructor({filesToIgnore}: Object = {}) {
+    this.filesToIgnore = filesToIgnore || [
```
rpl: This PR looks great!

The only thing that I'm not completely sure about is related to the `filesToIgnore` defaults and how its overriding will work (especially when it will be available through a config file or a command line option), and as this feature is not exposed to the user yet, most of this could be discussed and resolved in follow-ups if we prefer.

here are a couple of doubts/ideas:

- by looking into the test cases I though that `manifest.json` is not a great example of possible `filesToIgnore` customization, because we probably want to be sure it is never excluded by mistake (e.g. by always prepending `"!manifest.json"` to this array of patterns) or the resulting xpi could be invalid even if the extension source dir was completely valid (excluding the hypotetic filesToIgnore customization).

- how about excluding any hidden file by default? (e.g. by using `"**/.*"` instead of `"**/.git"`

- if we suppose that the addon developer can customize the excluded/included patterns somehow, how about composing the default patterns and the user patterns instead of a providing a full override by default? e.g. something like `[].concat(mandatory_patterns, user_patterns, defaults_patterns)`

kumar303: Good points! I fixed the first two. As for concatenating filter overrides, I left it as is. I think this will become more clear once we implement customizations -- I don't want to make it too flexible before then. The feature I added to override all default filter rules will probably still be necessary since it is valid to include nested XPIs inside an XPI. Some very large add-ons do this.

Issue #120, opened 2016-03-08
Hide logs from test output


kumar303:
Once we have a logger (https://github.com/mozilla/web-ext/issues/33) we can hide log output from the test run. However, I want to see log output when there is a failure. Hopefully we can accomplish that by creating a mocha reporter or something.

PR #160, opened 2016-03-28
Adds a logger and implements --verbose

kumar303:
* Adds a logger, implements `--verbose`, and converts all old console calls (fixes https://github.com/mozilla/web-ext/issues/33)
* Adds log capturing for tests so that only failing tests spew logs (fixes https://github.com/mozilla/web-ext/issues/120)

Example:

````
$ web-ext build
Building web extension from /Users/kumar/dev/bookmarker-ext/moz-webext
Your web extension is ready: /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.25.xpi
````

````
$ web-ext --verbose build
[cmd/build.js][info] Building web extension from /Users/kumar/dev/bookmarker-ext/moz-webext
[util/manifest.js][debug] Validating manifest at /Users/kumar/dev/bookmarker-ext/moz-webext/manifest.json
[cmd/build.js][debug] Not including file /Users/kumar/dev/bookmarker-ext/moz-webext/.DS_Store in ZIP archive
[cmd/build.js][debug] Not including file /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.22.xpi in ZIP archive
[cmd/build.js][debug] Not including file /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.23-fx.xpi in ZIP archive
[cmd/build.js][debug] Not including file /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.25-fx.xpi in ZIP archive
[cmd/build.js][debug] Not including file /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.25.xpi in ZIP archive
[cmd/build.js][info] Your web extension is ready: /Users/kumar/dev/bookmarker-ext/moz-webext/web-ext-artifacts/getting_started_example-1.2.25.xpi
````

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -11,6 +11,9 @@ import isDirectory from '../util/is-directory';
 import {promisify} from '../util/es6-modules';
 import {onlyErrorsWithCode, WebExtError} from '../errors';
 import {getPrefs as defaultPrefGetter} from './preferences';
+import {createLogger} from '../util/logger';
+
+const log = createLogger(__filename);
 
 
 export const defaultFirefoxEnv = {
@@ -25,7 +28,7 @@ export function run(
     profile: FirefoxProfile,
     {fxRunner=defaultFxRunner, firefoxBinary}: Object = {}): Promise {
 
-  console.log(`Running Firefox with profile at ${profile.path()}`);
+  log.info(`Running Firefox with profile at ${profile.path()}`);
   return fxRunner(
     {
       // if this is falsey, fxRunner tries to find the default one.
@@ -44,26 +47,26 @@ export function run(
       return new Promise((resolve) => {
         let firefox = results.process;
 
-        console.log(`Executing Firefox binary: ${results.binary}`);
-        console.log(`Executing Firefox with args: ${results.args.join(' ')}`);
+        log.debug(`Executing Firefox binary: ${results.binary}`);
+        log.debug(`Executing Firefox with args: ${results.args.join(' ')}`);
 
         firefox.on('error', (error) => {
           // TODO: show a nice error when it can't find Firefox.
           // if (/No such file/.test(err) || err.code === 'ENOENT') {
-          console.log(`Firefox error: ${error}`);
+          log.error(`Firefox error: ${error}`);
           throw error;
         });
 
         firefox.stderr.on('data', (data) => {
-          console.error(`stderr: ${data.toString().trim()}`);
+          log.error(`Firefox stderr: ${data.toString().trim()}`);
         });
 
         firefox.stdout.on('data', (data) => {
-          console.log(`stdout: ${data.toString().trim()}`);
+          log.debug(`Firefox stdout: ${data.toString().trim()}`);
         });
 
         firefox.on('close', () => {
-          console.log('Firefox closed');
+          log.debug('Firefox closed');
           resolve();
         });
       });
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: src/firefox/index.js [hunk 1]
```diff
@@ -44,26 +47,26 @@ export function run(
...
         });
 
         firefox.stderr.on('data', (data) => {
-          console.error(`stderr: ${data.toString().trim()}`);
+          log.error(`stderr: ${data.toString().trim()}`);
         });
 
         firefox.stdout.on('data', (data) => {
-          console.log(`stdout: ${data.toString().trim()}`);
+          log.info(`stdout: ${data.toString().trim()}`);
```
rpl: I'm not sure that we want to log the Firefox binary stdout data in the bunyan `INFO` log level,
how about change this into a `log.debug`?

rpl: Even when the logger is not in verbose mode, we probably want to make it clear when the logs come from the Firefox binary, how about add it here as part of the log message? e.g.

```
        firefox.stderr.on('data', (data) => {
          log.error(`stderr: ${data.toString().trim()}`);
          log.error(`${results.binary} stderr: ${data.toString().trim()}`);
        });

        firefox.stdout.on('data', (data) => {
          log.info(`stdout: ${data.toString().trim()}`);
          log.debug(`${results.binary} stdout: ${data.toString().trim()}`);
        });
```

src/firefox/index.js
```diff
@@ -44,26 +47,26 @@ export function run(
...
         });
 
         firefox.on('close', () => {
-          console.log('Firefox closed');
+          log.debug('Firefox closed');
```
rpl: imho, this is probably a better candidate for the `log.info` level than the above firefox stdout logs.

src/firefox/index.js
rpl: In the code coverage data, these two event handlers (`firefox.stderr.on('data', ...)` and `firefox.stdout.on('data', ...)`) are the only two lines not yet covered by tests in this module.

Now that we have a proper logger module, which can collect the logged messages, we can probably a new test case which tests how the logger behaves, and reach a full coverage of this module.

rpl:
@kumar303 I've added a couple of comments about the log events associated to the log messages generated by the Firefox instance during a `web-ext run`.

Anyway, this PR looks great! and now we can add the `no-console` eslint rule and prevent any further usage of the console object in the next patches. (the only file in which we probably want to disable the no-console eslint rule on is the test fixture in `tests/fixtures/minimal-web-ext/background-script.js`), e.g.:

```diff
diff --git a/.eslintrc b/.eslintrc
index 2b097ae..dfe2124 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -37,6 +37,7 @@
     "no-caller": 2,
     "no-cond-assign": [2, "except-parens"],
     "no-debugger": 2,
+    "no-console": 2,
     "no-empty": 2,
     "no-eval": 2,
     "no-extend-native": 2,
diff --git a/tests/fixtures/minimal-web-ext/background-script.js b/tests/fixtures/minimal-web-ext/background-script.js
index def0db5..2fd54ef 100644
--- a/tests/fixtures/minimal-web-ext/background-script.js
+++ b/tests/fixtures/minimal-web-ext/background-script.js
@@ -1 +1,3 @@
+/* eslint no-console:0 */
+
 console.log('background script loaded');
```

src/firefox/index.js
kumar303: hmm, on Mac this is a long string (`/Applications/Firefox.app/Contents/MacOS/firefox-bin`). I think making it `Firefox` should be good enough to indicate where that logging is coming from.

src/firefox/index.js
kumar303: for normal operation, Firefox will close because the user said so. Thus, I don't think it needs to be an info message. The error callback is already logged (in the case of a crash).

kumar303:
Good idea on the `no-console` linting. Thanks for the review!

Issue #193, opened 2016-04-14
Build XPI continuously as files are edited


kumar303:
As a first step to auto-reloading (https://github.com/mozilla/web-ext/issues/138), let's build continuously as files are edited. It could be run as:

    web-ext build --as-needed

This will provide the function necessary to continuously build an XPI file. However, auto-reloading will eventually not need to use XPI files.

kumar303:
documented: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#--as-needed

PR #197, opened 2016-04-18
Added build --as-needed to watch and rebuild

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/193

This lets you watch files and rebuild the extension, like:

    web-ext build --as-needed

It's not such an interesting feature itself but it is a stepping stone to re-installing an XPI which will be one way to reload changes in Firefox.

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -4,6 +4,7 @@ import minimatch from 'minimatch';
 import {createWriteStream} from 'fs';
 import streamToPromise from 'stream-to-promise';
 
+import defaultSourceWatcher from '../watcher';
 import {zipDir} from '../util/zip-dir';
 import getValidatedManifest from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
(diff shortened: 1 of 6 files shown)


src/cmd/build.js
```diff
@@ -30,31 +33,48 @@ export default function build(
...
+
+  return prepareArtifactsDir(artifactsDir)
+    .then(() => createPackage())
+    .then((result) => {
+      if (rebuildWhenSourceChanges) {
+        log.info('Rebuilding when files change...');
+        onSourceChange({
+          sourceDir, artifactsDir, onChange: () => {
+            return createPackage().catch((error) => {
+              throw error;
```
rpl: This error is not currently logged (it probably ends up into the node [uncaughtexception](https://nodejs.org/api/process.html#process_event_uncaughtexception).

we risk to don't warn the user of errors that he probably wants to know about, e.g.:
- if we during a "web-ext build --as-needed" session, we suddently remove the `web-ext-artifacts` dir, currently no error is shown in the web-ext logs.

we can even just log the error in addition to the `throw` (maybe with the stack trace information when it is running in verbose mode), e.g. something like:
```js
return createPackage().catch((error) => {
    log.error(`Error during build: ${error}`);
    log.debug(`Last error debug info: ${error.stack}`);
    throw error;
}
```

src/watcher.js
```diff
@@ -0,0 +1,38 @@
...
+  // TODO: support windows See:
+  // http://stackoverflow.com/questions/10021373/what-is-the-windows-equivalent-of-process-onsigint-in-node-js
+  process.on('SIGINT', () => watcher.close());
+  return watcher;
+}
+
+
+export function proxyFileChanges({artifactsDir, onChange, filePath}: Object) {
+  if (filePath.indexOf(artifactsDir) === 0) {
+    log.debug(`Ignoring change to: ${filePath}`);
```
rpl: It would be nice if the watched files could be filtered using the same file filter already used by the build command by default.

e.g. the build file filter currently ignores the hidden files (I mean files which start with a '.', like temp file wrong the editors or the .gitignore), but currently any changes on dotfiles generate an xpi rebuild even if they will not be included into the xpi.

src/watcher.js
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    log.info(`Changed: ${filePath}`);
```
rpl: is it possible to log here the type of change happened? (e.g. if the file change is related to a modified file, a new file or a removed file)

src/watcher.js
```diff
@@ -0,0 +1,38 @@
...
+
+export default function onSourceChange(
+    {sourceDir, artifactsDir, onChange}: Object): Watchpack {
+  // TODO: For network disks, we would need to add {poll: true}.
+  const watcher = new Watchpack();
+
+  const executeImmediately = true;
+  watcher.on('change', debounce((filePath) => {
+    proxyFileChanges({artifactsDir, onChange, filePath});
+  }, 1000, executeImmediately));
```
rpl: I think that is not important or mandatory, but it probably worth a mention:
I'm wondering if in some scenarios it could be helpful being able to configure the debounce time (e.g. in case the changes are generated from another build tool, like webpack or browserify).

src/watcher.js
```diff
@@ -0,0 +1,38 @@
...
+const log = createLogger(__filename);
+
+
+export default function onSourceChange(
+    {sourceDir, artifactsDir, onChange}: Object): Watchpack {
+  // TODO: For network disks, we would need to add {poll: true}.
+  const watcher = new Watchpack();
+
+  const executeImmediately = true;
+  watcher.on('change', debounce((filePath) => {
```
rpl: Nit: how about extracting the debounce callback for readability? (probably for the same reason, `executeImmediately` is defined at line 15 instead of being passed as a `true` literal value in the `debounce` parameters), e.g. something like:

```js
  const executeImmediately = true;
  const onFileChange = (filePath) => {
    proxyFileChanges({artifactsDir, onChange, filePath});
  };
  watcher.on('change', debounce(onFileChange, 1000, executeImmediately));
```

src/cmd/build.js
```diff
@@ -13,8 +14,10 @@ const log = createLogger(__filename);
 
 
 export default function build(
-    {sourceDir, artifactsDir}: Object,
-    {manifestData, fileFilter}: Object = {}): Promise {
+    {sourceDir, artifactsDir, asNeeded: rebuildWhenSourceChanges}: Object,
```
rpl: I find the renaming of `asNeeded` to `rebuildWhenSourceChanges` in the above function parameters a bit hard to read (and the flow annotations makes it even a bit harder then usual), I propose to use it directly as `asNeeded` (without renaming it), or to rename it to `rebuildWhenSourceChanges` into the function body.

src/cmd/build.js
```diff
@@ -30,31 +33,48 @@ export default function build(
...
-
-        return promisedStream
-          .then(() => {
-            log.info(`Your web extension is ready: ${extensionPath}`);
-            return {extensionPath};
-          });
-      })
-    );
+  if (!createPackage) {
+    createPackage = () => {
```
rpl: The `createPackage` parameter is probably needed to simplify the testing of this feature, how about refactoring this `createPackage` helper out of the `build` function body (mostly for readability reasons), e.g. I imagine that it could become something like:

```js
const defaultCreatePackage = (
  {resolveManifest, fileFilter, sourceDir, artifactsDir}: Object): Function => {
  return function() {
     // the same body of the original arrow function
  };
};

export default function build(
  ...
  if (!createPackage) {
    createPackage = defaultCreatePackage({
      resolveManifest,
      fileFilter,
      sourceDir,
      artifactsDir,
    });
  }
  ...
```

rpl:
During this review I noticed a couple of errors produced during the test run, but they are not related to this change (basically firefox-profile adds [two listeners to the process object](https://github.com/saadtazi/firefox-profile-js/blob/master/lib/firefox_profile.js#L144) that are then removed in the [deleteDir](https://github.com/saadtazi/firefox-profile-js/blob/master/lib/firefox_profile.js#L229) method):

```
(node) warning: possible EventEmitter memory leak detected. 11 exit listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at process.addListener (events.js:252:17)
    at /.../web-ext/node_modules/firefox-profile/lib/firefox_profile.js:144:13
    at Array.forEach (native)
    at new FirefoxProfile (/.../web-ext/node_modules/firefox-profile/lib/firefox_profile.js:143:22)
    at /.../web-ext/dist/webpack:/tests/test-firefox/test.firefox.js:262:23
    at /.../web-ext/dist/webpack:/src/util/temp-dir.js:9:5342
    at run (/.../web-ext/node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:89:22)
    at /.../web-ext/node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:102:28
    at flush (/.../web-ext/node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:14:5)
    at _combinedTickCallback (node.js:370:9)
    at process._tickDomainCallback (node.js:425:11)
```

and a similar one (same stack trace but sightly different error message)

```
(node) warning: possible EventEmitter memory leak detected. 11 SIGINT listeners added. Use emitter.setMaxListeners() to increase limit.
Trace...
```

src/cmd/build.js
kumar303: Good idea. I originally thought the thrown error would get logged (like default Node behavior) but for some reason it does not! I will add logging.

src/watcher.js
kumar303: oh yeah! good idea

kumar303: It uses the file filter now

src/watcher.js
kumar303: It is not currently possible using [watchpack](https://github.com/webpack/watchpack). If we were to use [chokidar](https://github.com/paulmillr/chokidar) directly we could do it but it seems like watchpack makes some useful optimizations (however, I'm not so sure that the optimizations apply to our exact use case).

src/watcher.js
kumar303: hmm, maybe. We can wait until someone needs it. It would be easy to add.

src/cmd/build.js
kumar303: this is an obscure ES6 deconstruction feature but, yes, it is a bit alien

kumar303:
> I noticed a couple of errors produced during the test run

I think I see what's happening. This may be fixable in web-ext: https://github.com/mozilla/web-ext/issues/211

kumar303:
Thanks, this was review was really helpful, especially in making the code more readable.

Issue #138, opened 2016-03-14
Automatically reload extension in Firefox as you edit code


kumar303:
When editing web extension source code, reload the extension in Firefox so you can test it as you save your source files. There might be a couple ways to do this. 

Firstly, the platform supports a special symlink file: If you put a text file in profile/extensions whose name is the addon id and the content is the absolute path to your addon source, you can reload it. 

To reload it, you just need to disable it, flush the cache, and re-enable it (see [flushJarCache() and flushStartupCache()](https://dxr.mozilla.org/mozilla-central/source/toolkit/mozapps/extensions/internal/XPIProvider.jsm#1495)). We will [add a button for this](https://bugzilla.mozilla.org/show_bug.cgi?id=1246030) but ideally web-ext could use the debugging API to send a message to Firefox to reload when necessary.

When reloaded, the add-on will see the shutdown and startup methods called, it won't see the uninstall/install called, but you probably don't want that for reload anyway

This reload trick won't work for XPI paths, only for directory paths.

kumar303:
Mossop: The add-on will see the shutdown and startup methods called, it won't see the uninstall/install called, but you probably don't want that for reload anyway

Also, this reload trick may not work for XPI paths. I think it's only for directory paths.

kumar303:
There are some features in `about:debugging` that activate hot reloading for files like stylesheets (see [bug 1257525](https://bugzilla.mozilla.org/show_bug.cgi?id=1257525)). When running web-ext from the command line, we'll want to trigger this same behavior. We may need to do it like:
* patch `about:debugging` to make remote actor calls instead of calling AddonManager APIs directly
* mimic the same remote actor calls in web-ext

kumar303:
All major bugs here are fixed! There are a few open bugs but they are for less important edge cases.

PR #205, opened 2016-04-21
Added remote Firefox reloading support

kumar303:
This is required for https://github.com/mozilla/web-ext/issues/138

### DIFF
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -1,0 +1,104 @@
+/* @flow */
+import {createLogger} from '../util/logger';
+import {WebExtError} from '../errors';
+import defaultFirefoxConnector from 'node-firefox-connect';
+
+const log = createLogger(__filename);
+
+
+export default function connect(
+    port: number = 6000,
+    {connectToFirefox=defaultFirefoxConnector}: Object = {}): Promise {
+  return connectToFirefox(port)
+    .then((client) => {
+      log.info('Connected to the Firefox remote debugger');
+      return new RemoteFirefox(client);
+    });
+}
+
+
+export class RemoteFirefox {
+  client: Object;
+  checkForAddonReloading: Function;
+  checkedForAddonReloading: boolean;
+  addonRequest: Function;
+  getInstalledAddon: Function;
+
+  constructor(client: Object) {
+    this.client = client;
+    this.checkedForAddonReloading = false;
+  }
+
+  disconnect() {
+    this.client.disconnect();
+  }
+
+  addonRequest(addon: Object, request: string): Promise {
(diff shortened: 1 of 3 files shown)


src/firefox/remote.js
```diff
@@ -0,0 +1,102 @@
+/* @flow */
+import {createLogger} from '../util/logger';
+import {WebExtError} from '../errors';
+import defaultFirefoxConnector from 'node-firefox-connect';
```
rpl: `node-firefox-connect` seems to be missing from the dependencies currently specified in the `package.json`.

src/firefox/remote.js
```diff
@@ -0,0 +1,102 @@
...
+
+const log = createLogger(__filename);
+
+
+export default function connect(
+    port: number = 6000,
+    {connectToFirefox=defaultFirefoxConnector}: Object = {}): Promise {
+  return connectToFirefox(port)
```
rpl: it could be reasonable to `log.info` a message to show the user when the tool is "connecting" to the Firefox Remote Debugging and when it is "connected".

It is something that doesn't happen to often, and so I think that it is something that could be reasonably logged at `info` level, and it could help the user to detect, and report the issue correctly, where the tool is stuck if it doesn't work correctly.

src/firefox/remote.js
```diff
@@ -0,0 +1,102 @@
...
+          }
+        });
+    }
+  }
+
+  reloadAddon(addonId: string): Promise {
+    return this.getInstalledAddon(addonId)
+      .then((addon) => this.checkForAddonReloading(addon))
+      .then((addon) => {
+        log.debug(`Reloading ${addon.id}`);
```
rpl: I would like to look this once this feature is exposed to the user through cli options (which is going to be part of the next patch on top of this one) to be sure about how it looks like from a user point of view, but I think that "Reloading the addon" is a log message that we should probably print at `info` level.

rpl:
@kumar303 besides the above comments (of which, the only mandatory one is the missing `node-firefox-connect` npm dependency in the `package.json, the other are mostly "nice to have" or seomthing which can be easily done as follow-ups) this PR looks great, I'm looking forward to the next one which will provide the new related cli options!

src/firefox/remote.js
kumar303: whoops! I extracted this from the bigger patch so that it would be easier to review

kumar303: fixed

src/firefox/remote.js
kumar303: Sounds like a good idea, added

PR #216, opened 2016-04-25
feat: Added --timeout option for signing

### LINKED ISSUE #163: Add signing timeout option
Add a command line option to configure the [timeout](https://github.com/mozilla/sign-addon/#programmatic-use) parameter of `signAddon()`. Maybe something like `web-ext sign --timeout ...`. This might be helpful for when AMO is overloaded which can still happen even though we've made improvements.

sharang-d:
Fixes #163

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -167,6 +167,10 @@ Example: $0 --help run.
         demand: true,
         type: 'string',
       },
+      'timeout' : {
+        describe: 'Number of milleseconds to wait before giving up',
+        type: 'number',
+      },
     })
     .command('run', 'Run the web extension', commands.run, {
       'firefox-binary': {
--- a/tests/test-cmd/test.sign.js
+++ b/tests/test-cmd/test.sign.js
@@ -17,6 +17,7 @@ describe('sign', () => {
       apiKey: 'AMO JWT issuer',
       apiSecret: 'AMO JWT secret',
       apiUrlPrefix: 'http://not-the-real-amo.com/api/v3',
+      timeout: 999,
     };
 
     const buildResult = {
@@ -102,6 +103,8 @@ describe('sign', () => {
                        stubs.signingConfig.apiSecret);
           assert.equal(signedAddonCall.apiUrlPrefix,
                        stubs.signingConfig.apiUrlPrefix);
+          assert.equal(signedAddonCall.timeout,
+                       stubs.signingConfig.timeout);
           assert.equal(signedAddonCall.xpiPath,
                        stubs.buildResult.extensionPath);
           assert.equal(signedAddonCall.id,
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -12,7 +12,7 @@ const log = createLogger(__filename);
 
 export default function sign(
     {verbose, sourceDir, artifactsDir, apiKey, apiSecret,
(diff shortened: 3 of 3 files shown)


tests/test-cmd/test.sign.js
```diff
@@ -17,6 +17,7 @@ describe('sign', () => {
       apiKey: 'AMO JWT issuer',
       apiSecret: 'AMO JWT secret',
       apiUrlPrefix: 'http://not-the-real-amo.com/api/v3',
+      timeout: 999,
```
sharang-d: Should this value be generated randomly?

sharang-d:
@kumar303 I'm pretty sure this would work now though I was unable to verify it manually. Couldn't find an addon source that I could test it with. I've created an addons.mozilla.org org account and generated the API keys. Can you point me to a addon source that I could use for testing this issue and possibly others as well?

kumar303:
I think you should be able to manually test it with any of these examples: https://github.com/mdn/webextensions-examples

by the way, congrats on [landing your first Web Extensions bug](https://wiki.mozilla.org/WeeklyUpdates/2016-04-25#Friends_of_Mozilla)! It was presented to all of Mozilla today. Welcome!

src/program.js
```diff
@@ -167,6 +167,11 @@ Example: $0 --help run.
         demand: true,
         type: 'string',
       },
+      'timeout' : {
+        describe: 'A length in millesconds to give up after if the add-on ' +
+                  'hasn\'t been signed',
```
kumar303: I think this message can be more concise:

Number of milleseconds to wait before giving up

tests/test-cmd/test.sign.js
kumar303: I think it's fine as is.

src/program.js
kumar303: Is this value `undefined` when not specified on the command line? It needs to be exactly `undefined`. It can't be `false` or `null`.

kumar303: maybe you need to add `default: undefined` to the yargs config but I'm not sure.

sharang-d: > I think this message can be more concise

I copied the message from the https://github.com/mozilla/sign-addon source and added in another word. I will go ahead and make the modification if you think that is better.

sharang-d: > Is this value undefined when not specified on the command line? It needs to be exactly undefined. It can't be false or null.

Verified: Works as expected. Logged the `timeout` using `typeof`. It is `undefined` when not specified.

sharang-d: > Number of milleseconds to wait before giving up. Default: 2 minutes.

I think setting the default here increases coupling with the https://github.com/mozilla/sign-addon library. Assuming that changing the timeout there would need a change here as well to maintain consistency/uniformity.

Let me know what you think @kumar303

sharang-d:
Added a new commit with just the help text change. I'll squash it later when all looks good.

kumar303:
Looks great. I tested it locally and it worked for me. This is ready to be squashed.

sharang-d:
Done! 😄

sharang-d:
> by the way, congrats on landing your first Web Extensions bug! It was presented to all of Mozilla today. Welcome!

Thanks, never knew something like this existed. 😄

PR #219, opened 2016-04-26
Run cmd watches source, reloads addon in Firefox

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/222

With this patch you can execute `web-ext run` and it will reload your extension in Firefox as you edit the source. If you want to disable it, you can do `web-ext run --no-reload`.

I'm sure that are still corner cases that will need addressing (I've filed [some bugs](https://github.com/mozilla/web-ext/labels/auto-reloading)) but this is probably OK for a first pass. Also, this has a build and reinstall step which can be optimized away later.

Some of the architecture in the `run.js` module feels awkward to me. Let me know if you have some ideas for how to simplify it.

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -28,7 +28,7 @@ export class Program {
   }
 
   command(name: string, description: string, executor: Function,
-          commandOptions: ?Object): Program {
+          commandOptions: Object = {}): Program {
     this.yargs.command(name, description, (yargs) => {
       if (!commandOptions) {
         return;
@@ -183,6 +183,10 @@ Example: $0 --help run.
                   'a new temporary profile will be created.',
         demand: false,
         type: 'string',
+      },
+      'no-reload': {
+        describe: 'Do not reload the extension as the source changes',
+        type: 'boolean',
       },
     });
 
--- a/tests/test-cmd/test.run.js
+++ b/tests/test-cmd/test.run.js
@@ -1,17 +1,46 @@
 /* @flow */
+import path from 'path';
+import {EventEmitter} from 'events';
+import deepcopy from 'deepcopy';
 import {describe, it} from 'mocha';
 import {assert} from 'chai';
+import sinon from 'sinon';
 
-import run from '../../src/cmd/run';
+import {ExtendableError} from '../../src/util/es6-modules';
+import run, {defaultWatcherCreator, defaultReloadStrategy, ExtensionRunner}
+  from '../../src/cmd/run';
 import * as firefox from '../../src/firefox';
-import {fake, fixturePath} from '../helpers';
+import {RemoteFirefox} from '../../src/firefox/remote';
+import {makeSureItFails, fake, fixturePath} from '../helpers';
+import {createLogger} from '../../src/util/logger';
+import {basicManifest} from '../test-util/test.manifest';
+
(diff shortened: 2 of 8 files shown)


src/program.js
```diff
@@ -184,6 +184,10 @@ Example: $0 --help run.
         demand: false,
         type: 'string',
       },
+      'no-reload': {
```
rpl: @kumar303 how about defaulting `noReload` to `true` until we handle the issues with the XPI file caching ([Bug 1268884](https://bugzilla.mozilla.org/show_bug.cgi?id=1268884) and `SIGINT` support on windows (webext issue #225)?

tests/test-cmd/test.run.js
```diff
@@ -28,48 +58,279 @@ describe('run', () => {
...
+    return cmd.run({noReload: false}).then(() => {
+      assert.equal(reloadStrategy.called, true);
+      const args = reloadStrategy.firstCall.args[0];
+      assert.equal(args.sourceDir, sourceDir);
+      assert.equal(args.artifactsDir, artifactsDir);
+      assert.typeOf(args.createRunner, 'function');
+    });
+  });
+
+  it('allows you to opt out of extension reloading', () => {
```
rpl: how about adding another test case which asserts the expected default behavior

e.g.:
```it('reload the extension on source changes by default', ...)``` or ```it('does not reload the extension on source changes by default')``` (which should be probably tested in the `test.program.js` file where we can test that the options is defined in `program.js` with the expected default value, because we already test the watcher or the reload strategy elsewhere)

File: src/program.js (hunk 1 above)
Comment (kumar303) [inline src/program.js]: the option would need to have a different name because you can't set a flag to false on the command line. I think it's ok if it's slightly broken since the tool is still experimental. I won't cut a new release of web-ext until reloading works better.

File: tests/test.watcher.js [hunk 3]
```diff
@@ -93,14 +93,23 @@ describe('watcher', () => {
...
       assert.equal(conf.onChange.called, true);
 
     });
 
+    it('filters out commonly unwanted files by default', () => {
+      const conf = {
+        ...defaults, shouldWatchFile: undefined,
+        onChange: sinon.spy(() => {}),
+      };
+      proxyFileChanges({...conf, filePath: '/somewhere/.git'});
```
rpl: how about adding explicitly an xpi path to the tested unwanted file patterns?

kumar303: I don't think it's necessary to re-test all of the logic already tested by the file filter tests, such as [ignoring XPIs](https://github.com/mozilla/web-ext/blob/master/tests/test-cmd/test.build.js#L181). Ideally, I wanted to do something more like `assert.isInstance(filter, build.FileFilter)` but I couldn't think of an easy way to do it.

kumar303:
I think this is ok to merge for now. We identified the improvements that need to be made:
- when supporting more reload strategies, we'll need a more flexible abstraction
- the build step is not technically necessary so it should be removed. It adds slowness and more code complexity because of how temp dirs need to be torn down (addressed in https://github.com/mozilla/web-ext/issues/239)
- this patch, as is, will fail against Nightly Firefox for some subtle edge cases, such as changing the name in the manifest before a reload

Issue #169, opened 2016-04-04
web-ext should log errors and usage info on unknown command line interface options


rpl:
STR:
- run web-ext with an unkown command line option:
  ```$ web-ext run --unknown-option```
- **expected behavior**: web-ext log an error message complaining that `--unknown-option` does not exist, print the usage info for the command and exit with a non-zero exit code
- **actual behavior**: the unknown cli option is ignored and no error logged

kumar303:
I think [yargs.strict()](http://yargs.js.org/docs/#methods-strict) is all we need to solve this!

sharang-d:
Tried that in #226 ..
Doesn't seem to help. Will probably take a closer look at this post #226.
In the meantime if anyone wants to take it up that is fine with me.

kumar303:
The fix for this is
````diff
diff --git a/src/program.js b/src/program.js
index 55902fc..8bc7645 100644
--- a/src/program.js
+++ b/src/program.js
@@ -34,7 +34,7 @@ export class Program {
       if (!commandOptions) {
         return;
       }
-      return yargs.options(commandOptions);
+      return yargs.strict().options(commandOptions);
     });
     this.commands[name] = executor;
     return this;
````
but it would be great to also add an automated test for this, if possible.

kumar303:
I filed https://github.com/yargs/yargs/issues/489 to make yargs more testable in-process

kumar303:
It looks like we can use `.exitProcess(false)` or `.fail(customHandler)` to write tests for this. More details: https://github.com/yargs/yargs/issues/489

PR #226, opened 2016-04-29
fix: Fixed error on unknown commands

### LINKED ISSUE #215: Unknown commands cause errors
Using any unknown command on the CLI causes a Node error.
Eg:
```
➜  bin git:(master) ✗ ./web-ext unknown

module.js:341
    throw err;
    ^

Error: Cannot find module '../dist/web-ext'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.Module._load (module.js:290:25)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (/Users/sharang/Data/open-source/mozilla/web-ext/bin/web-ext:5:1)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)
```

We should probably not do that and show help text to the user instead.

sharang-d:
Fixes #215

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -24,6 +24,7 @@ export class Program {
       yargsInstance = yargs(argv);
     }
     this.yargs = yargsInstance;
+    this.yargs.strict();
     this.commands = {};
   }
 


sharang-d:
@kumar303 Please verify. Is there a bot to assign issues to someone?

kumar303:
ok, I think this is the right thing to do. Unfortunately, I can't think of a way to test this because yargs doesn't provide control over how the process exits.

The bug you initially reported isn't actually a valid bug. Here's what I see before your patch:

````
$ web-ext unknown

unknown: WebExtError: unknown command: unknown
    at /Users/kumar/dev/web-ext/dist/webpack:/src/program.js:65:17
    at lib$es6$promise$$internal$$initializePromise (/Users/kumar/dev/web-ext/node_modules/es6-promise/dist/es6-promise.js:408:9)
    at new lib$es6$promise$promise$$Promise (/Users/kumar/dev/web-ext/node_modules/es6-promise/dist/es6-promise.js:581:60)
    at Program.run (/Users/kumar/dev/web-ext/dist/webpack:/src/program.js:58:12)
    at Object.main (/Users/kumar/dev/web-ext/dist/webpack:/src/program.js:190:18)
    at Object.<anonymous> (/Users/kumar/dev/web-ext/bin/web-ext:5:28)
    ...
````
And here is what I see after your patch:
````
$ web-ext unknown
Usage: web-ext [options] command
...

Commands:
  build  Create a web extension package from source
  sign   Sign the web extension so it can be installed in Firefox
  run    Run the web extension
  ...

Unknown argument: unknown
````

So I think your patch is an improvement :)

Also, we definitely need strict argument parsing so I'm going to merge this in. The main reason is that it will report unknown global options:
````
$ web-ext --not-an-option build
Usage: web-ext [options] command
....

Commands:
  build  Create a web extension package from source
  sign   Sign the web extension so it can be installed in Firefox
  run    Run the web extension
  ...

Unknown argument: not-an-option
````
Thanks!

I confirmed that this does not fix https://github.com/mozilla/web-ext/issues/169

PR #244, opened 2016-05-06
fix typo

andymckay:


kumar303:
Oh I see, trying to change it to the British spelling! jk, good catch, thanks :)

PR #245, opened 2016-05-06
log the version if verbose

### LINKED ISSUE #161: Log web-ext version
The program should start by logging the web-ext version. This will help when checking output in bug reports.

andymckay:
* maybe fixes #161 but I'm struggling on how to test this

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -51,7 +51,9 @@ export class Program {
     return this;
   }
 
-  run({throwError=false, systemProcess=process, logStream=defaultLogStream}
+  run(absolutePackageDir: string,
+      {throwError=false, systemProcess=process, logStream=defaultLogStream,
+       getVersion=version}
       : Object = {}): Promise {
     let argv = this.yargs.argv;
     let cmd = argv._[0];
@@ -65,6 +67,7 @@ export class Program {
           throw new WebExtError(`unknown command: ${cmd}`);
         }
         if (argv.verbose) {
+          log.info('Version:', getVersion(absolutePackageDir));
           logStream.makeVerbose();
         }
         resolve(runCommand);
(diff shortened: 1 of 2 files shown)


src/program.js
```diff
@@ -195,5 +196,5 @@ Example: $0 --help run.
       },
     });
 
-  return program.run(runOptions);
+  return program.run(runOptions, absolutePackageDir);
```
kumar303: since `absolutePackageDir` is a required argument, I'd like to see it come before the optional arguments since one might call it without options and `run(undefined, absolutePackageDir)` would look weird.

kumar303:
You could test it like this:

* add a new option to `run()` for the version callable, like `{getVersion=version}`
* write a test that passes in a mock, like `program.run({getVersion: sinon.stub(() => {})})`
* assert that it gets called with `absolutePackageDir` when verbose and not called when not verbose

It's not a perfect test but it will add some decent coverage. The `version()` function is tested independently so it doesn't need more coverage. Flow will help ensure the integrity of the argument types. 

It might be a little better to test with a mock version of `log.info()` but for some reason that feels more brittle to me. Also, `version()` introduces some expensive I/O believe it or not (or maybe TravisCI's I/O is just terrible).

src/program.js
andymckay: Makes sense.

andymckay:
Thanks, I was trying to avoid adding in more arguments to the function, but if that works for you. Was surprised I/O was that expensive on Travis.

kumar303:
Thanks, this looks good. The last step is to squash and also format the final commit for the change log, as shown here: https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages

andymckay:
Done

kumar303:
tina-fey-self-high-five

Issue #276, opened 2016-06-07
Environment variables not working with non-global options


prometh07:
[Command reference](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_sign) mentions the ability of using env vars for settings such as api-secret or api-key. However, the tool doesn't parse them.

kumar303:
Thanks for filing (I thought there was already an open bug but I can't find it). I think it's confusing behavior and will likely be fixed in yargs directly -- https://github.com/yargs/yargs/issues/486 . Until then, your workaround looks good.

PR #277, opened 2016-06-07
fix: Fixed using env variables for non-global options

prometh07:
#276

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -8,6 +8,7 @@ import {WebExtError} from './errors';
 import {createLogger, consoleStream as defaultLogStream} from './util/logger';
 
 const log = createLogger(__filename);
+const envPrefix = 'WEB_EXT';
 
 
 /*
@@ -34,7 +35,9 @@ export class Program {
       if (!commandOptions) {
         return;
       }
-      return yargs.options(commandOptions);
+      // Calling env will be unnecessary after
+      // https://github.com/yargs/yargs/issues/486 is fixed
+      return yargs.env(envPrefix).options(commandOptions);
     });
     this.commands[name] = executor;
     return this;
@@ -108,7 +111,7 @@ export function main(
     .usage(`Usage: $0 [options] command
 
 Option values can also be set by declaring an environment variable prefixed
-with \$WEB_EXT_. For example: $WEB_EXT_SOURCE_DIR=/path is the same as
+with $${envPrefix}_. For example: $${envPrefix}_SOURCE_DIR=/path is the same as
 --source-dir=/path.
 
 To view specific help for any given command, add the command name.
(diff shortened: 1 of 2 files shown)


src/program.js
```diff
@@ -34,7 +34,7 @@ export class Program {
       if (!commandOptions) {
         return;
       }
-      return yargs.options(commandOptions);
+      return yargs.env('WEB_EXT').options(commandOptions);
```
kumar303: I think the `WEB_EXT` string should be referenced from a constant so it can be shared with https://github.com/mozilla/web-ext/blob/master/src/program.js#L119 I typically don't like using constants for strings like this but it would be confusing if someone changed it in one place and not the other. Using a shared constant will prevent that from happening.

kumar303: Also, I think you should add a comment saying that this might be fixed in the future by https://github.com/yargs/yargs/issues/486

kumar303:
I know this is only a configuration change but since we have a couple layers of abstraction for commands, it would be great to add a test for this. Here's an idea for one:
````diff
diff --git a/tests/test.program.js b/tests/test.program.js
index a93153e..9b27c8f 100644
--- a/tests/test.program.js
+++ b/tests/test.program.js
@@ -66,6 +66,38 @@ describe('program.Program', () => {
       });
   });
 
+  it('reads option values from env vars in sub commands', () => {
+    // Set an env var that mimics web-ext cmd --some-opt=value
+    process.env.WEB_EXT_SOME_OPT = 'value';
+    let valueReceived;
+    const program = new Program(['cmd'])
+      .command('cmd', 'some command', ({someOpt}) => {
+        valueReceived = someOpt;
+      }, {
+        'some-opt': {
+          describe: 'example option',
+        },
+      });
+    return run(program, {throwError: false})
+      .then(() => {
+        assert.equal(valueReceived, 'value');
+        delete process.env.WEB_EXT_SOME_OPT;
+      });
+  });
+
+  it('exits 1 on a thrown error', () => {
+    let fakeProcess = fake(process);
+    let program = new Program(['cmd'])
+      .command('cmd', 'some command', () => {
+        throw new Error('this is an error from a command handler');
+      });
+    return run(program, {systemProcess: fakeProcess, throwError: false})
+      .then(() => {
+        assert.equal(fakeProcess.exit.called, true);
+        assert.equal(fakeProcess.exit.firstCall.args[0], 1);
+      });
+  });
+
   it('handles errors that have codes', () => {
 
     class ErrorWithCode extends Error {
````

**Comment by kumar303** [comment]
I think your patch will fix https://github.com/mozilla/web-ext/issues/203 as well. Can you check? If so, you'll get a two-fer.

**Comment by prometh07** [comment]
#203 works too.

File: src/program.js [hunk 2]
```diff
@@ -108,15 +111,15 @@ export function main(
     .usage(`Usage: $0 [options] command
 
 Option values can also be set by declaring an environment variable prefixed
-with \$WEB_EXT_. For example: $WEB_EXT_SOURCE_DIR=/path is the same as
+with $${envPrefix}_. For example: $${envPrefix}_SOURCE_DIR=/path is the same as
```
kumar303: good call!

kumar303:
looks good, thanks for the patch

PR #290, opened 2016-06-14
feat: Started ignoring node_modules folders by default

### LINKED ISSUE #269: Exclude node_modules from the built .xpi file
STR:
* Create simple web extension
* Run `npm install eslint` (or install some other stuff)
* Run `web-ext build`

Actual results: The resulting .xpi is huge (couple MB) and node_modules is the biggest contributor to that size.

Expected results: The .xpi is tiny and doesn't contain any unneeded files.

In issue #85, web-ext started to ignore some files but node_modules was not among them. I think it can be safely excluded from the .xpi since AFAIK Firefox won't be able to find the modules from there anyway.

If #131 and #176 get fixed soonish, I think we could leverage that to specify excludes. Otherwise, I'd add node_modules to the hardcoded list of excluded dirs while that feature is being implemented.

freaktechnik:
This fixes #269

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -100,6 +100,7 @@ export class FileFilter {
       '**/*.xpi',
       '**/*.zip',
       '**/.*', // any hidden file
+      '**/node_modules',
     ];
   }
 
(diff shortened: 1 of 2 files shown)


kumar303:
Looks great, thanks!

Issue #286, opened 2016-06-10
Support Firefox 48


kumar303:
The easiest way to land https://github.com/mozilla/web-ext/pull/282 was to require Firefox 49. We should try to add support back for Firefox 48, maybe with reloading disabled if that's easier. Firefox 48 was the first stable web extension release so it would be nice if web-ext supported it.

kumar303:
We need to document that to run an extension in Firefox 48, the command is `web-ext run --pre-install` and this will not support automatic add-on reloading.

kumar303:
documented: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Testing_in_Firefox_48

PR #309, opened 2016-06-21
fix: Re-enabled support for running extensions in Firefox 48

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/286

With this patch, here is an example of running in Firefox 48:

````
$ web-ext run
Running web extension from /Users/kumar/dev/bookmarker-ext/moz-webext
Running Firefox with profile at /var/folders/h7/rttdqkks7fl_txp9_rr4j4840000gn/T/67d00895-ed01-41fe-9a04-05b775c1d97c
Connected to the Firefox remote debugger

run: WebExtError: Temporary add-on installation is not supported in this version of Firefox (you need Firefox 49 or higher). For older Firefox versions, use --pre-install
    at /Users/kumar/dev/web-ext/dist/webpack:/src/cmd/run.js:160:13
    at /Users/kumar/dev/web-ext/dist/webpack:/src/errors.js:52:14
    at process._tickCallback (node.js:368:9)
````

At which point, you are prompted to run it like:

````
web-ext run --pre-install
````

This works :) It implies `--no-reload` since only temporarily installed add-ons can be reloaded.

I haven't added any optimizations (such as ignoring the remote debugger) but I think this patch is good enough to quickly support Firefox 48 without auto-reloading.

We should also start hiding the stacktrace for these usage exceptions (see https://github.com/mozilla/web-ext/issues/310).

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -50,7 +50,7 @@ function defaultPackageCreator(
 
 
 export default function build(
-    {sourceDir, artifactsDir, asNeeded}: Object,
+    {sourceDir, artifactsDir, asNeeded=false}: Object,
     {manifestData, fileFilter=new FileFilter(),
      onSourceChange=defaultSourceWatcher,
      packageCreator=defaultPackageCreator}
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -1,7 +1,10 @@
 /* @flow */
 import * as defaultFirefox from '../firefox';
 import defaultFirefoxConnector from '../firefox/remote';
-import {onlyErrorsWithCode} from '../errors';
+import {
+  onlyInstancesOf, onlyErrorsWithCode, RemoteTempInstallNotSupported,
+  WebExtError,
+} from '../errors';
 import {createLogger} from '../util/logger';
 import getValidatedManifest from '../util/manifest';
 import defaultSourceWatcher from '../watcher';
@@ -83,12 +86,18 @@ export function defaultFirefoxClient(
 
 
 export default function run(
-    {sourceDir, artifactsDir, firefoxBinary, firefoxProfile, noReload}: Object,
+    {sourceDir, artifactsDir, firefoxBinary, firefoxProfile,
+     preInstall=false, noReload=false}: Object,
     {firefoxClient=defaultFirefoxClient, firefox=defaultFirefox,
      reloadStrategy=defaultReloadStrategy}
     : Object = {}): Promise {
 
   log.info(`Running web extension from ${sourceDir}`);
+  if (preInstall) {
+    log.info('Disabled auto-reloading because it\'s not possible with ' +
+             '--pre-install');
+    noReload = true;
+  }
 
   function createRunner() {
     return getValidatedManifest(sourceDir)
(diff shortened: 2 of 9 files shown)


rpl:
@kumar303 This PR looks great to me, it makes web-ext reasonably good on Firefox 48 (not as awesome as it is on Firefox 49, but cool enough :-))

My only doubt is that instead of building the XPI, we could use the "Shadow Install"-mode (which is installing an addon by creating a text file which contains the full path to the add-on source dir), I took a brief look and it seems easy enough to achieve with minimal changes:

- Here is a link to a patch which add a "Shadow Install" version on top of this change: https://github.com/kumar303/web-ext/compare/firefox48...rpl:PR-309

I think that this strategy can probably provide a better "web-ext run" experience on Firefox 48, mainly because:

- there is no need to rebuild the xpi, and it reduces the complexity of this feature and remove any problem related to the xpi caching (which is an issue that is still present of Firefox < 49)

- most of the changes (e.g. changes in the popup) will be active without making anything special, and it will feel like the full live reload that is supported on Firefox >= 49

- the reload button will work as expected (because the source dir is always updated, even without watching for file changes and rebuilding the xpi) and it doesn't suffer of the xpi caching issues on Firefox < 49

rpl:
@kumar303 after digging a bit into MDN I found a page where this install method is explained (and it is actually called "Firefox extension proxy file"):

https://developer.mozilla.org/en-US/Add-ons/Setting_up_extension_development_environment#Firefox_extension_proxy_file

kumar303:
Ah! Great idea about doing the shadow install. I honestly forgot about it. I will switch it over to this approach.

I agree with all bullet points except there is a catch to the reload button -- as of a recent patch, it now requires the add-on's location object to be a TemporaryInstallLocation, which I think is only created by installTemporaryAddon().

Nice work digging up that MDN page!

rpl:
@kumar303 This new version looks great, but I noticed the following two issues (the first one happens consistently and I think it is pretty important, the second one is intermittent and it is not so critical):

- if "--pre-install" cli option is specified then we should probably implicitly activate "--no-reload" as well, or the following error will be raised in the console on every file change if we are running the add-on on Firefox >= 49 (because it supports the add-on reload only if it has been "temporarely installed"):

> WebExtError: reload response error: unknownError: error occurred while processing 'reload: Error: Only temporary add-ons can be reloaded

- by running web-ext on Firefox < 49 without the "--pre-install" cli option, sometimes it fails to remove the temporary profile and raise the following exception (it looks like a timing issue, and the issue probably happens if the Firefox instance is not yet closed when we remove the profile dir, and as additional context I'm testing this PR on Linux):

> % web-ext run --firefox-binary /usr/bin/firefox
> ...
> run: WebExtError: Temporary add-on installation is not supported in this version of Firefox (you need Firefox 49 or higher). For older Firefox versions, use --pre-install
    at /zfs-mordor/work/webext/tools/web-ext/dist/webpack:/src/cmd/run.js:160:13
    at /zfs-mordor/work/webext/tools/web-ext/dist/webpack:/src/errors.js:52:14
    at process._tickCallback (node.js:406:9)
>
> [firefox-profile] cannot delete profileDir on exit /tmp/eb6d9687-f613-433f-b912-2b7aa32203a3 { [Error: ENOTEMPTY: directory not empty, rmdir '/tmp/eb6d9687-f613-433f-b912-2b7aa32203a3']
  errno: -39,
  code: 'ENOTEMPTY',
  syscall: 'rmdir',
  path: '/tmp/eb6d9687-f613-433f-b912-2b7aa32203a3' }

kumar303:
Hmm, it looks like the 'cannot delete profileDir on exit' warning is buried in the firefox-profile module. I think that one would be hard to fix.

kumar303:
Good call on the implicit `--no-reload`. At first I wanted the developer to make an explicit decision but it doesn't make sense because reloading is simply not possible ;) I'll fix it.

PR #311, opened 2016-06-22
Show usage on missing or wrong command.

johannhof:
It's pretty annoying that you only get a stack trace when simply running `web-ext` without any arguments for the first time, so I added help messages.

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -120,7 +120,9 @@ Example: $0 --help run.
     .help('help')
     .alias('h', 'help')
     .env(envPrefix)
-    .version(() => version(absolutePackageDir));
+    .version(() => version(absolutePackageDir))
+    .demand(1)
+    .strict();
 
   program.setGlobalOptions({
     'source-dir': {


kumar303:
I think it's better to do this with `yargs.strict()`. Could you try it? It was started here but abandoned: https://github.com/mozilla/web-ext/issues/169 Also, it needs to work for incorrect sub-command arguments. Example: `web-ext build --not-an-option` should show an error.

kumar303:
Were you trying on web-ext 1.0.1 or one built against master? When I try against master it's pretty good but it needs this patch to handle sub-command options:
````diff
index 7b21b16..2f03ae4 100644
--- a/src/program.js
+++ b/src/program.js
@@ -37,7 +37,7 @@ export class Program {
       }
       // Calling env will be unnecessary after
       // https://github.com/yargs/yargs/issues/486 is fixed
-      return yargs.env(envPrefix).options(commandOptions);
+      return yargs.strict().env(envPrefix).options(commandOptions);
     });
     this.commands[name] = executor;
     return this;

````

kumar303:
^ that causes the test process to exit though because yargs exits the process. My idea for dealing with that was: https://github.com/mozilla/web-ext/issues/169#issuecomment-216341919

johannhof:
@kumar303 pushed another approach which works for me, anything wrong with that one? Tests seem to be passing.

johannhof:
The thing that looks ugly on master for me is simply typing `web-ext` without any sub-commands, but now it should look good in all cases.

kumar303:
Thanks! Looks good and the patch works for me. It doesn't address https://github.com/mozilla/web-ext/issues/169 but we can do that in a separate patch (it's a related but separate issue).

PR #314, opened 2016-06-23
fix: Missing command/option errors are reported better

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/169

I also noticed a test was indented wrong -- we didn't have an eslint indent rule. This adds an indent rule (sorry for the noise).

### DIFF
--- a/tests/test.program.js
+++ b/tests/test.program.js
@@ -2,10 +2,11 @@
 import {describe, it} from 'mocha';
 import path from 'path';
 import fs from 'mz/fs';
+import sinon from 'sinon';
 import {assert} from 'chai';
 import {spy} from 'sinon';
 
-import {version, main, Program} from '../src/program';
+import {defaultVersionGetter, main, Program} from '../src/program';
 import commands from '../src/cmd';
 import {onlyInstancesOf, WebExtError} from '../src/errors';
 import {fake, makeSureItFails} from './helpers';
@@ -20,7 +21,7 @@ describe('program.Program', () => {
     return program.run(
       absolutePackageDir, {
         systemProcess: fakeProcess,
-        throwError: true,
+        shouldExitProgram: false,
         ...options,
       });
   }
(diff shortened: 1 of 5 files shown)


kumar303:
ooh, since this patch added `yargs.exitProcess(false)` I was able to write a test to cover yargs' magic `--version` handling. This should bring line coverage up to 100% ✨

tests/test.program.js
```diff
@@ -202,17 +249,38 @@ describe('program.main', () => {
...
 
+  it('can get the program version', () => {
+    const fakeVersionGetter = sinon.spy(() => '<version>');
+    const fakeCommands = fake(commands, {
+      build: () => Promise.resolve(),
+    });
+    const projectRoot = '/pretend/project/root';
+    // For some reason, executing --version like this
+    // requires a command. In the real CLI, it does not.
+    return run(['--version', 'build'],
```
rpl: @kumar303 I looked a bit deeper into this one and I think that the different behavior is related to the `exitProcess` boolean:

when `web-ext` is executed from the command line `exitProcess` is true and `yargs` makes `web-ext` exit just before the commands are checked:

```
...
    return new Promise(
      (resolve) => {
        this.yargs.exitProcess(this._exitProcess);  // <<-- on `web-ext --verbose` yargs exit here
        argv = this.yargs.argv;
        cmd = argv._[0];
        if (cmd === undefined) {
          throw new WebExtError('No sub-command was specified in the args');
        }
...
```

rpl:
@kumar303 This PR looks good to me
 
I'm thinking that maybe `exitProcess` could be renamed into something that doesn't sounds/looks like `systemProcess` (which has a totally different meanings / roles / values), but I consider it a nit (or a matter of taste) and not a critical issue.

tests/test.program.js
kumar303: Hmm, I actually cannot reproduce this with an isolated yargs program. It's probably not worth spending too much time on it. Here's what I tried to mimic what web-ext does:
````javascript
var yargs = require('yargs');
yargs
  .help('help')
  .demand(1)
  .strict()
  .exitProcess(false)
  .version(function() {
    return "the version!"
  })
  .command('thing', '', function(yargs) {
  }, function() {
    console.log('running thing');
  })
  .argv;
````

Issue #322, opened 2016-06-27
Increase timeout before giving up on remote debugger connection


kumar303:
Increase the timeout before `web-ext run` gives up on establishing a connection to the remote debugger. One theory for https://github.com/mozilla/web-ext/issues/317 is that the connection is timing out before the Firefox debugger server can legitimately start up.

PR #324, opened 2016-06-27
fix: `web-ext run`: increased debugger connection timeout

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/322


rpl:
@kumar303 r+

question: do we currently print the current timeout time to the user? I'm thinking that with a longer default timeout, it could be reasonable to give to the user more informations about how much time is could be waiting for even when web-ext is not running in verbose mode.

kumar303:
I see what you mean. I added some logging so at least the user knows web-ext is trying to establish a connection.

rpl:
@kumar303 :+1: 

r+

Issue #177, opened 2016-04-07
Run / install web extensions without an ID


kumar303:
With optional IDs (https://github.com/mozilla/addons/issues/61), web-ext needs to auto-generate an ID to install and run the extension (the `run` command). This will be similar to how it already installs extensions; it currently reads the ID from the manifest first, creates a special file in the profile directory, and launches Firefox.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1262005#c10 for the required Firefox changes.

Related to https://github.com/mozilla/web-ext/issues/132

Issue #132, opened 2016-03-10
Build web extensions without an ID


andymckay:
In support of https://wiki.mozilla.org/WebExtensions/Applications, this will let you build an extension without an ID.

andymckay:
Tracking: mozilla/addons#61

andymckay:
We slightly changed the specification to not worry about mozilla.json. Instead we'll just cope with it being in one of two fields in manifest.json, see the spec for more.

kumar303:
For this first pass, only the build command can allow extensions without an ID. 

TODO:
* run extensions without an ID (https://github.com/mozilla/web-ext/issues/177)
* sign extensions without and ID (https://github.com/mozilla/web-ext/issues/178)

kumar303:
I don't think this needs any explicit documentation

Issue #178, opened 2016-04-07
Do not require an ID when signing an extension


kumar303:
Once https://github.com/mozilla/addons-server/issues/1890 lands on AMO, web-ext can be adjusted not to post an ID to the signing service.

Supports https://github.com/mozilla/addons/issues/61

* [x] Update `sign-addon` for no-ID support
* [x] Update `web-ext` for no-ID support
* [x] Release `sign-addon` and bump the `web-ext` dependency

kumar303:
The new API endpoint is documented here http://addons-server.readthedocs.io/en/latest/topics/api/signing.html

kumar303:
just hit an API error https://sentry.prod.mozaws.net/operations/olympia-dev/issues/339797/

kumar303:
> just hit an API error

This was bogus. I had the API endpoint wrong.

andymckay:
We shouldn't be generating a 500 from user data though, so there might be a fix we need there to return a 400 or something.

kumar303:
I'm pretty sure the 500 was from dirty dev data (multiple objects returned). I don't really think it's worth chasing down TBH. It doesn't happen when I use the right endpoint :)

kumar303:
This will introduce a new option, `web-ext sign --id ...`, that needs documenting

kumar303:
documented: 
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#--id
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Signing_extensions_without_an_explicit_ID

PR #326, opened 2016-06-27
handle extensions without IDs in build/run

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/132, fixes https://github.com/mozilla/web-ext/issues/177

Changes:
* `build`, `run`, and auto-reloading work for manifests without an ID
* `run --pre-install` will error if the manifest doesn't have an ID
* `sign` will error if the manifest doesn't have an ID but this is just temporary until https://github.com/mozilla/web-ext/issues/178 is fixed

### DIFF
--- a/src/util/manifest.js
+++ b/src/util/manifest.js
@@ -33,18 +33,11 @@ export default function getValidatedManifest(sourceDir: string): Promise {
         errors.push('missing "version" property');
       }
 
-      // Make sure the manifest defines a gecko id.
-      let idProps = ['applications', 'gecko', 'id'];
-      var propPath = '';
-      var objectToCheck = manifestData;
-
-      for (let nextProp of idProps) {
-        propPath = propPath ? `${propPath}.${nextProp}` : nextProp;
-        objectToCheck = objectToCheck[nextProp];
-        if (!objectToCheck) {
-          errors.push(`missing "${propPath}" property`);
-          break;
-        }
+      if (manifestData.applications && !manifestData.applications.gecko) {
+        // Since the applications property only applies to gecko, make
+        // sure 'gecko' exists when 'applications' is defined. This should
+        // make introspection of gecko properties easier.
+        errors.push('missing "applications.gecko" property');
       }
 
       if (errors.length) {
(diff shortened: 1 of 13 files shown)


src/util/manifest.js
```diff
@@ -33,17 +33,20 @@ export default function getValidatedManifest(sourceDir: string): Promise {
...
-      var propPath = '';
-      var objectToCheck = manifestData;
-
-      for (let nextProp of idProps) {
-        propPath = propPath ? `${propPath}.${nextProp}` : nextProp;
-        objectToCheck = objectToCheck[nextProp];
-        if (!objectToCheck) {
-          errors.push(`missing "${propPath}" property`);
-          break;
+      if (manifestData.applications) {
```
rpl: @kumar303 I'm assuming that it is possible that we want to omit the `id` and, at the same time, we want the `applications` section to enforce the `strict_min_version`, `strict_max_version` (or probably, even if it is rarely used, the `update_url`).

If the above assumption is true, we should change this validation accordingly.

kumar303: oh, huh, I didn't think about that. D'oh.

kumar303: thanks, I'll add a test for that and fix it

tests/test-util/test.manifest.js
```diff
@@ -35,6 +38,15 @@ describe('util/manifest', () => {
         })
     ));
 
+    it('allows manifests without IDs', () => withTempDir(
```
rpl: @kumar this is similar to the test cases that starts from line 136, how about move it near there? 

by reading the test description, it seems the same test from line 136, but it is actually different, in fact it tests that a manifest without the entire "applications" section is still valid, and so it could be changed to have a more explicit description (e.g. "allows manifests without the 'applications' property", and maybe renaming `manifestWithoutId` manifestWithoutApplications` as well?)

kumar303: It's different in that one tests a manifest without the `applications` property at all and the other tests for an `applications.gecko` property but with an undefined `applications.gecko.id` property. I updated one of the test titles to make this more clear.

rpl:
@kumar303 looks great! 

r+

Issue #119, opened 2016-03-08
Package also for chrome


Mte90:
Is possible add a parameter for build that package the extension for chrome in a zip file?
Basically the web-ext works also for chrome without the `applications` section in the manifest.json.
Actually to do that I remove that part in the manifest.json and create a zip to upload on the Chrome Web Store with the same files of the firefox version.

kumar303:
Yes, I'd like to add this feature. It would be great if the build step automatically converted the manifest for each platform.

bwinton:
I believe we're looking into [not adding the extra keys](https://wiki.mozilla.org/WebExtensions/Applications) into manifest.json, and instead having a “mozilla.json” file (or something similar), which should make this easier to implement.  :wink:

kumar303:
Chrome package format: https://developer.chrome.com/extensions/crx

Mte90:
Actually to package automatically for firefox and chrome I've written this https://github.com/Mte90/Script/blob/master/pack-ext.py

kumar303:
Does the extension load in Chrome? I don't see it writing the "magic number" to the beginning of the zip file which Chrome's docs say is mandatory: "Chrome requires this constant at the beginning of every .crx package."

Mte90:
the extension create the zip for the chrome web store and works without problems.
The web store require a zip package, the .crx is only outside the web store.

kumar303:
OK, I see.

tofumatt:
I think the answer to this is we should output `.zip` files by default, not `.xpi` files. [AMO now accepts `.zip`](https://github.com/mozilla/addons-server/issues/2727) files (converts them to XPI for Firefox) but for development Firefox will accept `.zip` temporary add-ons.

kumar303:
Oh yeah, good idea. https://github.com/mozilla/web-ext/issues/325

kumar303:
This has been fixed by https://github.com/mozilla/web-ext/pull/328

We're not explicitly supporting the `.crx` format. Instead we build to `.zip` which is supported by both Firefox and Chrome :)

kumar303:
@tofumatt kindly updated all the docs so I'm removing the label:
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference

Croydon:
@kumar303 I'm wondering, what is about the manifest?

[...]

PR #328, opened 2016-06-29
Output ZIP files instead of XPI files (fix #325)

### LINKED ISSUE #325: Build to zip files by default
Since [AMO supports zip files](https://github.com/mozilla/addons-server/issues/2727) as well as Firefox (as temporary installs) and Chrome, we might as well build to zip as the default rather than `.xpi`.

tofumatt:
Fixes #325

r?

I tested signing and it still a. works with the zip (as expected) and b. sends down an XPI, which is expected behaviour.

Tested `web-ext run` and everything still worked as normal too.


tofumatt:
I think this should also close #119, as it will package a ZIP file Firefox will load and (presumably) Chrome will load/accept.

tofumatt:
Just uploaded the ZIP file `web-ext build` output and the Chrome Store accepted the upload and it worked, so I'd say #119 is closed by this, yup.

screenshot 2016-06-29 13 29 38

kumar303:
Cool! I'm taking a look now. 

One thing we've been doing is making each commit message [auto-changelog-ready](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages) so could you rename the message to "feat: Output ZIP files instead of XPI files" ? 

Also, the commit message itself should not mention the github issue (only the pull request should) otherwise every time you rebase and push, it will spam the issue thread (which gets messy).

tofumatt:
Ten four, I shall fix these things.

tofumatt:
Fixed both things!

src/cmd/build.js
```diff
@@ -31,8 +31,9 @@ function defaultPackageCreator(
           filter: (...args) => fileFilter.wantFile(...args),
         })
         .then((buffer) => {
+          let extension = 'zip';
           let packageName = safeFileName(
-            `${manifestData.name}-${manifestData.version}.xpi`);
+            `${manifestData.name}-${manifestData.version}.${extension}`);
```
kumar303: might as well just make this `${manifestData.name}-${manifestData.version}.zip` -- no need for a variable

tofumatt: Oh yeah, you’re right. I was thinking of making the output extension an
option but then realised it’s unneeded. Shall remove.

tests/test-cmd/test.build.js
```diff
@@ -209,7 +209,7 @@ describe('build', () => {
       assert.equal(filter.wantFile('some.xpi'), true);
       assert.equal(filter.wantFile('manifest.json'), false);
     });
-    
+
```
kumar303: heh. I fixed the eslint rule for this in another PR that hasn't landed yet. I was like wut, how did that even get committed? 🎯

tofumatt: And my editor just deletes trailing whitespace so I didn't even notice I committed it 😄

kumar303:
> the Chrome Store accepted the upload

highfive

kumar303:
At first I was wondering if we should make an option to output to `.xpi` if desired but I can't really think of any reason to :)

Looks good!

tests/test-cmd/test.build.js
kumar303: no, your change is correct. Thanks.

kumar303:
oh, heh, I was waiting for fixing up the [extraneous variable](https://github.com/mozilla/web-ext/pull/328/files#r68977634). All good though, no need for an additional patch. Thanks for getting this in.

tofumatt:
Shoot! I made the commit and didn't push it. I am dumb.

I'll make another patch, I hate silly extra lines.

kumar303:
ok, if you do, prefix the commit message with `chore:` so that it gets excluded from the auto-changelog. It's not exactly a chore but hey.

tofumatt:
Noted! Thanks :-)

Issue #330, opened 2016-06-29
Require npm greater than 3.0.0 in package.json


kumar303:
We should specify that npm greater than 3.0.0 is required directly in package.json. With npm 2.x, installations seem to result in errors like this:

````
$ web-ext --version
Error: Cannot find module 'any-promise'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:227:19)
    at __webpack_require__ (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:22:30)
    at Object.<anonymous> (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:163:16)
    at __webpack_require__ (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:22:30)
    at Object.<anonymous> (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:57:12)
    at __webpack_require__ (/usr/local/lib/node_modules/web-ext/node_modules/sign-addon/dist/sign-addon.js:22:30)
````

kumar303:
tofumatt from IRC: "That should do what you want, yup"

PR #331, opened 2016-06-29
fix: Installation will now fail if your npm is not up to date

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/330


tofumatt:
r+

Issue #332, opened 2016-06-29
Relative source path causes NS_ERROR_FILE_UNRECOGNIZED_PATH error


kumar303:
If you run using a relative path to your extension source you'll get an error like this:

````
$ web-ext run --source-dir beastify/
run: WebExtError: installTemporaryAddon: Error: unknownError: Could not install add-on at 'beastify/': [Exception... "Component returned failure code: 0x80520001 (NS_ERROR_FILE_UNRECOGNIZED_PATH) [nsILocalFile.initWithPath]"  nsresult: "0x80520001 (NS_ERROR_FILE_UNRECOGNIZED_PATH)"  location: "JS frame :: resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/actors/addons.js :: AddonsActor<.installTemporaryAddon< :: line 23"  data: no]
    at /Users/kumar/dev/web-ext/dist/webpack:/src/firefox/remote.js:85:29
    at EventEmitter.extend.handleMessage (/Users/kumar/dev/web-ext/node_modules/firefox-client/lib/client.js:161:7)
    at EventEmitter.extend.readMessage (/Users/kumar/dev/web-ext/node_modules/firefox-client/lib/client.js:220:10)
    at EventEmitter.extend.onData (/Users/kumar/dev/web-ext/node_modules/firefox-client/lib/client.js:186:16)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Socket.Readable.push (_stream_readable.js:110:10)
    at TCP.onread (net.js:523:20)
````

tanettrimas:
Btw, I still experience this error when using webext plugin on webpack.

To solve this I have to do the following inside webpack.config.js:

`plugins: [
    new WebExtWebpackPlugin({ 
			sourceDir: path.resolve(__dirname, 'dist'), 
			startUrl: 'https://www.finn.no/realestate/homes/search.html'
		}),
]`

If you type `sourceDir: './dist'` you will get the error.

PR #333, opened 2016-06-29
fix: Relative `--source-dir` paths are handled correctly now

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/332


rpl:
@kumar303 r+

Nit: we could probably add one more test to assert that path.normalize is applied to both the source dir and the artifact dir paths.

kumar303:
Sure. Added.

PR #340, opened 2016-06-30
feat: `web-ext sign` now supports extensions without explicit IDs

kumar303:
Supports https://github.com/mozilla/web-ext/issues/178

This feature also requires the patch from https://github.com/mozilla/sign-addon/pull/114

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -1,19 +1,21 @@
 /* @flow */
+import path from 'path';
+import fs from 'mz/fs';
 import {signAddon as defaultAddonSigner} from '../util/es6-modules';
 
 import defaultBuilder from './build';
-import {InvalidManifest} from '../errors';
 import {withTempDir} from '../util/temp-dir';
-import getValidatedManifest from '../util/manifest';
+import {onlyErrorsWithCode, WebExtError} from '../errors';
+import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
 
 const log = createLogger(__filename);
-
+export const extensionIdFile = '.web-extension-id';
 
 export default function sign(
     {verbose, sourceDir, artifactsDir, apiKey, apiSecret,
-     apiUrlPrefix, timeout}: Object,
+     apiUrlPrefix, id, timeout}: Object,
     {build=defaultBuilder, signAddon=defaultAddonSigner,
      preValidatedManifest=null}: Object = {}): Promise {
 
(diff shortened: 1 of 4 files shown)


andymckay:
If I sign the add-on a second (or third or whatever) time, how do I specify the add-on id? Do I have to change the manifest, or could I specify an id on the command line as an argument?

kumar303:
Hmm, good point. As it stands now, each time you run the sign command it will create a new add-on which will start to make your DevHub add-on list pretty unbearable.

kumar303:
Good catch @andymckay ✋ fixed

src/cmd/sign.js
```diff
@@ -28,37 +27,42 @@ export default function sign(
...
-          verbose,
-          xpiPath: buildResult.extensionPath,
-          id: manifestData.applications.gecko.id,
-          version: manifestData.version,
-          downloadDir: artifactsDir,
-        }))
+        .then(({buildResult, manifestData}) => {
+          const manifestId = getManifestId(manifestData);
+          if (id && manifestId) {
+            log.warn(`Overriding manifest.json ID: ${id}`);
```
rpl: @kumar303 How about `Overriding manifest.json ID "${id}" with "${manifestId}"`?

rpl:
@kumar303 how about the following tweak to the strategy used for signing the updates of an add-on without an extension id in the manifest?

- warn the user when he submit a webextension without an extension id for the first time
- save the addon id from the first successful signed xpi in an hidden file in the root of the add-on source dir
- then we can use the hidden file to detect when the addon has been signed once, and even reuse it to automatically sign the next add-on XPIs with the same id, without the need for the user to remember to pass it to the sign command on every released update (and if the user re-creates its development environment, then it will be warned on the first web-ext sign, because the hidden file is missing in the newly created development environment)

kumar303:
Per our IRC conversation, I checked what AMO does in the case of a manifest.json and custom declared ID. It chooses the manifest.json ID so I made `web-ext` raise an error for that case (to avoid surprises!).

kumar303:
BTW, this is testable on https://addons-dev.allizom.org like `web-ext sign --api-key ... --api-secret ... --api-url-prefix https://addons-dev.allizom.org/api/v3`

kumar303:
@rpl really great idea! I added this feature. What it does is save a file to the artifacts directory. It will re-use the file if it exists but only as a fallback. Let me know how it looks.

src/cmd/sign.js
```diff
@@ -28,39 +30,105 @@ export default function sign(
...
-          id: manifestData.applications.gecko.id,
-          version: manifestData.version,
-          downloadDir: artifactsDir,
-        }))
         .then((signingResult) => {
           // All information about the downloaded files would have
           // already been logged by signAddon().
-          log.info(signingResult.success ? 'SUCCESS' : 'FAIL');
+          if (signingResult.success) {
+            log.info(`Extension ID: ${signingResult.id}`);
```
rpl: @kumar303 at line 74, it seems to me that in some cases the `id` property can be missing from the `signingResult` (my guess is that it could happen when the extension id is assigned statically in the `manifest.json`), and in that case the above line is going to print a "Extension ID: undefined" log, which can be confusing for the the user.

If the above assumptions are true, we can probably tweak it a bit, to ensure that it is going to always print a log which contains the `id` (even if it is the same statically assigned in the manifest.json)

rpl:
@kumar303 besides the above comment (and the doubt below), this PR looks great now!

My only remaining doubt is potential side-effects of saving the auto-generated ID in the "artifact dir vs the project root":

my feeling is that is some cases, the addon developer is going to have its own building steps (e.g. for babel, webpack or browserify etc.) and it is pretty common that one of those building steps could remove the entire artifact dir (e.g. at every full rebuild), and the saved auto-generated ID will be lost (and a new one will be generated on the next `web-ext sign` run)

Given that the auto-generated ID is not a sensible information (and any user can get the addon id on an existent add-on pretty easily), I will probably prefer if it could be saved as an hidden file in the root of the project, because it ensures that it will not be removed by a build automation script (and by making it an hidden file, `web-ext build` will not include it in the generated XPI).

It is probably even reasonable to assume that the addon developer wants to add and commit the hidden file in the repository of the add-on sources, so that any of the developer that should be able to re-sign the addon will have access to the currently assigned id (but it can be overriden from the command line, if they need to test it as a separate add-on)

src/cmd/sign.js
kumar303: Did you install the latest `sign-addon` from this branch into web-ext? https://github.com/mozilla/sign-addon/pull/114 What you're describing is the behavior before that patch is applied but not after.

kumar303:
Ah, true, good point. I will make it a hidden file in the source dir. I was a little unsure about this when working on it. I think the artifacts dir is more ephemeral.

rpl:
@kumar303 this PR looks great

r+

Issue #307, opened 2016-06-21
Add a functional test suite


kumar303:
There's no avoiding it, we need a functional test suite. There is just no way to cover all end user functionality with our unit test suite. The latest regression had to do with [missing yargs configuration](https://github.com/mozilla/web-ext/issues/304). There have been a few others like this.

Goals:
- test *only* things that cannot be covered in unit tests
- make them as fast as possible
- they cannot be flaky

kumar303:
another regression example: https://github.com/mozilla/web-ext/pull/380 (a call that failed without mock injection)

kumar303:
added in https://github.com/mozilla/web-ext/pull/469 !

PR #371, opened 2016-07-15
Fix `web-ext run`

yan12125:
Something not covered by unt tests :D

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -1,7 +1,7 @@
 /* @flow */
 import nodeFs from 'fs';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


kumar303:
Ugh, thanks. Sounds like https://github.com/mozilla/web-ext/issues/307 is getting more critical. Also, Flow usually catches import errors like this. Grr.

kumar303:
This will be on npm as soon as the tag is done building https://travis-ci.org/mozilla/web-ext/builds/145106876

yan12125:
Thanks! Just received version 1.2.4 and it's working fine.

Issue #204, opened 2016-04-20
Hook up the linter


andymckay:
I made some stupid mistakes with permissions the other day, writing an extension. Hooking up web-ext lint to point to the amo-linter might future proof my stupidity. Unlikely though.

kumar303:
Documented:
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Checking_for_code_lint
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_lint

PR #372, opened 2016-07-15
feat: Added `lint` command to validate extension source

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/204

### DIFF
--- a/src/cmd/index.js
+++ b/src/cmd/index.js
@@ -1,5 +1,6 @@
 /* @flow */
 import build from './build';
+import lint from './lint';
 import run from './run';
 import sign from './sign';
-export default {build, run, sign};
+export default {build, lint, run, sign};
--- a/src/program.js
+++ b/src/program.js
@@ -243,6 +243,37 @@ Example: $0 --help run.
         demand: false,
         type: 'boolean',
       },
+    })
+    .command('lint', 'Validate the web extension source', commands.lint, {
+      'output': {
+        alias: 'o',
+        describe: 'The type of output to generate',
+        type: 'string',
+        default: 'text',
+        choices: ['json', 'text'],
+      },
+      'metadata': {
+        describe: 'Output only metadata as JSON',
+        type: 'boolean',
+        default: false,
+      },
+      'pretty': {
+        describe: 'Prettify JSON output',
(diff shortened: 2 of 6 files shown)


src/cmd/index.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 import build from './build';
 import run from './run';
+import lint from './lint';
 import sign from './sign';
-export default {build, run, sign};
+export default {build, run, sign, lint};
```
muffinresearch: Looks like these are sorted with the exception of the addition.

src/program.js
```diff
@@ -243,6 +243,37 @@ Example: $0 --help run.
...
+        alias: 'o',
+        describe: 'The type of output to generate',
+        type: 'string',
+        default: 'text',
+        choices: ['json', 'text'],
+      },
+      'metadata': {
+        describe: 'Output only metadata as JSON',
+        type: 'boolean',
+        default: 'false',
```
muffinresearch: Shouldn't this be `false` not `'false'`

muffinresearch:
I wonder if there would be a way to import the yargs config from the linter directly so you can expose the linter as a command and get all the cli features in the linter? Then as the linter changes you should just get the added functionality and it would help avoid any potential breakage should something in the linter change.

If this is working for now then it makes sense to not block on it, however, if there's anything that can be done on the linter side to make integration better/easier in future feel free to add issues there.

r+wc

src/program.js
kumar303: oh, heh, copy/paste error. Thanks.

kumar303:
> I wonder if there would be a way to import the yargs config from the linter directly

I thought about this too. The upside, as you mentioned, is that new linter program options would magically appear in web-ext. The downside is that it would be hard to know if any of those magic additions would make sense for web-ext or not. For example, it might break web-ext or a new option may be something that web-ext should handle directly (such as how it handles `sourceDir`).

So, yeah, I'm not entirely sure. Copying and pasting the configuration into web-ext makes it very explicit which is nice. If a new option gets added to `linter` I envision someone filing a bug in web-ext saying "you don't support X." This doesn't seem so bad.

PR #382, opened 2016-07-20
Use ./node_modules/.bin/grunt directly instead of inline Node scripts in package.json

pdehaan:
Use ./node_modules/.bin/grunt directly instead of inline Node scripts in package.json

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,7 @@
   "parser": "babel-eslint",
   "env": {
     "node": true,
-    "es6": true,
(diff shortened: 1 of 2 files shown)


kumar303:
Oh yeah, good idea. Thanks!

Your commit message just needs some minor fix-ups for [changelog extraction](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages).

pdehaan:
Done.

PR #383, opened 2016-07-20
chore: Remove unused devDependencies

pdehaan:
Removing **eslint** and **shelljs** since I couldn't see where they were used (for ESLint, the dependency is included with **grunt-eslint** anyways).

Here's the current output from [**depcheck**](http://npm.im/depcheck), but everything else looks kosher since they're imported via babel or Grunt.

Although I'm now wondering if we need both **flow-bin** and **grunt-flowbin**.

```sh
$ depcheck

Unused devDependencies
* babel-istanbul-loader
* babel-plugin-transform-class-properties
* babel-plugin-transform-es2015-modules-commonjs
* babel-plugin-transform-flow-strip-types
* babel-preset-es2015
* babel-preset-stage-2
* flow-bin
* grunt-contrib-clean
* grunt-contrib-copy
* grunt-contrib-watch
* grunt-coveralls
* grunt-eslint
* grunt-flowbin
* grunt-mocha-test
* grunt-newer
* grunt-webpack
* mocha-multi
* webpack-dev-server
```


kumar303:
Ah, yes, `shelljs` was used once but no longer is. It can be removed. I don't think we need a direct dep on eslint either, not sure why that was there. I don't think `flow-bin` needs to be a direct dep. The `grunt-flowbin` project was extracted from `web-ext` so that's probably why.

pdehaan:
@kumar303 Want me to remove **flow-bin** devDep now, or leave that for a future PR?

kumar303:
Sure, if you don't mind, you can remove `flow-bin` in this patch too

pdehaan:
Done

PR #392, opened 2016-07-26
feat: Add file extension option to `web-ext build`

rhelmer:
Fixes https://github.com/mozilla/web-ext/issues/393

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -14,7 +14,7 @@ const log = createLogger(__filename);
 
 
 function defaultPackageCreator(
-    {manifestData, sourceDir, fileFilter, artifactsDir}) {
+    {manifestData, sourceDir, fileFilter, artifactsDir, fileExtension}) {
 
   return new Promise(
     (resolve) => {
@@ -32,8 +32,13 @@ function defaultPackageCreator(
           filter: (...args) => fileFilter.wantFile(...args),
         })
         .then((buffer) => {
+          let ext = 'zip';
+          // override file extension if appropriate
+          if (fileExtension) {
+            ext = fileExtension;
+          }
           let packageName = safeFileName(
-            `${manifestData.name}-${manifestData.version}.zip`);
+            `${manifestData.name}-${manifestData.version}.${ext}`);
           let extensionPath = path.join(artifactsDir, packageName);
           let stream = createWriteStream(extensionPath);
           let promisedStream = streamToPromise(stream);
@@ -51,7 +56,7 @@ function defaultPackageCreator(
 
 
 export default function build(
-    {sourceDir, artifactsDir, asNeeded=false}: Object,
+    {sourceDir, artifactsDir, asNeeded=false, extension}: Object,
     {manifestData, fileFilter=new FileFilter(),
      onSourceChange=defaultSourceWatcher,
      packageCreator=defaultPackageCreator}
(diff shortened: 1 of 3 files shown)


src/program.js
```diff
@@ -182,6 +182,10 @@ Example: $0 --help run.
           describe: 'Watch for file changes and re-build as needed',
           type: 'boolean',
         },
+        'extension': {
+          describe: 'Change the file extension from .zip to something else',
```
kumar303: To make it clear that the value should not include a dot, maybe add another sentence: "Example: --extension=xpi"

tests/test-cmd/test.build.js
```diff
@@ -37,6 +37,31 @@ describe('build', () => {
...
+    return withTempDir(
+      (tmpDir) =>
+        build({
+          sourceDir: fixturePath('minimal-web-ext'),
+          artifactsDir: tmpDir.path(),
+          extension: 'blah',
+        })
+        .then((buildResult) => {
+          assert.match(buildResult.extensionPath,
+                       /minimal_extension-1\.0\.blah$/);
```
kumar303: You could end the test here. I don't think you need to make assertions about the contents of the package since that's covered by other tests.

src/cmd/build.js
```diff
@@ -32,8 +32,13 @@ function defaultPackageCreator(
           filter: (...args) => fileFilter.wantFile(...args),
         })
         .then((buffer) => {
+          let ext = 'zip';
+          // override file extension if appropriate
+          if (fileExtension) {
+            ext = fileExtension;
```
kumar303: hmm. I'm a little concerned that someone would accidentally add a dot which could make the filename `package..xpi`. Perhaps the `fileExtension` value should be stripped of all non `[a-zA-Z0-9_-]` characters?

kumar303:
Thanks! I figured someone would ask for this eventually :) Comments added.

kumar303:
@rhelmer I guess you found a workaround but did you still want to get this in?

rhelmer:
Yep, sorry for the delay.

tests/test-cmd/test.build.js
```diff
@@ -37,6 +37,40 @@ describe('build', () => {
...
+          return buildResult.extensionPath;
+        })
+    );
+  });
+
+  it('throws an error when invalid extension is provided', () => {
+    let failed = false;
+    try {
```
kumar303: all callers will expect to treat `build()`'s return value as a promise so I think it's time to wrap the first part of that function in a promise:
````javascript
return new Promise(
  (resolve) => {
    if (!allowedExtension.test(fileExtension)) {
      throw new Error('...');
    }
    resolve();
  })
  .then(() => prepareArtifactsDir(artifactsDir));
````

@rpl is embarking on an epic patch to switch us to async/await! https://github.com/mozilla/web-ext/pull/411

Comment (kumar303) [inline tests/test-cmd/test.build.js]: after that change, here's [the pattern](https://github.com/rhelmer/web-ext/blob/3556526a4c2c1f357265982a8ab96cfb2b5ebfe9/tests/test-cmd/test.build.js#L203-L208) we use to check for promise rejections.

File: src/cmd/build.js [hunk 5]
```diff
@@ -60,8 +65,16 @@ export default function build(
   const rebuildAsNeeded = asNeeded; // alias for `build --as-needed`
   log.info(`Building web extension from ${sourceDir}`);
 
+  const fileExtension = extension; // alias for `build --extension <ext>`
+  let allowedExtension = /^[a-zA-Z0-9_-]+$/;
+  if (!allowedExtension.test(fileExtension)) {
+    throw Error('Only alphanumeric, -, _ characters are ' +
```
kumar303: this should throw [WebExtError](https://github.com/mozilla/web-ext/blob/master/src/errors.js#L8).

kumar303:
Now that we landed [improved static typing](https://github.com/mozilla/web-ext/pull/424), this patch will need a rebase and considerable cleanup. If you want to close it and just wait until someone needs the feature, that's fine. I am happy to review it again if you want to resurrect it though.

rhelmer:
@kumar303 well I am using it, happy to rebase and cleanup :)

kumar303:
Hi @rhelmer . We landed a big change on master that switches to async/await instead of promises. This should make your patch easier though :) The tests still use promises (for now).

kumar303:
We're getting a bit overwhelmed with pull requests so I'm trying to clear some mental space on the tracker. Feel free to re-open this if you feel like working on it!

PR #411, opened 2016-08-04
refactor: use async/await syntax

rpl:
The goal of this PR is discussing the following proposals:

- introducing `async / await` syntax to make some of the asynchronous functions nicer to read
- introducing more explicit flow types declarations to extend the kind of statical analysis that flow is able to
  make on the sources 

The first patch contains a set of experimental changes related to the usage of the `async / await` syntax to make long chains of promises more readable, by configuring babel to transpile them into code that can run across the various supported nodejs versions.
This patch ports to the `async / await` syntax only functions where it makes the code nicer and easier to read, and it doesn't port to the new syntax any of the test cases (mostly because it is better to keep the tests unmodified during such a change).   

~~The second patch introduces more explicit flow types to be able to give to flow the knowledge needed to check the code in much more detail, e.g. by declaring the types of the objects passed as params, signature of the callbacks etc.
The second patch can be augmented with much more explicit flow types definitions, but I wanted to give an initial taste of what it means in terms of the introduced changes before going further in this path. **(landed as #424)**~~

As a side note, [`async/await` has been recently moved into Stage 4](https://github.com/tc39/proposals/commit/e8c03544751b071b973064ef7d437a34ac6752cd), and so it seems that it will be officially part of the 2017 version of the standard.


rpl:
@kumar303 I've just updated this PR by rewriting any function that could benefit from the new async/await syntax, above the changes introduced by #424 

The coverall report is currently red because of a small helper that I added into "utils/errors.js", I'm going to write a new small test for it asap.

Besides the (temporary) coverall report failure, this PR should now ready for a review.

rpl:
@kumar303 I just added the new test case for the `isErrorWithCode` helper (as 406aadc) which makes coveralls to be green and happy as before ;-)

src/cmd/build.js
```diff
@@ -40,41 +40,34 @@ export type PackageCreatorParams = {
...
-        });
-    });
+  if (manifestData) {
+    id = getManifestId(manifestData);
+    log.debug(`Using manifest id=${id || '[not specified]'}`);
+  } else {
+    manifestData = await getValidatedManifest(sourceDir);
+  }
+
+  let buffer =  await zipDir(sourceDir, {
```
kumar303: nit: extra space after the equal sign. Maybe we could add a linting rule for that?

src/cmd/build.js
```diff
@@ -40,41 +40,34 @@ export type PackageCreatorParams = {
...
+    filter: (...args) => fileFilter.wantFile(...args),
+  });
+
+  let packageName = safeFileName(
```
kumar303: I would prefer `const` for all immutable vars because then the linter will tell us if they get mutated.

kumar303: I know you just copy/pasted but now you own it :)

src/cmd/build.js
```diff
@@ -40,41 +40,34 @@ export type PackageCreatorParams = {
...
+    `${manifestData.name}-${manifestData.version}.zip`);
+  let extensionPath = path.join(artifactsDir, packageName);
+  let stream = createWriteStream(extensionPath);
+  let promisedStream = streamToPromise(stream);
+
+  stream.write(buffer, () => stream.end());
+
+  await promisedStream;
```
kumar303: maybe this could just be `await streamToPromise(stream)`?

kumar303: I forget if it needs to be initialized before writing begins or not.

kumar303:
What happens if some code accidentally calls an `async` function without using the `await` keyword? Is there an error? If not, could we add a linting rule for that?

src/cmd/run.js
```diff
@@ -107,33 +108,33 @@ export function defaultFirefoxClient(
...
-        .catch((error) => {
+  async function establishConnection() {
+    var retries = 0;
+    var lastError;
+
+    while (retries <= maxRetries) {
+      try {
+        return await connectToFirefox();
+      } catch (error) {
+        if (isErrorWithCode('ECONNREFUSED', error)) {
```
kumar303: 👍 for `isErrorWithCode()`, that reads well

src/cmd/run.js
```diff
@@ -107,33 +108,33 @@ export function defaultFirefoxClient(
...
-              resolve(establishConnection());
-            }, retryInterval);
-          }
-        }))
-        .catch((error) => {
+  async function establishConnection() {
+    var retries = 0;
+    var lastError;
+
+    while (retries <= maxRetries) {
```
kumar303: you could probably get rid of `retries` if you wanted by doing:

````javascript
for (const retry of new Array(maxRetries)) {
  // ...
}
````

Comment (kumar303) [inline src/cmd/run.js]: bah! Nevermind. That doesn't give you a counter. Looks like there isn't a one-liner equivalent of Python's `range()`.

File: src/cmd/run.js [hunk 6]
```diff
@@ -107,33 +108,33 @@ export function defaultFirefoxClient(
...
+            `Retrying Firefox (${retries}); connection error: ${error}`);
+        } else {
           log.error(error.stack);
-          reject(error);
-        });
-    });
+          throw error;
+        }
+      }
+    }
```
kumar303: no recursion! yay, so much easier to read

src/cmd/run.js
```diff
@@ -186,87 +187,70 @@ export default function run(
...
+
+  runner = new ExtensionRunner({
+    sourceDir,
+    firefox,
+    firefoxBinary,
+    manifestData,
+    profilePath: firefoxProfile,
+  });
+
+  profile = await  runner.getProfile();
```
kumar303: nit: a couple extra spaces here

src/cmd/run.js
```diff
@@ -186,87 +187,70 @@ export default function run(
...
-    });
+
+      log.debug(
+        `Reloading extension when the source changes; id=${addonId}`);
+      reloadStrategy({
+        firefox: runningFirefox,
+        profile, client, sourceDir, artifactsDir, addonId,
+      });
+    }
+  }
```
kumar303: oh man, this function is beautiful now

src/cmd/sign.js
```diff
@@ -57,113 +57,115 @@ export default function sign(
...
+      }
+
+      if (manifestId) {
+        id = manifestId;
+      }
+
+      if (!id && idFromSourceDir) {
+        log.info(
+          'Using previously auto-generated extension ID: ' +
+          `${idFromSourceDir}`);
```
kumar303: this can probably fit on one line now that it's not nested in `then()` soup

src/cmd/sign.js
```diff
@@ -57,113 +57,115 @@ export default function sign(
...
-          return line;
-        }
-      });
-      let id = lines[0];
-      log.debug(`Found extension ID ${id} in ${filePath}`);
-      if (!id) {
-        throw new WebExtError(`No ID found in extension ID file ${filePath}`);
+
+  try {
+    const content = await fs.readFile(filePath);
```
kumar303: I feel like only this block should be wrapped in the try/catch to check for missing files:
````javascript
try {
  const content = await fs.readFile(filePath);
} catch(error) {
    log.debug(`No ID file found at: ${filePath}`);
    return;
  } else {
    throw error;
  }
}
````
What do you think? I suspect that the old code was just wrong.

File: src/errors.js [hunk 11]
```diff
@@ -95,3 +95,16 @@ export function onlyErrorsWithCode(
     return errorHandler(error);
   };
 }
+
+export function isErrorWithCode(
```
kumar303: I guess we have to wait until the tests are finished to delete `onlyErrorsWithCode`?

src/firefox/index.js
```diff
@@ -46,28 +46,31 @@ export type RemotePortFinderParams = {
...
-      client.disconnect();
-      // TODO: instead of throw an error, pick a new random port until
-      // one of them is available.
-      // https://github.com/mozilla/web-ext/issues/283
-      throw new WebExtError(
-        `Cannot listen on port ${portToTry} because it's in use`);
-    })
-    .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
+  try {
+    const client = await connectToFirefox(portToTry);
```
kumar303: I think you can move up the `catch` block to this line and return `portToTry`.

kumar303:
I'm excited about this! I got halfway through but I have to end for the day. I'll pick it up tomorrow.

src/cmd/build.js
pdehaan: FWIW, there are 2 new ESLint rules you can add:

- [`no-var`](eslint.org/docs/rules/no-var) &mdash; require `let` or `const` instead of `var`.
- [`prefer-const`](eslint.org/docs/rules/prefer-const) &mdash; Suggest using `const`.

```diff
diff --git a/.eslintrc b/.eslintrc
index 5589499..5264ac3 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -73,8 +73,10 @@
     "no-undef": 2,
     "no-underscore-dangle": 0,
     "no-unused-vars": 2,
+    "no-var": 2,
     "no-with": 2,
     "one-var": [2, "never"],
+    "prefer-const": 2,
     "prefer-template": 2,
     "quote-props": [1, "consistent-as-needed"],
     "quotes": [2, "single", "avoid-escape"],
```

After adding those 2 rules, I get `(125 errors, 0 warnings)` after running `npm run lint`.

src/cmd/build.js
pdehaan: I think the rule in question is [`no-multi-spaces`](http://eslint.org/docs/rules/no-multi-spaces).

rpl: :+1:

added `no-multi-spaces` to the enabled eslint rules and removed all the related linting errors in 358c26b

src/cmd/build.js
rpl: @pdehaan That's great, I tried to enable the above two eslint rules and let eslint to fix the errors and it seems to be able to fix almost every one of them (besides a small bunch).

Unfortunately these errors are spread all over the repo, and so my proposal is to choose one of the following approaches:

- enable the two eslint rules as warnings instead of errors, fix only the warnings that are into functions (or files) that we are already touching with this pull request and defer the other in a separate followup pull request (to be opened immediately after merging this one)

- defer this part ot the last patch of this pull request, where we enable the two rules as errors and fix all of them, once the async/await part of the change is ready (mainly to make the review process easier on the eyes :-))

@kumar303 how this sounds to you?

rpl:
> What happens if some code accidentally calls an async function without using the await keyword? Is > there an error? If not, could we add a linting rule for that?

@kumar303 sorry, I forgot to mention this in the summary of this pull request, and I totally agree that it is an important detail that needs to be discussed in this pull request:

An "async function" always returns a Promise, the returned promise is resolved to the value `return`-ed by the async function, if an exception is thrown in the body of the async function (and there is no try/catch to stop it), then the returned promise is rejected and the exception passed as parameter of the Promise error callback.

This is very nice, because it means that async function can be composed into Promise chains and mixed with non-async function which returns a Promise (e.g. it is completely valid to put the result of an async function into a `Promise.all` array), but it probably makes a bit harder to decide if the await keyword is missing by mistake or intentionally, e.g.:

```
let res = asyncFunction01(...);
let res2 = asyncFunction02(...);

await Promise.all([res, res2]);
```

versus

```
let res = await asyncFunction01(...);
...
let res2 = await asyncFunction02(...);
...
```

Nevertheless, there is an `eslint-plugin-async-await` npm package that has two rules that can be interesting, because they enforce a space where async/await can be confused for a function name, e.g.:

```
// This is valid for babel, but it can be confusing
let res = await(asyncFunction01(...)); 

// is what the `eslint-plugin-async-await` can help to enforce
let res = await (asyncFunction01(...));

**Comment by rpl** [comment]
> but it probably makes a bit harder to decide if the await keyword is missing by mistake or intentionally

to be completely fair, what I think is that probably eslint will not be able to detect this kind of mistakes... but I expect that most of them should be caught by the flow type checks :-D

e.g. Flow should be able to detect when we are trying to use a promise as the result of "awaiting a promise resolved value", as well as if we are trying to put into a `Promise.all` something that is not a promise, and if it doesn't detect it, now we know how to look into the flow coveraging info ;-)

File: src/cmd/run.js (hunk 8 above)
Comment (rpl) [inline src/cmd/run.js]: agree, this function and the `establishConnection` helper above are my favorites "rewrited to async functions" use cases :-D

(once I had done with them, I re-looked at the diff and I couldn't block some "tears of happiness", like a young groupie at the concert of their favorite band :-P)

File: src/cmd/build.js (hunk 3 above)
Comment (rpl) [inline src/cmd/build.js]: :+1:

We already do something similar in `"src/firefox/index.js"`, fixed in ff12c02d

File: src/cmd/run.js (hunk 5 above)
Comment (rpl) [inline src/cmd/run.js]: @kumar303 Actually, we can achieve it using a generator, e.g. something like:

```
function* range(max) {
  for (let i = 0; i < max; i++) {
    yield i;
  }
}

for (let el of range(3)) {
  console.log(el);
}
```

I rewrote it as a for-loop over a custom generator in cdcbc92, if we like it I can move the helper into a separate file and add a test to cover it, or I can rewrite it directly using the for loop (which is basically there, only hidden inside the generator, which is nice because it lets us to use `const` for the `retries` identifier for the for loop).

File: src/cmd/sign.js (hunk 10 above)
Comment (rpl) [inline src/cmd/sign.js]: @kumar303 yep, we can do it, but we need to move `content` declaration out of the try/catch block, or it will not be available to the rest of the code in the function.

I've applied this change in 777a8bf0

File: src/errors.js (hunk 11 above)
Comment (rpl) [inline src/errors.js]: I was thinking about it too:
my only concern is that it can be still useful if in the near future we need to catch errors into a promise chain and we don't want (or maybe we can't) use the async functions directly for some reasons that maybe we don't see right now.

**Comment by kumar303** [comment]
ok, if you think Flow will catch enough mistakes then that's fine.

Another idea I had: it seems like 99% of the time we will be using the `await` keyword. If that's correct, we could add the `eslint-plugin-async-await` rule to enforce the use of `await` for any `async` function. For any legitimate concurrency cases where we will use `Promise.all()` without an `await` keyword, we could tell the linter to allow these lines with `eslint-disable-next-line`.

File: src/cmd/run.js [hunk 13]
```diff
@@ -107,33 +108,37 @@ export function defaultFirefoxClient(
...
+  async function establishConnection() {
+    var lastError;
+
+    function* range(min: number, max: number, step: number = 1) {
+      for (let i = min; i <= max; i += step) {
+        yield i;
+      }
+    }
+
+    for (const retries of range(0, maxRetries)) {
```
kumar303: on a second look, I think this could just be the following (unless I'm missing something):

    for (let retries = 0; retries <= maxRetries; retries++) {
    }

src/firefox/remote.js
```diff
@@ -143,38 +143,35 @@ export class RemoteFirefox {
...
   }
 
-  checkForAddonReloading(
+  async checkForAddonReloading(
     addon: FirefoxRDPAddonActor
   ): Promise<FirefoxRDPAddonActor> {
     if (this.checkedForAddonReloading) {
       // We only need to check once if reload() is supported.
-      return Promise.resolve(addon);
+      return addon;
```
kumar303: I don't think `checkForAddonReloading` needs to return `addon` anymore. IIRC this was just to make chaining promises more convenient.

src/program.js
```diff
@@ -65,54 +65,52 @@ export class Program {
...
     this.shouldExitProgram = shouldExitProgram;
     let argv;
     let cmd;
 
-    return new Promise(
-      (resolve) => {
-        this.yargs.exitProcess(this.shouldExitProgram);
-        argv = this.yargs.argv;
+    this.yargs.exitProcess(this.shouldExitProgram);
+    argv = this.yargs.argv;
```
kumar303: I think this can just be `let argv = ...` since we no longer have to expose it to nested promise callback scopes.

src/program.js
```diff
@@ -65,54 +65,52 @@ export class Program {
...
-        resolve(runCommand);
-      })
-      .then((runCommand) => runCommand(argv))
-      .catch((error) => {
-        const prefix = cmd ? `${cmd}: ` : '';
-        log.error(`\n${prefix}${error.stack}\n`);
-        if (error.code) {
-          log.error(`${prefix}Error code: ${error.code}\n`);
-        }
+    cmd = argv._[0];
```
kumar303: This can also be `let cmd = ...` because it's not longer set in a nested scope.

src/util/manifest.js
```diff
@@ -16,46 +16,54 @@ export type ExtensionManifest = {
...
-        throw new InvalidManifest(
-          `Manifest at ${manifestFile} is invalid: ${errors.join('; ')}`);
-      }
-      return manifestData;
-    });
+
+  let manifestContents;
+
+  try {
+    manifestContents = await fs.readFile(manifestFile);
```
kumar303: No need to pre-declare this anymore. It can be `const manifestContents = ...`

src/util/manifest.js
```diff
@@ -16,46 +16,54 @@ export type ExtensionManifest = {
...
+  } catch (error) {
+    throw new InvalidManifest(
+      `Could not read manifest.json file at ${manifestFile}: ${error}`);
+  }
+
+  let manifestData;
+
+  try {
+    manifestData = JSON.parse(manifestContents);
```
kumar303: No need to pre-declare this anymore. It can be `const manifestData = ...`.

src/util/manifest.js
kumar303: oops, nevermind. I forgot that the try block is a nested scope.

src/util/manifest.js
kumar303: oops, nevermind. I forgot that the try block is a nested scope.

kumar303:
I finished reviewing :) This looks great, I only left minor comments. r+

I didn't have time to check out this branch and give it a spin. I'd suggest doing a quick smoke test of each command (if you haven't already).

src/firefox/index.js
```diff
@@ -47,36 +47,40 @@ export type RemotePortFinderParams = {
...
-        }
-      })
-      // The connection was refused so this port is good to use.
-     .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
-       return portToTry;
-     }));
+
+  let client;
+
+  while (retriesLeft >= 0) {
```
rpl: @kumar303 This is the "async/await"-based version of the new `defaultRemotePortFinder` (which is the bigger change after the rebase, and it worth a second look from an additional pair of eyes ;-))

src/firefox/index.js
```diff
@@ -47,36 +47,40 @@ export type RemotePortFinderParams = {
...
+    try {
+      client = await connectToFirefox(portToTry);
+      log.debug(`Remote Firefox port ${portToTry} is in use ` +
+                `(retries remaining: ${retriesLeft} )`);
```
kumar303: while you're at it, you could fix the extra space in the parentheses around `${retriesLeft}` :) no biggie though, just looks odd in the console.

src/firefox/index.js
kumar303: This looks great to me.

src/firefox/index.js
rpl: :+1: small tweak just pushed

PR #412, opened 2016-08-04
Fixed Flow coverage for imports of ES6 modules

kumar303:
* Fixed our `.flowconfig` and re-introduced coverage of our ES6 imports. Can't believe [how simple](https://github.com/facebook/flow/issues/1448#issuecomment-237404200) this fix was!
* I upgraded flow-bin to get the new `<PROJECT_ROOT>` thing. This also required adding a polymorphic argument to all Promise return types. However, note that I did not add strict argument types in all cases. @rpl has a better patch for that in the works.


package.json
```diff
@@ -78,13 +78,14 @@
     "copy-dir": "0.3.0",
     "coveralls": "2.11.12",
     "deepcopy": "0.6.3",
+    "flow-bin": "0.30.0",
```
kumar303: even though `grunt-flowbin` pulls this in for us, I'd like to keep it web-ext's `package.json` so that greenkeeper will auto-update our repo when new versions are released. It turns out web-ext was a bit behind.

src/cmd/run.js
```diff
@@ -1,4 +1,5 @@
 /* @flow */
+import type FirefoxProfile from 'firefox-profile';
```
rpl: @kumar303 I would prefer if we can keep the flow `import/export type`s clearly separated from the regular ES6 imports/exports,
mostly because I think that it would make much more clear that we are not importing/exporting code but flow types.

How about moving all the `import/export type` grouped after the group of ES6 module imports?

e.g.
``` 
import * as defaultFirefox from '../firefox';
import defaultFirefoxConnector from '../firefox/remote';
...
import defaultSourceWatcher from '../watcher';

// Import and export Flow types
import type FirefoxProfile from 'firefox-profile';
```

kumar303: good idea

rpl:
@kumar303 looks good to me, there is just a small nit in the above comment.

Then, as pointed out in the summary of this PR, I'm going to rebase #412 on the changes introduced by this PR and extend the flow types checking further ;-)

:+1: 

r+wc

Issue #310, opened 2016-06-21
Show usage exceptions without a stacktrace


kumar303:
Some web-ext functions throw exceptions that indicate incorrect usage. We should do a few things to handle these:

- create a subclass of `WebExtError` as `UsageError`
- update all relevant exceptions to use this new class (not all exceptions are usage errors)
- hide the stack trace for instances of `UsageError` (unless running with `--verbose`)

This will make the output easier to read and easier to act upon.

See https://github.com/mozilla/web-ext/pull/309 for examples.

PR #424, opened 2016-08-08
Test: add more flow types

rpl:
In this PR there are the changes related to the additional flow type checking (that are currently part of #411) extracted and extended with much more flow types checks.

Once we complete this PR, #411 will be rebased on the final version of these changes and it will be limited to the introduction of the `async/await` syntax (with the additional safety net provided by the additional static analysis that flow will be able to provide).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,13 @@
 {
   "parser": "babel-eslint",
+  "plugins": [
+    "flowtype"
+  ],
+  "settings": {
+    "flowtype": {
+      "onlyFilesWithFlowAnnotation": true
+    }
+  },
   "env": {
     "node": true,
     "es6": true
@@ -75,6 +83,10 @@
     "space-infix-ops": 0,
     "strict": [2, "never"],
     "valid-typeof": 2,
-    "wrap-iife": [2, "inside"]
+    "wrap-iife": [2, "inside"],
+
+    "flowtype/define-flow-type": 1,
+    "flowtype/use-flow-type": 1,
+    "flowtype/space-after-type-colon": [1, "always"]
   }
 }
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,6 +5,8 @@
 [options]
 module.system=node
 log.file=./artifacts/flow.log
+suppress_comment= \\(.\\|\n\\)*\\$FLOW_FIXME
+suppress_comment= \\(.\\|\n\\)*\\$FLOW_IGNORE 
 
 [ignore]
 <PROJECT_ROOT>/dist/.*
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "copy-dir": "0.3.0",
     "coveralls": "2.11.12",
     "deepcopy": "0.6.3",
+    "firefox-client": "0.3.0",
     "flow-bin": "0.30.0",
     "grunt": "1.0.1",
     "grunt-contrib-clean": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "copy-dir": "0.3.0",
     "coveralls": "2.11.12",
     "deepcopy": "0.6.3",
+    "eslint-plugin-flowtype": "2.4.1",
     "firefox-client": "0.3.0",
     "flow-bin": "0.30.0",
     "grunt": "1.0.1",
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -31,23 +31,6 @@ export type PackageCreatorParams = {
   artifactsDir: string,
 };
 
-export type BuildCmdParams = {
-  sourceDir: string,
-  artifactsDir: string,
-  asNeeded?: boolean,
-};
-
-export type BuildCmdOptions = {
-  manifestData?: ExtensionManifest,
-  fileFilter?: FileFilter,
-  onSourceChange?: OnSourceChangeFn,
-  packageCreator?: PackageCreatorFn,
-};
-
-export type FileFilterOptions = {
-  filesToIgnore?: Array<string>,
-};
-
 // Module internals & exports
 
 function defaultPackageCreator(
@@ -87,6 +70,18 @@ function defaultPackageCreator(
     });
 }
 
+export type BuildCmdParams = {
+  sourceDir: string,
+  artifactsDir: string,
+  asNeeded?: boolean,
+};
+
+export type BuildCmdOptions = {
+  manifestData?: ExtensionManifest,
+  fileFilter?: FileFilter,
+  onSourceChange?: OnSourceChangeFn,
+  packageCreator?: PackageCreatorFn,
+};
 
 export default function build(
   {sourceDir, artifactsDir, asNeeded=false}: BuildCmdParams,
@@ -128,6 +123,9 @@ export function safeFileName(name: string): string {
   return name.toLowerCase().replace(/[^a-z0-9\.-]+/g, '_');
 }
 
+export type FileFilterOptions = {
+  filesToIgnore?: Array<string>,
+};
 
 /*
  * Allows or ignores files when creating a ZIP archive.
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
 
 const log = createLogger(__filename);
 
+// Flow Types
+
+import type {OnSourceChangeFn} from '../watcher';
+import type {ExtensionManifest} from '../util/manifest';
+
+export type PackageCreatorFn =
+    (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
+
+export type ExtensionBuildResult = {
+  extensionPath: string,
+};
+
+export type PackageCreatorParams = {
+  manifestData?: ExtensionManifest,
+  sourceDir: string,
+  fileFilter: FileFilter,
+  artifactsDir: string,
+};
+
+export type BuildCmdParams = {
+  sourceDir: string,
+  artifactsDir: string,
+  asNeeded?: boolean,
+};
+
+export type BuildCmdExtraParams = {
+  manifestData?: ExtensionManifest,
+  fileFilter?: FileFilter,
+  onSourceChange?: OnSourceChangeFn,
+  packageCreator?: PackageCreatorFn,
+};
+
+export type FileFilterOptions = {
+  filesToIgnore?: Array<string>,
+};
+
+// Module internals & exports
 
 function defaultPackageCreator(
-    {manifestData, sourceDir, fileFilter, artifactsDir}) {
+  {manifestData, sourceDir, fileFilter, artifactsDir}: PackageCreatorParams
+): Promise<ExtensionBuildResult> {
 
   return new Promise(
     (resolve) => {
@@ -51,12 +89,13 @@ function defaultPackageCreator(
 
 
 export default function build(
-    {sourceDir, artifactsDir, asNeeded=false}: Object,
-    {manifestData, fileFilter=new FileFilter(),
-     onSourceChange=defaultSourceWatcher,
-     packageCreator=defaultPackageCreator}
-    : Object = {}): Promise<Object> {
-
+  {sourceDir, artifactsDir, asNeeded=false}: BuildCmdParams,
+  {
+    manifestData, fileFilter=new FileFilter(),
+    onSourceChange=defaultSourceWatcher,
+    packageCreator=defaultPackageCreator,
+  }: BuildCmdExtraParams = {}
+): Promise<ExtensionBuildResult> {
   const rebuildAsNeeded = asNeeded; // alias for `build --as-needed`
   log.info(`Building web extension from ${sourceDir}`);
 
@@ -96,7 +135,7 @@ export function safeFileName(name: string): string {
 export class FileFilter {
   filesToIgnore: Array<string>;
 
-  constructor({filesToIgnore}: Object = {}) {
+  constructor({filesToIgnore}: FileFilterOptions = {}) {
     this.filesToIgnore = filesToIgnore || [
       '**/*.xpi',
       '**/*.zip',
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -37,7 +37,7 @@ export type BuildCmdParams = {
   asNeeded?: boolean,
 };
 
-export type BuildCmdExtraParams = {
+export type BuildCmdOptions = {
   manifestData?: ExtensionManifest,
   fileFilter?: FileFilter,
   onSourceChange?: OnSourceChangeFn,
@@ -94,7 +94,7 @@ export default function build(
     manifestData, fileFilter=new FileFilter(),
     onSourceChange=defaultSourceWatcher,
     packageCreator=defaultPackageCreator,
-  }: BuildCmdExtraParams = {}
+  }: BuildCmdOptions = {}
 ): Promise<ExtensionBuildResult> {
   const rebuildAsNeeded = asNeeded; // alias for `build --as-needed`
   log.info(`Building web extension from ${sourceDir}`);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -10,15 +10,17 @@ import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
 
+
 const log = createLogger(__filename);
 
-// Flow Types
+
+// Import flow types.
 
 import type {OnSourceChangeFn} from '../watcher';
 import type {ExtensionManifest} from '../util/manifest';
 
-export type PackageCreatorFn =
-    (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
+
+// defaultPackageCreator types and implementation.
 
 export type ExtensionBuildResult = {
   extensionPath: string,
@@ -31,7 +33,8 @@ export type PackageCreatorParams = {
   artifactsDir: string,
 };
 
-// Module internals & exports
+export type PackageCreatorFn =
+    (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
 function defaultPackageCreator(
   {manifestData, sourceDir, fileFilter, artifactsDir}: PackageCreatorParams
@@ -69,6 +72,9 @@ function defaultPackageCreator(
         });
     });
 }
+
+
+// Build command types and implementation.
 
 export type BuildCmdParams = {
   sourceDir: string,
@@ -119,9 +125,14 @@ export default function build(
 }
 
 
+// safeFileName helper implementation.
+
 export function safeFileName(name: string): string {
   return name.toLowerCase().replace(/[^a-z0-9\.-]+/g, '_');
 }
+
+
+// FileFilter types and implementation.
 
 export type FileFilterOptions = {
   filesToIgnore?: Array<string>,
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -13,6 +13,10 @@ import {createLogger} from '../util/logger';
 
 const log = createLogger(__filename);
 
+export function safeFileName(name: string): string {
+  return name.toLowerCase().replace(/[^a-z0-9\.-]+/g, '_');
+}
+
 
 // Import flow types.
 
@@ -122,13 +126,6 @@ export default function build(
       }
       return result;
     });
-}
-
-
-// safeFileName helper implementation.
-
-export function safeFileName(name: string): string {
-  return name.toLowerCase().replace(/[^a-z0-9\.-]+/g, '_');
 }
 
 
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -41,7 +41,7 @@ export type LintCmdParams = {
   pretty?: boolean,
 };
 
-export type LintCmdExtraParams = {
+export type LintCmdOptions = {
   createLinter: LinterCreatorFn,
   fileFilter: FileFilter,
 };
@@ -56,7 +56,7 @@ export default function lint(
   {
     createLinter=defaultLinterCreator,
     fileFilter=new FileFilter(),
-  }: LintCmdExtraParams = {}
+  }: LintCmdOptions = {}
 ): Promise<void> {
   log.debug(`Running addons-linter on ${sourceDir}`);
   const linter = createLinter({
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -5,11 +5,59 @@ import {FileFilter} from './build';
 
 const log = createLogger(__filename);
 
+// Flow types
+
+export type LinterLogLevel = 'debug' | 'fatal';
+export type LinterOutputType = 'text' | 'json';
+
+export type LinterCreatorParams = {
+  config: {
+    logLevel: LinterLogLevel,
+    stack: boolean,
+    pretty?: boolean,
+    metadata?: boolean,
+    output?: LinterOutputType,
+    boring?: boolean,
+    selfHosted?: boolean,
+    shouldScanFile: (fileName: string) => boolean,
+    _: Array<string>,
+  },
+};
+
+export type Linter = {
+  run: () => Promise<void>,
+};
+
+export type LinterCreatorFn = (params: LinterCreatorParams) => Linter;
+
+
+export type LintCmdParams = {
+  sourceDir: string,
+  verbose?: boolean,
+  selfHosted?: boolean,
+  boring?: boolean,
+  output?: LinterOutputType,
+  metadata?: boolean,
+  pretty?: boolean,
+};
+
+export type LintCmdExtraParams = {
+  createLinter: LinterCreatorFn,
+  fileFilter: FileFilter,
+};
+
+// Module internals & exports
+
 export default function lint(
-    {verbose, sourceDir, selfHosted, boring, output,
-     metadata, pretty}: Object,
-    {createLinter=defaultLinterCreator, fileFilter=new FileFilter()}
-    : Object = {}): Promise<void> {
+  {
+    verbose, sourceDir, selfHosted, boring, output,
+    metadata, pretty,
+  }: LintCmdParams,
+  {
+    createLinter=defaultLinterCreator,
+    fileFilter=new FileFilter(),
+  }: LintCmdExtraParams = {}
+): Promise<void> {
   log.debug(`Running addons-linter on ${sourceDir}`);
   const linter = createLinter({
     config: {
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -3,7 +3,9 @@ import {createInstance as defaultLinterCreator} from 'addons-linter';
 import {createLogger} from '../util/logger';
 import {FileFilter} from './build';
 
+
 const log = createLogger(__filename);
+
 
 // Define the needed 'addons-linter' module flow types.
 
@@ -29,7 +31,8 @@ export type Linter = {
 
 export type LinterCreatorFn = (params: LinterCreatorParams) => Linter;
 
-// Define the './cmd/lint' flow types.
+
+// Lint command types and implementation.
 
 export type LintCmdParams = {
   sourceDir: string,
@@ -45,8 +48,6 @@ export type LintCmdOptions = {
   createLinter?: LinterCreatorFn,
   fileFilter?: FileFilter,
 };
-
-// Module internals & exports
 
 export default function lint(
   {
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -7,6 +7,8 @@ const log = createLogger(__filename);
 
 // Flow types
 
+// Define the needed 'addons-linter' module flow types.
+
 export type LinterOutputType = 'text' | 'json';
 
 export type LinterCreatorParams = {
@@ -29,6 +31,7 @@ export type Linter = {
 
 export type LinterCreatorFn = (params: LinterCreatorParams) => Linter;
 
+// Define the './cmd/lint' flow types.
 
 export type LintCmdParams = {
   sourceDir: string,
@@ -41,8 +44,8 @@ export type LintCmdParams = {
 };
 
 export type LintCmdOptions = {
-  createLinter: LinterCreatorFn,
-  fileFilter: FileFilter,
+  createLinter?: LinterCreatorFn,
+  fileFilter?: FileFilter,
 };
 
 // Module internals & exports
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -5,8 +5,6 @@ import {FileFilter} from './build';
 
 const log = createLogger(__filename);
 
-// Flow types
-
 // Define the needed 'addons-linter' module flow types.
 
 export type LinterOutputType = 'text' | 'json';
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -7,12 +7,11 @@ const log = createLogger(__filename);
 
 // Flow types
 
-export type LinterLogLevel = 'debug' | 'fatal';
 export type LinterOutputType = 'text' | 'json';
 
 export type LinterCreatorParams = {
   config: {
-    logLevel: LinterLogLevel,
+    logLevel: 'debug' | 'fatal',
     stack: boolean,
     pretty?: boolean,
     metadata?: boolean,
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -253,7 +253,7 @@ export default function run(
         log.debug('Extension auto-reloading has been disabled');
       } else {
         if (!addonId) {
-          throw new Error(
+          throw new WebExtError(
             'Unexpected missing addonId in the installAsTemporaryAddon result'
           );
         }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -249,15 +249,15 @@ export default function run(
         'versions, use --pre-install');
     }))
     .then(() => {
-      if (!addonId) {
-        throw new WebExtError(
-          'Missing addonId in the installAsTemporaryAddon result'
-        );
-      }
-
       if (noReload) {
         log.debug('Extension auto-reloading has been disabled');
       } else {
+        if (!addonId) {
+          throw new Error(
+            'Unexpected missing addonId in the installAsTemporaryAddon result'
+          );
+        }
+
         log.debug(
           `Reloading extension when the source changes; id=${addonId}`);
         reloadStrategy({
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -17,8 +17,14 @@ const log = createLogger(__filename);
 import type FirefoxProfile from 'firefox-profile';
 import type {OnSourceChangeFn} from '../watcher';
 import type Watchpack from 'watchpack';
-import type {EventEmitter} from 'events';
-import type {FirefoxConnectorFn, RemoteFirefox} from '../firefox/remote';
+import type {
+  FirefoxProcess,
+} from '../firefox/index';
+import type {
+  FirefoxConnectorFn, RemoteFirefox,
+  FirefoxRDPResponseAddon,
+} from '../firefox/remote';
+import type {ExtensionManifest} from '../util/manifest';
 
 export type WatcherCreatorParams = {
   addonId: string,
@@ -30,7 +36,7 @@ export type WatcherCreatorParams = {
 
 export type ReloadStrategyParams = {
   addonId: string,
-  firefox: EventEmitter,
+  firefox: FirefoxProcess,
   client: RemoteFirefox,
   profile: FirefoxProfile,
   sourceDir: string,
@@ -78,7 +84,7 @@ export function defaultReloadStrategy(
     createWatcher=defaultWatcherCreator,
   }: ReloadStrategyOptions = {}
 ): void {
-  let watcher;
+  let watcher: Watchpack;
 
   firefox.on('close', () => {
     client.disconnect();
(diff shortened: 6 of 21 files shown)


.flowconfig
```diff
@@ -5,6 +5,8 @@
 [options]
 module.system=node
 log.file=./artifacts/flow.log
+suppress_comment= \\(.\\|\n\\)*\\$FLOW_FIXME
+suppress_comment= \\(.\\|\n\\)*\\$FLOW_IGNORE
```
rpl: This `suppress_comment` options in the `".flowconfig"` are useful to ignore (and associate an helpful comment to) flow errors that are raised due to some of the tricks used in the tests or due to flow limitations.

I've defined `// $FLOW_IGNORE: <comment>` to ignore errors that are related to type checking errors that are used on purpose in the testing suite, and `// $FLOW_FIXME: <comment>` to ignore errors that we want to fix in followups (the latter is currently unused).

.travis.yml
```diff
@@ -30,3 +30,8 @@ deploy:
     secure: gMZR5u6VxrXFxTigPlaoCre+SpFWf2dp3nsjYqYfvSW7eVqy9WtZpxOxJaNMn4Fwo+U8BVz7u23sugC/gm6JwjhIjVb20dwHhKJd1rjo4U+1Dbf9IdGysNWE6FYi3IKr2HpB/lht2HFb3+X01bLgB0xcTY459HOorAIDinTvrMMlmilV0S+sp9MeYjbcNl+tT71raVf4O2AJqsI1cA5yDLc3lhoNd87l7aB/+3AR3Vvbt0HymuzZJBz5dY5txxw2m53jPIj4Dxj6RMwsJ7fuEESkM0Da26ViOEuK+4zvwt/HhUFeDveKyOZklX7tLHg9m5MhOrnRsRRvGTxi+jIGlyO9NHhKMaK7Cvlp3K9DzTSpyilNFjEzmAE6H6HiI/DfdrmvD365RJU1SE/3R77c/n3GROmka/m9g6dJXteScoA2kitFizjnkMwPdPunZP2OXl8kUGeA3VIIDF03JXTFPz+bjNslmO0u5Y4vuKBy3fVhoxMd0c8hb+okCbkFJSwHVM4qWmLPYarN8I1XxmcVuJWbrjnFo8EAgaFXwge4PoglZsmCo0oDbM7IC0DirCsCiq4Fj+U2mrIA7k/9EVQ/i/3t2NCZ5xDdTyQzhqRx78lJt2Wr2lhx7kImiZ8SiFrIsAjkLvU4P8zE66eOpfgMAHdDDQAX9PA08Twyp1IG8Bk=
   on:
     tags: true
+
+### cache node_modules/ dir between builds on travis
```
rpl: This change is unrelated to the changes needed to extend the flow type checks (and it is applied in its own atomic commit), but I noticed that a good amount of the build time is currently related to the dependency installation, and so caching the `node_modules/` dir cuts a good amount of the build time, and it is especially beneficial for the "nodejs 4" builds that are much more slower than the "nodejs 6" ones.

package.json
```diff
@@ -78,6 +78,7 @@
     "copy-dir": "0.3.0",
     "coveralls": "2.11.12",
     "deepcopy": "0.6.3",
+    "firefox-client": "0.3.0",
```
rpl: `"firefox-client"` is added to the `"devDependencies"` to be able to do an `import type FirefoxClient from 'firefox-client'` across npm versions (it is needed only in the devDependencies because the `import type` statements are removed from the final build).

tests/helpers.js
```diff
@@ -134,7 +134,8 @@ export function makeSureItFails(): Function {
  * assert.equal(fakeProcess.exit.called, true);
  *
  */
-export function fake(original: Object, methods: Object = {}): Object {
+ // $FLOW_IGNORE: fake can return any kind of object and fake a defined set of methods for testing.
+export function fake<T>(original: Object, methods: Object = {}): T {
```
rpl: The above change seems to be needed to convince flow that this `fake` test helper can return any kind of objects
(it helps to be able to use the tricks used in the testing suite without loosing the additional type checking in the actual code)

src/cmd/sign.js
```diff
@@ -13,12 +13,46 @@ import {createLogger} from '../util/logger';
...
+  sourceDir: string,
+  artifactsDir: string,
+  apiKey: string,
+  apiSecret: string,
+  apiUrlPrefix: string,
+  timeout: number,
+};
+
+export type SignCmdExtraParams = {
+  build?: typeof defaultBuilder,
```
rpl: Initially I added explicitly defined types for this kind of helper functions, but it seems that `typeof <function>` does a good job into enforcing the same kind of statical analysis on the passed functions.

.flowconfig
```diff
@@ -5,6 +5,8 @@
 [options]
 module.system=node
 log.file=./artifacts/flow.log
+suppress_comment= \\(.\\|\n\\)*\\$FLOW_FIXME
```
kumar303: What is the first part of this pattern matching? Could we keep it simple and make ``// \$FLOW_FIXME` ?

.travis.yml
kumar303: Good idea. Just a nit but all other comments in this file start with a single hash (`#`).

.travis.yml
```diff
@@ -30,3 +30,8 @@ deploy:
...
+cache:
+  directories:
+  - node_modules
```
kumar303: do you know what happens when a node dependency gets *removed* from `package.json`? Will the next test run still have the package in `node_modules`?

package.json
kumar303: Good catch. This was a flattened deps casualty (it's a sub dep defined in https://github.com/mozilla/node-firefox-connect/blob/master/package.json#L8)

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+  artifactsDir: string,
+};
+
+export type BuildCmdParams = {
+  sourceDir: string,
+  artifactsDir: string,
+  asNeeded?: boolean,
+};
+
+export type BuildCmdExtraParams = {
```
kumar303: this could be called `BuildCmdOptions`

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+  manifestData?: ExtensionManifest,
+  fileFilter?: FileFilter,
```
kumar303: This is not optional but a default will be provided. Does Flow complain when you remove the question mark? We would want Flow to catch any case where `null` was given for `fileFilter`.

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+  onSourceChange?: OnSourceChangeFn,
```
kumar303: also not optional. A default will be provided.

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+  packageCreator?: PackageCreatorFn,
```
kumar303: also not optional. A default will be provided.

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+};
+
+export type FileFilterOptions = {
```
kumar303: I know you placed all of the exported type definitions in one place but I think it would make the code easier to read if each signature type was directly above the function or class that it pertains to. This is because, in most cases, when you're reading a function, you will need to reference its signature type.

kumar303: Any objection to that? I'm personally not too worried about mixing the `type` operator with regular code -- babel takes care of stripping it out.

src/cmd/lint.js
```diff
@@ -5,11 +5,59 @@ import {FileFilter} from './build';
...
 const log = createLogger(__filename);
 
+// Flow types
+
+export type LinterLogLevel = 'debug' | 'fatal';
+export type LinterOutputType = 'text' | 'json';
+
+export type LinterCreatorParams = {
+  config: {
+    logLevel: LinterLogLevel,
```
kumar303: Should this just be `logLevel: 'debug' | 'fatal'`? I don't see `LinterLogLevel` referenced elsewhere.

src/cmd/lint.js
```diff
@@ -5,11 +5,59 @@ import {FileFilter} from './build';
...
+
 export default function lint(
-    {verbose, sourceDir, selfHosted, boring, output,
-     metadata, pretty}: Object,
-    {createLinter=defaultLinterCreator, fileFilter=new FileFilter()}
-    : Object = {}): Promise<void> {
+  {
+    verbose, sourceDir, selfHosted, boring, output,
+    metadata, pretty,
+  }: LintCmdParams,
```
kumar303: I find it unfortunate that you can't declare destructured types inline, thus the need to declare a separate type object and repeat all the keys. Judging from [this thread](https://github.com/facebook/flow/issues/235), I don't think it will happen anytime soon though -- [destructured rebinding](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names) is the culprit!

src/cmd/sign.js
kumar303: yeah, I like the `typeof` reference because it should make refactoring easier too

src/cmd/sign.js
```diff
@@ -13,12 +13,46 @@ import {createLogger} from '../util/logger';
...
+  signAddon?: typeof defaultAddonSigner,
+  preValidatedManifest?: ExtensionManifest,
+};
 
+export type SignCmdResult = {
```
kumar303: Yay, this makes me happy. I don't think we even had this documented in a single place before. Later on, we may want to add documenting comments to things like this since it could be consumed by an external web-ext library user.

src/cmd/build.js
rpl: no objections at all, I agree with you, and it was something that I was thinking a lot of times when I was working on these patches, then I decided to defer a decision to be able to discuss it with you during the review.

Given that it seems that we already agree even before starting to discuss it, I'm going to make this change in the next update of this PR ;-)

src/cmd/lint.js
rpl: yeah, we can (I was a bit unsure if I prefer the explicit type name or just add a union type in the logLevel type definition).

src/firefox/preferences.js
```diff
@@ -1,22 +1,26 @@
...
+export type FirefoxPreferences = {
+  [key: string]: bool | string | number,
+};
+
+// NOTE: 'thuderbird' prefs are not defined on purpose, it is used in the tests to
+// be sure that if flow checks are not running and web-ext is used as a library,
+// an unknown app name raises the expected exception at runtime.
+// Preferences Maps
 
+// $FLOW_IGNORE: suppress property 'thuderbird' Property not found
```
kumar303: Could we just put the `$FLOW_IGNORE` comment on the test code instead of in this module?

src/firefox/preferences.js
```diff
@@ -65,13 +69,13 @@ prefs.common = {
 };
 
 // Prefs specific to Firefox for Android.
-prefs.fennec = {
+const prefsFennec:FirefoxPreferences = {
```
kumar303: nit: space after `prefsFennec:` for consistency

src/firefox/preferences.js
```diff
@@ -65,13 +69,13 @@ prefs.common = {
...
   'browser.console.showInPanel': true,
   'browser.firstrun.show.uidiscovery': false,
 };
 
 // Prefs specific to Firefox for desktop.
-prefs.firefox = {
+const prefsFirefox:FirefoxPreferences = {
```
kumar303: nit: space after `prefsFirefox:` for consistency

src/cmd/lint.js
rpl: yes, it is very unfortunate :-(

anyway, I noticed that "declaring the types of this destructured objects", besides reducing mistakes, forces us to re-look and re-think the type of this "options" parameters (something like the kind of additional thinking that the testing suite forces us to do on the code that we wrote), and I found it an useful exercise.

.travis.yml
rpl: very good point!

I did't tried, but I think that it will behave exactly like an npm install in an previously initialized "node_modules/" dir.

I can see what you are thinking:
if we remove a dependency but there is still code that refers to the missing dependency, then it could happen that the failure is not caught by the travis build... that would be very annoying :-(

It worth to notice that the cache can be removed from the travis web ui, but nevertheless we have to remember to remove it manually when we remove a dependency from the package.json... which seems a perfect recipe to experience the above issue.

it would be probably better if we create a script which detect if any dependency has been removed from package.json and remove the "node_modules/" content in that case before the "npm install" step (e.g. by copying the last cached web-ext package.json in the cache dir itself).

Another potential issue that is coming to my mind is:

- is this cache shared by the nodejs 4 and nodejs 6 sub-jobs?
  can the different behavior of the npm version available in these two version create issues similar to the above?

src/cmd/build.js
rpl: uhm... that's a good point

I tried that and, if I recall correctly, if you explicitly pass null (or another incompatible type) then the issue will be caught by flow as expected. 

Nevertheless it worth a more deep check, there are some scenario that I'd like to check to be sure that they are handled correctly.

I'll take a look at this (and the other similar issue in the other review comments).

src/firefox/index.js
```diff
@@ -17,6 +17,82 @@ import {default as defaultFirefoxConnector, REMOTE_PORT} from './remote';
...
+  PreferencesAppName,
+  PreferencesGetterFn,
+} from './preferences';
+
+import type {ExtensionManifest} from '../util/manifest';
+
+export type FirefoxRunnerParams = {
+  binary?: string,
+  profile?: string,
+  "new-instance"?: boolean,
```
kumar303: we enforce single quotes everywhere else. Hmm, I guess eslint doesn't look at type declarations? Not really a big deal

src/firefox/index.js
```diff
@@ -17,6 +17,82 @@ import {default as defaultFirefoxConnector, REMOTE_PORT} from './remote';
...
+};
+
+export type FirefoxRunnerResults = {
+  process: ChildProcess,
+  binary: string,
+  args: Array<string>,
+};
+
+export type FirefoxRunnerFn =
+  (params: FirefoxRunnerParams) => FirefoxRunnerResults;
```
kumar303: I think the return should be `Promise<FirefoxRunnerResults>`

src/firefox/remote.js
```diff
@@ -4,15 +4,53 @@ import {RemoteTempInstallNotSupported, WebExtError} from '../errors';
...
+};
+
+export type FirefoxRDPResponseAny = Object;
+
+export type FirefoxRDPResponseAddon = {
+  addon: FirefoxRDPAddonActor,
+};
+
+export type FirefoxRDPResponseMaybe =
+  FirefoxRDPResponseError | FirefoxRDPResponseAddon | FirefoxRDPResponseAny;
```
kumar303: It looks like `FirefoxRDPResponseAny` could be replaced with `Object`. However, if the last check is `Object`, does it matter to specify any other object types? It's nice to document the error and add-on response types though.

src/firefox/remote.js
```diff
@@ -151,3 +196,10 @@ export class RemoteFirefox {
       });
   }
 }
+
+// NOTE: this fix an issue with flow and default exports (which currently
+// lose their type signatures) by explicitly declare the default export
+// signature. Reference: https://github.com/facebook/flow/issues/449
+declare function exports( // eslint-disable-line no-unused-vars
+  port: number, extraParams?: ConnectExtraParams
+): Promise<RemoteFirefox>;
```
kumar303: The issue you linked to seems to talk more about `declare module...` syntax. Is there a different issue to link to? Also, I think it would be helpful to move this underneath `connect` so that maybe it will be more obvious when we have to change the signature.

src/util/logger.js
```diff
@@ -2,19 +2,74 @@
...
+export type FATAL = 60;
+
+export type BunyanLogLevel =
+  TRACE | DEBUG | INFO | WARN | ERROR | FATAL;
+
+export type ConsoleStreamParams = {
+  verbose?: boolean,
+};
+
+export type ConsoleFormatParams = {
```
kumar303: I think `BunyanLogEntry` would be a better name

src/util/logger.js
```diff
@@ -2,19 +2,74 @@
...
+
+import type {Writable} from 'stream';
+import type {WriteStream} from 'tty';
+
+export type FakeProcess = {
+  stdout: Writable | WriteStream;
+};
+
+export type ConsoleExtraParams = {
+  localProcess?: FakeProcess,
```
kumar303: Could this be `localProcess? typeof process` ?

src/firefox/index.js
rpl: Actually, this is the signature of this function:

https://github.com/mozilla-jetpack/node-fx-runner/blob/master/lib/run.js#L21

defined in the 'fx-runner' dependency, which returns this object (which corresponds to the type defined as FirefoxRunnerResults) hehe:

https://github.com/mozilla-jetpack/node-fx-runner/blob/master/lib/run.js#L74

Probably we should rename the types in this file so that it is more clear (and/or add an inline couple of inline comments which point to the above lines in the fx-runner sources).

src/firefox/remote.js
rpl: ... that really small bit of haskell that I learned in the past is taking over from inside me :-)

I tried to find a better workaround than creating a type alias of Object, unfortunately this reduces the safety of the type check a bit (a good bit actually :-( ), but at least leave a trace of our intentions and it is a documentation of what the method should return. (basically it is a bit more useful for humans, it makes not a big difference for to the machines ;-)).

src/watcher.js
```diff
@@ -7,9 +7,38 @@ import {FileFilter} from './cmd/build';
 
 const log = createLogger(__filename);
 
+// Flow Types
+
+export type ShouldWatchFn = (filePath: string) => boolean;
+
+export type OnChangeFn = () => any;
```
kumar303: This should probably be `() => void` since the return value is never used.

src/firefox/remote.js
rpl: In the link the workaround seems to be applied into one of those external declaration files, I'm trying to avoid to put the type declarations in a separate file that is magically included in the flow checks, and it seems that we can put declarations for a module in the module itself instead by just skipping the "declare module" container.

I agree, I'm going to move all the type declarations near to where they are used.

tests/helpers.js
kumar303: does `$FLOW_IGNORE` cause it to completely ignore this line? Will these annotations be completely ignored? In other words, does the `<T>` polymorphic arg have any affect?

tests/test-cmd/test.lint.js
```diff
@@ -85,10 +86,15 @@ describe('lint', () => {
   it('passes through linter configuration', () => {
     const {lint, createLinter} = setUp();
     return lint({
+      // $FLOW_IGNORE: wrong type used for testing purpose
```
kumar303: This test should probably be re-written to pass in realistic values. It was asserting that a parameter sent to `lint()` gets passed to `createLinter()`. I guess my main concern was that passing in a default value could have caused a silent failure. In other words, if you pass in opposites of the defaults for each parameter then that should be enough to make the same assertion as before.

kumar303:
First review pass complete! My main suggestions:
* adopting the name `ThingOptions` instead of `ThingExtraParams`
* moving all signature type declarations directly above the function or method they pertain to

src/cmd/lint.js
kumar303: for me, the less variables I have to follow the better. If I'm looking for a type definition and it points to a variable then I have to look in two places to find the definition. Obviously you need to use a variable when the definition is shared by more than one place.

src/firefox/index.js
kumar303: > Actually, this is the signature of this function:
> 
> https://github.com/mozilla-jetpack/node-fx-runner/blob/master/lib/run.js#L21
> 
> defined in the 'fx-runner' dependency, which returns this object (which corresponds to the > type defined as FirefoxRunnerResults) hehe:

but the [first line you linked to](https://github.com/mozilla-jetpack/node-fx-runner/blob/master/lib/run.js#L21), shows that `runFirefox()` returns a promise, no?

The signature to me looks like:

````
export type FirefoxRunnerFn =
  (params: FirefoxRunnerParams) => Promise<FirefoxRunnerResults>;
````

File: src/firefox/remote.js (hunk 21 above)
Comment (kumar303) [inline src/firefox/remote.js]: Yeah, tracing the intention seems useful. It might be worth adding a comment here explaining that it exists more for documentation (otherwise, an astute reader may be tempted to correct it).

File: tests/helpers.js (hunk 4 above)
Comment (rpl) [inline tests/helpers.js]: The above "flow ignore" comment is needed to stop flow from complaining about type errors in the body of the `fake` function itself, but it doesn't seem to affect the checks of the polymorphic type argument, which is what convince flow that `fake` is actually what we use in tests to create any type we want to "fake".

File: src/cmd/lint.js (hunk 13 above)
Comment (rpl) [inline src/cmd/lint.js]: :+1:

File: src/firefox/preferences.js (hunk 17 above)
Comment (pdehaan) [inline src/firefox/preferences.js]: I'm not [even remotely] a Flow expert, but we may be able to use [**eslint-plugin-flowtype**](https://www.npmjs.com/package/eslint-plugin-flowtype) to enforce this. Specifically the [`flowtype/space-after-type-colon`](https://www.npmjs.com/package/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-space-after-type-colon) rule.

File: src/firefox/index.js (hunk 20 above)
Comment (rpl) [inline src/firefox/index.js]: ouch, you are definitely right! 
Probably I got the wrong assumption from the jsdoc comment above the function:

https://github.com/mozilla-jetpack/node-fx-runner/blob/master/lib/run.js#L19

thanks for triple-checking it!

File: src/firefox/preferences.js (hunk 17 above)
Comment (rpl) [inline src/firefox/preferences.js]: @pdehaan it seems a great idea, especially now that we are expanding the flow type definitions, I'll try it asap.

Comment (pdehaan) [inline src/firefox/preferences.js]: Adding the **eslint-plugin-flowtype** could definitely be a separate PR, but I'm a big fan of linting.
😻

File: src/firefox/index.js (hunk 20 above)
Comment (kumar303) [inline src/firefox/index.js]: I guess Flow never showed an error for your declaration because returning a value (not a promise) midway through the promise chain will just pass along that value transparently.

Comment (rpl) [inline src/firefox/index.js]: @kumar303 yeah, that was a thing that I definitely wanted to look before fixing it:

basically flow couldn't catch the above mistake in the type signature because, in the only place where we use it, we are returning its return value in the middle of a Promise chain, where
both the signatures would become a "Promise<FirefoxRunnerResults>" in any case (with both the right and the wrong type signature), on the contrary flow would be able to detect the wrong signature if we tried to use the return value directly.

File: src/cmd/build.js (hunk 9 above)
Comment (rpl) [inline src/cmd/build.js]: @kumar303 checked this again and it is working as expected, e.g.:

- if fileFilter is set to `null` (e.g. in test.build.js at line 85), flow raises a type error (because null in not compatible with the FileFilter type)
- if fileFilter is set to `undefined`, flow does not raise any type error (because the default value will be used, and it is a valid FileFilter for sure)

File: src/watcher.js (hunk 25 above)
Comment (rpl) [inline src/watcher.js]: unfortunately if we change this into `() => void`, then we need to remove the `return` from the `onChange` function defined in "cmd/build.js" and then "test.build.js" fails because it is currently using the return value of the `onChange` function.

File: .eslintrc [hunk 27]
```diff
@@ -75,6 +83,10 @@
     "space-infix-ops": 0,
     "strict": [2, "never"],
     "valid-typeof": 2,
-    "wrap-iife": [2, "inside"]
+    "wrap-iife": [2, "inside"],
+
+    "flowtype/define-flow-type": 1,
+    "flowtype/use-flow-type": 1,
+    "flowtype/space-after-type-colon": [1, "always"]
```
kumar303: cool

.travis.yml
```diff
@@ -30,3 +30,8 @@ deploy:
     secure: gMZR5u6VxrXFxTigPlaoCre+SpFWf2dp3nsjYqYfvSW7eVqy9WtZpxOxJaNMn4Fwo+U8BVz7u23sugC/gm6JwjhIjVb20dwHhKJd1rjo4U+1Dbf9IdGysNWE6FYi3IKr2HpB/lht2HFb3+X01bLgB0xcTY459HOorAIDinTvrMMlmilV0S+sp9MeYjbcNl+tT71raVf4O2AJqsI1cA5yDLc3lhoNd87l7aB/+3AR3Vvbt0HymuzZJBz5dY5txxw2m53jPIj4Dxj6RMwsJ7fuEESkM0Da26ViOEuK+4zvwt/HhUFeDveKyOZklX7tLHg9m5MhOrnRsRRvGTxi+jIGlyO9NHhKMaK7Cvlp3K9DzTSpyilNFjEzmAE6H6HiI/DfdrmvD365RJU1SE/3R77c/n3GROmka/m9g6dJXteScoA2kitFizjnkMwPdPunZP2OXl8kUGeA3VIIDF03JXTFPz+bjNslmO0u5Y4vuKBy3fVhoxMd0c8hb+okCbkFJSwHVM4qWmLPYarN8I1XxmcVuJWbrjnFo8EAgaFXwge4PoglZsmCo0oDbM7IC0DirCsCiq4Fj+U2mrIA7k/9EVQ/i/3t2NCZ5xDdTyQzhqRx78lJt2Wr2lhx7kImiZ8SiFrIsAjkLvU4P8zE66eOpfgMAHdDDQAX9PA08Twyp1IG8Bk=
   on:
     tags: true
+
+# cache node_modules/ dir between builds on travis
+cache:
+  directories:
+  - node_modules
```
kumar303: reminder to remove this before merge

src/cmd/build.js
```diff
@@ -12,9 +12,47 @@ import {createLogger} from '../util/logger';
...
+import type {ExtensionManifest} from '../util/manifest';
+
+export type PackageCreatorFn =
+    (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
+
+export type ExtensionBuildResult = {
+  extensionPath: string,
+};
+
+export type PackageCreatorParams = {
```
kumar303: were you going to move all of these? I think `BuildCmdParams` should go right above the `build()` function, etc.

src/cmd/run.js
```diff
@@ -111,13 +182,14 @@ export default function run(
...
-    .then((runner) => {
-      return runner.getProfile().then((profile) => {
+    .then((runner: ExtensionRunner) => {
+      return runner.getProfile().then((profile: FirefoxProfile) => {
         return {runner, profile};
       });
     })
     .then((config) => {
-      const {runner, profile} = config;
+      const runner: ExtensionRunner = config.runner;
```
kumar303: are you sure this type hint is necessary? I thought Flow would figure it out because it would be tracing the `runner` instance from https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/cmd/run.js#L177

src/cmd/run.js
```diff
@@ -111,13 +182,14 @@ export default function run(
...
+      const profile: FirefoxProfile = config.profile;
```
kumar303: I also would think this type is unnecessary because it traces the return value argument from https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/cmd/run.js#L186

src/cmd/run.js
```diff
@@ -9,15 +9,60 @@ import {createLogger} from '../util/logger';
...
+  FirefoxProcess,
+} from '../firefox/index';
+import type {
+  FirefoxConnectorFn, RemoteFirefox,
+  FirefoxRDPResponseAddon,
+} from '../firefox/remote';
+import type {ExtensionManifest} from '../util/manifest';
 
-const log = createLogger(__filename);
+export type WatcherCreatorParams = {
```
kumar303: I would prefer to see this directly above the [defaultWatcherCreator()](https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/cmd/run.js#L60) definition.

src/cmd/run.js
```diff
@@ -9,15 +9,60 @@ import {createLogger} from '../util/logger';
...
+  addonId: string,
+  client: RemoteFirefox,
+  sourceDir: string,
+  artifactsDir: string,
+  onSourceChange?: OnSourceChangeFn,
+};
+
+export type ReloadStrategyParams = {
```
kumar303: I'd prefer to see this directly above the [defaultReloadStrategy()](https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/cmd/run.js#L79) function.

src/cmd/run.js
```diff
@@ -9,15 +9,60 @@ import {createLogger} from '../util/logger';
...
+  artifactsDir: string,
+};
+
+export type WatcherCreatorFn = (params: WatcherCreatorParams) => Watchpack;
 
+export type ReloadStrategyOptions = {
+  createWatcher?: WatcherCreatorFn,
+};
+
+export type CreateFirefoxClientParams = {
```
kumar303: same thing here

src/cmd/run.js
```diff
@@ -168,7 +242,12 @@ export default function run(
         'of Firefox (you need Firefox 49 or higher). For older Firefox ' +
         'versions, use --pre-install');
     }))
-    .then(({firefox, profile, client, addonId}) => {
+    .then((config) => {
+      const firefox: FirefoxProcess = config.firefox;
+      const profile: FirefoxProfile = config.profile;
+      const client: RemoteFirefox = config.firefox;
+      const addonId: string = config.addonId;
```
kumar303: all these type hints makes the code highly verbose. If they aren't necessary, I'd like to avoid them. If they are necessary, maybe we need to refactor this code to pass around a class instance instead (in a later patch).

src/cmd/sign.js
```diff
@@ -13,12 +13,46 @@ import {createLogger} from '../util/logger';
...
+  success: boolean,
+  id: string,
+  downloadedFiles: Array<string>,
+};
+
+
+// Module internals & exports
```
kumar303: in this case, we do need to put type declarations below this line but elsewhere, I think there will not be a clean delineation. Maybe this comment marker should be removed everywhere? Just a nit.

src/cmd/sign.js
```diff
@@ -13,12 +13,46 @@ import {createLogger} from '../util/logger';
...
-export default function sign(
-    {verbose, sourceDir, artifactsDir, apiKey, apiSecret,
-     apiUrlPrefix, id, timeout}: Object,
-    {build=defaultBuilder, signAddon=defaultAddonSigner,
-     preValidatedManifest=null}: Object = {}): Promise<Object> {
+// Flow Types
+
+import type {ExtensionManifest} from  '../util/manifest';
+
+export type SignCmdParams = {
```
kumar303: nit: could we call these `SignParams` and `SignOptions`? The `sign()` function could be used by a library -- i.e. not as a command.

src/cmd/build.js
rpl: yep, I forgot to move it :-(  
I'm going to push a fix after this round of review comment.

src/firefox/index.js
```diff
@@ -17,6 +17,53 @@ import {default as defaultFirefoxConnector, REMOTE_PORT} from './remote';
...
+import type {
+  PreferencesAppName,
+  PreferencesGetterFn,
+} from './preferences';
+
+import type {ExtensionManifest} from '../util/manifest';
+
+// Declare the needed 'fx-runner' module flow types.
+
+export type FirefoxRunnerParams = {
```
kumar303: I'd prefer to see `FirefoxRunnerParams`, `FirefoxRunnerResults`, and `FirefoxRunnerFn` grouped above the [run()](https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/firefox/index.js#L112) command since they only applies to `FirefoxRunOptions`.

src/firefox/index.js
```diff
@@ -134,15 +210,19 @@ export function configureProfile(
...
 /*
  * Creates a new temporary profile and resolves with the profile object.
  *
  * The profile will be deleted when the system process exits.
  */
 export function createProfile(
-    {app, configureThisProfile=configureProfile}: Object = {})
-    : Promise<FirefoxProfile> {
-
+  {app, configureThisProfile=configureProfile}: CreateProfileParams = {}
```
kumar303: heh, I remember now when naming `configureThisProfile` that `defaultProfileConfigurator` didn't sound right :)

src/firefox/remote.js
```diff
@@ -4,32 +4,58 @@ import {RemoteTempInstallNotSupported, WebExtError} from '../errors';
...
+  error: {
+    message: string,
+  },
+};
+
+export type FirefoxRDPResponseAddon = {
+  addon: FirefoxRDPAddonActor,
+};
+
+// NOTE: this type aliases Object to catch any other possible reponse.
```
kumar303: typo: reponse

src/firefox/remote.js
```diff
@@ -4,32 +4,58 @@ import {RemoteTempInstallNotSupported, WebExtError} from '../errors';
...
+
+
+// Flow Types
+
+import type FirefoxClient from 'firefox-client';
+
+export type FirefoxConnectorFn =
+  (port?: number) => Promise<FirefoxClient>;
+
+export type ConnectOptions = {
```
kumar303: I would prefer to see this above the [connect()](https://github.com/rpl/web-ext/blob/d5083f42002b2dcb5e72f8c886997d0035c1c0c7/src/firefox/remote.js#L196) function

src/firefox/remote.js
```diff
@@ -151,3 +184,23 @@ export class RemoteFirefox {
       });
   }
 }
+
+
+// NOTE: this fix an issue with flow and default exports (which currently
```
kumar303: nit: this fixes an issue... (present tense)

src/firefox/remote.js
```diff
@@ -151,3 +184,23 @@ export class RemoteFirefox {
...
+// lose their type signatures) by explicitly declare the default export
```
kumar303: nit: by explicitly declaring the...

tests/test-cmd/test.run.js
```diff
@@ -281,6 +281,7 @@ describe('run', () => {
       const args = {
         addonId: 'some-addon@test-suite',
         client,
+        // $FLOW_IGNORE: fake a Firefox ChildProcess using an EventEmitter for testing reasons.
```
kumar303: I think we could solve this problem by switching:

````
import type {ChildProcess} from 'child_process';
````

to an [interface](https://flowtype.org/docs/quick-reference.html#interfaces):

````
interface ChildProcess {
  on: () => ...,
  stderr: {
    on: () => ...,
  },
  stdout: {
    on: () => ....,
  }
}
````

It's a little nicer to simply import the type directly from `child_process` but using interfaces will give us a lot of powerful flexibility in testing.

File: tests/test-util/test.logger.js [hunk 45]
```diff
@@ -65,36 +72,42 @@ describe('logger', () => {
     it('does not log debug packets unless verbose', () => {
       const log = new ConsoleStream({verbose: false});
       const localProcess = fakeProcess();
+      // $FLOW_IGNORE: fake process for testing reasons.
```
kumar303: could this also be solved with an [interface](https://flowtype.org/docs/quick-reference.html#interfaces)?

kumar303:
Looking good! I'm a little worried about the ignored test code. I am wondering if [interfaces](https://flowtype.org/docs/quick-reference.html#interfaces) will help.

src/cmd/sign.js
```diff
@@ -117,6 +155,8 @@ export function getIdFromSourceDir(sourceDir: string): Promise<string> {
 }
 
 
+// getIdFromSourceDir helper implementation.
```
kumar303: This comment doesn't seem right. I don't think these marker comments are necessary when there are no proceeding type definitions.

tests/test-cmd/test.run.js
kumar303: after discussing on IRC, we think an interface would be too verbose to cover this case

tests/test-firefox/test.remote.js
```diff
@@ -172,6 +172,7 @@ describe('firefox.remote', () => {
         const stubResponse = {requestTypes: ['reload']};
 
         const conn = makeInstance();
+        // $FLOW_IGNORE: override class method for testing reasons.
```
kumar303: I think when designing this type of thing in the future, we should not override properties but instead use a class instance variable to allow injection (or something like that).

kumar303:
Looks great, just some minor fixup suggestions.

src/cmd/sign.js
rpl: The one below is wrong for sure (it refers to the wrong function name).

I added them mostly to separate this helpers from the above group more clearly, but I agree that using just the "2 blank line" as a separator can be more then enough.

I'm going to remove these comments (the ones that are before an helper without any type defined)

rpl:
@kumar303 travis cache chage removed (deferred to a follow up), minor tweaks to the inline comments applied and added a new test for the "missing addonId in installTemporaryAddon result" scenario to recover (which makes coveralls happy again ;-))

NOTE: I opted to raise an Error instance instead of a WebExtError, mostly because it is an unexpected behavior that should never happen and if it happens it is better if it doesn't look as something that we expect to happen, and, even more important, it should produce a full stack trace in the console (for an easier diagnosis).

If the changes describe above are good, this is now ready to be squashed.

kumar303:
> I opted to raise an Error instance instead of a WebExtError

I was hoping to make all exceptions raised from within `web-ext` subclasses of `WebExtError`. I think this could be helpful for anyone who is trying to use `web-ext` as a library. 

My thoughts on hiding stack traces (https://github.com/mozilla/web-ext/issues/310) would only apply to *user-fixable* exceptions. For those cases, we would make and raise a new custom error, something like `WebExtUsageError`. The base error, `WebExtError`, would still always show stack traces.

rpl:
> I was hoping to make all exceptions raised from within web-ext subclasses of WebExtError. I think this could be helpful for anyone who is trying to use web-ext as a library. 

I totally agree, thanks for the additional details.

I just pushed a small change that raises the `WebExtError` (and added an assertion that the error raised is a WebExtError in the test case as well ;-)).

PR #434, opened 2016-08-16
fix: disabled caching during npm run lint

### LINKED ISSUE #339: `npm run lint` should not use file caching
The day to day development command, `npm start`, needs to use file caching for linting so that the watch-retest loop is fast. The `npm run lint` command uses the same caching -- we should prevent it from using any caching at all. This direct linting command should allow you to circumvent caching. 

Currently, the only way to escape caching is to `rm -fr node_modules/grunt-newer/.cache/*`. Sometimes the cache gets outdated in odd scenarios.

shubheksha:
Fixes #339 
The `.cache` directory is no longer created while running `npm run lint`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -35,7 +35,7 @@ module.exports = function(grunt) {
   grunt.registerTask('test', [
     'build-tests',
     'mochaTest',
-    'lint',
+    'newer:eslint',
     'flowbin:check',
     'check-for-smoke',
   ]);
@@ -53,7 +53,7 @@ module.exports = function(grunt) {
       grunt.log.writeln('task skipped because this version of Node is too old');
     } else {
       grunt.task.run([
-        'newer:eslint',
+        'eslint',
       ]);
     }
   });
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,7 +1,6 @@
 /*eslint prefer-template: 0*/
 var path = require('path');
 var spawn = require('child_process').spawn;
-var semver = require('semver');
 
 module.exports = function(grunt) {
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -35,7 +35,7 @@ module.exports = function(grunt) {
   grunt.registerTask('test', [
     'build-tests',
     'mochaTest',
-    'newer:eslint',
+    'eslint',
     'flowbin:check',
     'check-for-smoke',
   ]);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -48,9 +48,6 @@ module.exports = function(grunt) {
   grunt.registerTask('lint', 'checks for syntax errors', function() {
     if (process.env.SKIP_LINT) {
       grunt.log.writeln('lint task skipped because of $SKIP_LINT');
-    } else if (semver.satisfies(process.version, '< 4.0.0')) {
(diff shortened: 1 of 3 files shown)


kumar303:
This is off to a great start, thanks. I verified that the `npm run lint` command is working without the cache now 👍  

I went back and re-read the issue and remembered the main thing we need is for [npm start](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/package.json#L16) to use `newer:eslint`. The `npm start` command is the one we use in day to day development so it needs to be very fast. It will be executed every time you change a file so it needs to use the eslint cache for efficiency. 

The [npm start](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/package.json#L16) command is an alias for [grunt develop](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/Gruntfile.js#L43) which calls the subtask [watch:develop](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/Gruntfile.js#L45). This task is configured to [call grunt lint](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/tasks/watch.js#L19) but it should instead call `newer:eslint`.

As for [grunt test](https://github.com/shubheksha/web-ext/blob/b78621c16a7f258b32b27c1a975e071f7244211a/Gruntfile.js#L35), this command should not use the eslint cache (I was mistaken about that). This command runs once and exits so it's not important to use the cache.

shubheksha:
Thanks for the feedback! I will make the specified changes and open a new PR.

Gruntfile.js
```diff
@@ -35,7 +35,7 @@ module.exports = function(grunt) {
   grunt.registerTask('test', [
     'build-tests',
     'mochaTest',
-    'lint',
+    'eslint',
```
kumar303: this can continue to call the `lint` task, it's the same thing

Gruntfile.js
```diff
@@ -53,7 +53,7 @@ module.exports = function(grunt) {
       grunt.log.writeln('task skipped because this version of Node is too old');
```
kumar303: ooh, I just noticed that this function needs updating while you're working on it. We [no longer support](https://github.com/mozilla/web-ext/blob/master/.travis.yml#L3-L5) this old Node version so the if branch that uses `semver` can be deleted.

shubheksha: This seems to be breaking the builds. Did I do something wrong?

Gruntfile.js
```diff
@@ -48,12 +48,9 @@ module.exports = function(grunt) {
   grunt.registerTask('lint', 'checks for syntax errors', function() {
     if (process.env.SKIP_LINT) {
       grunt.log.writeln('lint task skipped because of $SKIP_LINT');
-    } else if (semver.satisfies(process.version, '< 4.0.0')) {
```
kumar303: At the top of the file, you can now remove the line that requires the `semver` module. The lint command in the test suite caught this error for you :) Also, you can now remove `semver` from `package.json` because it's not used anywhere else.

kumar303:
Thanks for all the fix-ups.

Issue #377, opened 2016-07-19
Write tests that more closely mimic the CLI


kumar303:
The current `web-ext` tests invoke commands directly as functions and pass in object properties for command line options.

For example, a command like `web-ext build --verbose` might be tested like this:

````javascript
import build from '../../src/cmd/build';

build({verbose: true}).then(() => {
  //...
});
````

We should test commands at a higher level so that the yargs configuration of the command line options are covered better. For example, the tests should look more like:

````javascript
import exec from './helpers';

exec('build', ['--verbose']).then(() => {
  //...
});
````

The tricky part is that some tests need to inject mock dependencies. For example:

````javascript
build({verbose: true}, {createXPI=myFakeXPICreator});
````

Implementation idea:
* Create a wrapper that runs the program object with [{shouldExitProgram: false}](https://github.com/mozilla/web-ext/blob/30c0b4edf132ff0a3c6df5d580940964e4eb09ee/tests/test.program.js#L24)
* To facilitate dependency injection, call `main()` with [defaultCommands](https://github.com/mozilla/web-ext/blob/30c0b4edf132ff0a3c6df5d580940964e4eb09ee/src/program.js#L129) but wrap the commands. For example, you might do `{build: injectDep(defaultCommands.build, {createXpi})}` so that the command callable is still the real callable, just with additional arguments.

PR #440, opened 2016-08-17
fix: --firefox-binary to --firefox

### LINKED ISSUE #116: Change --firefox-binary to --firefox

shubheksha:
Fixes #116 
For some reason, changing `firefox` to `firefoxApp` even on just [this](https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L167) line makes the tests fail.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -62,7 +62,7 @@ export function defaultWatcherCreator(
 
 export type ReloadStrategyParams = {
   addonId: string,
-  firefox: FirefoxProcess,
+  firefoxApp: FirefoxProcess,
   client: RemoteFirefox,
   profile: FirefoxProfile,
   sourceDir: string,
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -62,7 +62,7 @@ export function defaultWatcherCreator(
 
 export type ReloadStrategyParams = {
   addonId: string,
-  firefoxApp: FirefoxProcess,
+  firefoxProcess: FirefoxProcess,
   client: RemoteFirefox,
   profile: FirefoxProfile,
   sourceDir: string,
@@ -75,7 +75,7 @@ export type ReloadStrategyOptions = {
 
 export function defaultReloadStrategy(
   {
-    addonId, firefoxApp, client, profile, sourceDir, artifactsDir,
+    addonId, firefoxProcess, client, profile, sourceDir, artifactsDir,
   }: ReloadStrategyParams,
   {
     createWatcher=defaultWatcherCreator,
@@ -83,7 +83,7 @@ export function defaultReloadStrategy(
 ): void {
   let watcher: Watchpack;
 
-  firefoxApp.on('close', () => {
+  firefoxProcess.on('close', () => {
     client.disconnect();
     watcher.close();
   });
@@ -261,7 +261,7 @@ export default function run(
         log.debug(
           `Reloading extension when the source changes; id=${addonId}`);
         reloadStrategy({
-          firefoxApp: runningFirefox,
+          firefoxProcess: runningFirefox,
           profile, client, sourceDir, artifactsDir, addonId,
         });
       }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -75,7 +75,7 @@ export type ReloadStrategyOptions = {
 
 export function defaultReloadStrategy(
   {
-    addonId, firefox, client, profile, sourceDir, artifactsDir,
+    addonId, firefoxApp, client, profile, sourceDir, artifactsDir,
   }: ReloadStrategyParams,
   {
     createWatcher=defaultWatcherCreator,
@@ -83,7 +83,7 @@ export function defaultReloadStrategy(
 ): void {
   let watcher: Watchpack;
 
-  firefox.on('close', () => {
+  firefoxApp.on('close', () => {
     client.disconnect();
     watcher.close();
   });
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -146,25 +146,25 @@ export function defaultFirefoxClient(
 export type CmdRunParams = {
   sourceDir: string,
   artifactsDir: string,
-  firefoxBinary: string,
+  firefox: string,
   firefoxProfile: string,
   preInstall: boolean,
   noReload: boolean,
 };
 
 export type CmdRunOptions = {
-  firefox: typeof defaultFirefox,
+  firefoxApp: typeof defaultFirefox,
   firefoxClient: typeof defaultFirefoxClient,
   reloadStrategy: typeof defaultReloadStrategy,
 };
 
 export default function run(
   {
-    sourceDir, artifactsDir, firefoxBinary, firefoxProfile,
+    sourceDir, artifactsDir, firefox, firefoxProfile,
     preInstall=false, noReload=false,
   }: CmdRunParams,
   {
-    firefox=defaultFirefox,
+    firefoxApp=defaultFirefox,
     firefoxClient=defaultFirefoxClient,
     reloadStrategy=defaultReloadStrategy,
   }: CmdRunOptions = {}): Promise<Object> {
@@ -190,8 +190,8 @@ export default function run(
     .then((manifestData) => {
       return new ExtensionRunner({
         sourceDir,
+        firefoxApp,
         firefox,
-        firefoxBinary,
         manifestData,
         profilePath: firefoxProfile,
       });
@@ -261,11 +261,11 @@ export default function run(
         log.debug(
           `Reloading extension when the source changes; id=${addonId}`);
         reloadStrategy({
-          firefox: runningFirefox,
+          firefoxApp: runningFirefox,
           profile, client, sourceDir, artifactsDir, addonId,
         });
       }
-      return firefox;
+      return firefoxApp;
     });
 }
 
@@ -276,39 +276,39 @@ export type ExtensionRunnerParams = {
   sourceDir: string,
   manifestData: ExtensionManifest,
   profilePath: string,
-  firefox: typeof defaultFirefox,
-  firefoxBinary: string,
+  firefoxApp: typeof defaultFirefox,
+  firefox: string,
 };
 
 export class ExtensionRunner {
   sourceDir: string;
   manifestData: ExtensionManifest;
   profilePath: string;
-  firefox: typeof defaultFirefox;
-  firefoxBinary: string;
+  firefoxApp: typeof defaultFirefox;
+  firefox: string;
 
   constructor(
     {
-      firefox, sourceDir, manifestData,
-      profilePath, firefoxBinary,
+      firefoxApp, sourceDir, manifestData,
+      profilePath, firefox,
     }: ExtensionRunnerParams
   ) {
     this.sourceDir = sourceDir;
     this.manifestData = manifestData;
     this.profilePath = profilePath;
+    this.firefoxApp = firefoxApp;
     this.firefox = firefox;
-    this.firefoxBinary = firefoxBinary;
   }
 
   getProfile(): Promise<FirefoxProfile> {
-    const {firefox, profilePath} = this;
+    const {firefoxApp, profilePath} = this;
     return new Promise((resolve) => {
       if (profilePath) {
         log.debug(`Copying Firefox profile from ${profilePath}`);
-        resolve(firefox.copyProfile(profilePath));
+        resolve(firefoxApp.copyProfile(profilePath));
       } else {
         log.debug('Creating new Firefox profile');
-        resolve(firefox.createProfile());
+        resolve(firefoxApp.createProfile());
       }
     });
   }
@@ -320,8 +320,8 @@ export class ExtensionRunner {
   }
 
   installAsProxy(profile: FirefoxProfile): Promise<string|void> {
-    const {firefox, sourceDir, manifestData} = this;
-    return firefox.installExtension(
+    const {firefoxApp, sourceDir, manifestData} = this;
+    return firefoxApp.installExtension(
       {
         manifestData,
         asProxy: true,
@@ -332,7 +332,7 @@ export class ExtensionRunner {
   }
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
-    const {firefox, firefoxBinary} = this;
-    return firefox.run(profile, {firefoxBinary});
+    const {firefoxApp, firefox} = this;
+    return firefoxApp.run(profile, {firefox});
   }
 }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -1,5 +1,5 @@
 /* @flow */
-import * as defaultFirefox from '../firefox';
+import * as defaultFirefoxApp from '../firefox';
 import defaultFirefoxConnector from '../firefox/remote';
 import {
   onlyInstancesOf, onlyErrorsWithCode, RemoteTempInstallNotSupported,
@@ -153,7 +153,7 @@ export type CmdRunParams = {
 };
 
 export type CmdRunOptions = {
-  firefoxApp: typeof defaultFirefox,
+  firefoxApp: typeof defaultFirefoxApp,
   firefoxClient: typeof defaultFirefoxClient,
   reloadStrategy: typeof defaultReloadStrategy,
 };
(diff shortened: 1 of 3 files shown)


kumar303:
The patch you have here will cause [firefoxBinary](https://github.com/shubheksha/web-ext/blob/f614198eaaf8f91431da3b1b76f8c11955ffdf48/src/cmd/run.js#L163) to be undefined since yargs will no longer pass it in as such. Yargs would be passing in `firefox` but that is ignored since `run()` does not reference it in the destructor.

shubheksha:
What you say makes perfect sense, but how come all the test are passing in this case & it isn't being caught?

kumar303:
Ha, yeah, that is a very good question. This is actually a problem with the test suite that I plan to fix in https://github.com/mozilla/web-ext/issues/377

kumar303:
If you update the PR with the [changes you mentioned here](https://github.com/mozilla/web-ext/pull/439#issuecomment-240807838) then I could take a look.

shubheksha:
I tried a lot of different combinations of changes, actually. I tried changing all occurences inside the `run` function, then I tried to trace the what was getting undefined using the error messages, so I am kinda unsure of what to include in the PR. What do you suggest?

kumar303:
I think the way to fix it is to only change the two properties in the `run()` function and nothing else. You would change the `firefoxBinary` parameter to `firefox` and the `firefox` option to `firefoxApp`.  Renaming all the test code that invokes the `run()` function can be confusing. You can use the Flow type checker (which is part of the test suite invoked by `npm start`) to guide you.

I tried this quickly in your branch and this got me past most of the errors but didn't completely fix it. Does this help? Here is my diff:

````diff
index 0875c56..e8cfe95 100644
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -146,25 +146,25 @@ export function defaultFirefoxClient(
 export type CmdRunParams = {
   sourceDir: string,
   artifactsDir: string,
-  firefoxBinary: string,
+  firefox: string,
   firefoxProfile: string,
   preInstall: boolean,
   noReload: boolean,
 };
 
 export type CmdRunOptions = {
-  firefox: typeof defaultFirefox,
+  firefoxApp: typeof defaultFirefox,
   firefoxClient: typeof defaultFirefoxClient,
   reloadStrategy: typeof defaultReloadStrategy,
 };
 
 export default function run(
   {
-    sourceDir, artifactsDir, firefoxBinary, firefoxProfile,
+    sourceDir, artifactsDir, firefox, firefoxProfile,
     preInstall=false, noReload=false,
   }: CmdRunParams,
   {
-    firefox=defaultFirefox,
+    firefoxApp=defaultFirefox,
     firefoxClient=defaultFirefoxClient,
     reloadStrategy=defaultReloadStrategy,
   }: CmdRunOptions = {}): Promise<Object> {
@@ -190,8 +190,8 @@ export default function run(
     .then((manifestData) => {
       return new ExtensionRunner({
         sourceDir,
-        firefox,
-        firefoxBinary,
+        firefox: firefoxApp,
+        firefoxBinary: firefox,
         manifestData,
         profilePath: firefoxProfile,
       });
@@ -265,7 +265,7 @@ export default function run(
           profile, client, sourceDir, artifactsDir, addonId,
         });
       }
-      return firefox;
+      return firefoxApp;
     });
 }
 
diff --git a/tests/test-cmd/test.run.js b/tests/test-cmd/test.run.js
index 0dc2295..25e4a75 100644
--- a/tests/test-cmd/test.run.js
+++ b/tests/test-cmd/test.run.js
@@ -37,7 +37,7 @@ describe('run', () => {
       noReload: true,
     };
     let options = {
-      firefox: getFakeFirefox(),
+      firefoxApp: getFakeFirefox(),
       firefoxClient: sinon.spy(() => {
         return Promise.resolve(fake(RemoteFirefox.prototype, {
           installTemporaryAddon: () =>
@@ -77,7 +77,7 @@ describe('run', () => {
     let profile = {};
 
     const cmd = prepareRun();
-    const {firefox} = cmd.options;
     const firefoxClient = fake(RemoteFirefox.prototype, {
       installTemporaryAddon: () => Promise.resolve(tempInstallResult),
     });
@@ -91,8 +91,8 @@ describe('run', () => {
       assert.equal(install.called, true);
       assert.equal(install.firstCall.args[0], cmd.argv.sourceDir);
 
-      assert.equal(firefox.run.called, true);
-      assert.deepEqual(firefox.run.firstCall.args[0], profile);
+      assert.deepEqual(firefoxApp.run.firstCall.args[0], profile);
     });
   });
 
@@ -114,26 +114,25 @@ describe('run', () => {
   });
 
   it('passes a custom Firefox binary when specified', () => {
-    const firefoxBinary = '/pretend/path/to/Firefox/firefox-bin';
+    const firefox = '/pretend/path/to/Firefox/firefox-bin';
     const cmd = prepareRun();
-    const {firefox} = cmd.options;
 
-    return cmd.run({firefoxBinary}).then(() => {
-      assert.equal(firefox.run.called, true);
-      assert.equal(firefox.run.firstCall.args[1].firefoxBinary,
-                   firefoxBinary);
+    return cmd.run({firefoxApp}).then(() => {
+      assert.equal(firefoxApp.run.firstCall.args[1].firefox, firefox);
     });
   });
 
   it('passes a custom Firefox profile when specified', () => {
     const firefoxProfile = '/pretend/path/to/firefox/profile';
     const cmd = prepareRun();
-    const {firefox} = cmd.options;
 
     return cmd.run({firefoxProfile}).then(() => {
-      assert.equal(firefox.createProfile.called, false);
-      assert.equal(firefox.copyProfile.called, true);
-      assert.equal(firefox.copyProfile.firstCall.args[0],
+      assert.equal(firefoxApp.createProfile.called, false);
+      assert.equal(firefoxApp.copyProfile.called, true);
+      assert.equal(firefoxApp.copyProfile.firstCall.args[0],
                    firefoxProfile);
     });
   });
````

**Comment by shubheksha** [comment]
I didn't change anything in the tests, but that doesn't seem to make a difference, I get the same set of errors. I figured that the issue is mainly with this line - [createProfile()](https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L311). This is what is causing 7/10 tests to fail. Despite changing the name of the object to `firefoxApp`, it is `undefined`.

**Comment by shubheksha** [comment]
I fixed everything & created a PR, but the `flowbin:check` tests don't properly run on my system, hence I couldn't catch the errors which caused the builds to fail. I get an error, the details of which can be found in the screenshot. 
image
Why does this happen?

File: src/cmd/run.js [hunk 1]
```diff
@@ -62,7 +62,7 @@ export function defaultWatcherCreator(
 
 export type ReloadStrategyParams = {
   addonId: string,
-  firefox: FirefoxProcess,
+  firefoxApp: FirefoxProcess,
```
rpl: @shubheksha It is probably better to rename this to `firefoxProcess` (because it is of type [`FirefoxProcess`](https://github.com/mozilla/web-ext/blob/18f38c802ab8df05bb04df7ce504463042170687/src/firefox/index.js#L91), which is a type alias of [`ChildProcess`](https://nodejs.org/api/child_process.html#child_process_class_childprocess)), because it is not the same type of the other properties that you are rename to `firefoxApp` in this patch (in this patch most of the times it is associated to the helper module defined as ["src/firefox/index.js"](https://github.com/mozilla/web-ext/blob/18f38c802ab8df05bb04df7ce504463042170687/src/cmd/run.js#L2), which is a collection of helper functions used to create/copy a Firefox profile, install an extension into the profile, run a firefox instance etc.) and it is going to be helpful to avoid confusion in the future.

tests/test-cmd/test.run.js
```diff
@@ -308,7 +308,7 @@ describe('run', () => {
         addonId: 'some-addon@test-suite',
         client,
         // $FLOW_IGNORE: fake a Firefox ChildProcess using an EventEmitter for testing reasons.
-        firefox: new EventEmitter(),
+        firefoxApp: new EventEmitter(),
```
rpl: This is an EventEmitter used as a fake ChildProcess for testing reasons, it should be renamed to `firefoxProcess`, for the same reasons detailed in the above comment.

tests/test-cmd/test.run.js
```diff
@@ -330,9 +330,9 @@ describe('run', () => {
     }
 
     it('cleans up connections when firefox closes', () => {
-      const {firefox, client, watcher, reloadStrategy} = prepare();
+      const {firefoxApp, client, watcher, reloadStrategy} = prepare();
       reloadStrategy();
-      firefox.emit('close');
+      firefoxApp.emit('close');
```
rpl: this should be renamed to `firefoxProcess`, like the one above.

rpl:
@shubheksha I took a look at the patch and the previous comments attached to this pull request, and I added a couple of comments related to some minor changes.

Thanks for working on this patch, besides the goal of renaming the command line option, this is going to help to make the sources much more clear, there are currently too many things named `firefox` in the sources :-)

shubheksha:
@rpl, thank you so much for the detailed feedback, I really appreciate it! :) I made all the changes you proposed, however they were causing two tests to fail. Going through the errors, I figured that the same change needs to be made in three more places in addition to those you mentioned.

src/cmd/run.js
```diff
@@ -75,15 +75,15 @@ export type ReloadStrategyOptions = {
 
 export function defaultReloadStrategy(
   {
-    addonId, firefox, client, profile, sourceDir, artifactsDir,
+    addonId, firefoxApp, client, profile, sourceDir, artifactsDir,
```
shubheksha: This had to be changed.

tests/test-cmd/test.run.js
```diff
@@ -330,9 +330,9 @@ describe('run', () => {
     }
 
     it('cleans up connections when firefox closes', () => {
-      const {firefox, client, watcher, reloadStrategy} = prepare();
+      const {firefoxApp, client, watcher, reloadStrategy} = prepare();
```
shubheksha: This had to be changed too.

tests/test-cmd/test.run.js
```diff
@@ -330,9 +330,9 @@ describe('run', () => {
     }
 
     it('cleans up connections when firefox closes', () => {
-      const {firefox, client, watcher, reloadStrategy} = prepare();
```
shubheksha: And this one as well.

src/cmd/run.js
```diff
@@ -261,11 +261,11 @@ export default function run(
         log.debug(
           `Reloading extension when the source changes; id=${addonId}`);
         reloadStrategy({
-          firefox: runningFirefox,
```
shubheksha: It had to be changed here too!

tests/test-cmd/test.run.js
```diff
@@ -10,7 +10,7 @@ import {onlyInstancesOf, WebExtError, RemoteTempInstallNotSupported}
 import run, {
   defaultFirefoxClient, defaultWatcherCreator, defaultReloadStrategy,
 } from '../../src/cmd/run';
-import * as firefox from '../../src/firefox';
+import * as firefoxApp from '../../src/firefox';
```
rpl: @shubheksha how about renaming the above to `defaultFirefox`, like it is named in most of the sources?
it would make easier to guess that they are the same thing.

(to be completely honest, now it would be probably better to name it `defaultFirefoxApp`, here and in "src/cmd/run.js", to make it clear that it is the default implementation of what we usually call `firefoxApp`)

rpl:
@shubheksha like pointed out in the above comment, it would be nice to rename `defaultFirefox` into `defaultFirefoxApp`, now that we use it as the default value of `firefoxApp` instead of `firefox` (just to avoid a potential confusion), but besides that, This looks great now! :+1:

tests/test-cmd/test.run.js
shubheksha: @rpl, sorry if this is a stupid question, but won't I have to rename all occurrences is `test.run.js`? Since it is being used as a type in `run.js`, I changed it in a few places & it works, but the same isn't true for the test file because as far as I understood, it is being used there directly & changing one occurrence breaks a lot of tests.

rpl: @shubheksha don't worry, asking question is a good thing.

The very minimal change that I was suggesting require two additional identifier renamed in the test file:
```
diff --git a/tests/test-cmd/test.run.js b/tests/test-cmd/test.run.js
index 49dbd57..7293530 100644
--- a/tests/test-cmd/test.run.js
+++ b/tests/test-cmd/test.run.js
@@ -10,7 +10,7 @@ import {onlyInstancesOf, WebExtError, RemoteTempInstallNotSupported}
 import run, {
   defaultFirefoxClient, defaultWatcherCreator, defaultReloadStrategy,
 } from '../../src/cmd/run';
-import * as firefoxApp from '../../src/firefox';
+import * as defaultFirefox from '../../src/firefox';
 import {RemoteFirefox} from '../../src/firefox/remote';
 import {TCPConnectError, fakeFirefoxClient, makeSureItFails, fake, fixturePath}
   from '../helpers';
@@ -69,7 +69,7 @@ describe('run', () => {
       run: () => Promise.resolve(),
       ...implementations,
     };
-    return fake(firefoxApp, allImplementations);
+    return fake(defaultFirefox, allImplementations);
   }
```

Or, to rename `defaultFirefox` into `defaultFirefoxApp` (mainly because I think that `defaultFirefoxApp` immediately suggests that it is the default implementation of what we now call `firefoxApp` and not `firefox`) we need to apply some minor changes to both `"test.run.js"` and `"src/cmd/run.js"`, e.g. this seems to be the needed changes in the `"src/cmd/run.js"` (besides the slightly changed version of the above patch applied to `"test.run.js"`):

```
index 15e29d9..8c06f83 100644
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -1,5 +1,5 @@
 /* @flow */
-import * as defaultFirefox from '../firefox';
+import * as defaultFirefoxApp from '../firefox';
 import defaultFirefoxConnector from '../firefox/remote';
 import {
   onlyInstancesOf, onlyErrorsWithCode, RemoteTempInstallNotSupported,
@@ -153,7 +153,7 @@ export type CmdRunParams = {
 };

 export type CmdRunOptions = {
-  firefoxApp: typeof defaultFirefox,
+  firefoxApp: typeof defaultFirefoxApp,
   firefoxClient: typeof defaultFirefoxClient,
   reloadStrategy: typeof defaultReloadStrategy,
 };
@@ -164,7 +164,7 @@ export default function run(
     preInstall=false, noReload=false,
   }: CmdRunParams,
   {
-    firefoxApp=defaultFirefox,
+    firefoxApp=defaultFirefoxApp,
     firefoxClient=defaultFirefoxClient,
     reloadStrategy=defaultReloadStrategy,
   }: CmdRunOptions = {}): Promise<Object> {
@@ -276,7 +276,7 @@ export type ExtensionRunnerParams = {
   sourceDir: string,
   manifestData: ExtensionManifest,
   profilePath: string,
-  firefoxApp: typeof defaultFirefox,
+  firefoxApp: typeof defaultFirefoxApp,
   firefox: string,
 };

@@ -284,7 +284,7 @@ export class ExtensionRunner {
   sourceDir: string;
   manifestData: ExtensionManifest;
   profilePath: string;
-  firefoxApp: typeof defaultFirefox;
+  firefoxApp: typeof defaultFirefoxApp;
   firefox: string;

   constructor(
```

shubheksha:
Ah, I completely missed the fact that the return value from that function is what is being used in the rest of the program. For `run.js`, I got it working the first time itself.
Also - why is the build failing for node v6 due to a timeout? Did I do something wrong? It passed all the test on my system, I am using node v4 & the build also passed with that node version.

rpl:
@shubheksha I was wondering the same thing, and I tried to run the tests locally with nodejs v6 on your branch and they run correctly, as expected.

It looks like an unrelated issue, it is probably related to some issues on the travis ci build machine that was running the job because:

- the addon server signing request/response in the tests uses a mock, it doesn't really create an http request and so it should not be a source of intermittent failures

- I asked travis to re-run the job (by asking travis to rebuild the job failed on nodejs v6 unmodified, and on this second run, all the tests are completed successfully on nodejs v6 without any issue.

rpl:
@kumar303 This PR is now r+

@shubheksha :+1: thanks for it!

shubheksha:
@rpl, oh, okay!
So glad this is finally done! Thank you for all the help & guidance. :)

kumar303:
thanks @shubheksha and @rpl ! I just got back, thanks also for your patience.

Issue #438, opened 2016-08-17
Move --source-dir coercion into its option configuration


kumar303:
The `--source-dir` path needs to get coerced into a resolved path. We had a hack in place to do this while we waited on https://github.com/yargs/yargs/issues/535. Let's remove the hack!

The old way was to [coerce the value](https://github.com/mozilla/web-ext/blob/master/src/program.js#L82-L86) in `run()`.

The new way can be to use the [.coerce()](https://github.com/yargs/yargs#coercekey-fn) callback. We upgraded yargs for this feature in https://github.com/mozilla/web-ext/pull/435

shubheksha:
For this, I figured that I'll need to make changes [here](https://github.com/mozilla/web-ext/blob/master/src/program.js#L153). However, on going through the [docs](https://github.com/yargs/yargs#coercekey-fn), it isn't very clear how it'll be implemented as opposed to the way alternate usage options have been given for [choices](https://github.com/yargs/yargs#choiceskey-choices).

rpl:
> @rpl, can you give me some direction on this issue? 
@shubheksha sure, it seems that this new yargs `coerce` feature can be used to transform a value passed to a yargs option, to be able to try this change we need to:

- remove the old code that is currently handling this transformation: https://github.com/mozilla/web-ext/blob/master/src/program.js#L82-L86

- define a `coerce` function for the sourceDir option, using one of the supported syntax, e.g.

Defined as a property of the `option` config object:

```
program.setGlobalOptions({
    'source-dir': {
      alias: 's',
      ...
      type: 'string',
      coerce: (value) => {
         let newvalue = ....;
         ...;
         return newvalue;
      }
},
```

or using the coerce method: 

```
...
program.yargs.coerce('sourceDir', (value) => { ... });
...
```

For this PR we just need to transform the value using `path.resolve`, and so our `coerce` definition is probably even simpler of the above snippets.

shubheksha:
@rpl, thanks, got it! I was wondering if the syntax in the first snippet was valid or no even though it made sense in my head -- this cleared it up. I'll open a PR with the required changes soon.

shubheksha:
@rpl, I made all the required changes & everything works fine except for test which fails repeatedly & I am not being able to figure out why. [This](https://github.com/mozilla/web-ext/blob/master/tests/test.program.js#L280) line isn't giving the required output. I've attached the screenshot for reference. 
[...]

PR #453, opened 2016-08-24
refactor: use coerce function of yargs to resolve path

shubheksha:
Removed the hack which was in place to do this & replaced it with the `coerce` function to resolve the path of the source directory. Solves #438

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -79,12 +79,6 @@ export class Program {
         this.yargs.exitProcess(this.shouldExitProgram);
         argv = this.yargs.argv;
 
-        // Fix up sourceDir. This can hopefully move to option configuration
-        // soon. See https://github.com/yargs/yargs/issues/535
-        if (argv.sourceDir) {
-          argv.sourceDir = path.resolve(argv.sourceDir);
-        }
-
         cmd = argv._[0];
         if (cmd === undefined) {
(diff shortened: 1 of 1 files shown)


src/program.js
```diff
@@ -154,9 +148,11 @@ Example: $0 --help run.
       alias: 's',
       describe: 'Web extension source directory.',
       default: process.cwd(),
-      normalize: true,
       requiresArg: true,
       type: 'string',
+      coerce: (arg) => {
```
rpl: it looks like this can be reduced to:

```
...
coerce: path.resolve,
...
```

kumar303:
Thanks for the change. It feels much cleaner now 
le-happy

atsay:
Thanks for contributing Shubheksha, and welcome to the community! I've added your merges on our [recognition page](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#August_2016).

shubheksha:
@atsay, thank you so much, this made my day! :smile: I really hope to stay around and contribute a lot more! 
There is one more [PR](https://github.com/mozilla/web-ext/pull/434) I opened which got merged, could you include this as well? 
Thanks again!

atsay:
Done! Thanks for letting me know--we don't always catch everything, but we're working on ways to get better at it.

PR #468, opened 2016-09-06
fix: firefox binary path not passed in the run method exported by 'src/firefox/index'

rpl:
This PR fixes an issue with the optional Firefox binary path option and contains some tweaks to the related declared flow types, which shows how we can force flow to raise errors on this kind of issues (missing optional parameters in the `parameters` and `options` objects)

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -333,6 +333,6 @@ export class ExtensionRunner {
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
     const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefox});
+    return firefoxApp.run(profile, {firefoxBinary: firefox});
   }
 }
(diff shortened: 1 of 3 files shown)


src/cmd/run.js
```diff
@@ -333,6 +333,6 @@ export class ExtensionRunner {
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
     const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefox});
+    return firefoxApp.run(profile, {firefoxBinary: firefox || null});
```
kumar303: does this need to pass `null` to satisfy the flow type or something? Could it just be `{firefoxBinary: firefox}` ?

kumar303:
r+wc (with comments). I just had one question about passing `null`, otherwise it looks good.

src/cmd/run.js
rpl: @kumar303 yep, it was a small leftover.
Removed in the updated patch.

rpl:
@kumar303 the change suggested in the above comment has been applied and this pull request updated accordingly.

kumar303:
Thanks!

PR #469, opened 2016-09-07
test: add functional test suite

rpl:
This pull request contains the following proposals:

- move all the existent mocha tests into their own dir ("tests/unit")
- turn the smoke test into an additional mocha test suite ("tests/functional")
- add additional smoke tests which tests that the sub commands are not "catastrophically broken", by checking:
  - "run web-ext without any subcommand and the --help opition", which is the existent smoke test scenario
  - "run web-ext build/run/sign/lint on a simple test addon" (using two small nodejs scripts to fake Firefox and the Addon Server)
- change the order of the tasks executed on "npm run test" (lint, flow, unit, smoke)


rpl:
@kumar303 This PR is currently failing on travis "purposely" (it fails because of the issue fixed by #468), the test will pass as expected once this PR has been rebased on top of the fix applied by #468)

kumar303:
instead of `tests/smoke` (and corresponding identifiers) could we call it `tests/functional` ? Originally, calling it a "smoke test" made sense because there was one rudimentary check. I think now the checks are actually testing the functionality of the end product.

kumar303:
We could abbreviate it to `func` if that makes things any easier

Gruntfile.js
```diff
@@ -58,20 +55,9 @@ module.exports = function(grunt) {
...
-      result.on('close', function(code) {
-        grunt.log.writeln('web-ext exited: ' + code);
-        var succeeded = code === 0;
-        done(succeeded);
-      });
+      grunt.task.run([
+        'webpack:smoke',
+        'mochaTest:smoke',
+      ]);
     });
```
kumar303: This could be defined simpler now:
````
grunt.registerTask('check-for-smoke', [
  'webpack:smoke',
  '...',
]);
````

File: tests/smoke/test.cli.build.js [hunk 2]
```diff
@@ -0,0 +1,32 @@
+/* @flow */
+import {spawn} from 'child_process';
+import {describe, it} from 'mocha';
+
+import {
+  webExt, addonPath, artifactsPath,
+} from './common';
+
+describe('web-ext build', () => {
+  it('webext build --source-dir SRCDIR --artifacts-dir OUTDIR', () => {
```
kumar303: Sticking to [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development) makes things nice to read. Ideally, we should describe the behavior of `web-ext build` that we are testing and write it like a sentence in the call to `it(...)`. Does that pose any problems for the functional suite? I think something like it('builds and saves to a custom directory') would make sense here.

kumar303: if it makes more sense to add the full command line being tested then I'd suggest a few things:
* prefixing it with the command is redundant
* you could still adhere to a readable BDD descriptor like it('accepts: --source-dir SRCDIR --artifacts-dir OUTDIR')

tests/smoke/test.cli.build.js
```diff
@@ -0,0 +1,32 @@
...
+    return new Promise((resolve, reject) => {
+      const webextProcess = spawn(webExt, [
+        'build',
+        '--source-dir', addonPath,
+        '--artifacts-dir', artifactsPath,
```
kumar303: I know it's a bit more effort, but I would rather see `artifactsPath` done with a temp directory that gets torn down after the test (preferably also with an at-exit handler to guard against exceptions). You could maybe just use [TempDir](https://github.com/mozilla/web-ext/blob/0f87c9b986178db6d4e6a3e2d880cf2dfd7e873b/src/util/temp-dir.js#L51) actually. I'm worried that there could be some data leakage between tests if we rely on the project level artifact directory for this.

tests/smoke/test.cli.build.js
```diff
@@ -0,0 +1,32 @@
...
+      ]);
+
+      let errorData = '';
```
kumar303: this whole block of code looks like it can be wrapped up into a function that returns a promise, no?

tests/unit/test-cmd/test.build.js
```diff
@@ -5,8 +5,8 @@ import {it, describe} from 'mocha';
 import {assert} from 'chai';
 import sinon from 'sinon';
 
-import build, {safeFileName, FileFilter} from '../../src/cmd/build';
-import {withTempDir} from '../../src/util/temp-dir';
+import build, {safeFileName, FileFilter} from '../../../src/cmd/build';
```
kumar303: mmm, these relative paths are getting a little bit nuts. I filed https://github.com/mozilla/web-ext/issues/470 to fix it later.

kumar303:
should we run all test commands with `--verbose`? I think they would be easier to debug. Does that go to stdout or stderr? If it goes to stdout, it could just be merged into the `errorData` string.

tests/smoke/fake-amo-server.js
```diff
@@ -0,0 +1,36 @@
...
+      download_url: 'http://localhost:8989/some-signed-file-1.2.3.xpi',
+    }],
+  },
+  {},
+];
+
+var replyIndex = 0;
+
+http.createServer(function(req, res) {
+  //  console.log("REQ", req);
```
kumar303: this can be removed

tests/smoke/test.cli.sign.js
```diff
@@ -0,0 +1,38 @@
...
+import {describe, it} from 'mocha';
+
+import {
+  webExt, addonPath, artifactsPath, fakeServerPath,
+} from './common';
+
+describe('web-ext sign', () => {
+  it('webext lint --source-dir SRCDIR', () => {
+    return new Promise((resolve, reject) => {
+      const fakeServerProcess = spawn(fakeServerPath);
```
kumar303: cool idea!

tests/smoke/fake-amo-server.js
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
```
kumar303: It would be nice to stick a comment at the top of this file linking to http://addons-server.readthedocs.io/en/latest/topics/api/signing.html

tests/smoke/fake-amo-server.js
```diff
@@ -0,0 +1,36 @@
...
+    processed: true,
+    valid: true,
+    reviewed: true,
+    files: [{
+      signed: true,
+      download_url: 'http://localhost:8989/some-signed-file-1.2.3.xpi',
+    }],
+  },
+  {},
+];
```
kumar303: hmm, maybe we need to introduce some kind of spec file to generate this from [the source](http://addons-server.readthedocs.io/en/latest/topics/api/signing.html#get--api-v3-addons-[string-addon-id]-versions-[string-version]-(uploads-[string-upload-pk]-)) ? :) I'm just thinking out loud, may not be worth the effort.

kumar303:
thanks for whipping this into a mocha suite!

tests/functional/common.js
```diff
@@ -0,0 +1,126 @@
...
+const copyDirAsPromised = promisify(copyDir);
+export function withTempAddonDir(
+  {addonPath, runFromCwd}: TempAddonParams,
+  makePromise: TempAddonCallback,
+): Promise<any> {
+  return withTempDir((tmpDir) => {
+    const tempAddonDir = path.join(tmpDir.path(), 'tmp-addon-dir');
+    return copyDirAsPromised(addonPath, tempAddonDir)
+      .then(() => {
+        // NOTE: when "web-ext run" is spawned in functional tests,
```
rpl: @kumar303 This was a very weird issue that I still want to figure out (but in the meantime I workarounded it here):

if I change the current work dir of the functional test suite process, every tested web-ext command behaves as expected, but "web-ext run" doesn't understand "--no-reload" and it exits with the error message "Unknown argument: reload" (raised by yargs).

By tracing the parsed yargs argv (in both the cases: when the options are parsed correctly, and when yargs raises the above me error message), it seems that the issue is that yargs turns "--no-reload" into "reload: false" instead of "noReload: true".

kumar303: d'oh. That's a weird quirky feature of yargs. We had to override it in `package.json`; see https://github.com/mozilla/web-ext/commit/f93d68ab6a64ee3a322d5964c707a5a3cd36d287 . I guess that's why it stops working after you change directories.

kumar303: in hindsight, I think putting this in `package.json` was a mistake. We should move it to the main program. It looks like we can configure it [in a way similar to this](https://github.com/yargs/yargs/issues/531#issuecomment-227519201).

tasks/watch.js
```diff
@@ -13,9 +13,10 @@ module.exports = {
       '.flowconfig',
     ],
     tasks: [
-      'flowbin:status',
       'build-tests',
-      'mochaTest',
+      'mochaTest:unit',
+      'mochaTest:functional',
```
kumar303: I view `npm start` as a highly iterative development loop. It needs to be as fast as possible otherwise it's not so useful (and `npm test` is a better replacement). Thus, I think we should not run the functional tests in this loop. However, if you feel like they would still be useful then I would request to exclude them by default and include them only when a well-documented environment variable is present.

rpl:
@kumar303 

Time for a brief recap of the changes in this pull request:

- the main goal of this PR is to provide a functional test suite (#307) and a reasonable initial set of functional tests (currently one test case for web-ext command).
- these new functional tests are going to test more closely the web-ext CLI interface, by spawning the actual web-ext executable into another process through the nodejs `spawn` method from the child_process module.
- this is going to help achieving part of #377 goal (but with a different strategy).
- these tests are NOT going to test integration issues with the external components (e.g. incompatible changes in the Remote Debugging Protocol used to integrate the web-ext command with a Firefox instance or in the signing API provided by the addons-server will not be caught automatically by this test suites), these tests are using two simple nodejs scripts to mock a Firefox binary and the addon signing service.
- the test suite follows the BDD mocha syntax, but the description messages contain the command line syntax, to more easily recognize which combination of cli arguments are explicitly tested from this test suite and which are not.

The focus of this functional test suite (mostly like already stated in #307) is to cover what cannot be covered correctly and reliably with the unit tests, e.g.:

- ensuring the correct interaction between the various internal modules (which are tested as unit in the unit tests with mocks/shims of the other dependencies)
- ensuring the correct interaction between the web-ext internals and its npm dependencies (e.g. addons-linter, node-firefox-connect, yargs etc.)
- ensuring the correct behavior of the yargs cli arguments parsing (which should be easier and more reliable from this test suite)

Finally, this initial set of functional tests has brought to the light two issues (which have a proposed fix in some of the patches included in this pull request):

- fix on the `--no-reload` option (which is related to issues with the yargs internal config loading mechanisms, with a proposed fix in 82f288c and a follow up in 5a5e0a4)
- fix issues with the combination of the yargs `demand` usage and the `addons-linter` integrations and the fix for the `--no-reload` proposed in 82f288c (with a proposed fix in 5a5e0a4)

kumar303:
> ensuring the correct behavior of the yargs cli arguments parsing (which should be easier and more reliable from this test suite)

I don't think this needs to be a goal of the functional suite. I think we can accomplish this goal better in the unit tests after implementing https://github.com/mozilla/web-ext/issues/377

kumar303:
Aside from that ^ I agree with the goals!

src/program.js
```diff
@@ -19,15 +19,25 @@ export class Program {
...
-    }
+  constructor(
+    argv: ?Array<string>,
+    {absolutePackageDir=process.cwd()}: {absolutePackageDir?: string} = {}
+  ) {
+    // This allows us to override the process argv which is useful for
+    // testing.
+    // NOTE: process.argv.slice(2) removes the path to node and web-ext
+    // executables from the process.argv array.
+    argv = argv || process.argv.slice(2);
```
kumar303: could we just leave this as `undefined` when not overridden? That should trigger the `yargs` default behavior.

tests/functional/test.cli.build.js
```diff
@@ -0,0 +1,27 @@
...
+  withTempAddonDir, runCommand, reportRunCommandError,
+} from './common';
+
+describe('web-ext build', () => {
+  it('should accept: --source-dir SRCDIR',
+     () => withTempAddonDir({addonPath}, (srcDir, tmpDir) => {
+       const argv =  ['build', '--source-dir', srcDir, '--verbose'];
+       const cmd = runCommand(webExt, argv, {cwd: tmpDir});
+
+       return cmd.waitForExited.then(({exitCode, stdout, stderr}) => {
```
kumar303: nit: `cmd.waitForExit.then(...)` would read better to me

kumar303:
Looks great! r+wc

Issue #283, opened 2016-06-10
Discover an open remote debugger port


kumar303:
In the patch for https://github.com/mozilla/web-ext/issues/93, the remote debugger port selection was improved. It will fail if the port is already in use. We should go further and make it select a different open port if one is already in use.

kumar303:
There has been some architectural work done to prepare for this patch. As of now, the change should be simple. I can mentor anyone willing to try a patch for this. The work can be done in the [defaultRemotePortFinder()](https://github.com/mozilla/web-ext/blob/fa27fb0784e57a19a445139943c7ea3337c09139/src/firefox/index.js#L36-L40) function. Instead of throwing an error, a new port number should be chosen and tried in a loop until an open port is found.

Krizzu:
Hey @kumar303. This looks fairly simple, so I can try my best to do it.

kumar303:
> This looks fairly simple, so I can try my best to do it.

Great! Just let me know how I can help

> How to determinate when it is open and free to use? 

As you pointed out, the `ECONNREFUSED` error is probably enough to indicate that the port is free to use. Actually, the code is already in place to handle this situation. The only thing you'd need to implement is [where the error is thrown](https://github.com/mozilla/web-ext/blob/fa27fb0784e57a19a445139943c7ea3337c09139/src/firefox/index.js#L39). Instead of throwing an error, it needs to retry the promise in a loop.

Krizzu:
Hey @kumar303 
I think I rushed myself saying that looks easy. :-1: 
So I was thinking, if I should implement a function that will retry the promise and wait for it until it's rejected. Am I on the right way?

kumar303:
Heh, yeah, retry loops are never easy but this is a small piece of code and there are some other examples to follow. Take a look at how we [attempt connecting to Firefox in a retry loop](https://github.com/mozilla/web-ext/blob/96748e35ae5502d600b0255565b08f1c68e95355/src/cmd/run.js#L112-L140). You could do something similar to this.

Krizzu:
hey @kumar303 
Thanks for that. I come up with this: 

```
  let retriesCount = 20; 
  function portConnectionRetry(){
    if(retriesCount == 0) throw 'ERROR' //change here
    return new Promise((resolve, reject) => {
        connectToFirefox(portToTry)
        .then((client) => {
          log.debug(`Remote Firefox port ${portToTry} is in use. Trying next one`);
          client.disconnect();
          portToTry++;
[...]

PR #472, opened 2016-09-07
feat: discover an open remote debugger port

Krizzu:
#283 Feature


src/firefox/index.js
```diff
@@ -54,20 +54,41 @@ export function defaultRemotePortFinder(
...
+      );
+    }
+    return new Promise((resolve, reject) => {
+      connectToFirefox(portToTry)
+        .then((client) => {
+          log.debug(`Remote Firefox port ${portToTry} is in use.
+            Trying next one`);
+          client.disconnect();
+          portToTry++;
+          throw new WebExtError(
```
kumar303: instead of throwing the error here, you can just initiate the next retry like `return portConnectionRetry()`. If the retry number hits the max then throw an error.

src/firefox/index.js
```diff
@@ -54,20 +54,41 @@ export function defaultRemotePortFinder(
...
+          reject(err);
+        });
     })
-    .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
-      // The connection was refused so this port is good to use.
+    .then(() => {
       return portToTry;
-    }));
+    })
+    .catch(() => {
```
kumar303: you won't need this catch statement if you do the retry up above

kumar303:
Take a look at the [retry tests](https://github.com/mozilla/web-ext/blob/0f87c9b986178db6d4e6a3e2d880cf2dfd7e873b/tests/test-cmd/test.run.js#L353) for a similar function that needs to recover from errors and retry. This should help you write similar retry tests. What you will need to do is cover all the important cases: 

1. success on first try
2. success on second try
3. failure after too many retries.

src/firefox/index.js
Krizzu: Do You mean error throw at line 60 or 71?

Krizzu:
Hey, 
I'm kinda stuck. Everything works as it should, except for one test - to retry after first fail. My test comes to the point where I'm returning a port number after first failed test (I've put a log to see where is it), but then just before returning it from last **.then**, I receive a Timeout error.

kumar303:
Hi.
I don't see the tests added to your patch in this pull request. Can you add them? It would be easier if I could see exactly what you mean about the timeout so I can offer specific guidance.

By the way, the failure you are currently seeing is just a complaint about your commit message. It is saying to use `feat:` instead of `Feat:` as the prefix. You could fix this with [git commit --amend](https://help.github.com/articles/changing-a-commit-message/).

kumar303:
> You could fix this with git commit --amend.

Actually, I think you can maybe just push a new commit with a fixed commit message to stop seeing the error.

Krizzu:
Hello there and thank You for patience. 
Sorry for my bad, I created pull request here again. 
The line I'm talking about is the line 524 in test.firefox.js.
I thought that might be a problem with **client** that is being passed from success **connectToFirefox** function, so I added one module from *helpers*.

src/firefox/index.js
```diff
@@ -50,27 +50,39 @@ export function defaultRemotePortFinder(
...
-    .then((client) => {
-      log.debug(`Remote Firefox port ${portToTry} is in use`);
-      client.disconnect();
-      // TODO: instead of throw an error, pick a new random port until
-      // one of them is available.
-      // https://github.com/mozilla/web-ext/issues/283
-      throw new WebExtError(
-        `Cannot listen on port ${portToTry} because it's in use`);
+  let retriesCount = 10;
+  function portConnectionRetry(){
```
kumar303: this needs a space before the opening brace (just a nitpicky style thing)

src/firefox/index.js
```diff
@@ -50,27 +50,39 @@ export function defaultRemotePortFinder(
...
+    return new Promise((resolve, reject) => {
+      connectToFirefox(portToTry)
+        .then((client) => {
+          log.debug(`Remote Firefox port ${portToTry} is in use.`);
```
kumar303: it would be helpful to also log the current value of `retriesCount` here.

kumar303: you could do it as:

    log.debug(`Remote Firefox port ${portToTry} is in use (tries remaining: ${retriesCount}).`)

src/firefox/index.js
```diff
@@ -50,27 +50,39 @@ export function defaultRemotePortFinder(
...
-      log.debug(`Remote Firefox port ${portToTry} is in use`);
-      client.disconnect();
-      // TODO: instead of throw an error, pick a new random port until
-      // one of them is available.
-      // https://github.com/mozilla/web-ext/issues/283
-      throw new WebExtError(
-        `Cannot listen on port ${portToTry} because it's in use`);
+  let retriesCount = 10;
+  function portConnectionRetry(){
+    return new Promise((resolve, reject) => {
```
kumar303: There is no need to construct a promise object here. You can just return `connectToFirefox(portToTry)` because this returns a promise. After that change, you can remove your calls to `resolve()` and `reject()`. Replace `resolve(value)` with `return value` and replace `reject(error)` with `throw error`. This will make the promise chain easier to read.

Krizzu: That's what I thought in first place :tada: 
Going to change it soon

src/firefox/index.js
```diff
@@ -50,27 +50,39 @@ export function defaultRemotePortFinder(
...
+            return portConnectionRetry();
+          }
+          else {
+            throw new WebExtError('Too many retries on port search');
+          }
+        })
+       .catch(onlyErrorsWithCode('ECONNREFUSED', () =>{
+         resolve(portToTry);
+       }))
+       .catch((err) => {
```
kumar303: You can remove this catch block because it's not needed. For the final uncaught error case, the promise will already be rejected.

src/firefox/index.js
```diff
@@ -50,27 +50,39 @@ export function defaultRemotePortFinder(
...
     })
-    .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
-      // The connection was refused so this port is good to use.
-      return portToTry;
-    }));
+    .then((port) => {
+      return port;
+    });
+  }
+  return portConnectionRetry();
```
kumar303: instead of `portConnectionRetry()`, I think a more concise and descriptive name would be `tryToFindAnOpenPort()`

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
-          assert.equal(client.disconnect.called, true);
+    it('returns a port on first try', () => {
+      const connectToFirefox = sinon.spy(() => new Promise(
+        (resolve, reject) => {
+          reject(
+            new TCPConnectError('first call connection fails - port is free')
+          );
         }));
+      return findRemotePort({connectToFirefox})
+      .then((port) => {
```
kumar303: this block should be indented to show that it's an extension of the return statement

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
+        assert.equal(connectToFirefox.callCount, 1);
+        assert.isNumber(port);
+      });
     });
 
-    it('re-throws unexpected connection errors', () => {
-      const connectToFirefox = sinon.spy(
-        () => Promise.reject(new Error('not a connection error')));
+    it('cancel discovery after too many fails', () => {
```
kumar303: `it('cancels discovery...')`

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
+      const client = fake(RemoteFirefox.prototype);
+      const connectToFirefox = sinon.spy(() => new Promise(
+        (resolve) => {
+          resolve(client);
+        }));
```
kumar303: something like this is ok to put on one line:

    ... new Promise(
      (resolve) => resolve(client)
    );

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
       return findRemotePort({connectToFirefox})
```
kumar303: This function will try for 10 times before giving up. This will slow down the test suite unnecessarily so what we do here instead is to limit the number of times it can try before giving up (just for testing purposes). You would call it like this:

    return findRemotePort({connectToFirefox, retriesCount: 2})

You would then need to adjust the signature of the function to accept a custom `retriesCount` value. With this change, the test would only try twice instead of 10 times but it would still cover the behavior we want to test, i.e. the function will give up after X times.

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
-        .then(makeSureItFails())
-        .catch((error) => {
-          assert.match(error.message, /not a connection error/);
+      .catch((err) => {
```
kumar303: you also need to assert that the connector function was called X times, like:

    assert.equal(connectToFirefox.callCount, 2);

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
+        assert.equal(err, 'WebExtError: Too many retries on port search');
+      });
+    });
+
+    it('retries port discovery after first failure', () => {
+      const client = new RemoteFirefox(fakeFirefoxClient());
+      const connectToFirefox = sinon.spy((port) => {
+        return new Promise((resolve, reject) => {
+          if (port===6006) {
```
kumar303: relying on the port number will make this test fragile because the starting port number may be configured differently in the future. A better approach would be to use a counter variable and simply reject the *first* call to `connectToFirefox` but not the second call. For example:

````javascript
let callCount = 1
const connectToFirefox = sinon.spy((port) => new Promise((resolve, reject) => {
  if (callCount === 1) {
    // Pretend like the first attempt connects to a port in use:
    resolve(client);
  } else {
    // Pretend like the second attempt finds an open port:
    reject(new TCPConnectError('port is free'));
  }
  callCount ++;
});
````

File: tests/test-firefox/test.firefox.js [hunk 14]
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
+            reject(new TCPConnectError('port is free'));
+          }
+          else {
+            resolve(client);
+          }
         });
+      });
+      //to avoid test failing,
+      //I removed 'return' before calling findRemotePort.
```
kumar303: In the future, it's easier to make the test fail when you are still getting feedback on a patch. That way, I can look at the test failure and see exactly what you also see in your console.

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,53 @@ describe('firefox', () => {
...
+      findRemotePort({connectToFirefox})
+      .then((port) => {
+        assert.isNumber(port);
+        assert.equal(connectToFirefox.callCount, 2);
+        return port;
```
kumar303: you won't need to return `port` since this is the end of the test

tests/test-firefox/test.firefox.js
kumar303: Since I can't see the test failure I'm not sure what the exact problem was but I think if you use a counter it will be more straight forward. I'll take another look after that change.

kumar303:
Thanks for following up! I left some detailed comments, mostly about style but also about simplifying the code which will make it easier to figure out. If you are still getting test failures, push a change that breaks the test. That way I can see the actual error and try to help.

tests/test-firefox/test.firefox.js
Krizzu: Haha, I've missed that one. :smile:

tests/test-firefox/test.firefox.js
Krizzu: shouldn't `callCount++` be before calling `resolve` or `reject` ?

Something like this?
```
let callCount = 0;
const connectToFirefox = sinon.spy((port) => new Promise((resolve, reject) => {
 callCount ++;
  if (callCount === 1) {
    // Pretend like the first attempt connects to a port in use:
    resolve(client);
  } else {
    // Pretend like the second attempt finds an open port:
    reject(new TCPConnectError('port is free'));
  }
});
```

kumar303: yes, good catch

src/firefox/index.js
```diff
@@ -49,28 +50,33 @@ export type RemotePortFinderFn =
...
-      throw new WebExtError(
-        `Cannot listen on port ${portToTry} because it's in use`);
-    })
-    .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
-      // The connection was refused so this port is good to use.
-      return portToTry;
-    }));
+  function tryToFindAnOpenPort() {
+    return connectToFirefox(portToTry)
+        .then((client) => {
```
kumar303: nit: this block should only be indented two spaces

src/firefox/index.js
```diff
@@ -49,28 +50,33 @@ export type RemotePortFinderFn =
...
+          log.debug(`Remote Firefox port ${portToTry} is in use
+            (retries remaining: ${retriesLeft} )`);
```
kumar303: This will show up as a line break with lots of whitespace in the log. To avoid that you can concatenate to strings. It's not pretty but I think this will make a better log:
````javascript
log.debug(
  `Remote Firefox port ${portToTry} is in use ` +
  `(retries remaining: ${retriesLeft} )`);
````

File: src/firefox/index.js [hunk 18]
```diff
@@ -49,28 +50,33 @@ export type RemotePortFinderFn =
...
-    })
-    .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
-      // The connection was refused so this port is good to use.
-      return portToTry;
-    }));
+  function tryToFindAnOpenPort() {
+    return connectToFirefox(portToTry)
+      .then((client) => {
+        log.debug(`Remote Firefox port ${portToTry} is in use
+          (retries remaining: ${retriesLeft} )`);
```
kumar303: This will show up as a line break with lots of whitespace in the log. To avoid that you can concatenate to strings. It's not pretty but I think this will make a better log:
````javascript
log.debug(
  `Remote Firefox port ${portToTry} is in use ` +
  `(retries remaining: ${retriesLeft} )`);
````

File: src/firefox/index.js [hunk 19]
```diff
@@ -49,28 +50,33 @@ export type RemotePortFinderFn =
 export function defaultRemotePortFinder(
   {
     portToTry=REMOTE_PORT,
+    retriesLeft=10,
```
kumar303: I like this variable name better 👍

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,51 @@ describe('firefox', () => {
...
+      });
+    });
+
+    it('retries port discovery after first failure', () => {
+      const client = new RemoteFirefox(fakeFirefoxClient());
+      let callCount = 0;
+      const connectToFirefox = sinon.spy(() => {
+        callCount++;
+        return new Promise((resolve, reject) => {
+          if (callCount===2) {
```
kumar303: nit: add spaces around infix operators, like `callCount === 2`

tests/test-firefox/test.firefox.js
```diff
@@ -480,29 +481,51 @@ describe('firefox', () => {
...
-      return findRemotePort({connectToFirefox})
-        .then(makeSureItFails())
-        .catch((error) => {
-          assert.match(error.message, /not a connection error/);
+    it('cancels search after too many fails', () => {
+      const client = fake(RemoteFirefox.prototype);
+      const connectToFirefox = sinon.spy(() => new Promise(
+        (resolve) => resolve(client)));
+      return findRemotePort({connectToFirefox, retriesLeft: 2})
+      .catch((err) => {
```
kumar303: nit: this also needs indenting to indicate that it's a continuation of the return statement

kumar303:
Looking good! I commented about nitpicky style changes that we need :)

I'm not sure where you have a conflict. You should be able to pull from master, merge master into your branch, and push the merge after fixing conflicts. Let me know if you need help with that.

Krizzu:
I've pulled from master, there was one conflict, but solved it.

kumar303:
Oops, another patch landed on master that you'll need to merge in and resolve a conflict for. It was just a style change for default parameters so it should be easy.

Also, let me know if you need help fixing the changelog error. You can use something like [git rebase --interactive](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages) to rename the offending commit message.

kumar303:
Also, I tested out this patch and it works! :) I opened an instance of Nightly on a conflicting port and `web-ext` chose a new port.

src/firefox/index.js
```diff
@@ -49,28 +50,33 @@ export type RemotePortFinderFn =
...
+        portToTry++;
+        if (retriesLeft > 0) {
+          retriesLeft--;
+          return tryToFindAnOpenPort();
+        }
+        else {
+          throw new WebExtError('Too many retries on port search');
+        }
+      })
+     .catch(onlyErrorsWithCode('ECONNREFUSED', () => {
```
kumar303: I would like to see the original comment put back here:

    // The connection was refused so this port is good to use.

The reason is that it's confusing to look at such an obtuse error code and see the recovery from it as a successful result. The comment explains why.

Krizzu:
The problem is, I have only last 2 commits, so I can't modify the first one :disappointed:

kumar303:
oh, I just remembered that when I 'squash and merge' this PR into master, the commit with the misformatted message will disappear so it's all good.

Thanks again for all your work on this!

kumar303:
by the way, I noticed that you did this work on the master branch of your fork. This will make it complicated to sync up with the web-ext master now that I've merged the PR. In the future, it will be easier to [create git topic branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow) and request pull requests from there.

Krizzu:
Thanks for the everything, I've learned a lot!
Stay awesome :+1:

atsay:
Thanks for your contribution, Krizzu! I've added it [here](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#September_2016).

PR #474, opened 2016-09-08
fix: show an error if an argument is passed to the lint command

### LINKED ISSUE #398: Show an error when passing an argument to the lint command
If you run `web-ext lint ./some-file.xpi` then it will not lint `some-file.xpi`. It ignores command arguments. However, the `addons-linter` accepts arguments so this could be confusing. We should raise an error if an argument is given.

See https://github.com/mozilla/web-ext/issues/396

shubheksha:
Fixes #398 
Also added a test for checking if an argument is passed to a sub-command.

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -38,6 +38,7 @@ export class Program {
         return;
       }
       return yargs
+        .demand(0, 0, 'This command does not take any arguments')
         .strict()
         .exitProcess(this.shouldExitProgram)
         // Calling env() will be unnecessary after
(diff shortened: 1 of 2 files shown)


kumar303:
looks great, thanks!

I think one day we might create a command that requires an argument. But that day may never come ;) We can make this more configurable if the day does come.

atsay:
Yay! I've noted your contribution [here](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#September_2016).

PR #475, opened 2016-09-08
fix: test case checking for unknown sub-command options doesn't work

### LINKED ISSUE #473: Test case checking for unknown sub-command options doesn't work as expected
https://github.com/mozilla/web-ext/blob/master/tests/test.program.js#L218
The test case here takes the argument `['thing --nope']` to test for passing an unknown option to the sub-command. However, this is considered as one command and not `--nope` being passed to the command `thing`.

shubheksha:
Fixes #473


kumar303:
great catch, thanks

PR #483, opened 2016-09-13
Require spaces around infix operators

### LINKED ISSUE #482: Require spaces around infix operators (for web-ext code)
The majority of web-ext code uses spaces around infix operators but this is not enforced. Let's enforce it by adding this rule to `.eslintrc`:
````
"space-infix-ops": "error"
````

This will require fixing up some code after you add the rule, mostly around default parameter assignment.

To test the rule, do: `npm run lint`

### PR DESCRIPTION
**Pull request by pdehaan**
Fixes #482

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -81,7 +81,7 @@
     "semi": [2, "always"],
     "space-before-blocks": [2, "always"],
     "space-before-function-paren": [2, "never"],
-    "space-infix-ops": 0,
+    "space-infix-ops": 2,
     "strict": [2, "never"],
     "valid-typeof": 2,
     "wrap-iife": [2, "inside"],
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -48,8 +48,8 @@ export type RemotePortFinderFn =
 
 export function defaultRemotePortFinder(
   {
-    portToTry=REMOTE_PORT,
-    connectToFirefox=defaultFirefoxConnector,
+    portToTry = REMOTE_PORT,
+    connectToFirefox = defaultFirefoxConnector,
   }: RemotePortFinderParams = {}
 ): Promise<number> {
   log.debug(`Checking if remote Firefox port ${portToTry} is available`);
@@ -115,8 +115,8 @@ export type FirefoxRunOptions = {
 export function run(
   profile: FirefoxProfile,
   {
-    fxRunner=defaultFxRunner,
-    findRemotePort=defaultRemotePortFinder,
+    fxRunner = defaultFxRunner,
+    findRemotePort = defaultRemotePortFinder,
     firefoxBinary, binaryArgs,
   }: FirefoxRunOptions = {}
 ): Promise<FirefoxProcess> {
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: src/firefox/index.js [hunk 1]
```diff
@@ -310,8 +310,12 @@ export type InstallExtensionParams = {
...
  */
 export function installExtension(
-  {asProxy=false, manifestData, profile, extensionPath}: InstallExtensionParams
-): Promise<any> {
+  {
+    asProxy = false,
+    manifestData,
+    profile,
+    extensionPath,
+  }: InstallExtensionParams): Promise<any> {
```
pdehaan: I autofixed everything with `$ ./node_modules/.bin/eslint {tasks,tests,src,*.js} --fix`, but this was the one place where the _new_ line length now exceeded 80 and therefore threw an error.

kumar303: Cool. It reads a bit better on separate lines anyway.

.eslintrc
```diff
@@ -81,7 +81,7 @@
     "semi": [2, "always"],
     "space-before-blocks": [2, "always"],
     "space-before-function-paren": [2, "never"],
-    "space-infix-ops": 0,
+    "space-infix-ops": 2,
```
kumar303: oh, ha, I forgot we even added it. I guess it was because we liked doing default assignment in Python style. I realize we need to enforce it for things like `if (thing===other)` which is more important though.

kumar303:
Thanks! You will need to amend the commit here so that message can be [changelog-compatible](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages). I suggest going with `style:` or `chore:`, doesn't really matter.

.eslintrc
pdehaan: Sorry, I also went with the executive decision to use `2` instead of `"error"` for consistency with other rules.

> The majority of web-ext code uses spaces around infix operators but this is not enforced. Let's enforce it by adding this rule to .eslintrc:

> ```yaml
"space-infix-ops": "error"
```
> &mdash; via https://github.com/mozilla/web-ext/issues/482#issue-176728517

**Comment by pdehaan** [comment]
Amended and pushed, let's see if I did the right tweak.

Since we fail PRs on invalid commit messages, we may want to see if we can do that near the top of the script block so we fail fast and hard. I think my builds took 2.5 and 6 minutes before failing, since they had to run all the tests and other stuff first.

Not sure if it's as easy as reversing the order, or if we want to use the `before_script` hook, per https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle 

```yaml
before_script:
# Run changelog-lint but only on newer versions of Node (because of syntax errors)
- if [[ ${TRAVIS_NODE_VERSION:0:1} -ge "4" ]]; then
    npm run changelog-lint;
  fi

script:
- COVERAGE=y npm test
```

kumar303:
> Since we fail PRs on invalid commit messages, we may want to see if we can do that near the top of the script block so we fail fast and hard. 

Good idea! Could you file a bug for that?

kumar303:
Thanks again for the patch

pdehaan:
> Good idea! Could you file a bug for that?

Done. Ref #484

PR #485, opened 2016-09-13
Fail fast on changelog linting

### LINKED ISSUE #484: Fail fast and hard on changelog violations on Travis
Ref: https://github.com/mozilla/web-ext/pull/483#issuecomment-246828045

---

> Since we fail PRs on invalid commit messages, we may want to see if we can do that near the top of the script block so we fail fast and hard. I think my builds took 2.5 and 6 minutes before failing, since they had to run all the tests and other stuff first.

> Not sure if it's as easy as reversing the order, or if we want to use the `before_script` hook, per https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle 

> ```yaml
before_script:
# Run changelog-lint but only on newer versions of Node (because of syntax errors)
- if [[ ${TRAVIS_NODE_VERSION:0:1} -ge "4" ]]; then
    npm run changelog-lint;
  fi

> script:
- COVERAGE=y npm test
```

### PR DESCRIPTION
**Pull request by pdehaan**
Fixes #484

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,16 @@ before_install:
 - git remote set-branches origin master && git fetch
 - git checkout master
 - git checkout -
-script:
-- COVERAGE=y npm test
+
+before_script:
 # Run changelog-lint but only on newer versions of Node (because of syntax errors)
 - if [[ ${TRAVIS_NODE_VERSION:0:1} -ge "4" ]]; then
     npm run changelog-lint;
   fi
+
+script:
+- COVERAGE=y npm test
+
 after_script: npm run publish-coverage
 notifications:
   irc:

### CODE REVIEW

**Comment by pdehaan** [comment]
Now, I said "fast", but it doesn't seem very fast:

build__1406_-_mozilla_web-ext_-_travis_ci

**Node.js: 4** seems to take 5 minutes to fail, 270s (4.5 minutes) of that seems to be the `npm install` step.

**Node.js: 6** fails much quicker, at 1.5 minutes, 70s of that is `npm install`.

---

My [previous PR](https://travis-ci.org/mozilla/web-ext/builds/159696932) where I messed up my commit message and the linting and tests all ran, looked roughly like:

**Node.js: 4** 6 min, 4 sec
**Node.js: 6** 2 min, 27 sec

So moving the the changelog linting doesn't save TOO much time, since the biggest penalty seems to be the `npm install` step anyways, and tests only take an extra minute.

Not sure if worth proceeding, or if there is value in showing the linting and test errors and THEN fail on changelog violations.

**Comment by pdehaan** [comment]
build__1407_-_mozilla_web-ext_-_travis_ci

Not very inspiring time savings... Only other alternative I can think of, is a weird bit of `before_install` logic where we explicitly install the 2 conventional changelog modules and then run the linter before installing the rest of the modules.

```yaml
before_install:
- npm i conventional-changelog-{cli,lint} -g
# Run changelog-lint but only on newer versions of Node (because of syntax errors)
- if [[ ${TRAVIS_NODE_VERSION:0:1} -ge "4" ]]; then
    npm run changelog-lint;
  fi
```

But the upside is that it may fail within 20s instead of installing _ALL_ dependencies (for 5 minutes) and then failing.

kumar303:
Ah, yes, I forgot about this npm install problem. We should try and fix it.

We kicked around the idea of using [TravisCI's npm cache](https://docs.travis-ci.com/user/caching/) by adding this to `.travis.yml`:
````
cache:
  directories:
    - node_modules
````

My one concern was that it won't deal with deleted dependencies very well. However, I think we should just add it and see how it goes because the build time needs to be fast. @rpl what do you think?

kumar303:
I think you should add the npm caching. Let's see how it goes!

kumar303:
If you get a chance to pick this back up, it still looks useful. Thanks for getting it started.

kumar303:
We're getting a bit overwhelmed with pull requests so I'm trying to clear some mental space on the tracker. Feel free to re-open this if you feel like working on it!

Issue #198, opened 2016-04-19
Building an XPI with only localizations makes an ugly file name


kumar303:
Try running `web-ext build` in https://github.com/mdn/webextensions-examples/tree/master/notify-link-clicks-i18n

You get `_msg_extensionname_-1.0.xpi` which is ugly.

kumar303:
The fix is pretty simple: Adjust the build command to [generate a locale aware filename](https://github.com/mozilla/web-ext/blob/2c06dbd892feb02bd69a0345bee2b4d2d98233b1/src/cmd/build.js#L63-L64). I think it would be reasonable to generate the filename using the default locale of the manifest when the extension is localized.

kumar303:
fixed in https://github.com/mozilla/web-ext/pull/487 🌞

PR #487, opened 2016-09-14
fix: Building an XPI with only localizations makes an ugly file name

shubheksha:
A possible fix for #198


shubheksha:
It works when I run `web-ext build` in https://github.com/mdn/webextensions-examples/tree/master/notify-link-clicks-i18n.
However, the test fails due to `manifestData.default_locale` being undefined and hence it cannot be concatenated with `_locales` via `path.join()`. I am not sure if the tests take care of the `_locales` directory.

tests/unit/test-util/test.manifest.js
```diff
@@ -19,6 +19,7 @@ export const basicManifest = {
       id: 'basic-manifest@web-ext-test-suite',
     },
   },
+  default_locale: 'en',
```
kumar303: you shouldn't need to add this because the field is optional

kumar303:
> the test fails due to manifestData.default_locale being undefined

I don't see this error in the TravisCI results. Did you push all the code you have locally?

shubheksha:
That was happening due to #483. I just pulled from upstream & rebased my branch on top of the changes, so the error should be displayed in the results now.

src/cmd/build.js
```diff
@@ -60,8 +61,12 @@ function defaultPackageCreator(
           filter: (...args) => fileFilter.wantFile(...args),
         })
         .then((buffer) => {
+          let messageData: any =  readFileSync(
+             path.join(sourceDir, '_locales', manifestData.default_locale,
```
kumar303: In translation, the Flow error is saying "`manifestData.default_locale` could be undefined. This type is incompatible with `path.join(...parts: Array<string>)` (specifically, it's incompatible with the `string` part).

Does that help you pinpoint the problem you need to fix?

src/cmd/build.js
```diff
@@ -60,8 +61,21 @@ function defaultPackageCreator(
...
         .then((buffer) => {
+          let messageData: any;
+          if (manifestData.default_locale) {
+            messageData = readFileSync(
+               path.join(sourceDir, '_locales', manifestData.default_locale,
+                'messages.json'));
+          }
+          let extensionName: string;
+          if (messageData) {
+            extensionName = JSON.parse(messageData).extensionName.description;
```
jtojnar: Shouldn’t this be `JSON.parse(messageData).extensionName.message`?

jtojnar: Also `name` does not always have to be `__MSG_extensionName__` – I, for example, use `__MSG_addon_name__`. Maybe this whole issue would better be handled in when generating [manifestData](https://github.com/mozilla/web-ext/blob/2c06dbd892feb02bd69a0345bee2b4d2d98233b1/src/util/manifest.js)

jtojnar: `name` can even be something like `__MSG_hello__foo __MSG_bar_baz__ qux` so using something like the following code should cover it.

```javascript
messageData = ...
manifestData.name = manifestData.name.replace(/__MSG_(.+?)__/g, function(matches) {
	if (messageData[matches[1]] && messageData[matches[1]].message) {
		return messageData[matches[1]].message;
	} else {
		return matches[0];
	}
});
```

Standard8: I'd happened to start playing around with this before seeing this PR. For the matching I was using `/__MSG_([A-Za-z0-9@_]+?)__/` as that's what Firefox actually uses.

You'll also want to use the second argument to the function that's passed to `replace` rather than `matches[1]`.

jtojnar: I tried to grep `__MSG_` in Firefox source but could not find it so I had to verify it experimentally.

You are right about the callback arguments, probably some habit I brought from PHP world.

Standard8: For reference, its here: https://dxr.mozilla.org/mozilla-central/rev/8a494adbc5cced90a4edf0c98cffde906bf7f3ae/toolkit/components/extensions/ExtensionUtils.jsm#799

kumar303: ah, yes, good catch. The `description` explains what the field is that is being localized. The `message` provides the value for that field. https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization

kumar303: For the purpose of a default filename (well, you can't customize the name *yet*), we just have to guess at which field works best. I think it makes sense to pick `messageData.extensionName`.

kumar303:
Hi @shubheksha. We landed a big change on master that switches to using the [async/await](https://ponyfoo.com/articles/understanding-javascript-async-await) strategy instead of promises. You'll have to re-apply your patch using the new strategy. I would suggest leaving this PR open (as reference) and starting a new PR in a new branch. Let me or @rpl know if you need further guidance.

kumar303:
This is looking good. Next, it needs a test case to prove that an extension with a localized manifest will have its built extension named correctly.

Take a look at [this existing test](https://github.com/shubheksha/web-ext/blob/b0871c7ffb3759aacf151149162d8596ca5e2146/tests/unit/test-cmd/test.build.js#L19-L27) (and look closely at the code I highlighted). You can make a new test with a new fixture to test the localized filename.

src/cmd/build.js
jtojnar: @kumar303 I do not understand why would we have to guess. Translating the name is straightforward as I have outlined in the code above.

tests/fixtures/minimal-localizable-web-ext/_locales/en/messages.json
```diff
@@ -0,0 +1,20 @@
...
+      "id": "minimal-localizable-example@web-ext-test-suite"
+    }
+  },
+  "browser_action": {},
+  "permissions": [
+    "tabs"
+  ],
+  "background": {
+    "scripts": ["background-script.js"]
+  }
```
jtojnar: Why are there the lines 6–19?

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +37,25 @@ describe('build', () => {
...
+        build({
+          sourceDir: fixturePath('minimal-localizable-web-ext'),
+          artifactsDir: tmpDir.path(),
+        })
+        .then((buildResult) => {
+          assert.match(buildResult.extensionPath,
+                       /minimal_localizable_extension-1\.0\.zip$/);
+          return buildResult.extensionPath;
+        })
+        .then((extensionPath) => zipFile.open(extensionPath))
```
jtojnar: What does this line do?

tests/fixtures/minimal-localizable-web-ext/manifest.json
```diff
@@ -0,0 +1,19 @@
...
+      "id": "minimal-example@web-ext-test-suite"
+    }
+  },
+  "browser_action": {},
+  "permissions": [
+    "tabs"
+  ],
+  "background": {
+    "scripts": ["background-script.js"]
+  },
```
jtojnar: Why `permissions`, `browser_action` or `applications`?

src/cmd/build.js
```diff
@@ -55,9 +55,21 @@ async function defaultPackageCreator(
...
+  let extensionName: string;
+  if (messageData) {
+    extensionName = JSON.parse(messageData).extensionName.message;
+  }
+  else {
+    extensionName = manifestData.name;
+  }
   let packageName = safeFileName(
-    `${manifestData.name}-${manifestData.version}.zip`);
+            `${extensionName}-${manifestData.version}.zip`);
```
kumar303: nit: this is indented too far. It only needs to be two spaces.

src/cmd/build.js
```diff
@@ -55,9 +55,21 @@ async function defaultPackageCreator(
...
-
+  let messageData: any;
+  if (manifestData.default_locale) {
+    messageData = readFileSync(
+     path.join(sourceDir, '_locales', manifestData.default_locale,
+     'messages.json'));
+  }
+  let extensionName: string;
+  if (messageData) {
+    extensionName = JSON.parse(messageData).extensionName.message;
```
kumar303: Since you need to add a new test case, it will help to break this block of code into a separate "unit" so it can be tested in isolation more easily. To do that, move all of this code to a new `async` function and call it like this:

````javascript
let extensionName: string;
if (manifestData.default_locale) {
  extensionName = await getDefaultLocalizedName({manifestData, sourceDir});
} else {
  extensionName = manifestData.name;
}
````
This will allow you to test for JSON syntax errors without calling the entire `build()` function.

File: src/cmd/build.js [hunk 9]
```diff
@@ -55,9 +55,21 @@ async function defaultPackageCreator(
   let buffer = await zipDir(sourceDir, {
     filter: (...args) => fileFilter.wantFile(...args),
   });
-
+  let messageData: any;
+  if (manifestData.default_locale) {
+    messageData = readFileSync(
+     path.join(sourceDir, '_locales', manifestData.default_locale,
+     'messages.json'));
```
kumar303: This needs to be indented to show that it is a continuation of `path.join()`

src/cmd/build.js
kumar303: @jtojnar oh I see! I misread your comment at first. 

@shubheksha yes, your patch needs to be updated with a regular expression [like this one](https://github.com/mozilla/web-ext/pull/487#discussion_r79083775) so that it can support more than just the `extensionName` key.

kumar303:
Be sure to address [this concern](https://github.com/mozilla/web-ext/pull/487#discussion_r79917127) -- the comment provides an example for how to support more than just `extensionName`.

src/cmd/build.js
jtojnar: @kumar303 also wouldn’t it be better to localize the `extensionName` directly when generating the [manifestData](https://github.com/mozilla/web-ext/blob/2c06dbd892feb02bd69a0345bee2b4d2d98233b1/src/util/manifest.js) instead of during build and possibly in other places in the future?

kumar303: I think we could maybe move it there in the future. It will be easier to review a patch with a standalone function first though. The problem with adding it to `getValidatedManifest()` is not every part of jpm needs this localization code (which does some I/O so it's not cheap). We'll probably need to return an object and give it a method like `manifest.getExtensionName()`.

src/cmd/build.js
```diff
@@ -55,21 +61,35 @@ async function defaultPackageCreator(
...
+    let extensionName: string;
+
+    if (manifestData.default_locale) {
+      messageData = await fs.readFile(path.join(sourceDir, '_locales',
+                                    manifestData.default_locale,
+                                    'messages.json'));
+    }
+    extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/,
+                    function(match, messageName) {
+                      if (messageData[messageName]
```
shubheksha: This branch is never true. My guess is that the value of messageData is still undefined when this line is executed. As a result, `match` which is `_msg_extensionname_` is being returned always & making the test fail.

jtojnar: Maybe also test for `messageData`. (`messageData && messageData[messageName]`)

shubheksha:
@kumar303 - I believe I will have to add a test for extensions which aren't localized to include that one statement which is decreasing the coverage. Or can it be included in some existing test?

src/cmd/build.js
```diff
@@ -40,11 +42,43 @@ export type PackageCreatorParams = {
...
+  manifestData: ExtensionManifest,
+  messageData: any,
+};
+
+export function getDefaultLocalizedName(
+  {manifestData, messageData}: LocalizedNameParams
+  ): string {
+  let extensionName: string;
+
+  if (messageData) {
```
kumar303: This variable is not optional (because of how `LocalizedNameParams` typed it) so you don't need to check for its presence in an if statement. It is a required variable.

src/cmd/build.js
```diff
@@ -40,11 +42,43 @@ export type PackageCreatorParams = {
...
+    try {
+      messageData = JSON.parse(messageData);
+    }
+    catch (err) {
+      err.message = 'The JSON file is malformed';
+      let error = new WebExtError(err.message);
```
kumar303: In this situation you want to preserve the original error but also make the error more readable. I would suggest an approach like this:

    throw new WebExtError(`The JSON file is malformed: ${err}`);

This will throw something like "The JSON file is malformed: SyntaxError: unexpected token at column 45"

src/cmd/build.js
```diff
@@ -56,8 +90,17 @@ async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
+  let messageData: any;
+  if (manifestData.default_locale) {
+    messageData = await fs.readFile(path.join(sourceDir, '_locales',
```
kumar303: The `readFile()` function needs to move into `getDefaultLocalizedName()` because when you throw an error about malformed JSON, the error should include the filename. This will help the developer track down and fix the JSON syntax error more easily.

At first I thought we could avoid it but it needs to include the filename. Unfortunately, this means testing for syntax errors is a bit harder. You will have to pass it a real file instead of a string of JSON. You can use `withTempDir` to write a temporary file for the test. It would look something like:

````javascript
it('checks locale file for malformed json', () => {
  return withTempDir(
    (tmpDir) => {
      const messageFile = path.join(tmpDir, 'messages.json');
      return fs.writeFile(messageFile, '{"simulated:" "json syntax error"'))
        .then(() => {
          // assertion code
        });
    }
  );
});
````

File: src/cmd/build.js [hunk 14]
```diff
@@ -40,11 +42,43 @@ export type PackageCreatorParams = {
...
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
+export type LocalizedNameParams = {
+  manifestData: ExtensionManifest,
+  messageData: any,
+};
+
+export function getDefaultLocalizedName(
+  {manifestData, messageData}: LocalizedNameParams
+  ): string {
```
kumar303: it should look like:
````
export function getDefaultLocalizedName(
  {manifestData, messageData}: LocalizedNameParams
): string {
````

File: src/cmd/build.js [hunk 15]
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+}
+
 export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
+export async function getDefaultLocalizedName(
+  {messageFile, manifestData}: LocalizedNameParams
+): Promise<string> {
+
+  let messageData: any;
```
kumar303: try to only use `any` if you have a really good reason :) The `messageData` should always be a string but some internal Node functions return Buffer instances (which are compatible with strings) so I think it should be `messageData: string|Buffer`

src/cmd/build.js
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+  }
+
+  try {
+    messageData = JSON.parse(messageData);
+  }
+  catch (err) {
+    throw new WebExtError(`The JSON file is malformed: ${err}`);
+  }
+  extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
+                      (match, messageName) => {
```
kumar303: nit: this block should only be indented two spaces.

src/cmd/build.js
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+                        log.info(messageData[messageName]);
```
kumar303: this was probably just for development so you can remove it

src/cmd/build.js
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+                        if (!(messageData[messageName]
+                            && messageData[messageName].message)) {
+                          let error = new WebExtError
```
kumar303: always use `const` for variables unless you intend to modify them

src/cmd/build.js
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+    messageData = JSON.parse(messageData);
+  }
+  catch (err) {
+    throw new WebExtError(`The JSON file is malformed: ${err}`);
+  }
+  extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
+                      (match, messageName) => {
+                        log.info(messageData[messageName]);
+                        if (!(messageData[messageName]
+                            && messageData[messageName].message)) {
```
kumar303: this needs to be indented to convey the nested parentheses statement better. I'd suggest this:
````
if (!(messageData[messageName]
      && ...)) {
````
This helps the reader see that `&&` is inside the parenthesis

File: src/cmd/build.js [hunk 20]
```diff
@@ -37,14 +39,56 @@ export type PackageCreatorParams = {
...
+                          let error = new WebExtError
+                        ('The locale file does not have the correct format');
```
kumar303: This should give the developer a clue as to which key was missing by providing the key value in the error message. Something like ``The locale file ${messageFile} is missing a message for key: ${messageName}``

src/cmd/build.js
```diff
@@ -56,8 +100,19 @@ async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
+  let messageFile: string;
+  let extensionName: string = manifestData.name;
+
+  if (manifestData.default_locale) {
+    messageFile = path.join(sourceDir, '_locales',
+                                    manifestData.default_locale,
```
kumar303: nit: this only needs to be indented two spaces unless you want to align it to the opening parenthesis of `path.join(` for readability.

src/cmd/build.js
```diff
@@ -56,8 +100,19 @@ async function defaultPackageCreator(
...
+                                    'messages.json');
+    log.info(messageFile);
+    extensionName = await getDefaultLocalizedName
+                                      ({messageFile, manifestData});
```
kumar303: always break lines after an opening parenthesis (not before), like this:
````
getDefaultLocalizedName(
  {messageFile, manifestData});
````
This helps the reader follow the nesting easier

File: tests/unit/test-cmd/test.build.js [hunk 23]
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
     );
   });
 
+  it('gives the correct name to an extension that is not localized', () => {
```
kumar303: I don't think you need this test. The same assertion happens in https://github.com/shubheksha/web-ext/blob/8cd7c6659f244094f079e4ddd1ccd46e16bdc5a5/tests/unit/test-cmd/test.build.js#L28-L35 (although, perhaps the title of that test could be updated to say so)

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+        })
+    );
+  });
+
+  it('checks locale file for malformed json', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const messageFileName = path.join(tmpDir.path(), 'messages.json');
+        fs.writeFileSync(messageFileName,
+                '{"simulated:" "json syntax error"');
```
kumar303: nit: only indent two spaces

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+        return getDefaultLocalizedName({
+          messageFile: messageFileName,
+          manifestData: manifestWithoutApps,
+        })
+        .then(makeSureItFails())
```
kumar303: This indentation is misleading because `.then()` is on the same level as the `return` block yet it's a continuation of the `return` statement. I suggest writing it like this:
````
return getDefaultLocalizedName(
  {
    messageFile: '...',
  })
  .then(makeSureItFails());
````
I know it's not pretty :) It will all be converted to async/await soon

File: tests/unit/test-cmd/test.build.js [hunk 26]
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+        .catch((error) => {
+          assert.equal(
+            error instanceof WebExtError, true);
```
kumar303: use [assert.instanceOf()](http://chaijs.com/api/assert/#method_instanceof)

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+          assert.match(error.message, /The JSON file is malformed/);
```
kumar303: I would like to see some of the caught and re-thrown error here to test that the code includes the original error, like:
````
 /The JSON file is malformed: the original error.../
````

File: tests/unit/test-cmd/test.build.js [hunk 28]
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+        });
+      }
+    );
+  });
+
+  it('checks locale file for incorrect format', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const messageFileName = path.join(tmpDir.path(), 'messages.json');
+        fs.writeFileSync(messageFileName,
```
kumar303: Since it's not totally obvious, I'd suggest putting a comment above this like "This is missing the 'message' key"

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +44,94 @@ describe('build', () => {
...
+      }
+    );
+  });
+
+  it('throws an error if the locale file doesn\'t exist', () => {
+    return getDefaultLocalizedName({
+      messageFile: '/path/to/non-existent-dir/messages.json',
+      manifestData: manifestWithoutApps,
+    })
+    .then(makeSureItFails())
```
kumar303: same indentation nit applies here 🌵

tests/unit/test-util/test.manifest.js
```diff
@@ -21,6 +21,16 @@ export const basicManifest = {
   },
 };
 
+export const basicLocalizedManifest = {
```
kumar303: since this is only needed in `test.build.js` I say define it there (for now)

tests/unit/test-util/test.manifest.js
```diff
@@ -21,6 +21,16 @@ export const basicManifest = {
...
+  name: '__MSG_name__',
```
kumar303: I think `__MSG_extensionName__` would be more realistic

src/cmd/build.js
```diff
@@ -37,14 +39,55 @@ export type PackageCreatorParams = {
   artifactsDir: string,
 };
 
+export type LocalizedNameParams = {
+  messageFile: string,
+  manifestData: ExtensionManifest,
+}
+
 export type PackageCreatorFn =
```
kumar303: This used to be defined right above `defaultPackageCreator` so you will need to move it back there. This helps group the type definitions with the functions they relate to.

src/cmd/build.js
```diff
@@ -37,14 +39,55 @@ export type PackageCreatorParams = {
...
+): Promise<string> {
+
+  let messageData: string|Buffer;
+  let extensionName: string = manifestData.name;
+
+  try {
+    messageData = await fs.readFile(messageFile);
+  }
+  catch (err) {
+    log.info(err);
```
kumar303: this can be removed since the error will be thrown

src/cmd/build.js
```diff
@@ -37,14 +39,55 @@ export type PackageCreatorParams = {
...
+    throw new WebExtError(`${err}`);
+  }
+
+  try {
+    messageData = JSON.parse(messageData);
+  }
+  catch (err) {
+    throw new WebExtError(`The JSON file is malformed: ${err}`);
```
kumar303: This can be a `UsageError` since it's fixable. This message needs to include the filename so that the developer can track down and fix the issue. The same comment applies to the error up above so I think you can just group these together like:
````
try {
  messageData = JSON.parse(await fs.readFile(messageFile));
} catch (error) {
  throw new UsageError(`Error reading or parsing message file ${messageFile}: ${error}`);
}
````
I also changed `err` to `error` because I think it's easier to read. Usually abbreviations are fine but only if they significantly cut down on the number of characters needed otherwise.

File: src/cmd/build.js [hunk 35]
```diff
@@ -37,14 +39,55 @@ export type PackageCreatorParams = {
...
+  }
+  extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
+    (match, messageName) => {
+      if (!(messageData[messageName]
+            && messageData[messageName].message)) {
+        const error = new WebExtError
```
kumar303: when adding a line break in a function call, I'd like to see a break on the opening parenthesis because it indicates that the block is a continuation of the line before it:
````
const error = new WebExtError(
  'The locale file messages.json...'
);
````

File: src/cmd/build.js [hunk 36]
```diff
@@ -37,14 +39,55 @@ export type PackageCreatorParams = {
...
+          ('The locale file messages.json ' +
+            `for key: ${messageName}`);
```
kumar303: I don't understand this message. I think it should be something more like this, right? `The locale file ${messageFile} is missing key ${messageName}`

Also, it can be a `UsageError` since this is fixable by the developer. This error means they tried to use a localization key that hasn't been defined.

src/cmd/build.js
```diff
@@ -56,8 +99,17 @@ async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
+  let messageFile: string;
+  let extensionName: string = manifestData.name;
+
+  if (manifestData.default_locale) {
+    messageFile = path.join(sourceDir, '_locales',
```
kumar303: I think you can declare this variable here since it's not used outside of the if statement anymore. Like: `let messageFile = path.join...`

tests/unit/test-cmd/test.build.js
```diff
@@ -5,11 +5,37 @@ import {it, describe} from 'mocha';
...
+import {basicManifest,
+  manifestWithoutApps,
+} from '../test-util/test.manifest';
+import {WebExtError} from '../../../src/errors';
+import {createLogger} from '../../../src/util/logger';
 
+const basicLocalizedManifest = {
+  name: '__MSG_extensionName__',
+  version: '0.0.1',
+  applications: {
```
kumar303: you can remove the `applications` part since it's not required and not needed for the test

tests/unit/test-cmd/test.build.js
```diff
@@ -5,11 +5,37 @@ import {it, describe} from 'mocha';
...
+    gecko: {
+      id: 'basic-manifest@web-ext-test-suite',
+    },
+  },
+};
+
+const basicLocalizedManifestWithRepeatingPattern = {
```
kumar303: this object and also `basicLocalizedManifest` are only used once in each of their respective tests. That means they are not shared code and thus do not need to be defined at the top of the test suite file. You should define them in each test where they are needed so that someone reading the code can see everything in one place.

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+        .then((buildResult) => {
+          assert.match(buildResult.extensionPath,
+                       /name_of_the_extension-1\.0\.zip$/);
+          return buildResult.extensionPath;
+        })
+    );
+  });
+
+  it('gives the correct name to a localized' +
+    'extension with repeating pattern in manifest', () => {
```
kumar303: You could shorten this to make it under the 80 char limit to: `it('handles repeating localization keys')`

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+    return withTempDir(
+      (tmpDir) => {
+        const messageFileName = path.join(tmpDir.path(), 'messages.json');
+        //This is missing the 'message' key
+        fs.writeFileSync(messageFileName,
+          `{"extensionName": {
+              "message": "example extension",
+              "description": "example extension"
```
kumar303: I would call this key `example description` just to make sure the code doesn't accidentally access `messageData.description` one day.

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+            }
+          }`);
+        return getDefaultLocalizedName({
+          messageFile: messageFileName,
+          manifestData: basicLocalizedManifestWithRepeatingPattern,
+        })
+          .then((result) => {
+            assert.match(result, /example extensionexample extension/);
```
kumar303: this would be easier to read if it had a space (or something) in it: `example extension example extension`

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+        return getDefaultLocalizedName({
+          messageFile: messageFileName,
+          manifestData: basicLocalizedManifest,
+        })
+          .then(makeSureItFails())
+          .catch((error) => {
+            assert.instanceOf(error, WebExtError);
+            assert.match(
+              error.message,
+              /The locale file .*messages\.json .* key: extensionName/);
```
kumar303: yep, you got it!

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+          });
+      }
+    );
+  });
+
+  it('throws an error if the locale file doesn\'t exist', () => {
```
kumar303: you could cheat and say `does not` to avoid an escaped apostrophe :)

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +63,103 @@ describe('build', () => {
...
+    return getDefaultLocalizedName({
+      messageFile: '/path/to/non-existent-dir/messages.json',
+      manifestData: manifestWithoutApps,
+    })
+    .then(makeSureItFails())
```
kumar303: this needs to be indented two spaces to show that it's a continuation of the call to `getDefaultLocalizedName()`

src/cmd/build.js
```diff
@@ -37,14 +39,48 @@ export type PackageCreatorParams = {
...
+  catch (error) {
+    throw new UsageError(
+      `Error reading or parsing file ${messageFile}: ${error}`);
+  }
+  extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
+    (match, messageName) => {
+      if (!(messageData[messageName]
+            && messageData[messageName].message)) {
+        const error = new UsageError(
+          'The locale file messages.json ' +
```
kumar303: This needs to be `${messageFile}` so that the developer sees the exact path to the file that has the missing key

src/cmd/build.js
```diff
@@ -56,8 +92,16 @@ async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
+  let extensionName: string = manifestData.name;
+
+  if (manifestData.default_locale) {
+    let messageFile = path.join(sourceDir, '_locales',
+                            manifestData.default_locale, 'messages.json');
```
kumar303: only indent new blocks by two spaces unless you're aligning it with an opening paren or brace.

kumar303: oh, I think this just shifted after you added the `let`

tests/fixtures/minimal-localizable-web-ext/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "manifest_version": 2,
+  "name": "__MSG_extensionName__",
+  "description": "__MSG_extensionDescription__",
+  "version": "1.0",
+  "applications": {
```
kumar303: You should probably remove `applications` here as well since it's not needed for the test.

tests/unit/test-cmd/test.build.js
```diff
@@ -5,11 +5,17 @@ import {it, describe} from 'mocha';
 import {assert} from 'chai';
 import sinon from 'sinon';
 
-import build, {safeFileName, FileFilter} from '../../../src/cmd/build';
+import build, {safeFileName, FileFilter,
+                getDefaultLocalizedName} from '../../../src/cmd/build';
```
kumar303: this  indentation is off

tests/unit/test-cmd/test.build.js
```diff
@@ -5,11 +5,17 @@ import {it, describe} from 'mocha';
...
 import {withTempDir} from '../../../src/util/temp-dir';
 import {fixturePath, makeSureItFails, ZipFile} from '../helpers';
-import {basicManifest, manifestWithoutApps} from '../test-util/test.manifest';
+import {basicManifest,
```
kumar303: line breaks like this need to remain consistent. Either keep it on one line or break all elements into the next line like:
````
import {
  basicManifest,
  manifestWithoutApps,
} from '../test-util/test.manifest';
````

File: tests/fixtures/minimal-localizable-web-ext/_locales/en/messages.json [hunk 51]
```diff
@@ -0,0 +1,16 @@
+{
+  "extensionName": {
+    "message": "Name of the extension",
+    "description": "This is an example of a minimal extension that does nothing"
+  },
+
+  "version": "1.0",
```
kumar303: why does this include `version`, `applications`, and `background`? They are not needed. It only needs `extensionName` or possibly some other localization keys.

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+          return buildResult.extensionPath;
+        })
+    );
+  });
+
+  it('handles repeating localization keys', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const messageFileName = path.join(tmpDir.path(), 'messages.json');
+        //This is missing the 'message' key
```
kumar303: this comment is incorrect

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+        fs.writeFileSync(messageFileName,
+          `{"extensionName": {
+              "message": "example extension ",
```
kumar303: heh, you are sneaky ;) but I think it would make more sense to put the space here instead: `'__MSG_extensionName__ __MSG_extensionName__'`

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+              "description": "example description"
+            }
+          }`);
+
+        const basicLocalizedManifestWithRepeatingPattern = {
```
kumar303: this is a really long name. How about `manifestWithRepeatingPattern`?

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+        return getDefaultLocalizedName({
+          messageFile: messageFileName,
+          manifestData: manifestWithoutApps,
+        })
+          .then(makeSureItFails())
+          .catch((error) => {
+            assert.instanceOf(error, UsageError);
+            assert.match(
+              error.message,
+              /Error .* file .*messages\.json: SyntaxError: Unexpected string/);
```
kumar303: Nice

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+  it('throws an error if the locale file does not exist', () => {
+    return getDefaultLocalizedName({
+      messageFile: '/path/to/non-existent-dir/messages.json',
+      manifestData: manifestWithoutApps,
+    })
+      .then(makeSureItFails())
+      .catch((error) => {
+        log.info(error);
+        assert.instanceOf(error, WebExtError);
+        assert.match(error.message, /ENOENT: no such file or directory/);
```
kumar303: This also needs to make sure that `messages.json` is included in the error text

tests/unit/test-cmd/test.build.js
```diff
@@ -37,6 +43,112 @@ describe('build', () => {
...
+
+  it('throws an error if the locale file does not exist', () => {
+    return getDefaultLocalizedName({
+      messageFile: '/path/to/non-existent-dir/messages.json',
+      manifestData: manifestWithoutApps,
+    })
+      .then(makeSureItFails())
+      .catch((error) => {
+        log.info(error);
+        assert.instanceOf(error, WebExtError);
```
kumar303: specifically, it should be an instance of UsageError

src/cmd/build.js
```diff
@@ -56,8 +92,16 @@ async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
+  let extensionName: string = manifestData.name;
+
+  if (manifestData.default_locale) {
```
kumar303: After this if statement, I think we need to add a debugging statement that will show up in `--verbose` mode, like:

    log.debug('Manifest declared default_locale, localizing extension name');

This will help developers understand why they might see errors about a missing a messages file.

src/cmd/build.js
shubheksha: How do I handle the error message in that case?

src/cmd/build.js
```diff
@@ -37,14 +39,48 @@ export type PackageCreatorParams = {
...
+    throw new UsageError(
+      `Error reading or parsing file ${messageFile}: ${error}`);
+  }
+  extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
+    (match, messageName) => {
+      if (!(messageData[messageName]
+            && messageData[messageName].message)) {
+        const error = new UsageError(
+          `The locale file ${messageFile} ` +
+            `is missing key: ${messageName}`);
```
kumar303: I see that you indented this to show continuation but it actually wouldn't need indenting since it's not a *new* block. It's just an extension of the argument being passed to `UsageError()`.

No need to fix this since it's so minor, just something to keep in mind.

Issue #220, opened 2016-04-26
Also check errno when catching errors having a specific code


kumar303:
There is a [helper](https://github.com/mozilla/web-ext/blob/master/src/errors.js#L70) in web-ext which is used like:

    thing().then(...).catch(onlyErrorsWithCode('ECONNREFUSED', ...));

It looks for `error.code` but in NodeJS `error.errno` is an [alias](https://nodejs.org/api/errors.html#errors_system_errors) for this. The `onlyErrorsWithCode()` function should also check `error.errno`.

wagnerand:
See https://wiki.mozilla.org/Add-ons/Contribute/AMO/Code if this is your first web-ext bug you work on.

Mentor for this bug is @kumar303.

Issue #335, opened 2016-06-29
flaky test: build lets you rebuild when files change


kumar303:
I see this a lot. Restarting the build always fixes it:

````
  1) build lets you rebuild when files change:

     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
````

https://github.com/mozilla/web-ext/blob/master/tests/test-cmd/test.build.js#L90

PR #488, opened 2016-09-15
fix: Added errno for errors with specific codes

rackstar17:
Fixes https://github.com/mozilla/web-ext/issues/220

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -83,8 +83,10 @@ export function onlyErrorsWithCode(
     let throwError = true;
 
     if (Array.isArray(codeWanted) && codeWanted.indexOf(error.code) !== -1) {
-      throwError = false;
-    } else if (error.code === codeWanted) {
+      if (codeWanted.indexOf(error.errno) !== -1) {
+        throwError = false;
+      }
+    } else if (error.code === codeWanted && error.errno === codeWanted) {
       throwError = false;
     }
 
--- a/src/errors.js
+++ b/src/errors.js
@@ -78,7 +78,8 @@ export function onlyInstancesOf(
  *
  */
 export function onlyErrorsWithCode(
-    codeWanted: string | Array<string>, errorHandler: Function): Function {
+    codeWanted: (string|number) | Array<string|number>,
+      errorHandler: Function): Function {
   return (error) => {
     let throwError = true;
 
--- a/src/errors.js
+++ b/src/errors.js
@@ -83,10 +83,8 @@ export function onlyErrorsWithCode(
     let throwError = true;
 
     if (Array.isArray(codeWanted)) {
-      if (codeWanted.indexOf(error.code) !== -1) {
-        throwError = false;
-      }
-      if (codeWanted.indexOf(error.errno) !== -1) {
+      if (codeWanted.indexOf(error.code) !== -1 ||
+          codeWanted.indexOf(error.errno) !== -1) {
         throwError = false;
       }
     }
--- a/src/errors.js
+++ b/src/errors.js
@@ -82,11 +82,15 @@ export function onlyErrorsWithCode(
   return (error) => {
     let throwError = true;
 
-    if (Array.isArray(codeWanted) && codeWanted.indexOf(error.code) !== -1) {
+    if (Array.isArray(codeWanted)) {
+      if (codeWanted.indexOf(error.code) !== -1) {
+        throwError = false;
+      }
       if (codeWanted.indexOf(error.errno) !== -1) {
         throwError = false;
       }
-    } else if (error.code === codeWanted && error.errno === codeWanted) {
+    }
+    else if (error.code === codeWanted || error.errno === codeWanted) {
       throwError = false;
     }
 
--- a/tests/unit/test.errors.js
+++ b/tests/unit/test.errors.js
@@ -34,9 +34,11 @@ describe('errors', () => {
 
     class ErrorWithCode extends Error {
       code: string;
+      errno: string;
       constructor() {
         super('pretend this is a system error');
         this.code = 'SOME_CODE';
+        this.errno = 'SOME_CODE';
       }
     }
 
@@ -44,6 +46,13 @@ describe('errors', () => {
       return Promise.reject(new ErrorWithCode())
         .catch(onlyErrorsWithCode('SOME_CODE', (error) => {
           assert.equal(error.code, 'SOME_CODE');
+        }));
+    });
+
+    it('catches errors having a error no', () => {
+      return Promise.reject(new ErrorWithCode())
+        .catch(onlyErrorsWithCode('SOME_CODE', (error) => {
+          assert.equal(error.errno, 'SOME_CODE');
         }));
     });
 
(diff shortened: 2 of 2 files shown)


kumar303:
Hi. Thanks for getting this started!

I think it would be easier to start this by first adding a test that expects `errno` to be caught. You could model the test on this one: https://github.com/mozilla/web-ext/blob/ee197c4a0e045c282cb676284f2c3fb988c4b828/tests/unit/test.errors.js#L43

The test suite output is going to be messy when you work on this feature because the function, `onlyErrorsWithCode()`, is used by many other tests. To hide the irrelevant failures, you can run only the tests for this specific function like this:

    ./node_modules/.bin/mocha -r babel-core/register tests/unit/test.errors.js

rackstar17:
@kumar303 Hi. I tried to do what you said but I can't resolve the travis build issues 
I also tried to run the test.errors.js file as you mentioned and locally all the tests were passed for test.errors.js . Can you help me with it ?

kumar303:
The test failure may have been a timeout caused by slowness in the TravisCI system, not with your patch. I have restarted the test to see if it still fails repeatedly.

src/errors.js
```diff
@@ -82,9 +82,15 @@ export function onlyErrorsWithCode(
...
     let throwError = true;
 
-    if (Array.isArray(codeWanted) && codeWanted.indexOf(error.code) !== -1) {
-      throwError = false;
-    } else if (error.code === codeWanted) {
+    if (Array.isArray(codeWanted)) {
+      if (codeWanted.indexOf(error.code) !== -1) {
+        throwError = false;
+      }
+      if (codeWanted.indexOf(error.errno) !== -1) {
```
kumar303: You can group these statements to make the code more concise and less repetitive. Example:
````javascript
if (codeWanted.indexOf(error.code) !== -1 ||
    codeWanted.indexOf(error.errno) !== -1) {
  throwError = false;
}

File: tests/unit/test.errors.js [hunk 2]
```diff
@@ -34,9 +34,11 @@ describe('errors', () => {
 
     class ErrorWithCode extends Error {
       code: string;
+      errno: string;
```
kumar303: instead of editing this class, I think you need to create a new class named something like `ErrorWithErrno` and add `errno` to that one only. Otherwise, the existing tests for error codes will not be as effective.

kumar303:
> ... may have been a timeout caused by slowness in the TravisCI system

Yes, this was it. I restarted the build and the tests are passing now. Sorry for the confusion! This doesn't happen too often but it's a [known problem](https://github.com/mozilla/web-ext/issues/335) with this test.

src/errors.js
pdehaan: You could _possibly_ even use [`Array#includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), which I think is supported in Firefox 43+.

kumar303: NodeJS is our target platform with 4.0 being the minimum. I just checked and sadly 4.0 doesn't have this function.

rackstar17: I added the || code to reduce the repetitiveness of if statements but there was this issue of not having more than 80 columns in a single line. 
It could be done this way, didn't think of that earlier .  
Done.

tests/unit/test.errors.js
rackstar17: Done

tests/unit/test.errors.js
```diff
@@ -40,13 +40,28 @@ describe('errors', () => {
       }
     }
 
+    class ErrorWithErrno extends Error {
+      errno: string;
```
kumar303: this should be `errno: number`

tests/unit/test.errors.js
```diff
@@ -40,13 +40,28 @@ describe('errors', () => {
...
+      constructor() {
+        super('pretend this is a system error');
+        this.errno = 'SOME_CODE';
```
kumar303: this should be something like `this.errno = 53`, which means [ENOTEMPTY](https://github.com/rvagg/node-errno)

tests/unit/test.errors.js
```diff
@@ -40,13 +40,28 @@ describe('errors', () => {
...
     it('catches errors having a code', () => {
       return Promise.reject(new ErrorWithCode())
         .catch(onlyErrorsWithCode('SOME_CODE', (error) => {
           assert.equal(error.code, 'SOME_CODE');
         }));
     });
 
+    it('catches errors having a error no', () => {
+      return Promise.reject(new ErrorWithErrno())
+        .catch(onlyErrorsWithCode('SOME_CODE', (error) => {
```
kumar303: and then this would need to be `onlyErrorsWithCode(53, ...)`

tests/unit/test.errors.js
```diff
@@ -65,6 +80,13 @@ describe('errors', () => {
         }));
     });
 
+    it('catches errors having one of many errno', () => {
+      return Promise.reject(new ErrorWithErrno())
+        .catch(onlyErrorsWithCode(['OTHER_CODE', 'SOME_CODE'], (error) => {
```
kumar303: and this would need updating too

tests/unit/test.errors.js
```diff
@@ -76,6 +98,17 @@ describe('errors', () => {
         });
     });
 
+    it('throws errors that are not in an array of errno', () => {
```
kumar303: I don't think you need to include this test. It doesn't introduce any new coverage since the previous one already checks that an array of mismatching codes does not catch any error at all.

tests/unit/test.errors.js
rackstar17: Done.

tests/unit/test.errors.js
rackstar17: Done.

tests/unit/test.errors.js
rackstar17: Done.

tests/unit/test.errors.js
rackstar17: Done.

tests/unit/test.errors.js
rackstar17: Removed.

rackstar17:
@kumar303 Hi, the travis checks have failed . Can you take a look ?
Thanks.

rackstar17:
Ah , I see now .
The thing you suggested to make errno: number , the onlyErrorsWithCode expects it's parameter to be string or array of strings and here we are passing number .

kumar303:
Yes, exactly, we use [Flow](https://flowtype.org/) to incorporate static typing into the code. You'll need to specify that [codeWanted](https://github.com/rackstar17/web-ext/blob/57d9881fedcbba68c5893a2db55772189fef4011/src/errors.js#L81) can take an array of either strings *or* numbers. It's a bit cryptic due to polymorphism so here is the typing fix:

    codeWanted: string | Array<string|number>

rackstar17:
@kumar303 Hi, 
The tests have passed , Can you review it ?
Thanks .

atsay:
Thanks for your contribution, @rackstar17! I've added it to our [recognition page](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#September_2016).

PR #509, opened 2016-09-20
refactor: Show usage exceptions without a stacktrace

### LINKED ISSUE #310: Show usage exceptions without a stacktrace

shubheksha:
Fixes #310 partially.
- Created a UsageError class 
- Converted some of the tests from WebExtError to UsageError

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -13,9 +13,19 @@ export class WebExtError extends ExtendableError {
 
 
 /*
+ * The class for errors that can be fixed by the developer.
+ */
+export class UsageError extends WebExtError {
+  constructor(message: string) {
+    super(message);
+  }
+}
+
+
+/*
  * The manifest for the extension is invalid (or missing).
  */
-export class InvalidManifest extends WebExtError {
+export class InvalidManifest extends UsageError {
   constructor(message: string) {
     super(message);
   }
@@ -26,12 +36,6 @@ export class InvalidManifest extends WebExtError {
  * The remote Firefox does not support temporary add-on installation.
  */
 export class RemoteTempInstallNotSupported extends WebExtError {
-  constructor(message: string) {
-    super(message);
-  }
-}
-
-export class UsageError extends WebExtError {
   constructor(message: string) {
     super(message);
   }
--- a/src/errors.js
+++ b/src/errors.js
@@ -31,6 +31,12 @@ export class RemoteTempInstallNotSupported extends WebExtError {
   }
 }
 
+export class UsageError extends WebExtError {
+  constructor(message: string) {
+    super(message);
+  }
+}
+
 
 /*
  * Sugar-y way to catch only instances of a certain error.
--- a/src/firefox/preferences.js
+++ b/src/firefox/preferences.js
@@ -1,5 +1,5 @@
 /* @flow */
-import {UsageError} from '../errors';
+import {WebExtError} from '../errors';
 
 
 // Flow Types
@@ -110,7 +110,7 @@ export function getPrefs(
 ): FirefoxPreferences {
   const appPrefs = prefs[app];
   if (!appPrefs) {
-    throw new UsageError(`Unsupported application: ${app}`);
+    throw new WebExtError(`Unsupported application: ${app}`);
   }
   return {
     ...prefsCommon,
--- a/src/firefox/preferences.js
+++ b/src/firefox/preferences.js
@@ -1,5 +1,5 @@
 /* @flow */
-import {WebExtError} from '../errors';
+import {UsageError} from '../errors';
 
 
 // Flow Types
@@ -110,7 +110,7 @@ export function getPrefs(
 ): FirefoxPreferences {
   const appPrefs = prefs[app];
   if (!appPrefs) {
-    throw new WebExtError(`Unsupported application: ${app}`);
+    throw new UsageError(`Unsupported application: ${app}`);
   }
   return {
     ...prefsCommon,
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -1,9 +1,10 @@
 /* @flow */
 import {createLogger} from '../util/logger';
-import {RemoteTempInstallNotSupported,
-        UsageError,
-        WebExtError,
-        } from '../errors';
+import {
+  RemoteTempInstallNotSupported,
+  UsageError,
+  WebExtError,
+} from '../errors';
 import defaultFirefoxConnector from 'node-firefox-connect';
 
 
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -1,6 +1,9 @@
 /* @flow */
 import {createLogger} from '../util/logger';
-import {RemoteTempInstallNotSupported, WebExtError} from '../errors';
+import {RemoteTempInstallNotSupported,
+        UsageError,
+        WebExtError,
+        } from '../errors';
 import defaultFirefoxConnector from 'node-firefox-connect';
 
 
@@ -138,7 +141,7 @@ export class RemoteFirefox {
         }
         log.debug(
           `Remote Firefox has these addons: ${addons.map((a) => a.id)}`);
-        throw new WebExtError(
+        throw new UsageError(
           'The remote Firefox does not have your extension installed');
       });
   }
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -109,7 +109,7 @@ export class RemoteFirefox {
           {to: response.addonsActor, type: 'installTemporaryAddon', addonPath},
           (response) => {
             if (response.error) {
-              return reject(new UsageError(
+              return reject(new WebExtError(
                 'installTemporaryAddon: Error: ' +
                 `${response.error}: ${response.message}`));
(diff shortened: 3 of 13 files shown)


src/program.js
```diff
@@ -103,7 +103,9 @@ export class Program {
       await runCommand(argv);
     } catch (error) {
       const prefix = cmd ? `${cmd}: ` : '';
-      log.error(`\n${prefix}${error.stack}\n`);
+      if (!error instanceof UsageError) {
```
kumar303: The coverage report says the `log` statement within this if branch isn't covered but we definitely have a test that covers it. This makes me think the JS interpreter is parsing `!error` before passing the value to `instanceof`. Try this instead and let's see if the code coverage goes back to 100%:
````javascript
if (!(error instanceof UsageError)) {
}
````

File: src/program.js [hunk 2]
```diff
@@ -103,7 +103,9 @@ export class Program {
...
+        log.error(`\n${prefix}${error.stack}\n`);
+      }
       if (error.code) {
         log.error(`${prefix}Error code: ${error.code}\n`);
       }
```
kumar303: comment 1

kumar303: comment 2

src/program.js
```diff
@@ -103,7 +103,9 @@ export class Program {
       await runCommand(argv);
     } catch (error) {
       const prefix = cmd ? `${cmd}: ` : '';
-      log.error(`\n${prefix}${error.stack}\n`);
+      if (!error instanceof UsageError) {
+        log.error(`\n${prefix}${error.stack}\n`);
```
kumar303: you also need to log this line if `argv.verbose` is true, regardless of whether it's a `UsageError` or not.

src/errors.js
```diff
@@ -31,6 +31,12 @@ export class RemoteTempInstallNotSupported extends WebExtError {
   }
 }
 
+export class UsageError extends WebExtError {
```
kumar303: this needs a comment at the top summarizing that this error class is for something that the developer will be able to fix when they see it in the console

kumar303: Also, since `InvalidManifest` is something the developer can fix, you should make this a `UsageError` by changing its base class like:
````javascript
export class InvalidManifest extends UsageError {
}
````

**Comment by shubheksha** [comment]
@kumar303, I addressed all your feedback in the last commit. Please take a look, let me know if I missed something.

File: src/firefox/preferences.js [hunk 5]
```diff
@@ -110,7 +110,7 @@ export function getPrefs(
 ): FirefoxPreferences {
   const appPrefs = prefs[app];
   if (!appPrefs) {
-    throw new WebExtError(`Unsupported application: ${app}`);
+    throw new UsageError(`Unsupported application: ${app}`);
```
kumar303: This should be a `WebExtError` because the developer can't select the type of Firefox application yet. When we support Android this might change but this is an internally used function anyway so probably not.

src/firefox/remote.js
```diff
@@ -106,7 +109,7 @@ export class RemoteFirefox {
           {to: response.addonsActor, type: 'installTemporaryAddon', addonPath},
           (response) => {
             if (response.error) {
-              return reject(new WebExtError(
+              return reject(new UsageError(
```
kumar303: This should be a `WebExtError` because the developer can't recover from it. It indicates a serious problem internal to Firefox's remote debugger.

src/firefox/remote.js
```diff
@@ -138,7 +141,7 @@ export class RemoteFirefox {
         }
         log.debug(
           `Remote Firefox has these addons: ${addons.map((a) => a.id)}`);
-        throw new WebExtError(
+        throw new UsageError(
```
kumar303: This should be a `WebExtError` because the installation process was done automatically by `web-ext`, not the developer. If the extension was not installed, it means there was a bug in `web-ext`. It is unlikely that this error will ever happen.

src/program.js
```diff
@@ -88,11 +88,11 @@ export class Program {
     const cmd = argv._[0];
 
     if (cmd === undefined) {
-      throw new WebExtError('No sub-command was specified in the args');
+      throw new UsageError('No sub-command was specified in the args');
     }
     let runCommand = this.commands[cmd];
     if (!runCommand) {
-      throw new WebExtError(`unknown command: ${cmd}`);
+      throw new UsageError(`unknown command: ${cmd}`);
```
kumar303: while you're in here could you capitalize the first word so it reads "Unknown command: ${cmd}" ? That should be fixed.

src/util/artifacts.js
```diff
@@ -12,7 +12,7 @@ export async function prepareArtifactsDir(
   try {
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
-      throw new WebExtError(`${artifactsDir} is not a directory`);
+      throw new UsageError(`${artifactsDir} is not a directory`);
```
kumar303: It would be easier for developers to understand this if you change the message to: `--artifacts-dir=${artifactsDir} (this value is not a directory)`

kumar303:
I'm writing a note to myself that after this lands, we need to check all open pull requests to make sure they don't need to use the new `UsageError`.

src/firefox/remote.js
kumar303: This one still needs fixing

src/util/artifacts.js
```diff
@@ -12,7 +12,8 @@ export async function prepareArtifactsDir(
   try {
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
-      throw new WebExtError(`${artifactsDir} is not a directory`);
+      throw new UsageError(`artifacts-dir=${artifactsDir} 
+                          (this value is not a directory)`);
```
kumar303: this line should not have so much whitespace in it because it will print like that in the console which will be harder to read. You can get around the line length limit in a couple ways. Like this:
````
throw new UsageError(
  `artifacts-dir=${artifactsDir} (this value is not a directory)`);
````
or like:
````
throw new UsageError(`artifacts-dir=${artifactsDir} ` +
                     `(this value is not a directory)`);
````

File: src/firefox/remote.js [hunk 11]
```diff
@@ -1,6 +1,9 @@
 /* @flow */
 import {createLogger} from '../util/logger';
-import {RemoteTempInstallNotSupported, WebExtError} from '../errors';
+import {RemoteTempInstallNotSupported,
+        UsageError,
+        WebExtError,
+        } from '../errors';
```
kumar303: To remain consistent with the rest of the code, this should be formatted like:
````
import {
  RemoteTempInstallNotSupported,
  UsageError,
  WebExtError,
} from '../errors';
````

File: src/program.js [hunk 12]
```diff
@@ -103,7 +103,9 @@ export class Program {
       await runCommand(argv);
     } catch (error) {
       const prefix = cmd ? `${cmd}: ` : '';
-      log.error(`\n${prefix}${error.stack}\n`);
+      if (!(error instanceof UsageError) || argv.verbose) {
+        log.error(`\n${prefix}${error.stack}\n`);
+      }
```
kumar303: this needs an `else` block that prints the error without any stack information. For example:

    log.error(`\n${prefix}${error}\n`);

kumar303: it may be helpful to use an intermediate variable (like `errorMessage`) so that we only add the `prefix` once.

src/program.js
```diff
@@ -88,11 +88,11 @@ export class Program {
     const cmd = argv._[0];
 
     if (cmd === undefined) {
```
kumar303: None of the exceptions below this line will be caught in the `try` block so their stack traces will not be hidden. I suggest moving the `try` block up above this line so that they are all caught.

src/program.js
```diff
@@ -114,6 +115,11 @@ export class Program {
         throw error;
       }
     }
+
+    if (argv.verbose) {
```
kumar303: was this causing an error? It should go back to the top of the `try` block because it needs to happen before any other log statement is called.

src/util/artifacts.js
```diff
@@ -12,7 +12,8 @@ export async function prepareArtifactsDir(
   try {
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
-      throw new WebExtError(`${artifactsDir} is not a directory`);
+      throw new UsageError(
+        `artifacts-dir=${artifactsDir} (this value is not a directory)`);
```
kumar303: This should be `--artifacts-dir` (with the leading dashes) so that it's clearly a command line option

src/program.js
shubheksha: Nope, this wasn't causing any error at all, it was something completely different.

PR #517, opened 2016-09-22

### LINKED ISSUE #481: print a development version when using a custom web-ext build
For released versions of web-ext you can type this:

    web-ext --version

And you will see something like `1.4.0` or whatever the current version is.

For development, you should be able to type the same thing and see a development version based on the current commit. For example:

    master-c760a77b83cb732ea2202e00194c0b6492ba2632

This would be the auto-generated version if you had done `npm run build` from https://github.com/mozilla/web-ext/commit/c760a77b83cb732ea2202e00194c0b6492ba2632

This involves the following tasks:
* Adjust [getVersion()](https://github.com/mozilla/web-ext/blob/c760a77b83cb732ea2202e00194c0b6492ba2632/src/program.js#L118) to return a development version
* Ensure that the release build includes a real version number and the development build includes a git commit hash
* Add a test [similar to this one](https://github.com/mozilla/web-ext/blob/c760a77b83cb732ea2202e00194c0b6492ba2632/tests/test.program.js#L315) that verifies this new version feature

harshitkgupta:
Fixes #481

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
 - git checkout master
 - git checkout -
 script:
-- COVERAGE=y npm test
+- COVERAGE=y NODE_ENV=production npm test
 # Run changelog-lint but only on newer versions of Node (because of syntax errors)
 - if [[ ${TRAVIS_NODE_VERSION:0:1} -ge "4" ]]; then
     npm run changelog-lint;
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "eslint-plugin-flowtype": "2.18.2",
     "firefox-client": "0.3.0",
     "flow-bin": "0.32.0",
+    "git-rev-sync": "1.6.0",
     "grunt": "1.0.1",
     "grunt-contrib-clean": "1.0.0",
     "grunt-contrib-copy": "1.0.0",
--- a/src/program.js
+++ b/src/program.js
@@ -119,9 +119,14 @@ export class Program {
 
 
 export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
+  if (process.env.NODE_ENV === 'production') {
+    let packageData: any = readFileSync(
     path.join(absolutePackageDir, 'package.json'));
-  return JSON.parse(packageData).version;
+    return JSON.parse(packageData).version;
+  } else {
+    var git = require('git-rev-sync');
+    return `${git.branch()}-${git.long()}`;
+  }
 }
 
 
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -313,7 +313,8 @@ describe('program.main', () => {
 
 describe('program.defaultVersionGetter', () => {
 
-  it('returns the package version', () => {
+  it('returns the production package version', () => {
+    process.env.NODE_ENV = 'production';
     let root = path.join(__dirname, '..', '..');
     let pkgFile = path.join(root, 'package.json');
     return fs.readFile(pkgFile)
@@ -321,6 +322,15 @@ describe('program.defaultVersionGetter', () => {
         assert.equal(defaultVersionGetter(root),
                      JSON.parse(pkgData).version);
       });
+  });
+
+  it('returns the development package version', () => {
+    process.env.NODE_ENV = 'development';
+    let root = path.join(__dirname, '..', '..');
+    let git = require('git-rev-sync');
+    var devVersion = `${git.branch()}-${git.long()}`;
+    assert.equal(defaultVersionGetter(root),
+                     devVersion);
   });
 
 });


tests/unit/test.program.js
```diff
@@ -313,7 +313,8 @@ describe('program.main', () => {
 
 describe('program.defaultVersionGetter', () => {
 
-  it('returns the package version', () => {
+  it('returns the production package version', () => {
+    process.env.NODE_ENV = 'production';
```
kumar303: You can't set a value directly on `process` because that's a global object shared by all other functions in the test suite. You may be setting an environment variable here that would cause another unrelated test to break.

To fix this, you can use a [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) strategy which means you would inject a fake process object into the `defaultVersionGetter()` function for the purpose of testing. This is safe because the process object is a stable interface that we can depend on.

This change requires two things.

First, change the signature of the version function to accept a `process` object to use. *By default* it will use the global `process` object:
````javascript
export function defaultVersionGetter(
  absolutePackageDir: string, 
  {localProcess = process} = {}
): string {
  if (localProcess.env.NODE_ENV === 'production') {
    // ...
  }
}
````

Secondly, update this test to pass in a fake `process` that simulates what a real object would look like. Example:
````javascript
// This doesn't have to implement all process methods, just what we care about.
const fakeProcess = {   
  env: {
    NODE_ENV: 'production',   
  },
};
assert.equal(
  defaultVersionGetter(root, {localProcess: fakeProcess}),
  JSON.parse(pkgData).version);
````

File: tests/unit/test.program.js [hunk 2]
```diff
@@ -323,4 +324,13 @@ describe('program.defaultVersionGetter', () => {
       });
   });
 
+  it('returns the development package version', () => {
+    process.env.NODE_ENV = 'development';
```
kumar303: If you remove the modifications to this global object then you won't need to set this env var here

tests/unit/test.program.js
```diff
@@ -323,4 +324,13 @@ describe('program.defaultVersionGetter', () => {
...
+    let root = path.join(__dirname, '..', '..');
+    let git = require('git-rev-sync');
```
kumar303: this should be imported at the top of the file

tests/unit/test.program.js
```diff
@@ -323,4 +324,13 @@ describe('program.defaultVersionGetter', () => {
       });
   });
 
+  it('returns the development package version', () => {
+    process.env.NODE_ENV = 'development';
+    let root = path.join(__dirname, '..', '..');
```
kumar303: to reduce repetition, set this to the top of the `describe()` function for this group of tests so that it can be referenced by both tests

tests/unit/test.program.js
```diff
@@ -323,4 +324,13 @@ describe('program.defaultVersionGetter', () => {
...
+    let git = require('git-rev-sync');
+    var devVersion = `${git.branch()}-${git.long()}`;
```
kumar303: Always declare variables with `const` (constant) when you don't need to modify them. This will help catch any unintended modifications.

tests/unit/test.program.js
```diff
@@ -323,4 +324,13 @@ describe('program.defaultVersionGetter', () => {
...
+    assert.equal(defaultVersionGetter(root),
+                     devVersion);
```
kumar303: `devVersion` should line up visually with the opening of the parenthesis in `assert.equal()` from the line above

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
 - git checkout master
 - git checkout -
 script:
-- COVERAGE=y npm test
+- COVERAGE=y NODE_ENV=production npm test
```
kumar303: In addition to this change, please document this new behavior in the [build section](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#build-web-ext). I would suggest adding a new paragraph that says:

By default, `npm run build` creates a development build of `web-ext`. To create a production build, use the `NODE_ENV` variable like this:

    NODE_ENV=production npm run build

package.json
```diff
@@ -84,6 +84,7 @@
     "eslint-plugin-flowtype": "2.18.2",
     "firefox-client": "0.3.0",
     "flow-bin": "0.32.0",
+    "git-rev-sync": "1.6.0",
```
kumar303: do you know if `git-rev-sync` works on Windows or not?

src/program.js
```diff
@@ -119,9 +119,14 @@ export class Program {
 
 
 export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
+  if (process.env.NODE_ENV === 'production') {
+    let packageData: any = readFileSync(
```
harshitkgupta: let packageData: string = readFileSync(
                           ^ Buffer. This type is incompatible with string

kumar303: ah, right you are. It should be:

    let packageData: Buffer = readFileSync(...)

Let me know if you have any other questions about the feedback I left.

kumar303:
Hi @harshitkgupta , thanks for getting this patch started. Hopefully it was an interesting experience to learn about the `web-ext` codebase. Another developer had some time to work on this feature so we have closed issue https://github.com/mozilla/web-ext/issues/481 . Thanks again for your interest in the `web-ext` project.

PR #519, opened 2016-09-22
chore: Run the node security platform tool in continuous integration

### LINKED ISSUE #513: Run the node security platform tool in continuous integration
It would be helpful to run the [nsp](https://github.com/nodesecurity/nsp) command as part of our [continuous integration check](https://github.com/mozilla/web-ext/blob/master/.travis.yml) so that we can be sure we're never shipping any known vulnerabilities in `web-ext`.

This should be as simple as adding `npm check` to our `.travis.yml`.

So far so good:
````
$ nsp check
(+) No known vulnerabilities found
````

### PR DESCRIPTION
**Pull request by pdehaan**
Fixes #513

### CODE REVIEW

File: .nsprc [hunk 1]
```diff
@@ -0,0 +1,4 @@
+{
+  // https://github.com/nodesecurity/nsp#exceptions
+  "exceptions": []
+}
```
pdehaan: Technically not really "used" at the moment, but just some pre-emptive scaffolding for when there is a vulnerability in one of the nested dependencies and you can't fix it immediately but you also don't want your build system perma-failing until a un-vuln module is released.

.travis.yml
```diff
@@ -8,6 +8,10 @@ before_install:
 - git remote set-branches origin master && git fetch
 - git checkout master
 - git checkout -
+
+after_install:
+- npm run nsp-check # -- --warn-only
```
pdehaan: Similar to the empty `exceptions` Array above, we could add `--warn-only` flag here if we want `nsp` to always exit with an exit code of 0 (temporarily or permanently).

package.json
```diff
@@ -70,8 +71,8 @@
     "babel-plugin-transform-class-properties": "6.11.5",
     "babel-plugin-transform-es2015-modules-commonjs": "6.14.0",
     "babel-plugin-transform-flow-strip-types": "6.14.0",
-    "babel-preset-es2015": "6.14.0",
     "babel-plugin-transform-runtime": "6.15.0",
+    "babel-preset-es2015": "6.14.0",
```
pdehaan: I didn't do this, but I assume npm auto-sorted the dependencies when I installed `nsp` as a devDep.

package.json
```diff
@@ -18,6 +18,7 @@
     "lint": "grunt lint",
     "test": "grunt test",
     "publish-coverage": "grunt coveralls",
+    "nsp-check": "nsp check -o summary",
```
pdehaan: Since this is an npm script, you can still do `npm run nsp-check` locally to check for vulnerable dependencies, not limited to just running on CI (and doesn't require a globally installed module).

pdehaan:
Current status:

```
1.57s $ npm run nsp-check
> web-ext@1.4.0 nsp-check /home/travis/build/mozilla/web-ext
> nsp check -o summary
(+) No known vulnerabilities found
```

(TIL: Travis doesn't have a "after_install" hook, which is why this PR originally did nothing.)

.travis.yml
```diff
@@ -8,13 +8,18 @@ before_install:
 - git remote set-branches origin master && git fetch
 - git checkout master
 - git checkout -
+
+before_script:
+- npm run nsp-check # -- --warn-only
```
kumar303: looks like this comment was just for testing, can you remove it?

package.json
kumar303: yep, might have been hand edited at some point

.travis.yml
pdehaan: Ah, actually I left that in.
The way nsp+CI works is, once a new vulnerability is found it may fail your CI, and your Travis will remain broken until a newer version of that vulnerable dependency is released. This could take hours/days/weeks. So we'd have two obvious solutions:

- Uncomment the `-- --warn-only` bit, so nsp still runs, but won't exit with an exit code !== 0.
- Add the vulnerable module exception to the .nsprc file so we essentially ignore the warning.

I'm happy to remove the comment above, but just wanted to explain what the comment was for.

kumar303: ok, I see. No one will ever see warnings in the CI build so they are not useful. I suggest adding a comment above this command that says "If this command fails and you can't fix it, file an issue and add an exception to .nsprc"

pdehaan: Done. Pushed.

PR #523, opened 2016-09-24
fix: Support FIREFOX_BINARY again for `web-ext run`

### LINKED ISSUE #512: Support FIREFOX_BINARY again for `web-ext run`
In https://github.com/mozilla/web-ext/pull/440 we shortened the `--firefox-binary` option to `--firefox`. After thinking more about it, I think we should also alias this option to `--firefox-binary` so that developers can continue to use the `FIREFOX_BINARY` environment variable if they wish. See the `alias` key of the [options()](http://yargs.js.org/docs/#methods-optionkey-opt) call.

shubheksha:
Fixes #512 
- Created another alias `firefox-binary`
- Added a test for the new alias


src/program.js
```diff
@@ -218,7 +218,7 @@ Example: $0 --help run.
       })
     .command('run', 'Run the web extension', commands.run, {
       'firefox': {
-        alias: 'f',
+        alias: ['f', 'firefox-binary'],
```
kumar303: looks good

tests/functional/test.cli.run.js
```diff
@@ -47,3 +47,44 @@ describe('web-ext run', () => {
          });
        }));
 });
+
+describe('web-ext run', () => {
```
kumar303: Functional tests are very expensive because they take a lot longer than unit tests. Anytime we add a functional test we have to ask, can this be tested just as well in unit test? For this case I think the answer is yes. You can make a test that passes in a stub version of the `run` function and asserts that `--firefox-binary` is passed to the function like `run({firefox: '/the/path/'})`. You can model it after this test: https://github.com/mozilla/web-ext/blob/master/tests/unit/test.program.js#L283-L294

shubheksha: Doesn't [this](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-cmd/test.run.js#L116) test do exactly that?

kumar303: No, that test does other things:
* It invokes the `cmd.run()` function directly so it doesn't cover any command line option parsing
* It is making sure that the `firefoxApp` object (which represents the Firefox application process) is getting called correctly which is something separate

tests/unit/test.program.js
```diff
@@ -307,6 +307,19 @@ describe('program.main', () => {
                      process.cwd() + path.sep);
       });
   });
+  it('passes the path of a firefox binary when specified', () => {
```
kumar303: nit: this line should have one line of whitespace above it to show that `it()` is a new function separate from the function above it

tests/unit/test.program.js
```diff
@@ -307,6 +307,19 @@ describe('program.main', () => {
...
+    const fakeCommands = fake(commands, {
+      run: () => Promise.resolve(),
+    });
+    return run(
+      ['run', '--firefox-binary', '/path/to/firefox-binary'],
+      {commands: fakeCommands})
+      .then(() => {
+        assert.equal(fakeCommands.run.called, true);
+        assert.equal(fakeCommands.run.firstCall.args[0].firefoxBinary,
```
kumar303: you should assert that the call to `run()` passes `firefox`, not `firefoxBinary` because [firefox is the variable](https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L162) used by the command.

kumar303: this test needs to make sure that `--firefox-binary` gets treated just like `--firefox`

PR #547, opened 2016-10-01
refactor: rename the run test helper to execProgram

### LINKED ISSUE #545: Rename run() test helper to execWebExt()
It can be considered as an enhancement.

There is a [run handler](https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L160) function which invokes `web-ext run` and there is a [run](https://github.com/mozilla/web-ext/blob/master/src/program.js#L79) used as a test helper for `program.js` & invokes `web-ext`. It makes it confusing to read the code and understand the flow.

The `run` test helper should be renamed to `execWebExt()`

shubheksha:
Fixes #545

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -76,7 +76,7 @@ export class Program {
     return this;
   }
 
-  async run(absolutePackageDir: string,
+  async execWebExt(absolutePackageDir: string,
       {systemProcess = process, logStream = defaultLogStream,
        getVersion = defaultVersionGetter, shouldExitProgram = true}
       : Object = {}): Promise<void> {
@@ -284,5 +284,5 @@ Example: $0 --help run.
       },
     });
 
-  return program.run(absolutePackageDir, runOptions);
+  return program.execWebExt(absolutePackageDir, runOptions);
 }
--- a/src/program.js
+++ b/src/program.js
@@ -76,7 +76,7 @@ export class Program {
     return this;
   }
 
-  async execWebExt(absolutePackageDir: string,
+  async execute(absolutePackageDir: string,
       {systemProcess = process, logStream = defaultLogStream,
        getVersion = defaultVersionGetter, shouldExitProgram = true}
       : Object = {}): Promise<void> {
@@ -284,5 +284,5 @@ Example: $0 --help run.
       },
     });
 
-  return program.execWebExt(absolutePackageDir, runOptions);
+  return program.execute(absolutePackageDir, runOptions);
 }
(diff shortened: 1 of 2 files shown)


tests/unit/test.program.js
```diff
@@ -60,7 +60,8 @@ describe('program.Program', () => {
       .command('cmd', 'some command', () => {
         throw new Error('this is an error from a command handler');
       });
-    return run(program, {systemProcess: fakeProcess, shouldExitProgram: true})
+    return execWebExt(program, {systemProcess: fakeProcess,
+                                  shouldExitProgram: true})
```
kumar303: nit: this indent needs to align vertically with the opening brace, like:
````
return execWebExt(program, {systemProcess: fakeProcess,
                            shouldExitProgram: true})
````

File: src/program.js [hunk 2]
```diff
@@ -76,7 +76,7 @@ export class Program {
     return this;
   }
 
-  async run(absolutePackageDir: string,
+  async execWebExt(absolutePackageDir: string,
```
kumar303: In this case the web-ext part is implied. I think it would make sense to call this `execute` because it gets called like `program.execute()` which makes more sense.

tests/unit/test.program.js
```diff
@@ -15,10 +15,10 @@ import {ConsoleStream} from '../../src/util/logger';
 
 describe('program.Program', () => {
 
-  function run(program, options = {}) {
+  function execWebExt(program, options = {}) {
```
kumar303: I realize now that what this function does is execute a program object (not necessarily the "web-ext" program). Therefore, I think it would make more sense to call it `execProgram()`.

tests/unit/test.program.js
```diff
@@ -60,7 +60,8 @@ describe('program.Program', () => {
       .command('cmd', 'some command', () => {
         throw new Error('this is an error from a command handler');
       });
-    return run(program, {systemProcess: fakeProcess, shouldExitProgram: true})
+    return execProgram(program, {systemProcess: fakeProcess,
+                                shouldExitProgram: true})
```
kumar303: The indent needs to align to somewhere after the brace of the opening statement. It needs to give a visual cue to the reader about which block is still open. Example:
````
return execProgram(program, {systemProcess: fakeProcess,
                             shouldExitProgram: true})
````
This would also work to give visual indication of which block is still open:
````
return execProgram(program, {
  systemProcess: fakeProcess,
  shouldExitProgram: true,
})
````
or this:
````
return execProgram(program, 
                   {systemProcess: fakeProcess, shouldExitProgram: true})
````

Issue #507, opened 2016-09-20
Adding a command to automatically generate updateManifest.json like jpm does?


Azeirah:
It would be extremely helpful to have a command to generate an updateManifest.json within this web-ext tool, just like how jpm is able to do it. Right now, I'm using a custom script, but I feel it should be part of web-ext.

Is there any reason why web-ext doesn't automatically generate an updateManifest.json file during the `sign` task like jpm does it?

kumar303:
> Right now, I'm using a custom script

Why do you need a script for this? Is it helpful because it pre-populates the extension ID? I'd be curious to look at this because I can't think of anything else that is possible to generate automatically.

Just to provide more context for other readers, updates for WebExtensions are different from SDK based add-ons (made with `jpm`). This [article](https://developer.mozilla.org/en-US/Add-ons/Updates) explains how updates work. The first step is to add an [update_url](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/applications) to the WebExtension manifest.

Azeirah:
> Why do you need a script for this?

I'm using a custom script because I want to automate the process from `git push`, on my webextension to automatically deploy the new version on my website if the whole pipeline -- test, build.. and then deploy -- is passed.

I am still working on this. The custom script is indeed based on the article you mentioned.

Am I wrong in assuming that it's possible to automate this process?

1. Add `update_url` to `manifest.json`
2. Create initial `updateManifest.json` file, and upload it to my webserver
3. Whenever I git push, check if version in manifest.json has changed
4. If so, fetch the old updateManifest.json
5. Add the details of the new version to the new updateManifest.json
6. Upload the new updateManifest.json to my web server

All variables, such as the extension ID, web server URL, will be defined within the pipeline itself.

kumar303:
Ah, thanks for the details. I understand this better.

Yes, I think the [build](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build) command should produce `updateManifest.json` if the manifest includes `update_url`.

I don't think `web-ext` should try to push `updateManifest.json` to a remote server though. The details for how to do that are going to vary from server to server so it seems out of scope.

kumar303:
Note that the `sign` command calls `build` automatically

kumar303:
It may be easier to start this as a standalone command like `web-ext update-info` (or something)

Azeirah:
[...]

PR #552, opened 2016-10-03
Add feature to automatically generate updateManifest files from previous updateManifests

Azeirah:
https://github.com/mozilla/web-ext/issues/507

This pull request is not ready. Tests need to be written, documentation needs to be updated, and a short code-review would be really helpful. I've done some ad-hoc testing with this command 

## Usage and testing:

This feature expects a manifest.json to exist. For it to generate a new updateManifest, an older one needs to exist on a web server somewhere. This is configured in the manifest.json's `applications.gecko.update_url` property. For my ad-hoc tests I've simply used my personal web-server.

## The command that I used

`./web-ext sign --api-key {YOUR-USER} --api-secret {YOUR-API} --api-url-prefix https://addons-dev.allizom.org/api/v3 --update-link https://example.com/{xpiFileName}`

## What I did

- [x] src/cmd/sign.js, add function generateNewUpdateManifest(...)
    - [x] Fetch an updateManifest.json from the location pointed to by manifest.applications.gecko.update_url
    - [x] Error messages (show the user an error message, and exit with error code)
        - [x] the old, fetched updateManifest.json is invalid
        - [x] unable to fetch updateManifest.json due to connection error
        - [x] argument passed to --update-link flag is invalid (== is missing `{xpiFileName}`)
    - [x] Warning messages
        - [x] The user passed the --update-link flag, but manifest.json lacks an applications.gecko.update_url property
        - [x] The fetched updateManifest.json had no previous release of an addon with this extension-id. Warn the user that if this should not be the case to check her extension-id
    - [x] Export the generated update manifest to the artifacts directory
- [x] all code complies to flow
- [x] all code complies to the linter
- [ ] test cases?
- [x] program.js, add --update-link param to `sign` .command
- [ ] documentation
    - [x] add description of the --update-link param
    - [ ] add explanation on how to auto-generate updateManifest

## What still needs to be done:

1. Automated tests
2. Code review
3. Additions to the official documentation

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -2,6 +2,8 @@
 import path from 'path';
 import {fs} from 'mz';
 import defaultAddonSigner from 'sign-addon';
+import {httpFetchFile} from '../util/net';
+import url from 'url';
 
 import defaultBuilder from './build';
 import {withTempDir} from '../util/temp-dir';
@@ -32,6 +34,7 @@ export type SignParams = {
   apiSecret: string,
   apiUrlPrefix: string,
   timeout: number,
+  updateLink: string
 };
 
 export type SignOptions = {
@@ -49,7 +52,7 @@ export type SignResult = {
 export default function sign(
   {
     verbose, sourceDir, artifactsDir, apiKey, apiSecret,
-    apiUrlPrefix, id, timeout,
+    apiUrlPrefix, id, timeout, updateLink,
   }: SignParams,
   {
     build = defaultBuilder, signAddon = defaultAddonSigner,
@@ -107,6 +110,19 @@ export default function sign(
         downloadDir: artifactsDir,
       });
 
+      if (updateLink && updateLink.length > 0 && signingResult.success) {
+        let extensionID;
+        if (signingResult.id) {
+          extensionID = signingResult.id;
+        } else {
+          extensionID = id;
+        }
+        await generateNewUpdateManifest(
+          extensionID, manifestData.version, signingResult.downloadedFiles[0],
+          artifactsDir, manifestData, updateLink
+        );
+      }
+
       if (signingResult.id) {
         await saveIdToSourceDir(sourceDir, signingResult.id);
       }
@@ -125,6 +141,103 @@ export default function sign(
   );
 }
 
+async function generateNewUpdateManifest(
+  id: ?string, newVersion: string, XPIPath: string, artifactsDir,
+  manifestData: Object, updateLink: string
+) {
+  let oldUpdateManifest;
+  let oldUpdateManifestData: Object;
+  let statusCode;
+  let newUpdateManifest;
+  let updateManifestFileName = '';
+
+  if (manifestData !== undefined &&
+      manifestData.applications !== undefined &&
+      manifestData.applications.gecko !== undefined &&
+      manifestData.applications.gecko.update_url !== undefined) {
+    let parsed = url.parse(manifestData.applications.gecko.update_url);
+    if (parsed !== undefined && parsed.pathname !== undefined) {
+      updateManifestFileName = path.basename(parsed.pathname);
+    }
+
+    try {
+      [oldUpdateManifest, statusCode] = await httpFetchFile(
+        manifestData.applications.gecko.update_url);
+      if (statusCode < 200 || statusCode > 290) {
+        throw new WebExtError(
+          `Failed to retrieve ${updateManifestFileName}\n` +
+          `http statusCode error, server responded with: ${statusCode}`);
+      }
+    } catch (e) {
+      throw new WebExtError(
+        `Was unable to download ${updateManifestFileName}\nerror is: ${e}`);
+    }
+
+    try {
+      oldUpdateManifestData = JSON.parse(oldUpdateManifest);
+    } catch (e) {
+      throw new WebExtError(
+        `Unable to parse ${updateManifestFileName} file located at ` +
+        `${manifestData.applications.gecko.update_url}`);
+    }
+
+    if (updateLink.indexOf('{xpiFileName}') === -1) {
+      throw new WebExtError(
+        'Unable to parse --update-link url, please use {xpiFileName} ' +
+        'as a substitute for the XPI file');
+    }
+
+    let newVersion = {
+      version: manifestData.version,
+      update_link: updateLink.replace('{xpiFileName}', path.basename(XPIPath)),
+    };
+
+    if (oldUpdateManifestData) {
+      if (oldUpdateManifestData.addons === undefined) {
+        oldUpdateManifestData.addons = {};
+      }
+
+      if (id && oldUpdateManifestData.addons[id]) {
+        // append a new release to the updateManifest
+        if (oldUpdateManifestData.addons[id].updates) {
+          oldUpdateManifestData.addons[id].updates.push(newVersion);
+        }
+      } else {
+        // first release of this addon (or maybe the extension-id was botched? be careful!)
+        let addonName = manifestData.name || 'your application';
+        log.warn(`Creating first release of ${addonName}`);
+        log.warn(
+          'If this is not the actual first release, check your extension-id');
+        oldUpdateManifestData.addons[id] = {};
+        oldUpdateManifestData.addons[id].updates = [newVersion];
+      }
+
+      // the oldUpdateManifest has been updated, it's ready to be saved
+      newUpdateManifest = JSON.stringify(oldUpdateManifestData, null, 4);
+    } else {
+      throw new WebExtError('The .addons property is missing in the ' +
+        'retrieved updateManifest.json file');
+    }
+
+    return new Promise(function(resolve, reject) {
+      log.info('Saving to', path.join(artifactsDir, updateManifestFileName));
+      fs.writeFile(
+        path.join(artifactsDir, updateManifestFileName),
+        newUpdateManifest,
+        function(error) {
+          if (error) {
+            reject(error);
+          }
+          resolve();
+        });
+    });
+  } else {
+    throw new WebExtError(
+      'update-link was passed, but the manifest.json ' +
+      'does not have the applications.gecko.update_url property ' +
+      'therefore, no updateManifest could be generated');
+  }
+}
 
 export async function getIdFromSourceDir(
   sourceDir: string
@@ -160,7 +273,6 @@ export async function getIdFromSourceDir(
 
   return id;
 }
-
 
 export async function saveIdToSourceDir(sourceDir: string, id: string)
     : Promise<void> {
--- a/src/program.js
+++ b/src/program.js
@@ -215,6 +215,14 @@ Example: $0 --help run.
           describe: 'Number of milliseconds to wait before giving up',
           type: 'number',
         },
+        'update-link': {
+          describe:
+            'A URL to where your XPI files will be hosted. ' +
+            'ex: https://example.com/plugins/{xpiFileName} ' +
+            'if you select this option, an updateManifest ' +
+            'file will be generated ' +
+            'in the web-ext-artifacts folder',
+        },
       })
     .command('run', 'Run the web extension', commands.run, {
       'firefox': {
--- a/src/util/net.js
+++ b/src/util/net.js
@@ -1,0 +1,20 @@
+/* @flow */
+let http = require('http');
+let https = require('https');
+
+/**
+ * Fetch a file from a given url
+ * Returns a promise resolving to a tuple of [body, statusCode]
+ * Will throw an error on connection errors
+ */
+export function httpFetchFile(url: string): Promise<[string, number]> {
+  return new Promise((resolve, reject) => {
+    const lib = url.startsWith('https') ? https : http;
+    const request = lib.get(url, (response) => {
+      const body = [];
+      response.on('data', (chunk) => body.push(chunk));
+      response.on('end', () => resolve([body.join(''), response.statusCode]));
+    });
+    request.on('error', (err) => reject(err));
+  });
+}
--- a/tests/unit/test-cmd/test.sign.js
+++ b/tests/unit/test-cmd/test.sign.js
@@ -25,6 +25,7 @@ describe('sign', () => {
       apiSecret: 'AMO JWT secret',
       apiUrlPrefix: 'http://not-the-real-amo.com/api/v3',
       timeout: 999,
+      updateLink: '',
     };
 
     const buildResult = {


src/cmd/sign.js
```diff
@@ -107,6 +110,19 @@ export default function sign(
         downloadDir: artifactsDir,
       });
 
+      if (updateLink && updateLink.length > 0 && signingResult.success) {
```
Azeirah: updateLink.length > 0 is unnecessary, overlooked this one

src/cmd/sign.js
```diff
@@ -125,6 +141,103 @@ export default function sign(
...
+    return new Promise(function(resolve, reject) {
+      log.info('Saving to', path.join(artifactsDir, updateManifestFileName));
+      fs.writeFile(
+        path.join(artifactsDir, updateManifestFileName),
+        newUpdateManifest,
+        function(error) {
+          if (error) {
+            reject(error);
+          }
+          resolve();
```
Azeirah: Should probably return something, `{success: Boolean, newUpdateManifestPath: String, ...?}`

src/cmd/sign.js
```diff
@@ -107,6 +110,19 @@ export default function sign(
...
 
+      if (updateLink && updateLink.length > 0 && signingResult.success) {
+        let extensionID;
+        if (signingResult.id) {
+          extensionID = signingResult.id;
+        } else {
+          extensionID = id;
+        }
+        await generateNewUpdateManifest(
+          extensionID, manifestData.version, signingResult.downloadedFiles[0],
```
kumar303: Keep in mind that `signingResult.downloadedFiles` is an array of many files. I'm not entirely sure if newer WebExtensions will have multiple files but historically AMO would sign one file for desktop, one file for Android, and so on (this was mostly for old add-ons with binary components though). In any case, I suppose we should try to support an array of files?

src/cmd/sign.js
```diff
@@ -125,6 +141,103 @@ export default function sign(
...
+      // the oldUpdateManifest has been updated, it's ready to be saved
+      newUpdateManifest = JSON.stringify(oldUpdateManifestData, null, 4);
+    } else {
+      throw new WebExtError('The .addons property is missing in the ' +
+        'retrieved updateManifest.json file');
+    }
+
+    return new Promise(function(resolve, reject) {
+      log.info('Saving to', path.join(artifactsDir, updateManifestFileName));
+      fs.writeFile(
```
kumar303: you can just `await fs.writeFile()` (it returns a promise since we're using [mz/fs](https://github.com/normalize/mz))

kumar303:
Looks like you're on the right track! Since there are a lot of error cases, I'd suggest breaking it into smaller functions so that it's easier to test.

src/cmd/sign.js
Azeirah: I understand that's the case, but how would that fit inside an updateManifest.json file? Here's what I don't get:

id: 123456789
version: 1.2
android: someGeneratedXPI-ANDROID-v1.2.xpi
desktop: someGeneratedXPI-DESKTOP-v1.2.xpi

The older updateManifest.json file looks like this
```
{
    "addons": {
        "123456789": {
            "updates": []
        }
    }
}
```

So let's update it with the given data

```
{
    "addons": {
        "123456789": {
            "updates": [
                {
                    "version": "1.2",
                    "update_link": "https://example.com/someGeneratedXPI-ANDROID-v1.2.xpi"
                },
                {
                    // also version 1.2?
                    // this shouldn't be possible, right?
                    // the android XPI has the same extensionID, no?
                    "version": "1.2",
                    "update_link": "https://example.com/someGeneratedXPI-DESKTOP-v1.2.xpi"
                }
            ]
        }
    }
}
```

Azeirah:
Thanks for the feedback, I'm working on it!

Azeirah:
I've pushed a version where I split-up the function into three parts.

src/cmd/sign.js
```diff
@@ -107,6 +117,27 @@ export default function sign(
...
+        // flow thinks that oldUpdateManifestData can be potentially undefined
+        // this is wrong however, the if-case guarding both the `oldUpdateManifestData = `
+        // statement and this function call are both checking
+        // the exact same thing, which doesn't change
+        // anywhere in the code. Basically, `if (A) then define B` and later on `if (A) then use B`
+        // which flow doesn't seem to understand
+        let oldUpdateManifestDataMakeFlowHappy =
+          oldUpdateManifestData == null ? {} : oldUpdateManifestData;
+        await createNewUpdateManifest(
+          extensionID, signingResult.downloadedFiles[0], artifactsDir,
```
kumar303: ok, I looked more into the case where multiple files could exist in `signingResult.downloadedFiles` and this is just for legacy (non-webextension) add-ons. So, yeah, it should never happen. However, I'd like to see something that checks the length of this array and throws an error if it's longer than one. The error should just explain how such a situation is not supported.

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+ * Generates a new updateManifest.json file
+ * It will add an entry to a previously fetched updateManifest.json file
+ * for each generated XPI file.
+ * The result will be stored in the artifacts directory
+ */
+// jslint incorrectly reports that this return statement should be a yield.
+// eslint-disable-next-line
+async function createNewUpdateManifest(
+  id: ?string, XPIPath: string, artifactsDir,
+  manifestData: Object, updateLink: string, oldUpdateManifestData: Object
```
kumar303: All arguments should be keyword args to make them more readable and easier to change later on. This is a simple change using ES6 destructors (just add braces):
````javascript
async function createNewUpdateManifest({
  id: ?string, XPIPath: string, artifactsDir, ...
}) {
````

File: src/cmd/sign.js [hunk 7]
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+          `Failed to retrieve ${updateManifestFileName}\n` +
+          `http statusCode error, server responded with: ${statusCode}`);
+      }
+    } catch (e) {
+      throw new WebExtError(
+        `Was unable to download ${updateManifestFileName}\nerror is: ${e}`);
+    }
+
+    try {
+      resolve(JSON.parse(oldUpdateManifest));
```
kumar303: Why do you need `resolve()` here? I think if you declare `fetchUpdateManifest` as `async` then you can simply do `return JSON.parse(oldUpdateManifest)` on this line.

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+    } catch (e) {
+      throw new WebExtError(
+        `Unable to parse ${updateManifestFileName} file located at ` +
+        `${manifestData.applications.gecko.update_url}`);
```
kumar303: This messages needs to include the original error message and try to avoid one character variables. I would suggest calling it `catch(error)`

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+ * Validates all data needed to generate a new updatemanifest
+ * if some data is missing/corrupt, the manifest can't be generated
+ */
+function prevalidateUpdateManifestParams(manifestData, updateLink) {
+  // ..first validate if the manifest has the right properties
+  // manifestData.applications.gecko.updateUrl
+  // note: `var == null` is equal to
+  // `(variable === undefined || variable === null)`
+  // see: https://stackoverflow.com/questions/2647867
+  if (manifestData == null ||
```
kumar303: the top-level `sign()` function accepts the `id` parameter which gets a value when someone does `web-ext sign --id=custom@example`. This validation check looks like it is not honoring that `id` variable (for when it's not undefined).

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
   );
 }
 
+/**
+ * Fetches an updateManifest.json file from a remote web server
+ */
+function fetchUpdateManifest(manifestData: Object): Object {
```
kumar303: The Flow return value should be `Promise<Object>` but it would be better if you could define the parts of the manifest object you'll need and specify a type for it. See this example: https://github.com/mozilla/web-ext/blob/master/src/util/manifest.js#L14-L18

After a custom definition, the return value would become `Promise<UpdateManifest>`.

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+
+    try {
+      [oldUpdateManifest, statusCode] = await httpFetchFile(
+        manifestData.applications.gecko.update_url);
+      if (statusCode < 200 || statusCode > 290) {
+        throw new WebExtError(
+          `Failed to retrieve ${updateManifestFileName}\n` +
+          `http statusCode error, server responded with: ${statusCode}`);
+      }
+    } catch (e) {
```
kumar303: I'd like to see this spelled out as `error` not `e`.

src/cmd/sign.js
```diff
@@ -107,6 +117,27 @@ export default function sign(
...
+          extensionID = id;
+        }
+        // flow thinks that oldUpdateManifestData can be potentially undefined
+        // this is wrong however, the if-case guarding both the `oldUpdateManifestData = `
+        // statement and this function call are both checking
+        // the exact same thing, which doesn't change
+        // anywhere in the code. Basically, `if (A) then define B` and later on `if (A) then use B`
+        // which flow doesn't seem to understand
+        let oldUpdateManifestDataMakeFlowHappy =
+          oldUpdateManifestData == null ? {} : oldUpdateManifestData;
```
kumar303: If you can push a change that shows the Flow error, I could take a look. I would rather not see extra variable assignments like this. I also can't spot a situation where it would be null but maybe it's because you didn't define the `Promise` return value from `fetchUpdateManifest`, I don't know.

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+    let parsed = url.parse(manifestData.applications.gecko.update_url);
+    if (parsed == null || parsed.pathname == null) {
+      // the most elaborate flow workaround
+      let applications =
+        manifestData.applications == null ?
+        {gecko: manifestData.applications.gecko} : manifestData.applications;
+      let gecko =
+        applications.gecko == null ?
+        {gecko: applications.gecko} : manifestData.applications.gecko;
+      let update_url =
```
kumar303: all assignments must be camel case

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+  }
+
+  // ..lastly check if the passed updateLink parameter is correct
+  if (updateLink.indexOf('{xpiFileName}') === -1) {
+    throw new WebExtError(
+      'Unable to parse --update-link url, please use {xpiFileName} ' +
+      'as a substitute for the XPI file');
+  }
+
+  return true;
```
kumar303: this return value doesn't look like it's used by anything

src/cmd/sign.js
```diff
@@ -125,6 +156,170 @@ export default function sign(
...
+  }
+
+  // the oldUpdateManifest has been updated, it's ready to be saved
+  // prettify it so it's readable to the maintainers
+  newUpdateManifest = JSON.stringify(oldUpdateManifestData, null, 4);
+
+  return fs.writeFile(
+    path.join(artifactsDir, updateManifestFileName),
+    newUpdateManifest,
+    function(error) {
```
kumar303: you can't use the error callback here since you are returning a promise -- i.e. the caller needs be notified about the error too. You can use chaining instead:
````javascript
return fs.writeFile(...)
  .catch((error) => {
    throw new WebExtError(
      `Was unable to write updated ${addonName}: ${error}`);
  });
````

File: src/program.js [hunk 16]
```diff
@@ -215,6 +215,14 @@ Example: $0 --help run.
           describe: 'Number of milliseconds to wait before giving up',
           type: 'number',
         },
+        'update-link': {
+          describe:
+            'A URL to where your XPI files will be hosted. ' +
+            'ex: https://example.com/plugins/{xpiFileName} ' +
```
kumar303: New sentences should start with a capital letter. I think it would also be helpful to spell out `Example: `

src/program.js
```diff
@@ -215,6 +215,14 @@ Example: $0 --help run.
...
+            'if you select this option, an updateManifest ' +
+            'file will be generated ' +
+            'in the web-ext-artifacts folder',
```
kumar303: Say `in the --artifacts-dir folder` instead since this value is customizable.

src/util/net.js
```diff
@@ -0,0 +1,20 @@
+/* @flow */
+let http = require('http');
+let https = require('https');
+
+/**
+ * Fetch a file from a given url
+ * Returns a promise resolving to a tuple of [body, statusCode]
+ * Will throw an error on connection errors
+ */
+export function httpFetchFile(url: string): Promise<[string, number]> {
```
kumar303: This code would be easier to write (and easier to test) if you use the [request](https://github.com/request/request) module. I don't mind adding it as a dependency.

kumar303:
Hi @Azeirah. There's no rush on this but I just wanted to check in and see if you had any questions? Thanks for your continued work on it -- I think this will be a useful feature!

Azeirah:
Hey @kumar303, I ran into some walls while developing this feature. I came here expecting that it would be a breeze to implement the whole thing, and that was partially true, the feature wasn't too hard to implement. Everything else was a bit overwhelming for me, complying to fairly strict coding standards (flux), learning several new ES6 features and learning how to write tests. Now that some time has passed I'll try to tackle the issue again.

The feature itself would be very useful for me as well, so I will continue work on this. I'll let you know if I run into any troubles.

kumar303:
I understand completely. Each patch takes a bit more time on this project, especially in writing the tests. We require automated tests to make sure we don't regress the features accidentally. Let me know if you need any help.

kumar303:
Hi @Azeirah . Would it help to try landing a smaller patch for this feature? It still seems like a nice feature to have. Let me know if you are stuck on anything.

src/cmd/sign.js
Azeirah: I'm not sure what you mean here, the `prevalidateUpdateManifestParams` function doesn't check *anything* related to the id, it is left alone completely.
The right id value gets selected earlier on in the code

```js
if (id && manifestId) {
  throw new WebExtError(
    `Cannot set custom ID ${id} because manifest.json ` +
    `declares ID ${manifestId}`);
}

if (manifestId) {
  id = manifestId;
}

if (!id && idFromSourceDir) {
  log.info(
    `Using previously auto-generated extension ID: ${idFromSourceDir}`);
  id = idFromSourceDir;
}

if (!id) {
  log.warn('No extension ID specified (it will be auto-generated)');
}
```

And this id later gets passed on to the `createNewUpdateManifest` function

```js
let signingResult = await signAddon({...});
[...]
let extensionID;
if (signingResult.id) {
  extensionID = signingResult.id;
} else {
  extensionID = id;
}
[...]
await createNewUpdateManifest(
  extensionID, signingResult.downloadedFiles[0], artifactsDir,
  manifestData, updateLink, oldUpdateManifestData,
);
```

I'm probably just not completely getting what you mean here.

Azeirah:
@kumar303 I've picked it up again, just finished doing my exams at school so I have some time again. Apologies for the little communication

src/cmd/sign.js
Azeirah: I think it's an issue with flow, I reproduced the issue in a very simple file

```js
// @flow
function printAmountOfFruits(amountOfFruits: number) {
    console.log(amountOfFruits);
}
let areApplesNeeded = false;

let amountOfFruits;
if (areApplesNeeded) {
    amountOfFruits = 10;
}

if (areApplesNeeded) {
    printAmountOfFruits(amountOfFruits);
}
```

Flow reports

```
test-flow.js:13
 13:     printAmountOfFruits(amountOfFruits);
                             ^^^^^^^^^^^^^^ possibly uninitialized variable. This type is incompatible with the expected param type of
  2: function printAmountOfFruits(amountOfFruits: number) {
                                                  ^^^^^^ number


Found 1 error
```

In this case, if the code `printAmountOfFruits(amountOfFruits)` runs, the variable `amountOfFruits` is defined with 100% certainty. Flow doesn't seem to understand this however.

It's the exact same case as with the `oldUpdateManifest`. Unfortunately, the workaround I used previously doesn't really work anymore if you take into account the other remark you made about wanting to use keyword arguments, since I created a type definition for the `createNewUpdateManifest` arguments. The workaround now looks like

```js
// note that we assign an empty object instead of leaving the variable
// undefined because of a problem with flow, see #{Link to this comment} 
let oldUpdateManifestData = {
  addons: {},
};
```

Which works, but it's just weird, it's a skeleton-structure to please flow, instead of having any real purpose within the code.

kumar303: That may be a Flow bug but it's bad practice to shadow an outer-scope variable in your local scope. In fact, I would like to turn on the [lint rule](http://eslint.org/docs/rules/no-shadow) to prevent this in `web-ext`. To fix it, rename your local variable:
````javascript
function printAmountOfFruits(localAmountOfFruits: number) {
    console.log(localAmountOfFruits);
}
````

kumar303: I'm going to turn this on here https://github.com/mozilla/web-ext/pull/768

src/cmd/sign.js
kumar303: yeah, sorry, I have no idea what I was talking about either. I guess ignore it for now. I will do another review when you've addressed the other change requests anyway.

src/cmd/sign.js
Azeirah: @kumar303 Ah yeah, I didn't consider the shadowing. Still, the flow problem persists.

Azeirah:
I've pushed the changes where I fix the issues you pointed out.

kumar303:
Thanks! Can you merge in master and resolve the conflicts? That will make it easier for me to review and test.

shubheksha:
Hey @Azeirah, would it be possible to pick this up any time soon again?

rpl:
Closing as "state: do not merge", the issue is still open and technically not covered yet by web-ext itself, unfortunately this PR has been branched out quite some time ago and it is unlikely to be resumed from this old branch.

PR #554, opened 2016-10-05
test: introducing flowtype exact object types

rpl:
Flow supports a new syntax for exact object types, starting from v.0.32.0:
- [Release notes Flow v. 0.32.0](https://github.com/facebook/flow/releases/tag/v0.32.0)

Using the new `{| prop: type |}` syntax, flow checks that the objects of this type contains only exactly the defined properties, and using this additional restriction (where it is appropriate), flow will be able to catch additional classes of issues early.

This PR contains an initial proposal of the changes needed to introduce this new flowtype feature, and it is composed of:
- a temporary change in the npm dependency to be able to use the proposed change to grunt-flowbin from kumar303/grunt-flowbin#41
- minor fixes related to flow 0.33.0 (the flowchecks were currently running on flow 0.30.0, see the above grunt-flowbin PR for a rationale)
- "exact objects types" introduced only in src/cmd/build.js (as an initial evaluation of the impact of the change)
- fix new flowtype validation error caught by flow in tests/unit/test-cmd/test.build.js, related to the above change in src/cmd/build.js types

### DIFF
--- a/src/util/manifest.js
+++ b/src/util/manifest.js
@@ -14,6 +14,11 @@ const log = createLogger(__filename);
 export type ExtensionManifest = {
   name: string,
   version: string,
+  applications: {
+    gecko: {
+      id: string
+    }
+  }
 };
 
 export default async function getValidatedManifest(
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -43,8 +43,9 @@ describe('build', () => {
         // Make sure a manifest without an ID doesn't throw an error.
         return build({
           sourceDir: fixturePath('minimal-web-ext'),
-          manifestData: manifestWithoutApps,
           artifactsDir: tmpDir.path(),
+        }, {
+          manifestData: manifestWithoutApps,
         });
       }
     );
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "grunt-contrib-watch": "1.0.0",
     "grunt-coveralls": "1.0.1",
     "grunt-eslint": "19.0.0",
-    "grunt-flowbin": "0.0.6",
+    "grunt-flowbin": "rpl/grunt-flowbin#fix/flow-bin-dep",
     "grunt-mocha-test": "0.13.2",
     "grunt-newer": "1.2.0",
     "grunt-webpack": "1.0.15",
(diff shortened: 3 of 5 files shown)


src/util/manifest.js
```diff
@@ -14,6 +14,11 @@ const log = createLogger(__filename);
 export type ExtensionManifest = {
   name: string,
   version: string,
+  applications: {
```
kumar303: shouldn't this be `applications?` since it's optional?

tests/unit/test-cmd/test.build.js
```diff
@@ -43,8 +43,9 @@ describe('build', () => {
         // Make sure a manifest without an ID doesn't throw an error.
         return build({
           sourceDir: fixturePath('minimal-web-ext'),
-          manifestData: manifestWithoutApps,
           artifactsDir: tmpDir.path(),
+        }, {
+          manifestData: manifestWithoutApps,
```
kumar303: I'm sold! :) This is a great catch.

rpl:
> Do you want to start with smaller patches like this, maybe one or two commands at a time?

@kumar303 exactly, that's the reason why I only did it on a single file.
my proposed plan is to turn this PR into a preparation to next "exact object types" sub-"pull requests",
with just this two changes:
  - the updated grunt-flowbin dependency (once  kumar303/grunt-flowbin#41 is completed and relesed)
  - the minimal fixes needed by flow 0.33

Then move the commit that applies the exact object types to the `src/cmd/build` module (and its tests) into a new PR, and then gradually open new ones related to the other modules.

How about creating a new issue to specifically track the progress of this group of pull requests?

kumar303:
The new version of `grunt-flowbin` is here https://www.npmjs.com/package/grunt-flowbin (I meant the version to be `0.1.0`, heh, oops).

kumar303:
> How about creating a new issue to specifically track the progress of this group of pull requests?

Yes, that would be helpful

src/util/manifest.js
kumar303: I changed this to `applications?` already upstream in https://github.com/mozilla/web-ext/pull/639

kumar303:
@shubheksha beat you to it! I believe https://github.com/mozilla/web-ext/pull/726 covers everything that was added here, right?

rpl:
@kumar303 yes, definitely! :raised_hands: 

@shubheksha Thank you so much! :heart:

PR #561, opened 2016-10-08
fix: Fixed commit format verification prompt for Windows users

### LINKED ISSUE #551: Support local commit verification for Windows users
After https://github.com/mozilla/web-ext/issues/515, Windows users will have to commit like `git commit --no-verify` since the pre-commit hook we added does not support Windows. Let's make it support Windows!

berraknil:
Fixes #551

Added the [git-hooks-win](https://www.npmjs.com/package/git-hooks-win) package to package.json under developer dependencies. This package has zero dependencies and supports the current model of format verification prompt that is available to Mac OS users for Windows users as well. 

This was the error message I got before adding the package

before

And after installing the package, prompt works correctly

after

Tested on Windows 10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "eslint-plugin-flowtype": "2.20.0",
     "firefox-client": "0.3.0",
     "flow-bin": "0.33.0",
-    "git-hooks": "1.1.1",
     "git-hooks-win": "1.1.11",
     "grunt": "1.0.1",
     "grunt-contrib-clean": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "firefox-client": "0.3.0",
     "flow-bin": "0.33.0",
     "git-hooks": "1.1.1",
+    "git-hooks-win": "1.1.11",
     "grunt": "1.0.1",
     "grunt-contrib-clean": "1.0.0",
     "grunt-contrib-copy": "1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     "firefox-client": "0.3.0",
     "flow-bin": "0.33.0",
     "git-hooks": "1.1.1",
+    "git-hooks-win": "1.1.11",
```
kumar303: Good find! This looks really promising and it looks like it will also work on Mac. However, since the installation of this model installs a hook script, I think we also need to remove the old `git-hooks` dependency, right?

berraknil: Good point, I didn't want to mess with the original package since I don't have a mac to try if this one works properly on Mac too. If it does we should remove the other one and keep the windows compatible  ,otherwise I'll try one of the few other solutions I looked into (like grunt-git-hooks) which are a bit more complicated.

package.json
```diff
@@ -87,6 +87,7 @@
     "firefox-client": "0.3.0",
     "flow-bin": "0.33.0",
     "git-hooks": "1.1.1",
```
kumar303: We can remove this. I just tried it on Mac and it works just fine with only the `git-hooks-win` package.

atsay:
Hi Berrak, thanks for the fix, and welcome! I've added this fix to our [recognition page](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#October_2016).

PR #565, opened 2016-10-10

### LINKED ISSUE #481: print a development version when using a custom web-ext build

saintsebastian:
Fixes #481


tests/unit/test.program.js
```diff
@@ -329,15 +331,26 @@ describe('program.main', () => {
...
-  it('returns the package version', () => {
-    let root = path.join(__dirname, '..', '..');
+  it('returns the package version in production', () => {
     let pkgFile = path.join(root, 'package.json');
     return fs.readFile(pkgFile)
-      .then((pkgData) => {
-        assert.equal(defaultVersionGetter(root),
-                     JSON.parse(pkgData).version);
-      });
+    .then((pkgData) => {
```
kumar303: Just a minor problem here: this block should be indented two spaces to show that `.then()` is a continuation of the call to `readFile()`.

tests/unit/test.program.js
```diff
@@ -329,15 +331,26 @@ describe('program.main', () => {
...
+        },
+      };
+      assert.equal(defaultVersionGetter(root, testProcess),
+                   JSON.parse(pkgData).version);
+    });
+  });
+
+  it('returns git commit information in development', () => {
+    const commit = `${git.branch()}-${git.long()}`;
+    assert.equal(defaultVersionGetter(root),
```
kumar303: This should be called using a `testProcess` object just like before. Whenever you have a test like this that relies on a specific state of a dependency (i.e. `process`) you need to explicitly set up that state, even though it's supposed to be the default state. Something like this:
````javascript
const testProcess = {
  // Empty env to simulate not declaring NODE_ENV.
  env: {},
};
defaultVersionGetter(root, testProcess)
````

File: src/program.js [hunk 3]
```diff
@@ -124,10 +126,19 @@ export class Program {
...
-export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
-    path.join(absolutePackageDir, 'package.json'));
-  return JSON.parse(packageData).version;
+export function defaultVersionGetter(
+  absolutePackageDir: string,
+  localProcess: Object = process
+): string {
+  if (localProcess.env.NODE_ENV === 'production') {
+    log.debug('Getting the version from package.json');
```
kumar303: oops, the code needs this fix to make the log statements show up in `--verbose` mode:
````diff
index eb9e8ae..5bf90ae 100644
--- a/src/program.js
+++ b/src/program.js
@@ -92,8 +92,8 @@ export class Program {
     let runCommand = this.commands[cmd];
 
     if (argv.verbose) {
-      log.info('Version:', getVersion(absolutePackageDir));
       logStream.makeVerbose();
+      log.info('Version:', getVersion(absolutePackageDir));
     }
 
     try {
````

**Comment by kumar303** [comment]
One thing I can think of is to define a magic global at compile time (when `npm run build` executes), using the [DefinePlugin](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin). You could add this to `webpack.config.js`:
````javascript
new webpack.DefinePlugin({
    WEBEXT_BUILD_ENV: JSON.stringify(process.env.NODE_ENV || 'development'),
})
````

File: tests/unit/test.program.js [hunk 4]
```diff
@@ -329,15 +331,27 @@ describe('program.main', () => {
 
 
 describe('program.defaultVersionGetter', () => {
-
-  it('returns the package version', () => {
-    let root = path.join(__dirname, '..', '..');
+  let root = path.join(__dirname, '..', '..');
+  let testProcess = {
```
kumar303: This shouldn't be shared be a shared variable because it makes the tests unpredictable (say, if they were run in reverse order). I'd say keep it defined as a local constant within each respective test function.

src/program.js
```diff
@@ -123,11 +125,21 @@ export class Program {
   }
 }
 
+declare var WEBEXT_BUILD_ENV: string;
```
kumar303: It might be worth adding a comment above this explaining how this is generated by webpack and that you can see how it's generated in `webpack.config.js`.

tests/unit/test.program.js
```diff
@@ -166,6 +168,7 @@ describe('program.Program', () => {
 
   it('configures the logger when verbose', () => {
     const logStream = fake(new ConsoleStream());
+    let version = spy();
```
kumar303: always use `const` if the variable does not need to be modified

src/program.js
```diff
@@ -123,11 +125,21 @@ export class Program {
...
 
-export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
-    path.join(absolutePackageDir, 'package.json'));
-  return JSON.parse(packageData).version;
+export function defaultVersionGetter(
+  absolutePackageDir: string,
+  localEnv: string = WEBEXT_BUILD_ENV
```
kumar303: Since this is an optional argument, it should be defined as a keyword argument instead. This help makes the calls easier to read and it makes it easy to add more options later on. Example:
````javascript
export function defaultVersionGetter(
  absolutePackageDir: string,
  {localEnv: string = WEBEXT_BUILD_ENV} = {}
): string {
````

File: tests/unit/test.program.js [hunk 8]
```diff
@@ -175,7 +178,7 @@ describe('program.Program', () => {
     });
     program.command('thing', 'does a thing', () => {});
 
-    return execProgram(program, {logStream})
+    return execProgram(program, {getVersion: version, logStream})
```
kumar303: it looks like this can just be `getVersion: spy()` since the `version` variable isn't used elsewhere

src/program.js
```diff
@@ -123,11 +125,22 @@ export class Program {
...
+// A global variable generated by DefinePlugin, generated in webpack.config.js
+declare var WEBEXT_BUILD_ENV: string;
 
-export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
-    path.join(absolutePackageDir, 'package.json'));
-  return JSON.parse(packageData).version;
+export function defaultVersionGetter(
+  absolutePackageDir: string,
+  {localEnv: string = WEBEXT_BUILD_ENV} = {}
```
kumar303: Oh, sorry, I gave you the wrong example code. The lint error is correct and it needs to be fixed. The problem is actually an obscure ES6 vs Flow problem so let me offer the solution to help you move forward.

First, define a type for the option object that needs to be passed to `defaultVersionGetter()`, like this:

````javascript
type versionGetterOptions = {
  localEnv?: string,
};
````
Next, fix the definition of `defaultVersionGetter` to reference the new type:
````javascript
export function defaultVersionGetter(
  absolutePackageDir: string,
  {localEnv = WEBEXT_BUILD_ENV}: versionGetterOptions = {}
): string {
````
After this change, you'll need to update all the calls to `defaultVersionGetter()`. The flow checks in `npm start` should help guide you on what to change and where.

File: src/program.js [hunk 10]
```diff
@@ -123,11 +125,27 @@ export class Program {
...
 }
 
+// A global variable generated by DefinePlugin, generated in webpack.config.js
+declare var WEBEXT_BUILD_ENV: string;
 
-export function defaultVersionGetter(absolutePackageDir: string): string {
-  let packageData: any = readFileSync(
-    path.join(absolutePackageDir, 'package.json'));
-  return JSON.parse(packageData).version;
+//A defintion of type of argument for defaultVersionGetter
```
kumar303: Just for future note: comments typically begin with a space elsewhere in the codebase, like `// A definition...`

tests/unit/test.program.js
```diff
@@ -329,15 +331,23 @@ describe('program.main', () => {
...
-  it('returns the package version', () => {
-    let root = path.join(__dirname, '..', '..');
+  it('returns the package version in production', () => {
     let pkgFile = path.join(root, 'package.json');
     return fs.readFile(pkgFile)
       .then((pkgData) => {
-        assert.equal(defaultVersionGetter(root),
-                     JSON.parse(pkgData).version);
+        const testBuildEnv = {localEnv: 'production'};
+        assert.equal(defaultVersionGetter(root, testBuildEnv),
```
kumar303: You could make this a bit more concise like:
````
assert.equal(defaultVersionGetter(root, {localEnv: 'production'})
````
Not a big deal though, don't worry about changing it.

saintsebastian: I was on the fence about doing that, thanks for clarification.

saintsebastian: Also I don't mind fixing these, I've committed them to my branch in case you have time to merge them as well.

kumar303: hmm, it's a bit tricky to pick those from a different branch. No worries! They aren't important changes but thanks.

atsay:
I've added your contribution to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#October_2016)--thanks, and welcome to our project!

PR #568, opened 2016-10-10

saintsebastian:
fixes #487


kumar303:
To make it easier to review, let's add new changes to the original pull request, https://github.com/mozilla/web-ext/pull/565

PR #577, opened 2016-10-15
fix: Don't log info about the built extension when signing

### LINKED ISSUE #181: Don't log info about the built extension when signing
When signing an extension, web-ext will first build the extension in a temp directory. When running in non-verbose mode, there is no reason to output information about this since it's an implementation detail. This *extension is ready* log line should be hidden when not running with `--verbose`:

````
$ web-ext sign --api-key ... --api-secret ...
Building web extension from /Users/kumar/dev/bookmarker-ext/moz-webext
Your web extension is ready: /var/folders/h7/rttdqkks7fl_txp9_rr4j4840000gn/T/tmp-web-ext-1114CJEXjz51BXXp/getting_started_example-1.2.25.xpi
Validating add-on [..........................................................................................................................................................................................................................................................]
Validation results: https://addons.mozilla.org/en-US/developers/upload/01f733ef20b443fcaa5973e6860d1326
Downloading signed files: 100%
Downloaded:
    ./web-ext-artifacts/getting_started_example-1.2.25-fx.xpi
SUCCESS
````

### PR DESCRIPTION
**Pull request by saintsebastian**
Fix #181 
Manually tested with example addon

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -79,9 +79,8 @@ export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
 async function defaultPackageCreator(
-  {manifestData, sourceDir, fileFilter, artifactsDir,
-    showReadyMessage}: PackageCreatorParams
-): Promise<ExtensionBuildResult> {
+  {manifestData, sourceDir, fileFilter, artifactsDir, showReadyMessage,
+}: PackageCreatorParams): Promise<ExtensionBuildResult> {
   let id;
   if (manifestData) {
     id = getManifestId(manifestData);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -37,6 +37,7 @@ export type PackageCreatorParams = {
   sourceDir: string,
   fileFilter: FileFilter,
   artifactsDir: string,
+  showReadyMessage: boolean
 };
 
 export type LocalizedNameParams = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/cmd/sign.js [hunk 1]
```diff
@@ -69,7 +69,8 @@ export default function sign(
       }
 
       let [buildResult, idFromSourceDir] = await Promise.all([
-        build({sourceDir, artifactsDir: tmpDir.path()}, {manifestData}),
+        build({sourceDir, artifactsDir: tmpDir.path()}, {manifestData,
+          showReadyMessage: false}),
```
kumar303: nitpicky request (sorry): when breaking a block into a new line, the break needs to be consistent. In this case the block is `{manifestData, showReadyMessage: false}`. You could make it break consistently in a couple different ways:

Here is an example of keeping all block elements on the same line:
````javascript
build({sourceDir, artifactsDir: tmpDir.path()}, {
  manifestData, showReadyMessage: false,
}),
````
Here is an example of moving the entire block to a new line:
````javascript
build({sourceDir, artifactsDir: tmpDir.path()}, 
      {manifestData, showReadyMessage: false}),
````
This type of consistency is hard to enforce with automatic lint rules so sorry about slowing you down to fix that. It's important to fix these kinds of things because otherwise other contributors may become distracted by the inconsistency when working on an unrelated feature.

Comment (saintsebastian) [inline src/cmd/sign.js]: @kumar303 no problem, good to learn the something new

File: src/cmd/build.js [hunk 2]
```diff
@@ -78,7 +79,8 @@ export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
 async function defaultPackageCreator(
-  {manifestData, sourceDir, fileFilter, artifactsDir}: PackageCreatorParams
+  {manifestData, sourceDir, fileFilter, artifactsDir,
+    showReadyMessage}: PackageCreatorParams
```
kumar303: oops, I completely missed this one. This is also formatted inconsistently because the block is only partially broken into the next line. An easy fix would be:
````javascript
async function defaultPackageCreator({
  manifestData, sourceDir, fileFilter, artifactsDir, showReadyMessage
}: PackageCreatorParams): Promise<ExtensionBuildResult> {
````

PR #586, opened 2016-10-17
Sign command exits with zero status when signing fails

### LINKED ISSUE #585: Sign command exits with zero status when signing fails

Bug

1. Run `web-ext sign` to sign an add-on with version X
2. Run `web-ext sign` again with the same version in the manifest. Signing will fail and the command will print "FAIL" as the last output
3. Check the exit status of the `web-ext sign` command (eg. via `echo $?` in bash). See that it is zero.

When signing fails, the command should exit with a non-zero status, so that `make` or a deployment shell script can react accordingly.

### Version information

(Please fill this in for bug reports.)

web-ext 1.5.0
node v6.8.1
Ubuntu 16.04

shubheksha:
Fixes #585 
- throw a `UsageError` when the sign command fails to sign an add on
- modified the current test which makes the test fail to check for the error
- verified the output of `$ echo $?`, it prints 1

@kumar303, will this work? I don't think we need a wrapper or am I missing something?

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -120,6 +120,8 @@ export default function sign(
         log.info('SUCCESS');
       } else {
         log.info('FAIL');
+        throw new UsageError(
+            'Add on could not be signed, exiting with status 1');
       }
 
       return signingResult;
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -5,7 +5,7 @@ import defaultAddonSigner from 'sign-addon';
 
 import defaultBuilder from './build';
 import {withTempDir} from '../util/temp-dir';
-import {isErrorWithCode, UsageError} from '../errors';
+import {isErrorWithCode, UsageError, WebExtError} from '../errors';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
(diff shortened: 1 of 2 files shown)


src/cmd/sign.js
```diff
@@ -120,6 +120,8 @@ export default function sign(
         log.info('SUCCESS');
       } else {
         log.info('FAIL');
+        throw new UsageError(
+            'Add on could not be signed, exiting with status 1');
```
kumar303: nit: this only needs to be indented two spaces 🍟

src/cmd/sign.js
```diff
@@ -120,6 +120,8 @@ export default function sign(
         log.info('SUCCESS');
       } else {
         log.info('FAIL');
+        throw new UsageError(
```
kumar303: I thought more about this and I think it should be `WebExtError` because it may not always be something that the developer can fix. For example, signing might fail because there was a 500 response from the API, we really can't be sure.

tests/unit/test-cmd/test.sign.js
```diff
@@ -248,8 +248,11 @@ describe('sign', () => {
             success: false,
           }),
         })
-        .then((result) => {
-          assert.equal(result.success, false);
+        .then(makeSureItFails())
+        .catch((error) => {
+          assert.instanceOf(error, UsageError);
+          assert.match(error.message,
+            /Add on could not be signed, exiting with status 1/);
```
kumar303: I think the error message can just be this:

    The WebExtension could not be signed

A couple reasons:
* the exist status isn't so important
* for consistency, we try to refer to add-ons as WebExtensions

PR #588, opened 2016-10-17
chore: improve linting rules for brace styles

kumar303:
I added a few rules after I had to be an annoying reviewer on [this patch](https://github.com/mozilla/web-ext/pull/577) :) 

* Added [brace-style](http://eslint.org/docs/rules/brace-style) to catch some if/else consistency problems I've been seeing lately (I actually haven't been nagging about these in reviews)
* Added deeper [indent](http://eslint.org/docs/rules/indent) rules. These are nits that I've commented about on reviews.
* Added [object-property-newline](http://eslint.org/docs/rules/object-property-newline). This is an attempt to address inconsistent new lines in property lists but doesn't fully address it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -54,6 +54,7 @@
         "body": 1
       }
     }],
+    "key-spacing": 2,
     "keyword-spacing": 2,
     "max-len": [2, 80, 2, {"ignoreComments": true}],
     "new-cap": [2, {"capIsNewExceptions": ["Deferred"]}],
--- a/.eslintrc
+++ b/.eslintrc
@@ -39,7 +39,20 @@
     "dot-notation": [2, {"allowKeywords": true}],
     "eqeqeq": [2, "allow-null"],
     "guard-for-in": 0,
-    "indent": [2, 2],
+    "indent": [2, 2, {
+      "SwitchCase": 1,
+      "VariableDeclarator": 1,
+      "outerIIFEBody": 1,
+      "MemberExpression": 1,
+      "FunctionDeclaration": {
+        "parameters": 1,
+        "body": 1
+      },
+      "FunctionExpression": {
+        "parameters": 1,
+        "body": 1
+      }
+    }],
     "keyword-spacing": 2,
     "max-len": [2, 80, 2, {"ignoreComments": true}],
     "new-cap": [2, {"capIsNewExceptions": ["Deferred"]}],
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,6 +31,7 @@
   "rules": {
     "arrow-parens": 2,
     "block-scoped-var": 0,
+    "brace-style": ["error", "1tbs", {"allowSingleLine": false}],
     "camelcase": 0,
     "comma-dangle": [2, "always-multiline"],
     "comma-spacing": 2,
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,7 +31,7 @@
   "rules": {
     "arrow-parens": 2,
     "block-scoped-var": 0,
-    "brace-style": ["error", "1tbs", {"allowSingleLine": false}],
+    "brace-style": [2, "1tbs", {"allowSingleLine": false}],
     "camelcase": 0,
     "comma-dangle": [2, "always-multiline"],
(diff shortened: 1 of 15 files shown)


kumar303:
@rpl could you take a look? 

The one I can't seem to find a rule for is to prevent this from happening:
````javascript
foo({bar: 1,
  baz: 2,
});
````
for consistency it should be one of these three:
````javascript
foo({bar: 1, baz: 2});
foo({
  bar: 1, baz: 2,
});
foo({
  bar: 1, 
  baz: 2,
});
````
For some reason the [object-curly-newline](http://eslint.org/docs/rules/object-curly-newline) rule has too many false positives.

File: .eslintrc [hunk 1]
```diff
@@ -31,6 +31,7 @@
   "rules": {
     "arrow-parens": 2,
     "block-scoped-var": 0,
+    "brace-style": ["error", "1tbs", {"allowSingleLine": false}],
```
rpl: I would prefer to use `2` instead of `"error"`, here and on the new `"object-property-newline"` eslint rules configuration, for consistency with the rest of the rules.

.eslintrc
```diff
@@ -76,6 +90,9 @@
     "no-underscore-dangle": 0,
     "no-unused-vars": 2,
     "no-with": 2,
+    "object-property-newline": ["error", {
```
rpl: Same here ("error" -> 2)

.eslintrc
```diff
@@ -39,7 +40,20 @@
...
+      "FunctionDeclaration": {
+        "parameters": 1,
+        "body": 1
+      },
+      "FunctionExpression": {
+        "parameters": 1,
+        "body": 1
+      }
+    }],
     "keyword-spacing": 2,
```
rpl: An additional rule that we can add here is ["key-spacing"](http://eslint.org/docs/rules/key-spacing), I tried to add it in our config (without changing anything in the rule default behavior) and it caught a couple of spacing errors in `src/firefox/preferences.js` and `src/program.js`.

src/program.js
```diff
@@ -78,10 +80,11 @@ export class Program {
     return this;
   }
 
-  async execute(absolutePackageDir: string,
-      {systemProcess = process, logStream = defaultLogStream,
-       getVersion = defaultVersionGetter, shouldExitProgram = true}
-      : Object = {}): Promise<void> {
+  async execute(
+    absolutePackageDir: string,
+    {systemProcess = process, logStream = defaultLogStream,
```
rpl: Nit, imho this kind of function declaration can be more readable written as follow (but as a matter of taste, it is not a required change for this pull request, just an opinion):

```js
async execute(
  absolutePackageDir: string,
  {
    systemProcess = process, logStream = defaultLogStream,
    ...
  }: Object = {}
): Promise<void> {
  ...
}
```

tests/unit/test-cmd/test.build.js
```diff
@@ -28,37 +28,39 @@ describe('build', () => {
...
-          return buildResult.extensionPath;
-        })
-        .then((extensionPath) => zipFile.open(extensionPath))
-        .then(() => zipFile.extractFilenames())
-        .then((fileNames) => {
-          fileNames.sort();
-          assert.deepEqual(fileNames,
-                           ['background-script.js', 'manifest.json']);
-        })
+        build(
```
rpl: Nit, for chains of promises I usually opt for the following style (which seems to be compatible with the eslint rules enforced in this branch, but, as the above, it is a matter of taste and it is not a mandatory required change for this review):

```js
    return withTempDir(
      (tmpDir) =>
        build({
          sourceDir: fixturePath('minimal-web-ext'),
          artifactsDir: tmpDir.path(),
        }).then((buildResult) => {
          assert.match(buildResult.extensionPath,
                       /minimal_extension-1\.0\.zip$/);
          return buildResult.extensionPath;
        }).then(
          (extensionPath) => zipFile.open(extensionPath)
        ).then(
          () => zipFile.extractFilenames()
        ).then((fileNames) => {
          fileNames.sort();
          assert.deepEqual(fileNames,
                           ['background-script.js', 'manifest.json']);
        })
    );
```

Issue #575, opened 2016-10-14
web-ext --version Error: Cannot find module 'regenerator-runtime


onazaro:
### Is this a bug or feature request?
Bug

### What is the current behavior?
Right after web-ext istallation, try:
`web-ext --version`

And get the error message:
`Error: Cannot find module 'regenerator-runtime`
`    at Function.Module._resolveFilename (module.js:325:15)`
`    at Function.Module._load (module.js:276:25)`
`    at Module.require (module.js:353:17)`
`    at require (internal/module.js:12:17)`
`    at Object.module.exports.exports.__esModule (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\webpack:\external "
regenerator-runtime":1:1)`
`    at __webpack_require__ (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\webpack:\webpack\bootstrap 96f6c32d219c1
eb025b0:19:1)`
`    at Object.<anonymous> (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\webpack:\~\babel-runtime\regenerator\inde
x.js:1:1)`
`    at __webpack_require__ (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\webpack:\webpack\bootstrap 96f6c32d219c1
eb025b0:19:1)`
`    at Object.<anonymous> (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\webpack:\src\program.js:9:215)`
`    at Object.<anonymous> (C:\Documents and Settings\Администратор\Application Data\npm\node_modules\web-ext\dist\web-ext.js:148:31)`

Prior to it, have two warnings installing web-ext:
`npm WARN engine web-ext@1.5.0: wanted: {"node":">=4.0.0","npm":">=3.0.0"} (current: {"node":"4.6.0","npm":"2.15.9"})`
`npm WARN optional dep failed, continuing fsevents@1.0.14`

### What is the expected or desired behavior?
Print web-ext version.

### Version information

(Please fill this in for bug reports.)

* Output of `web-ext --version`: n/a
* Output of `node --version`: v4.6.0
* Output of `npm --version`: 2.15.9
* Firefox version: 47.0
* Your OS and version: Windows XP SP3 32

kumar303:
[...]

PR #591, opened 2016-10-20
fix: Fixed `Error: Cannot find module 'regenerator-runtime'`

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/575

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "minimatch": "3.0.3",
     "mz": "2.4.0",
     "node-firefox-connect": "1.2.0",
+    "regenerator-runtime": "0.9.5",
     "sign-addon": "0.2.0",
     "source-map-support": "0.4.5",
     "stream-to-promise": "2.2.0",


kumar303:
@rpl does this look right to you? Should I also move `babel-runtime` to `dependencies` ?

kumar303:
The CI [tests](https://github.com/mozilla/web-ext/blob/master/.travis.yml#L4) were not catching this because they do `npm install` within the project (and that doesn't reproduce the bug). Oddly, I could not reproduce it from a source build even after removing all `node_modules` and doing `npm install --production`.

rpl:
@kumar303 I've also took a look into the content of `babel-runtime`, there is another dependency loaded from it (`core-js`), but it seems to be currently unused.

I would say that probably "`babel-runtime` added as a direct dependency" can be reasonable, because what the final "web-ext built file" needs at runtime can probably change over the time, based on which features of the babel runtime are enabled and used in the sources.

In #593, as a additional "safety net" for this kind of issues, I tried to put together an additional "smoke test" which should be able to better catch them during travis builds (the script currently refuses to run if the TRAVIS environment var is not defined, so that a developer do not accidentaly destroy his local `node_modules/` dir during his regular local development and testing)

Unfortunately travis seems to have some issues and it is not running automatically the tests on that pull request (I'll keep on eye on the travis status page and I'll repush it to force a build once the current travis issues are solved, so that we can check that the smoke test fails on the current master, and then pass with this fix applied).

PR #593, opened 2016-10-21
test: run a minimal set of 'production mode' smoke tests on travis.

rpl:
This PR contains a small shell script that run an additional minimal set of smoke tests on travis
(and only on travis, because it is going to destroy the `node_modules` dir to recereate it using `npm install --production`).

This additional smoke test should be able to catch issues related to missing dependencies when `web-ext` is installed in production mode (`npm install --production`), as in #575.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ before_script:
 script:
 - COVERAGE=y NODE_ENV=production npm test
 - npm run changelog-lint
+# Run a minimal set of smoke tests in production mode (npm install --production)
+- ./tests/smoke-test-production.sh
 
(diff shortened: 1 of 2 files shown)


rpl:
... that time when you really want a test to fail and it passes instead... how disappointing :-)

locally the test failed as expected, but I tried it on an environment with npm 2.14.12 (as in #575 comments), while the standard travis "nodejs 4" environment provides npm 2.15.9.

I'm going to try it locally with npm 2.15.9

rpl:
Closing, as this has been superseded and fixed in #1082

PR #594, opened 2016-10-21
fix: Format JSON parse errors so that they provide file context

### LINKED ISSUE #491: Format JSON parse errors so that they provide file context
I was trying to build an add-on with a manifest.json that started off like:

``` json
{
    "applications": {
      "gecko": {
        "id": "example-addon-repo@mozilla.org",
          "strict_min_version": "48.0",
          "strict_max_version": "50.*",
      }
    },
```

Running `web-ext build` gave:

``` shell
Building web extension from /Users/mark/dev/example-addon-repo

build: InvalidManifest: Error parsing manifest.json at /Users/mark/dev/example-addon-repo/manifest.json: SyntaxError: Unexpected token } in JSON at position 175
    at /Users/mark/dev/example-addon-repo/node_modules/web-ext/dist/webpack:/src/util/manifest.js:9:1805
```

This is confusing. At a minimum, it should give you the row/column of the unexpected token. Preferably, it might also point you to the possibility of an extra comma.

shubheksha:
Fixes #491 
Used the [json-parse](https://github.com/sindresorhus/parse-json) to display line column number for providing context for syntax errors in JSON files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "node-firefox-connect": "1.2.0",
     "regenerator-runtime": "0.10.0",
     "parse-json": "2.2.0",
-    "regenerator-runtime": "0.9.5",
     "sign-addon": "0.2.0",
     "source-map-support": "0.4.6",
     "stream-to-promise": "2.2.0",
--- a/package.json
+++ b/package.json
@@ -56,8 +56,10 @@
     "git-rev-sync": "1.8.0",
     "minimatch": "3.0.3",
     "mz": "2.6.0",
-    "node-firefox-connect": "1.2.0",
+    "node-firefox-connect": "1.2.0",< HEAD
     "regenerator-runtime": "0.10.0",
+    "parse-json": "2.2.0",
+    "regenerator-runtime": "0.9.5",ded parse-json package to package.json
     "sign-addon": "0.2.0",
     "source-map-support": "0.4.6",
     "stream-to-promise": "2.2.0",
--- a/package.json
+++ b/package.json
@@ -56,10 +56,10 @@
     "git-rev-sync": "1.8.0",
     "minimatch": "3.0.3",
     "mz": "2.6.0",
-    "node-firefox-connect": "1.2.0",< HEAD
+    "node-firefox-connect": "1.2.0",
     "regenerator-runtime": "0.10.0",
     "parse-json": "2.2.0",
-    "regenerator-runtime": "0.9.5",ded parse-json package to package.json
+    "regenerator-runtime": "0.9.5",
     "sign-addon": "0.2.0",
     "source-map-support": "0.4.6",
     "stream-to-promise": "2.2.0",
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -4,7 +4,7 @@ import minimatch from 'minimatch';
 import {createWriteStream} from 'fs';
 import streamToPromise from 'stream-to-promise';
 import {fs} from 'mz';
-import parseJson from 'parse-json';
+import parseJSON from 'parse-json';
 
 import defaultSourceWatcher from '../watcher';
 import {zipDir} from '../util/zip-dir';
@@ -60,14 +60,23 @@ export async function getDefaultLocalizedName(
 ): Promise<string> {
 
   let messageData: LocalizedMessageData;
+  let messageContents: string|Buffer;
   let extensionName: string = manifestData.name;
 
   try {
-    messageData = parseJson(await fs.readFile(messageFile));
+    messageContents = await fs.readFile(messageFile);
   } catch (error) {
-    error.fileName = messageFile;
-    throw new UsageError(error);
+    throw new UsageError(
+      `Could not read messages.json file at ${messageFile}: ${error}`);
   }
+
+  try {
+    messageData = parseJSON(messageContents, messageFile);
+  } catch (error) {
+    throw new UsageError(
+      `Error parsing manifest.json at ${messageFile}: ${error}`);
+  }
+
   extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
     (match, messageName) => {
       if (!(messageData[messageName]
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -60,7 +60,7 @@ export async function getDefaultLocalizedName(
 ): Promise<string> {
 
   let messageData: LocalizedMessageData;
-  let messageContents: string|Buffer;
+  let messageContents: string | Buffer;
   let extensionName: string = manifestData.name;
 
   try {
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -4,6 +4,7 @@ import minimatch from 'minimatch';
 import {createWriteStream} from 'fs';
 import streamToPromise from 'stream-to-promise';
 import {fs} from 'mz';
+import parseJson from 'parse-json';
 
 import defaultSourceWatcher from '../watcher';
 import {zipDir} from '../util/zip-dir';
@@ -62,10 +63,11 @@ export async function getDefaultLocalizedName(
   let extensionName: string = manifestData.name;
 
   try {
-    messageData = JSON.parse(await fs.readFile(messageFile));
-  } catch (error) {
-    throw new UsageError(
-      `Error reading or parsing file ${messageFile}: ${error}`);
+    messageData = parseJson(await fs.readFile(messageFile));
+  }
+  catch (error) {
+    error.fileName = messageFile;
+    throw new UsageError(error);
   }
   extensionName = manifestData.name.replace(/__MSG_([A-Za-z0-9@_]+?)__/g,
     (match, messageName) => {
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -64,8 +64,7 @@ export async function getDefaultLocalizedName(
 
   try {
     messageData = parseJson(await fs.readFile(messageFile));
-  }
-  catch (error) {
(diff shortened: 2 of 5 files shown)


src/cmd/build.js
```diff
@@ -4,6 +4,7 @@ import minimatch from 'minimatch';
 import {createWriteStream} from 'fs';
 import streamToPromise from 'stream-to-promise';
 import {fs} from 'mz';
+import parseJson from 'parse-json';
```
rpl: Nit: I know that `parseJson` is how this import is named in the examples from 'parse-json` readme file, but I would prefer it named as `parseJSON`.

src/cmd/build.js
```diff
@@ -53,10 +54,10 @@ export async function getDefaultLocalizedName(
   let extensionName: string = manifestData.name;
 
   try {
-    messageData = JSON.parse(await fs.readFile(messageFile));
+    messageData = parseJson(await fs.readFile(messageFile));
   } catch (error) {
-    throw new UsageError(
-      `Error reading or parsing file ${messageFile}: ${error}`);
+    error.fileName = messageFile;
```
rpl: Instead of setting the `error.fileName` property on the raised error, we should use the JSON file name as the second parameter of the `parseJSON` call (e.g. `parseJSON(content, fileName)`), by looking at the examples in the 'parse-json' README, it seems that this is how the file name is supposed to be officially associated to the raised exception (the same is needed for the other `parseJSON` calls in "src/util/manifest.js").

src/cmd/build.js
```diff
@@ -53,10 +54,10 @@ export async function getDefaultLocalizedName(
...
+    throw new UsageError(error);
```
rpl: `UsageError` constructor is supposed to receive a `string` error message as its only parameter, while the above line set the error message to an error object, and so it should be changed into something like: 
```js
  throw new UsageError(`Error parsing locale file at ${messageFile}: ${error}`);
```

src/cmd/build.js
```diff
@@ -53,10 +54,10 @@ export async function getDefaultLocalizedName(
   let extensionName: string = manifestData.name;
 
   try {
-    messageData = JSON.parse(await fs.readFile(messageFile));
+    messageData = parseJson(await fs.readFile(messageFile));
```
rpl: This line can raise error for two different reasons:

- the `await fs.readFile(...)` part rejects (e.g. because the file doesn't exist or it cannot be read because of file permissions issues)
- the `parseJSON(...)` part rejects (e.g. because the file is not valid JSON)

In `src/util/manifest.js` we currently handle this two exceptions separately:

```js
  let manifestContents;

  try {
    manifestContents = await fs.readFile(manifestFile);
  } catch (error) {
    throw new InvalidManifest(
      `Could not read manifest.json file at ${manifestFile}: ${error}`);
  }

  let manifestData;

  try {
    manifestData = parseJSON(manifestContents, messageFile);
  } catch (error) {
    throw new InvalidManifest(
      `Error parsing manifest.json at ${manifestFile}: ${error}`);
  }
```

we should do something similar here (but with `UsageError` instead of `InvalidManifest`)

tests/unit/test-cmd/test.build.js
```diff
@@ -104,7 +104,7 @@ describe('build', () => {
             assert.instanceOf(error, UsageError);
             assert.match(
               error.message,
-              /Error .* file .*messages\.json: SyntaxError: Unexpected string/);
+              /Unexpected token '"' .* in .*messages.json/);
```
rpl: How about splitting this assertion for readability, e.g.:

```js
assert.match(error.message, /Unexpected token '"' at 1:15/);
assert.include(error.message, messageFileName);
```

also, we should test the same change in "tests/unit/test-util/test.manifest.js", e.g. something like:
```diff
diff --git a/tests/unit/test-util/test.manifest.js b/tests/unit/test-util/test.manifest.js
index 166a276..18f8a95 100644
--- a/tests/unit/test-util/test.manifest.js
+++ b/tests/unit/test-util/test.manifest.js
@@ -70,7 +70,8 @@ describe('util/manifest', () => {
           .then(() => getValidatedManifest(tmpDir.path()))
           .then(makeSureItFails())
           .catch(onlyInstancesOf(InvalidManifest, (error) => {
-            assert.match(error.message, /Error parsing manifest\.json/);
+            assert.match(error.message, /Error parsing manifest\.json at /);
+            assert.include(error.message, 'Unexpected token \' \' at 2:49');
             assert.include(error.message, manifestFile);
           }));
       }
```

shubheksha:
@rpl, so sorry this was stalled for such a long time. I've addressed everything you mentioned. There were merge conflicts in a couple of files and I tried to do my best to solve them (hopefully) without messing anything up. :smile:

rpl:
@shubheksha no worries, we are going to have a lot of time to work together on web-ext in the next months ;-)

I'm looking into the updated version right now.

package.json
```diff
@@ -58,6 +58,8 @@
     "mz": "2.6.0",
     "node-firefox-connect": "1.2.0",
     "regenerator-runtime": "0.10.0",
+    "parse-json": "2.2.0",
+    "regenerator-runtime": "0.9.5",
```
rpl: "regenerator-runtime" is included twice (it is already on line 60), we should remove the "0.9.5" version and move "parse-json" to keep the dependencies ordered by name.

src/cmd/build.js
```diff
@@ -59,14 +60,23 @@ export async function getDefaultLocalizedName(
...
+  let messageContents: string | Buffer;
   let extensionName: string = manifestData.name;
 
   try {
-    messageData = JSON.parse(await fs.readFile(messageFile));
+    messageContents = await fs.readFile(messageFile);
   } catch (error) {
     throw new UsageError(
-      `Error reading or parsing file ${messageFile}: ${error}`);
+      `Could not read messages.json file at ${messageFile}: ${error}`);
```
rpl: how about changing this message to `Error reading messages.json file at ...`?

src/cmd/build.js
```diff
@@ -59,14 +60,23 @@ export async function getDefaultLocalizedName(
...
   }
+
+  try {
+    messageData = parseJSON(messageContents, messageFile);
+  } catch (error) {
+    throw new UsageError(
+      `Error parsing manifest.json at ${messageFile}: ${error}`);
```
rpl: This should be `Error parsing messages.json`

rpl: it looks like we can remove the `at ${messageFile}` part here, because the full path name is part of the error raised by parseJSON

rpl: @kumar303 do we want to create an explicit subclass of UsageError for this error scenario? (e.g. InvalidMessages)

src/util/manifest.js
```diff
@@ -45,8 +46,9 @@ export default async function getValidatedManifest(
   let manifestData;
 
   try {
-    manifestData = JSON.parse(manifestContents);
+    manifestData = parseJSON(manifestContents, manifestFile);
   } catch (error) {
+    error.fileName = manifestFile;
```
rpl: we should remove this `error.fileName = manifestFile` from the `catch` block.

src/util/manifest.js
```diff
@@ -45,8 +46,9 @@ export default async function getValidatedManifest(
...
     throw new InvalidManifest(
       `Error parsing manifest.json at ${manifestFile}: ${error}`);
```
rpl: we can remove the `at ${manifestFile}` part here, because the full path name is now part of the error raised by parseJSON

tests/unit/test-cmd/test.build.js
```diff
@@ -102,9 +102,8 @@ describe('build', () => {
           .then(makeSureItFails())
           .catch((error) => {
             assert.instanceOf(error, UsageError);
-            assert.match(
-              error.message,
-              /Error .* file .*messages\.json: SyntaxError: Unexpected string/);
+            assert.match(error.message, /Unexpected token '"' at 1:15/);
```
rpl: we should add an additional assertion here to check that the error message starts with the expected message, e.g. (using '^' in the regular expression):
```js
assert.match(error.message, /^Error parsing messages.json/);
```

tests/unit/test-cmd/test.build.js
```diff
@@ -150,7 +149,7 @@ describe('build', () => {
         assert.instanceOf(error, UsageError);
         assert.match(
           error.message,
-          /Error .* file .*messages\.json: .*: no such file or directory/);
+          /Error: ENOENT: no such file or directory, open .*messages.json/);
```
rpl: we should add an assertion that check how our error message starts (e.g. using a regexp like `/^Error reading messages.json/`) and also check that the full path of the messages.json file is included in the error message (like the `assert.include(error.message, messageFileName)` above

src/cmd/build.js
```diff
@@ -59,14 +60,23 @@ export async function getDefaultLocalizedName(
...
     throw new UsageError(
-      `Error reading or parsing file ${messageFile}: ${error}`);
+      `Error reading messages.json file at ${messageFile}: ${error}`);
   }
+
+  try {
+    messageData = parseJSON(messageContents, messageFile);
+  } catch (error) {
+    throw new UsageError(
+      `Error parsing messages.json ${error}`);
```
rpl: nit, this is probably short enough now to fit in line 76

kumar303:
@shubheksha thanks, this looks great! @rpl thanks for reviewing.

> do we want to create an explicit subclass of UsageError for this new error scenario? (e.g. InvalidMessages)

I can't think of any concrete reason to make a subclass. We can always add it later if we think of a concrete reason.

PR #613, opened 2016-10-26
Support warnings-as-errors

josteink:
This patch adds support for using `--warnings-as-errors` as a parameter for the lint command.

This enables the functionality provided in the PR https://github.com/mozilla/addons-linter/pull/1016 and partially resolves the issue https://github.com/mozilla/addons-linter/issues/1014.

Once `addons-linter` has PR accepted and merged, [packages.json](https://github.com/mozilla/web-ext/blob/master/package.json#L47) will also need to have the package-version bumped to fully resolve and close the issue.

### DIFF
--- a/tests/unit/test-cmd/test.lint.js
+++ b/tests/unit/test-cmd/test.lint.js
@@ -65,6 +65,22 @@ describe('lint', () => {
     });
   });
 
+  it('passes warningsAsErrors to the linter', () => {
+    const {lint, createLinter} = setUp();
+    return lint({warningsAsErrors: true}).then(() => {
+      const config = createLinter.firstCall.args[0].config;
+      assert.equal(config.warningsAsErrors, true);
+    });
+  });
+
+  it('passes warningsAsErrors undefined to the linter', () => {
+    const {lint, createLinter} = setUp();
+    return lint({}).then(() => {
+      const config = createLinter.firstCall.args[0].config;
+      assert.equal(config.warningsAsErrors, undefined);
+    });
+  });
+
   it('configures the linter when verbose', () => {
     const {lint, createLinter} = setUp();
     return lint({verbose: true}).then(() => {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "opera"
   ],
   "dependencies": {
-    "addons-linter": "0.15.6",
+    "addons-linter": "0.15.10",
     "babel-polyfill": "6.16.0",
     "babel-runtime": "6.11.6",
     "bunyan": "1.8.4",
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -16,6 +16,7 @@ export type LinterCreatorParams = {
     logLevel: 'debug' | 'fatal',
     stack: boolean,
     pretty?: boolean,
+    warningsAsErrors?: boolean,
     metadata?: boolean,
     output?: LinterOutputType,
     boring?: boolean,
@@ -42,6 +43,7 @@ export type LintCmdParams = {
   output?: LinterOutputType,
   metadata?: boolean,
   pretty?: boolean,
+  warningsAsErrors?: boolean,
 };
 
 export type LintCmdOptions = {
@@ -52,7 +54,7 @@ export type LintCmdOptions = {
 export default function lint(
   {
     verbose, sourceDir, selfHosted, boring, output,
-    metadata, pretty,
+    metadata, pretty, warningsAsErrors,
   }: LintCmdParams,
   {
     createLinter = defaultLinterCreator,
@@ -65,6 +67,7 @@ export default function lint(
       logLevel: verbose ? 'debug' : 'fatal',
       stack: Boolean(verbose),
       pretty,
+      warningsAsErrors,
       metadata,
       output,
       boring,
--- a/src/program.js
+++ b/src/program.js
@@ -300,6 +300,12 @@ Example: $0 --help run.
         type: 'boolean',
         default: false,
       },
+      'warnings-as-errors': {
+        describe: 'Treat warnings as errors',
+        alias: 'w',
+        type: 'boolean',
+        default: false,
+      },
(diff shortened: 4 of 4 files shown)


josteink:
I originally just tested that it worked as expected against the patched `addon-signer` from the command-line, but I can absolutely respect the desire for automated tests.

I'll be 100% honest and admit that I don't fully understand how those tests are composed and how they work in this project, and I don't have a webpack-decoding capable debugger to dig into it with...

But with that said, I've added some code which (to me) does look reasonable and which "tests green". Feel free to let me know what you think about it. Does it look correct? Does it suffice?

if so, we just need to wait for `addon-signer` to accept my other patch and bump their version and we should be good to go :)

kumar303:
> I'll be 100% honest and admit that I don't fully understand how those tests are composed and how they work in this project

The `npm start` command [as documented](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#develop-all-the-things) will run the entire test suite. Let me know if you run into any trouble with it.

tests/unit/test-cmd/test.lint.js
```diff
@@ -96,13 +104,16 @@ describe('lint', () => {
       boring: 'boring flag',
       // $FLOW_IGNORE: wrong type used for testing purpose
       selfHosted: 'self-hosted flag',
+      // $FLOW_IGNORE: wrong type used for testing purpose
+      warningsAsErrors: 'warnings-as-errors flag',
```
kumar303: you don't need to add this parameter to the test because the "pass through" is already covered by the first test you added. Also, this pass through test is kind of silly.

tests/unit/test-cmd/test.lint.js
```diff
@@ -65,6 +65,14 @@ describe('lint', () => {
     });
   });
 
+  it('passes warningsAsErrors to the linter', () => {
+    const {lint, createLinter} = setUp();
+    return lint({warningsAsErrors: true}).then(() => {
+      const config = createLinter.firstCall.args[0].config;
+      assert.equal(config.warningsAsErrors, true);
+    });
+  });
```
kumar303: this looks great! I think it would make sense to add one more test that does `lint()` without specifying the warnings flag and asserts the linter is called with `warningsAsErrors = undefined` (or should it be false?)

josteink:
Like i tried to say: I've added some tests and they all run fine. I'm just not 100% confident about how all the bits and pieces which the tests are built from fit together, so I cannot know for certain that what I'd like to test gets tested. 

If you could look over my commits and double check that you think it makes sense, that would be greatly appreciated.

josteink:
Ok.  Thanks for the feedback.

I agree the pass-through tests looks kind of silly but I decided to add them just for the sake of consistency. I'll add another test to verify undefined behavior.

Thanks again for your help.

kumar303:
Also, if you merge in the latest master, I fixed the lint error (which is unrelated to your patch).

josteink:
And there everything in `addons-linter` should be cleared, merged and released.

Once the [update addons-linter PR](https://github.com/mozilla/web-ext/pull/621) is merged, everything in this PR should be good to go too.

tests/unit/test-cmd/test.lint.js
kumar303: I think you can remove this entirely. Or is there any reason to include it? Do the tests fail?

tests/unit/test-cmd/test.lint.js
```diff
@@ -96,13 +112,16 @@ describe('lint', () => {
...
+      // $FLOW_IGNORE: wrong type used for testing purpose
+      warningsAsErrors: 'warnings-as-errors flag',
     }).then(() => {
       const config = createLinter.firstCall.args[0].config;
       assert.equal(config.pretty, 'pretty flag');
       assert.equal(config.metadata, 'metadata flag');
       assert.equal(config.output, 'output value');
       assert.equal(config.boring, 'boring flag');
       assert.equal(config.selfHosted, 'self-hosted flag');
+      assert.equal(config.warningsAsErrors, 'warnings-as-errors flag');
```
kumar303: you can remove this too (unless it's needed?)

josteink:
I've now fixed up everything as you've asked me to, and added a `-w` shorthand parameter for the feature as well. Still... We need `addons-linter` fixed before this PR is going anywhere.

And nothing seems to be happening there. Should I (or someone) file a PR to revert the yargs-changes causing this breakage?

kumar303:
> We need addons-linter fixed before this PR is going anywhere.

Yeah, I'll wait on https://github.com/mozilla/addons-linter/issues/1023 before merging

> Should I (or someone) file a PR to revert the yargs-changes causing this breakage?

I haven't had a chance to look at it. @tofumatt have you had time to take a look? 

If you can see anything in the yargs changelog that stands out, let us know. I'm not sure if it's a yargs regression or not. I suspect it has something to do with how `web-ext` uses the linter [programmatically](https://github.com/mozilla/web-ext/blob/master/src/cmd/lint.js#L63) and how both programs use yargs. We fixed one issue related to this already. It was that the linter was trying to parse the command line input a second time.

tofumatt:
I haven't figured it out yet though I've looked a bit. I'll have another
look tonight or this weekend if I can though!

- tofumatt

josteink:
I've just taken a cursory glance at this issue, and it didn't get anywhere.

I do have quite limited time at hand these days, so if you could dig deeper into this, that would really be appreciated.

josteink:
Just tested this patch with latest version of `addons-linter` (0.15.10) and everything now works as it should :)

josteink:
Just tested this patch with latest version of `addons-linter` (0.15.10) and everything now works as it should :)

This PR should now be ready for merge.

kumar303:
Yep, I bumped our linter dep with https://github.com/mozilla/web-ext/pull/632

kumar303:
Thanks for all your work on this feature!

PR #635, opened 2016-11-08
refactor: Make the help text more accurate for --warnings-as-errors

kumar303:


kumar303:
@josteink do you think this is a better help text? I wanted to make it clear that warnings will still be warnings in the output.

josteink:
I think the existing text is fine, but I can see how it may be considered insufficient/confusing.

Adding your suggestion to the existing text would IMO be a better option than removing the existing text all together.

josteink:
Yeah that looks good to me. No objections.

kumar303:
ok, I added the original text. It's not really accurate to say that warnings are errors though because they will be only called "warnings" in the report.

kumar303:
Thanks for the review

PR #651, opened 2016-11-27
feat: Added a command to show the documentation

### LINKED ISSUE #636: Add a command to show the documentation

feature

Just like `jpm docs` (which is ironically undocumented) we should add a command that opens the [web-ext documentation](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext).

When you execute `web-ext docs` it should open the docs [like this](https://github.com/mozilla-jetpack/jpm/blob/master/bin/jpm#L57-L63).

niharikak101:
Fixes #636


kumar303:
Here are some tips for adding a test:

Read through the developer guide for how to run the tests: https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md

Take a look at how the [lint command is tested](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-cmd/test.lint.js). You will probably end up writing tests very similar to this. The lint command is similar because it's simple: it just passes some command line parameters to an external library. We *can trust* that the external `addons-linter` library works so we can replace it with a [sinon.stub()](http://sinonjs.org/docs/#stubs) for testing purposes.

Similarly, we trust that the `open` library works. To pass in a stub object for testing you'll need to adjust the design of the function a little bit. Take a look at how the [lint function is defined](https://github.com/mozilla/web-ext/blob/master/src/cmd/lint.js#L54) and how [it's hooked up to the command callback](https://github.com/mozilla/web-ext/blob/master/src/program.js#L290) for ideas. Let me know if you have any questions.

niharikak101:
@kumar303 I understand! I'll definitely try to understand everything, and implement it as soon as I can. Will let you know if I have any questions. Thank you :)

niharikak101:
@kumar303 Hey! I'm working on the changes now, and was wondering if we could discuss my approach?

kumar303:
Hi @niharikak101 , we can definitely discuss your approach. Did you have any specific questions about my feedback in https://github.com/mozilla/web-ext/pull/651#issuecomment-263708806 ? That explains how to change your approach to make it testable. All patches require tests so that other collaborators don't accidentally break your feature.

niharikak101:
@kumar303 Thank you!

I've managed to define the function, and hooked it up to the command callback (much like you suggested). I've run tests on it, and that seems to be working.
I'm a bit unsure about the test cases: Should we be testing whether the URL is correct?

kumar303:
If you follow the links in https://github.com/mozilla/web-ext/pull/651#issuecomment-263708806 there are examples of doing the same thing elsewhere. You'll need to write some tests that make sure `open()` is called with the right URL. This will ensure that none of the "glue" code breaks in the future. In other words, we just need a test to make sure that `web-ext docs` calls the external `open` function in the correct manner.

src/program.js
```diff
@@ -408,7 +410,8 @@ Example: $0 --help run.
         type: 'boolean',
         default: false,
       },
-    });
+    })
+    .command('docs', 'Open the web-ext documentation', commands.docs, {});
```
kumar303: maybe let's make this more explicit: "Open the web-ext documentation in a browser"

tests/unit/test-cmd/test.docs.js
```diff
@@ -0,0 +1,23 @@
...
+    const docsResult = sinon.stub(openDocs, 'openURL');
+    return docsResult;
+  }
+
+  it('passes the correct URL to docs', () => {
+    const docsResult = setUp();
+    const expectedURL = 'https://developer.mozilla.org/en-US/Add-ons' +
+       '/WebExtensions/Getting_started_with_web-ext';
+    docs();
+    assert.equal(docsResult.calledWith((expectedURL)), true);
```
kumar303: I prefer to see an explicit check of the arguments because when it fails it tends to be a little bit easier to diagnose:
````js
assert.ok(docsResult.called);
assert.equal(docsResult.firstCall.args[0], expectedUrl);
````

File: tests/unit/test-cmd/test.docs.js [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+describe('docs', () => {
+
+  function setUp() {
+    const docsResult = sinon.stub(openDocs, 'openURL');
+    return docsResult;
+  }
+
+  it('passes the correct URL to docs', () => {
+    const docsResult = setUp();
+    const expectedURL = 'https://developer.mozilla.org/en-US/Add-ons' +
```
kumar303: nitpicky comment: for consistency with the rest of web-ext, make the camel case for this look like `expectedUrl`.

tests/unit/test-cmd/test.docs.js
```diff
@@ -0,0 +1,28 @@
...
+    assert.ok(openUrl.called);
+    assert.equal(openUrl.firstCall.args[0], expectedUrl);
+    });
+  });
+
+  it('throws an error when open fails', () => {
+    const openUrl = sinon.spy(
+      (callback) => callback(new Error('pretends this is an error from open()'))
+    );
+      return defaultDocsCommand({}, {openUrl}).catch((error) => {
```
kumar303: This looks correct but in case `catch()` is never called you'll want this test to fail. We do that like this, for example:
````js
import {makeSureItFails} from '../helpers';

return defaultDocsCommand(...)
  .then(makeSureItFails())
  .catch(...);
````

File: src/cmd/docs.js [hunk 5]
```diff
@@ -0,0 +1,27 @@
...
+const log = createLogger(__filename);
+
+type DocsOptions = {
+	openUrl?: typeof defaultUrlOpener,
+}
+
+export default function docs(
+  params: Object, {openUrl = defaultUrlOpener}: DocsOptions = {}
+	) {
+  const url = 'https://developer.mozilla.org/en-US/Add-ons' +
```
kumar303: If you move this to the top as:
````js
export const url = '...';
````
then you could import it in your test like:
````js
import defaultDocsCommand, {url} from '../../../src/cmd/docs';
````
and then you wouldn't have to copy/paste it to make the mock assertion.

niharikak101:
Hi @kumar303, can I get a review on this?

caitmuenster:
Thank you so much, @niharikak101! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#March_2017). 

I'd love to help you set up your profile on [mozillians.org](https://mozillians.org/en-US/)! Could you send me your email address (either here or to cneiman [at] mozilla [dot] com)? I'll send you an invite and vouch for your contribution to the project. :) 

Welcome onboard!

Issue #88, opened 2016-03-01
Install/run a web extension using custom preferences


kumar303:
A run/install command was added in https://github.com/mozilla/web-ext/issues/6 but it only launches Firefox with default preferences. Allow the caller to specify custom preferences. This would be a more lightweight way of customizing Firefox as opposed to creating a custom profile (https://github.com/mozilla/web-ext/issues/69).

kumar303:
I think this should be a pretty easy change actually, so I'm putting back the good first bug label. It only needs:
* A new repeatable option like `--pref="key:value" --pref="key2:value2" ...`
* An update to [firefox.configureProfile()](https://github.com/mozilla/web-ext/blob/master/src/firefox/index.js#L93) to add each preference
* Tests
I am happy to mentor this.

kumar303:
Here is a good use case: Let's say you want to test your extension in 30 different locales. Instead of making 30 profiles, you could just run it like `web-ext --pref=general.useragent.locale:fr-FR` for each locale.

wagnerand:
Please refer to https://wiki.mozilla.org/Add-ons/Contribute/AMO/Code if this is your first contribution.

If you want to work on this issue and need more information on how or where to fix it, please comment here and we will be happy to provide more details.

Mentor: @kumar303

davidamidon:
Hello, I would like to try my hand at this fix. Although I will certainly require direction, I am very eager to participate.

kumar303:
Hi @davidamidon ! I'm here to help, let me know if you have any questions about how to get started.

davidamidon:
@kumar303 Thank you! Should I ask in here or is there a better place? I've got everything installed I think. For this fix the code would go in /src/firefox/index.js?

kumar303:
Yes, this is a good place to ask questions. The locations are as follows:

* Custom preferences are applied to profiles [here](https://github.com/mozilla/web-ext/blob/0d1b82454ec5d3a6a58ddeb3e4c66432e6fbad77/src/firefox/index.js#L216).
* New command line options are defined [here](https://github.com/mozilla/web-ext/blob/0d1b82454ec5d3a6a58ddeb3e4c66432e6fbad77/src/program.js#L223).
* The configuration of profiles is [tested here](https://github.com/mozilla/web-ext/blob/0d1b82454ec5d3a6a58ddeb3e4c66432e6fbad77/tests/unit/test-firefox/test.firefox.js#L289).

davidamidon:
[...]

PR #658, opened 2016-12-05
feat: install and run an extension with custom Firefox preferences  

saintsebastian:
Current commit without repeated option and with preferences assignment commented. Since there are so many changes for this one, I'd like to know if it is the correct approach so far.

Fixes https://github.com/mozilla/web-ext/issues/88

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -283,7 +283,7 @@ export class ExtensionRunner {
   constructor(
     {
       firefoxApp, sourceDir, manifestData,
-      profilePath, firefox, customPrefs,
+      profilePath, firefox, customPrefs = {},
     }: ExtensionRunnerParams
   ) {
     this.sourceDir = sourceDir;
@@ -291,9 +291,7 @@ export class ExtensionRunner {
     this.profilePath = profilePath;
     this.firefoxApp = firefoxApp;
     this.firefox = firefox;
-    if (customPrefs) {
-      this.customPrefs = customPrefs;
-    }
+    this.customPrefs = customPrefs;
   }
 
   getProfile(): Promise<FirefoxProfile> {
@@ -301,7 +299,7 @@ export class ExtensionRunner {
     return new Promise((resolve) => {
       if (profilePath) {
         log.debug(`Copying Firefox profile from ${profilePath}`);
-        resolve(firefoxApp.copyProfile(profilePath, customPrefs));
+        resolve(firefoxApp.copyProfile(profilePath, {customPrefs}));
       } else {
         log.debug('Creating new Firefox profile');
         resolve(firefoxApp.createProfile({customPrefs}));
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -16,6 +16,7 @@ const log = createLogger(__filename);
 
 // Import objects that are only used as Flow types.
 import type FirefoxProfile from 'firefox-profile';
+import type {FirefoxPreferences} from '../firefox/preferences';
 import type {OnSourceChangeFn} from '../watcher';
 import type Watchpack from 'watchpack';
 import type {
@@ -151,7 +152,7 @@ export type CmdRunParams = {
   firefoxProfile: string,
   preInstall: boolean,
   noReload: boolean,
-  pref: {},
+  customPrefs?: FirefoxPreferences,
 };
 
 export type CmdRunOptions = {
@@ -163,7 +164,8 @@ export type CmdRunOptions = {
 export default async function run(
   {
     sourceDir, artifactsDir, firefox, firefoxProfile,
-    preInstall = false, noReload = false, pref = {},
+    preInstall = false, noReload = false,
+    customPrefs,
   }: CmdRunParams,
   {
     firefoxApp = defaultFirefoxApp,
@@ -189,7 +191,6 @@ export default async function run(
   let addonId;
 
   let manifestData = await getValidatedManifest(sourceDir);
-  let custom = pref;//TODO is not such a good idea maybe
 
   runner = new ExtensionRunner({
     sourceDir,
@@ -197,7 +198,7 @@ export default async function run(
     firefox,
     manifestData,
     profilePath: firefoxProfile,
-    custom,
+    customPrefs,
   });
 
   profile = await runner.getProfile();
@@ -256,7 +257,6 @@ export default async function run(
     }
   }
 
-  log.info(pref.toString()); //TODO don't need
   return firefoxApp;
 }
 
@@ -269,7 +269,7 @@ export type ExtensionRunnerParams = {
   profilePath: string,
   firefoxApp: typeof defaultFirefoxApp,
   firefox: string,
-  custom : {},
+  customPrefs?: FirefoxPreferences,
 };
 
 export class ExtensionRunner {
@@ -278,12 +278,12 @@ export class ExtensionRunner {
   profilePath: string;
   firefoxApp: typeof defaultFirefoxApp;
   firefox: string;
-  custom: {};
+  customPrefs: FirefoxPreferences;
 
   constructor(
     {
       firefoxApp, sourceDir, manifestData,
-      profilePath, firefox, custom,
+      profilePath, firefox, customPrefs,
     }: ExtensionRunnerParams
   ) {
     this.sourceDir = sourceDir;
@@ -291,18 +291,20 @@ export class ExtensionRunner {
     this.profilePath = profilePath;
     this.firefoxApp = firefoxApp;
     this.firefox = firefox;
-    this.custom = custom;
+    if (customPrefs) {
+      this.customPrefs = customPrefs;
+    }
   }
 
   getProfile(): Promise<FirefoxProfile> {
-    const {firefoxApp, profilePath, custom} = this;
+    const {firefoxApp, profilePath, customPrefs} = this;
     return new Promise((resolve) => {
       if (profilePath) {
         log.debug(`Copying Firefox profile from ${profilePath}`);
-        resolve(firefoxApp.copyProfile(profilePath, custom));
+        resolve(firefoxApp.copyProfile(profilePath, customPrefs));
       } else {
         log.debug('Creating new Firefox profile');
-        resolve(firefoxApp.createProfile(custom));
+        resolve(firefoxApp.createProfile({customPrefs}));
       }
     });
   }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -151,6 +151,7 @@ export type CmdRunParams = {
   firefoxProfile: string,
   preInstall: boolean,
   noReload: boolean,
+  pref: {},
 };
 
 export type CmdRunOptions = {
@@ -162,7 +163,7 @@ export type CmdRunOptions = {
 export default async function run(
   {
     sourceDir, artifactsDir, firefox, firefoxProfile,
-    preInstall = false, noReload = false,
+    preInstall = false, noReload = false, pref = {},
   }: CmdRunParams,
   {
     firefoxApp = defaultFirefoxApp,
@@ -188,6 +189,7 @@ export default async function run(
   let addonId;
 
   let manifestData = await getValidatedManifest(sourceDir);
+  let custom = pref;//TODO is not such a good idea maybe
 
   runner = new ExtensionRunner({
     sourceDir,
@@ -195,6 +197,7 @@ export default async function run(
     firefox,
     manifestData,
     profilePath: firefoxProfile,
+    custom,
   });
 
   profile = await runner.getProfile();
@@ -253,6 +256,7 @@ export default async function run(
     }
   }
 
+  log.info(pref.toString()); //TODO don't need
   return firefoxApp;
 }
 
@@ -265,6 +269,7 @@ export type ExtensionRunnerParams = {
   profilePath: string,
   firefoxApp: typeof defaultFirefoxApp,
   firefox: string,
+  custom : {},
 };
 
 export class ExtensionRunner {
@@ -273,11 +278,12 @@ export class ExtensionRunner {
   profilePath: string;
   firefoxApp: typeof defaultFirefoxApp;
   firefox: string;
+  custom: {};
 
   constructor(
     {
       firefoxApp, sourceDir, manifestData,
-      profilePath, firefox,
+      profilePath, firefox, custom,
     }: ExtensionRunnerParams
   ) {
     this.sourceDir = sourceDir;
@@ -285,17 +291,18 @@ export class ExtensionRunner {
     this.profilePath = profilePath;
     this.firefoxApp = firefoxApp;
     this.firefox = firefox;
+    this.custom = custom;
   }
 
   getProfile(): Promise<FirefoxProfile> {
-    const {firefoxApp, profilePath} = this;
+    const {firefoxApp, profilePath, custom} = this;
     return new Promise((resolve) => {
       if (profilePath) {
         log.debug(`Copying Firefox profile from ${profilePath}`);
-        resolve(firefoxApp.copyProfile(profilePath));
+        resolve(firefoxApp.copyProfile(profilePath, custom));
       } else {
         log.debug('Creating new Firefox profile');
-        resolve(firefoxApp.createProfile());
+        resolve(firefoxApp.createProfile(custom));
       }
     });
   }
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -219,10 +219,13 @@ export function configureProfile(
   Object.keys(prefs).forEach((pref) => {
     profile.setPreference(pref, prefs[pref]);
   });
-  Object.keys(customPrefs).forEach((custom) => {
-    profile.setPreference(custom, customPrefs[custom]);
-  });
-  log.info(customPrefs ? JSON.stringify(customPrefs) : 'EMPTY');
+  if (Object.keys(customPrefs).length > 0) {
+    log.info(`Setting custom Firefox preferences:
+      ${JSON.stringify(customPrefs, null, ' ')}`);
+    Object.keys(customPrefs).forEach((custom) => {
+      profile.setPreference(custom, customPrefs[custom]);
+    });
+  }
   profile.updatePreferences();
   return Promise.resolve(profile);
 }
@@ -245,7 +248,7 @@ export async function createProfile(
   {
     app,
     configureThisProfile = configureProfile,
-    customPrefs,
+    customPrefs = {},
   }: CreateProfileParams = {},
 ): Promise<FirefoxProfile> {
   const profile = new FirefoxProfile();
@@ -257,8 +260,9 @@ export async function createProfile(
 
 export type CopyProfileOptions = {
   app?: PreferencesAppName,
-  copyFromUserProfile?: Function,
   configureThisProfile?: ConfigureProfileFn,
+  copyFromUserProfile?: Function,
+  customPrefs?: FirefoxPreferences,
 };
 
 /*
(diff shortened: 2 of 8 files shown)


saintsebastian:
For some reason even though I've worked from a branch some of my commits are added here as well.

kumar303:
Hi @saintsebastian ! Thanks for the patch. @rpl could you take a look?

src/program.js
```diff
@@ -251,6 +251,12 @@ Example: $0 --help run.
         demand: false,
         type: 'boolean',
       },
+      'pref': {
+        describe: 'Launch firefox with custom preferences. Lightweight ' +
+                  'alternative to creating custom profile.',
+        demand: false,
+        type: 'object',
```
saintsebastian: After reading doc for some time I understand that it would look like `"pref.dotted.name=value"` and years will automatically turn this into `pref = {dotted: {name:value}}`, hence repeated option like `"pref.name=value pref.anotherName=value"` will just be a nested object handled my yargs. That's really cool! 
However I'm still not sure about when the type detection is happening. Numbers should be detected automatically as the link you've shared suggests, as I understand coercing has those capabilities and we don't need to know in advance which properties are used http://yargs.js.org/docs/#methods-coercekey-fn. I imagine it looks g something like this
 ```
yargs.coerce('pref', function (obj) {
  for (var in obj) {
    if (!isNaN(Number(obj.var))) {
      obj.var == Number(obj.var);
    } else if (obj.var === 'true' || obj.var === 'false) {
       obj.var == Boolean(obj.var);
    }
  }
})
```
Let me know if that is what you had in mind

File: src/firefox/index.js [hunk 2]
```diff
@@ -193,6 +193,7 @@ export type ConfigureProfileOptions = {
 
 export type ConfigureProfileFn = (
   profile: FirefoxProfile,
+  customs: {},
```
saintsebastian: So this is the last step in `configureProfile`, which I haven't properly tested yet.
```
 Object.keys(customPrefs).forEach((custom) => {
  profile.setPreference(custom, customPrefs[custom]);
});
```
Is using a defined flow type and yargs parsing of values suffice for control over what properties user assigns? At the moment command happily takes any property name (and number of them)

src/cmd/build.js
```diff
@@ -110,7 +111,9 @@ async function defaultPackageCreator(
 
   await streamToPromise(stream);
 
-  log.info(`Your web extension is ready: ${extensionPath}`);
+  if (showReadyMessage) {
```
rpl: Hehe, no worries, managing git branches can be tricky sometimes and git commands/workflows are not always very intuitive :-)
(and after a bit of practice everything looks simpler than it really is ;-))

Sometimes it is simpler to rename the old branch (`git branch -m OLDNAME NEWNAME`), switch to a new pristine branch, named like the one pushed on github (`git checkout master && git checkout -b OLDNAME`), and start again (and maybe cherry-picking what is still useful from the previous branch)

But in my opinion, there is a lot of value also into learning how to rewrite the existent history in the branch that we are working on (so that we feel confortable that we can manage every situation that we can find ourselves while we are working on a git branch), and so I'd like to give you more information of the topic :-D

There are many ways to get rid of commits from a branch in git, my favorite one is the interactive rebase, e.g. we can ask git to let us rebase interactively the last 4 commits using the command:

```
$ git rebase -i HEAD~4
```

and it will open our configured editor (the one in the EDITOR environment variable or the standard one of our system) with the
list of the commits that we are going to rebase:

```
pick 062d62a fix: Don't log info about the built extension when signing
pick 10f8b7b fix: the code consistency for issue 181
pick 7edc47f fix: another code inconsistency for issue 181
pick 77626b8 fix: issue 88 first commit
```

In the opened editor, if you remove the first three lines, save and exit, git will remove the correspondent commits from the branch you are working on.

The following are a couple of additional resources on the topic that you may find interesting/useful:

- http://learngitbranching.js.org/ has a nice interactive tutorial on both "reverting single commit" and "interactive rebasing", with nice diagrams to help you visualizing what is happening in the git repo
- https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase has a short description and a diagram
- https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History has more detailed informations on how the git interactive rebase works

saintsebastian: Thanks! That was surprisingly easy to do! I'll definitely read on that to avoid mishaps

src/firefox/index.js
rpl: it is hard to be completely sure that a custom property will be completely valid/recognized by Firefox (e.g. if you open the "about:config" url in a Firefox tab you can briefly take a look at the one defined in your Firefox profile), what we can check is probably that:

- preference name:
  - it should have at least one "."
  - it should always start with a lower case letter
(actually in the preferences names used in Firefox everything on the left of the first `"."` is always composed only by lowercase letters)
  - it should contain only valid characters (e.g. lowercase and uppercase letters, numbers, and a couple of special characters like `.` , `-`, `_` and `*`)
- preference value: 
  - it can only be integers, booleans or strings (everything that is not a boolean or an integer)

Most of the above validations on the preference names seems to be doable with regular expressions (https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions) that are still readable from us human beings :-) 

(we can write some really amazing/crazy things with regex syntaxes but, at least in my own opinion, if it is possible, it is better to use an amount of them that make the final expression still reasonably readable)

On the bright side, if someone really needs this feature, he usually knows pretty well what he is doing, and so the above validations should be more then enough (maybe some of them are even too much).

About the "where it would be better to validate the custom preferences", in general I think that is always better to validate the external inputs only on the "border" of our software systems (which means as near as possible to where the external input has been received) so that we are not going to check them everywhere.

In this case a good position to validate the custom preferences is probably inside the `execute` method of the `Program` class :
https://github.com/mozilla/web-ext/blob/master/src/program.js#L83

(it could be probably nicer if we define and export the helper function that will validate these `customPrefs` object in the `src/firefox/preferences.js` module and then import and call it from the `execute` method in the `src/program.js` module)

src/program.js
```diff
@@ -251,6 +251,12 @@ Example: $0 --help run.
         demand: false,
         type: 'boolean',
       },
+      'pref': {
+        describe: 'Launch firefox with custom preferences. Lightweight ' +
+                  'alternative to creating custom profile.',
+        demand: false,
```
rpl: we should also add `requiresArg: true` here 
(http://yargs.js.org/docs/#requiresArg).

src/program.js
rpl: I took a brief look at this alternative `--pref.key.subkey=value` syntax, but I'm unsure about it, in particular if it makes this feature more easily implementable and/or if it makes this command line option more or less readable for the user.

The usage on the command line would make the command to looks like:

```
$ web-ext run --pref.browser.search.region=IT \           
              --pref.general.useragent.locale=it-IT
```

while I was thinking about something like the following:

```
$ web-ext run --pref "browser.search.region=IT" \           
              --pref "general.useragent.locale=it-IT"
```

@kumar303 what do you think about the above two command line syntaxes for this `--pref` option?

rpl: But besides the command line syntax, using the `coerce` method seems like a good idea to me (and it is also in a better position than what I suggested in one of my comment above, it is as near as possible to where the external input enters our system), the code in `src/program.js` will become something like:

```
'pref': {
  describe: 'Launch firefox with custom preferences...',
  demand: false,
  requiresArg: true,
  type: 'string',
}
```

And in the helper function we can incapsulate all the details related to validate and parse the command line option, e.g. if we opt for the syntax that I was suggesting above:

- if value received does not represent a valid preference, we can `throw new UsageError("Invalid Preference...");` and web-ext will not run the command and it will show the usage help and the error message raised from our helper
  
- we can split key and value and return an object, e.g.
``` 
  let [key, value] = value.split("=");

  // ... detect and convert integer and boolean values, 
  // and fallback to a string value 

  // The following is a weird but useful ES6 syntax for object
  // property names which are computed:
  return {[`${key}`]: value};
```

For the coercion of the value, what I'm thinking of is not too far from the above (there are some typos in the snippet that you pasted above ;-) but no worries, we will get into the details once we are reviewing the actual commit, and typos are pretty common when sketching example code in a comment), e.g. to detect and convert an integer, I was thinking of something like:

```
if ( value === `${parseInt(value)}`) {
  value = parseInt(value, 10);
}
```

Because we want to be sure that we only convert the value to a number if it is an integer (and `parseInt("123.123", 10)` returns `123` and ignore the rest, and `parseInt` will try to detect the radix if it is not specified explicitly, you can find more details about the `parseInt` behavior on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt).

saintsebastian:
I am not entirely happy with everything. Here are my concerns
* insufficient testing, currently the feature is only tested within `configureProfile` testing suit, please let me know if I can and should improve that
* Every time I tried to add `customPrefs?: FirefoxPreferences` to `CopyProfileOptions` the entire type would break and Flow would refer to following as a source. For this reason the typing is inconsistent (`ConfigureProfileOptions`  and `CreateProfileParams` already include it) and I would live to try to fix it
```
export type FirefoxPreferences = {
  [key: string]: bool | string | number,
};
```

Looking forward for any input and suggestions!

src/firefox/index.js
```diff
@@ -235,10 +241,14 @@ export type CreateProfileParams = {
  * The profile will be deleted when the system process exits.
  */
 export async function createProfile(
-  {app, configureThisProfile = configureProfile}: CreateProfileParams = {}
+  {
+    app,
+    configureThisProfile = configureProfile,
+    customPrefs,
```
rpl: it looks like this should be `customPrefs = {}` like in `configureProfile`.

src/firefox/index.js
```diff
@@ -264,11 +274,12 @@ export type CopyProfileOptions = {
  */
 export async function copyProfile(
   profileDirectory: string,
+  customPrefs?: FirefoxPreferences,
```
saintsebastian: It makes perfect sense now. The flow error really confused me, even though that was a simple bug.

saintsebastian: Thanks for resolving this!

src/cmd/run.js
```diff
@@ -268,29 +273,33 @@ export class ExtensionRunner {
...
-      profilePath, firefox,
+      profilePath, firefox, customPrefs,
     }: ExtensionRunnerParams
   ) {
     this.sourceDir = sourceDir;
     this.manifestData = manifestData;
     this.profilePath = profilePath;
     this.firefoxApp = firefoxApp;
     this.firefox = firefox;
+    if (customPrefs) {
```
saintsebastian: Originally I've put the if statement there because flow doesn't like undefined assignment, now i've added default `customPrefs = {}` value instead.

src/firefox/index.js
```diff
@@ -264,11 +274,12 @@ export type CopyProfileOptions = {
...
   {
+    app,
```
saintsebastian: Sure! I sorted them in function definition as well for consistency, will apply the same order in tests.

saintsebastian:
The biggest thing left to is tests for changes I've made to` program.execute())` (I am a bit lost on what's the best place and approach to do it)  and clean up in some tests I've missed. Let me know if you find anything else

src/firefox/preferences.js
```diff
@@ -117,3 +117,33 @@ export function getPrefs(
     ...appPrefs,
   };
 }
+
+// Yargs coerce function for run command custom-prefs
+export function coerceCLICustomPreference(
```
rpl: To make flow happy about the type checking of this function, it needs a small rewrite:
```
export function coerceCLICustomPreference(
  cliPrefs: string | Array<string>
): FirefoxPreferences {
  let customPrefs = {};

  if (!Array.isArray(cliPrefs)) {
    cliPrefs = [cliPrefs];
  }

  for (let pref of cliPrefs) {
    let [key, value] = pref.split('=');
    // Check the property key.
    if (/[^\w{@}.-]/.test(key)) {
    }

    if (value === `${parseInt(value)}`) {
      value = parseInt(value, 10);
    } else if (value === 'true' || value === 'false') {
      value = (value === 'true');
    }

    customPrefs[`${key}`] = value;
  }

  return customPrefs;
}
```

The function esplicitly return a `FirefoxPreferences` type, and flow is able using its type inference capability that the `for...of` loop produce an object and its properties have the expected value types (boolean, number or string), which is the definition of the type FirefoxPreferences.

src/firefox/index.js
```diff
@@ -208,16 +210,22 @@ export function configureProfile(
...
   // operate, such as disabling signatures.
-  // TODO: support custom preferences.
-  // https://github.com/mozilla/web-ext/issues/88
   let prefs = getPrefs(app);
   Object.keys(prefs).forEach((pref) => {
     profile.setPreference(pref, prefs[pref]);
   });
+  if (Object.keys(customPrefs).length > 0) {
+    log.info(`Setting custom Firefox preferences:
+      ${JSON.stringify(customPrefs, null, ' ')}`);
```
rpl: The JSON.stringify "pretty print" mode is usually achieved with: `JSON.stringify(customPrefs, null, 2)`

Also, how about reformatting the above two lines into:
```
const customPrefsStr = JSON.stringify(...);
log.info(`Setting custom Firefox preferences: ${customPrefsStr}`);
```

saintsebastian:
Sorry for that mess leftover in tests from a time when customPrefs were not an optional parameter.

src/firefox/preferences.js
```diff
@@ -117,3 +119,38 @@ export function getPrefs(
...
+      throw new UsageError(`Invalid custom preference name: ${key}`);
+    }
+
+    if (value === `${parseInt(value)}`) {
+      value = parseInt(value, 10);
+    } else if (value === 'true' || value === 'false') {
+      value = (value === 'true');
+    }
+
+    let defaultProps = ['devtools.debugger.remote-enabled',
```
rpl: This can be defined using `const`, I would also rename it to make it immediately clear its role (e.g. something around "non overridable pref names" or "blacklisted custom pref names") and move out of this function, in the first part of this module (as an exported module const, which would also be helpful for the related tests).

Also, this array should be formatted like in:

```
const varName = [
  ..., // max 80 cols
  ..., // go to the new line if the line above is already too long 
  ..., // and add a "dangle comma" at the end (which prevents syntax errors if/when 
  ..., // a new element is appended).  
];
```

src/firefox/preferences.js
```diff
@@ -117,3 +119,38 @@ export function getPrefs(
...
+      'devtools.debugger.prompt-connection', 'xpinstall.signatures.required'];
+    if (defaultProps.includes(key)) {
+      log.info(`Setting '${key}' is important for the work of WebExtensions` +
+               ' and can not be customized');
```
rpl: we can use `continue;` after the `log.warn` above so that we don't need an `else` block.

src/firefox/preferences.js
```diff
@@ -117,3 +119,38 @@ export function getPrefs(
...
+    if (value === `${parseInt(value)}`) {
+      value = parseInt(value, 10);
+    } else if (value === 'true' || value === 'false') {
+      value = (value === 'true');
+    }
+
+    let defaultProps = ['devtools.debugger.remote-enabled',
+      'devtools.debugger.prompt-connection', 'xpinstall.signatures.required'];
+    if (defaultProps.includes(key)) {
+      log.info(`Setting '${key}' is important for the work of WebExtensions` +
```
rpl: `log.warn` is probably better here.
I would also use a shorter message, something like `'${key}' preference is used internally and it cannot be customized.`.

src/program.js
```diff
@@ -93,6 +94,8 @@ export class Program {
 
     const argv = this.yargs.argv;
     const cmd = argv._[0];
+    //Command line option (pref) renamed for internal use (customPref)
```
rpl: nit, add an empty line between line 96 and this comment, and add a space between // and Command (and a period, `.`, at the end of the comment).

tests/unit/test-firefox/test.firefox.js
```diff
@@ -345,6 +345,22 @@ describe('firefox', () => {
...
+        const customPrefs = {'extensions.checkCompatibility.nightly': true};
+        return firefox.configureProfile(profile, {customPrefs})
+          .then((profile) => fs.readFile(path.join(profile.path(), 'user.js')))
+          .then((prefFile) => {
+            // Check for custom pref set by configureProfile().
+            assert.include(prefFile.toString(),
+                           '"extensions.checkCompatibility.nightly", true');
+            // Check that one of the default preferences is set as well
+            assert.include(prefFile.toString(),
+                           '"devtools.debugger.remote-enabled", true');
```
rpl: :+1: perfect!

tests/unit/test-firefox/test.preferences.js
```diff
@@ -34,4 +36,49 @@ describe('firefox/preferences', () => {
...
+      assert.equal(typeof prefs['valid.preference'], 'number');
+    });
+
+    it('converts float values', () => {
+      const prefs = coerceCLICustomPreference('valid.preference=4.55');
+      assert.equal(prefs['valid.preference'], '4.55');
+    });
+
+    it('does not allow certain default preferences to be customized', () => {
+      const prefs = coerceCLICustomPreference('xpinstall.signatures.required');
```
rpl: how about exporting the array of the `non overridable` preferences from the preferences module, so that we can test that none of that array of preferences can be overridden?

tests/unit/test.program.js
```diff
@@ -327,6 +327,21 @@ describe('program.main', () => {
...
+      run: () => Promise.resolve(),
+    });
+    return execProgram(
+      ['run', '--pref', 'prop=true', '--pref', 'prop2=value2'],
+      {commands: fakeCommands})
+      .then(() => {
+        const {customPrefs} = fakeCommands.run.firstCall.args[0];
+        assert.isObject(customPrefs);
+        assert.equal(customPrefs.prop, true);
+        assert.equal(customPrefs.prop2, 'value2');
```
rpl: :+1:

tests/unit/test.program.js
```diff
@@ -327,6 +327,20 @@ describe('program.main', () => {
       });
   });
 
+  it('converts custom prefernces into object', () => {
```
rpl: This test description still contains a typo ;-)

tests/unit/test-firefox/test.preferences.js
```diff
@@ -34,4 +36,54 @@ describe('firefox/preferences', () => {
...
+      assert.equal(prefs['valid.preference'], '4.55');
+    });
+
+    it('does not allow certain default preferences to be customized', () => {
+      const nonChangeablePrefs = nonOverridablePreferences.map((prop) => {
+        return prop += '=true';
+      });
+      const prefs = coerceCLICustomPreference(nonChangeablePrefs);
+      for (var pref of nonChangeablePrefs) {
+        assert.equal(typeof(prefs[pref]), 'undefined');
```
rpl: We should change this assertion so that it use the specific chai assertion and it prints an helpful message if it fails:
```
assert.isUndefined(prefs[pref], `${pref} should be undefined`);
```

tests/unit/test-firefox/test.preferences.js
```diff
@@ -34,4 +36,54 @@ describe('firefox/preferences', () => {
...
+      const prefs = coerceCLICustomPreference('valid.preference=4.55');
+      assert.equal(prefs['valid.preference'], '4.55');
+    });
+
+    it('does not allow certain default preferences to be customized', () => {
+      const nonChangeablePrefs = nonOverridablePreferences.map((prop) => {
+        return prop += '=true';
+      });
+      const prefs = coerceCLICustomPreference(nonChangeablePrefs);
+      for (var pref of nonChangeablePrefs) {
```
rpl: we should use `let` here instead of `var`:
```
for (let pref of ...) {

}
```

tests/unit/test-firefox/test.preferences.js
```diff
@@ -34,4 +36,54 @@ describe('firefox/preferences', () => {
...
+    });
+
+    it('converts boolean values', () => {
+      const prefs = coerceCLICustomPreference('valid.preference=true');
+      assert.equal(prefs['valid.preference'], true);
+    });
+
+    it('converts number values', () => {
+      const prefs = coerceCLICustomPreference('valid.preference=455');
+      assert.equal(typeof prefs['valid.preference'], 'number');
```
rpl: `assert.equal(prefs['valid.preference'], 455)`

kumar303:
Thanks Elvina! I'm excited for this feature.

highfive

Issue #640, opened 2016-11-18
Add lint rule to prefer constants over variables


kumar303:
### Is this a bug or feature request?

feature

### What is the current behavior?

When contributors submit pull requests, @kumar303 always tells them to use `const` instead of `let` when they are not modifying the variable.

### What is the expected or desired behavior?

We should let the robots tell them this instead! Let's add this rule and fix up the code:
````diff
diff --git a/.eslintrc b/.eslintrc
index 24680cd..66c4e3c 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -100,6 +100,7 @@
       "allowMultiplePropertiesPerLine": true
     }],
     "one-var": [2, "never"],
+    "prefer-const": 2,
     "prefer-template": 2,
     "quote-props": [1, "consistent-as-needed"],
     "quotes": [2, "single", "avoid-escape"],````

pragyes31:
Hi,

I will work on this issue.

Thanks,
Rahul

kumar303:
great, thanks! Let me know if you run into any trouble before submitting a pull request.

shubheksha:
@kumar303, wouldn't this require changing the variables which aren't being reassigned anywhere to const? Eslint will keep complaining otherwise as is evident in #661

kumar303:
@shubheksha yes, exactly. @aniketkudale, sorry for the late reply. I just now had a chance to look at your patch and I pointed out some additional changes needed.

kewisch:
@pragyes31 are you still working on this bug? I'd love to hear back from you!

pragyes31:
@kewisch I am. Will submit the PR probably by this weekend.

PR #671, opened 2016-12-13
chore: Added prefer-const rule to eslint and corresponding changes

pragyes31:
Added prefer-const rule to eslint and corresponding changes.

Fixes https://github.com/mozilla/web-ext/issues/640

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -84,14 +84,16 @@ export function defaultReloadStrategy(
     createWatcher = defaultWatcherCreator,
   }: ReloadStrategyOptions = {}
 ): void {
-  let watcher: Watchpack;
+  const watcher: Watchpack = (
+  createWatcher({addonId, client, sourceDir, artifactsDir})
+);
 
   firefoxProcess.on('close', () => {
     client.disconnect();
     watcher.close();
   });
 
-  watcher = createWatcher({addonId, client, sourceDir, artifactsDir});
+
 }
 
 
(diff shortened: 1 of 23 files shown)


src/cmd/run.js
```diff
@@ -181,23 +183,19 @@ export default async function run(
...
-  let runner;
-  let profile;
-  let client;
-  let runningFirefox;
-  let addonId;
-
-  let manifestData = await getValidatedManifest(sourceDir);
-
-  runner = new ExtensionRunner({
+  const runner = new ExtensionRunner({
```
rpl: Hi @pragyes31 I took a brief look at this and it looks like the changes on `src/cmd/run.js` have a number of issues, even besides the issues that flow has detected on Travis (https://travis-ci.org/mozilla/web-ext/jobs/183711790#L692).

I would suggest to split this change in 2 or 3 commits to make the review easier (and, as a side effect, it would make easier also a rebase of the branch if needed):

- 1 commit related to the change to the eslintrc file
- 1 commit with the fixes that eslint is able to apply by running it on the project with the `--fix` command line option
- 1 additional commit related to the linting errors that cannot be fixed by eslint itself and have to be fixed manually (because they are probably the one that needs a more detailed review)

kumar303:
Thanks for taking the time to start a patch for this. Let us know if we can assist in any further development you'd like to do. Since we're pretty far along in the review of another patch for the same issue (https://github.com/mozilla/web-ext/pull/700), I'm closing this one.

PR #674, opened 2016-12-14
feat: show desktop notifications on addon reload

### LINKED ISSUE #511: Show a reloading error notification in the browser
User story: "As a developer who is using [web-ext run](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run), I want to see a [notification](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/notifications) in my browser when there is a reloading error. A reloading error, such as a syntax error in my manifest, will cause the reload to fail which means I may be staring at old, unreloaded code without knowing it. This would be really confusing."

**Let's show a browser notification of reload errors in addition to console errors.**

When a reload error happens, you can see the error in the console (where web-ext was started from) but this could be hidden if the developer is only toggling between their editor and the web browser while developing their extension. 

One way to implement this is to install a helper web extension that sits on the browser and listens for web-ext errors. It could use the [notification](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/notifications) API to show errors.

saintsebastian:
I've installed a package which seemed well thought-through and maintained. However, I am not sure about how effective catching the error is currently. 

fixes #511

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "minimatch": "3.0.3",
     "mz": "2.6.0",
     "node-firefox-connect": "1.2.0",
+    "node-notifier": "4.6.1",
     "regenerator-runtime": "0.10.1",
     "parse-json": "2.2.0",
     "sign-addon": "0.2.0",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "mz": "2.6.0",
     "node-firefox-connect": "1.2.0",
     "node-notifier": "4.6.1",
-    "regenerator-runtime": "0.10.1",
     "parse-json": "2.2.0",
     "regenerator-runtime": "0.10.1",
     "sign-addon": "0.2.0",
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -54,12 +54,6 @@ export function defaultWatcherCreator(
       log.debug(`Reloading add-on ID ${addonId}`);
 
       return client.reloadAddon(addonId)
-        .then(() => {
-          desktopNotifications({
-            titleString: 'web-ext run: reload',
-            messageString: `Reloaded add-on ID ${addonId}`,
-          });
-        })
         .catch((error) => {
           log.error(error.stack);
           desktopNotifications({
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -2,6 +2,7 @@
 import type FirefoxProfile from 'firefox-profile';
 import type Watchpack from 'watchpack';
 
+import {desktopNotifications as notifier} from '../util/desktop-notifier';
 import * as defaultFirefoxApp from '../firefox';
 import defaultFirefoxConnector from '../firefox/remote';
 import {
@@ -12,7 +13,6 @@ import {
 import {createLogger} from '../util/logger';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import defaultSourceWatcher from '../watcher';
-import {NotificationCenter as NC} from 'node-notifier';
 // Import objects that are only used as Flow types.
 import type {FirefoxPreferences} from '../firefox/preferences';
 import type {OnSourceChangeFn} from '../watcher';
@@ -26,9 +26,6 @@ import type {
 import type {ExtensionManifest} from '../util/manifest';
 
 const log = createLogger(__filename);
-const notifier = new NC({
-  withFallback: true,
-});
 
 // defaultWatcherCreator types and implementation.
 
@@ -38,7 +35,7 @@ export type WatcherCreatorParams = {
   sourceDir: string,
   artifactsDir: string,
   onSourceChange?: OnSourceChangeFn,
-  messenger?: NC,
+  desktopNotifications?: typeof notifier,
 };
 
 export type WatcherCreatorFn = (params: WatcherCreatorParams) => Watchpack;
@@ -47,7 +44,7 @@ export function defaultWatcherCreator(
   {
     addonId, client, sourceDir, artifactsDir,
     onSourceChange = defaultSourceWatcher,
-    messenger = notifier,
+    desktopNotifications = notifier,
   }: WatcherCreatorParams
  ): Watchpack {
   return onSourceChange({
@@ -55,16 +52,19 @@ export function defaultWatcherCreator(
     artifactsDir,
     onChange: () => {
       log.debug(`Reloading add-on ID ${addonId}`);
-      messenger.notify({
-        title: 'Started reloading',
-        message: `Reloading add-on ID ${addonId}`,
-      });
+
       return client.reloadAddon(addonId)
+        .then(() => {
+          desktopNotifications({
+            titleString: 'web-ext run: reload',
+            messageString: `Reloaded add-on ID ${addonId}`,
+          });
+        })
         .catch((error) => {
           log.error(error.stack);
-          messenger.notify({
-            title: 'Error occured',
-            message: error.message,
+          desktopNotifications({
+            titleString: 'web-ext run: error occured',
+            messageString: error.message,
           });
           throw error;
         });
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -13,8 +13,6 @@ import {createLogger} from '../util/logger';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import defaultSourceWatcher from '../watcher';
 import {NotificationCenter as NC} from 'node-notifier';
-
-
 // Import objects that are only used as Flow types.
 import type {FirefoxPreferences} from '../firefox/preferences';
 import type {OnSourceChangeFn} from '../watcher';
@@ -57,9 +55,6 @@ export function defaultWatcherCreator(
     artifactsDir,
     onChange: () => {
       log.debug(`Reloading add-on ID ${addonId}`);
-      var notifier = new NC({
-        withFallback: true,
-      });
       messenger.notify({
         title: 'Started reloading',
         message: `Reloading add-on ID ${addonId}`,
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -12,6 +12,9 @@ import {
 import {createLogger} from '../util/logger';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import defaultSourceWatcher from '../watcher';
+import {NotificationCenter as NC} from 'node-notifier';
+
+
 // Import objects that are only used as Flow types.
 import type {FirefoxPreferences} from '../firefox/preferences';
 import type {OnSourceChangeFn} from '../watcher';
@@ -25,6 +28,9 @@ import type {
 import type {ExtensionManifest} from '../util/manifest';
 
 const log = createLogger(__filename);
+const notifier = new NC({
+  withFallback: true,
+});
 
 // defaultWatcherCreator types and implementation.
 
@@ -34,6 +40,7 @@ export type WatcherCreatorParams = {
   sourceDir: string,
   artifactsDir: string,
   onSourceChange?: OnSourceChangeFn,
+  messenger?: NC,
 };
 
 export type WatcherCreatorFn = (params: WatcherCreatorParams) => Watchpack;
@@ -42,6 +49,7 @@ export function defaultWatcherCreator(
   {
     addonId, client, sourceDir, artifactsDir,
     onSourceChange = defaultSourceWatcher,
+    messenger = notifier,
   }: WatcherCreatorParams
  ): Watchpack {
   return onSourceChange({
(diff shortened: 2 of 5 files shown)


saintsebastian:
Merging and rebasing doesn't seem to help, even after merging with upstream. Merging helps, but want to avoid committing all the changed files. What's the best approach at the moment?

src/cmd/run.js
```diff
@@ -44,16 +49,25 @@ export function defaultWatcherCreator(
...
     sourceDir,
     artifactsDir,
     onChange: () => {
       log.debug(`Reloading add-on ID ${addonId}`);
+      messenger.notify({
+        title: 'Started reloading',
+        message: `Reloading add-on ID ${addonId}`,
+      });
       return client.reloadAddon(addonId)
         .catch((error) => {
```
rpl: It would be probably better to notify the user that the addon has been reloaded successfully (the success scenario) instead of sending a notification before starting the reload, so that we alway show 1 notification per addon reload (with the current implementation we are going to show 1 notification when the addon reload successfully and 2 notifications on reloading errors).

We can achieve it by running the success notify in a `then` callback before the above `catch`, eg.

```
return client.reloadAddon(addonId)
  .then(() => {
     // notify addon successfully reloaded! 
     // this is only called if the promise returned by reloadAddon has not been rejected
  })
  .catch((error) => {
     // notify addon reloading error! 
     // this is called when the promise returned by reloadAddon has been rejected, or if the above callback
     // raises an exception.
  });
```

src/cmd/run.js
```diff
@@ -42,16 +44,28 @@ export function defaultWatcherCreator(
...
  ): Watchpack {
   return onSourceChange({
     sourceDir,
     artifactsDir,
     onChange: () => {
       log.debug(`Reloading add-on ID ${addonId}`);
+
       return client.reloadAddon(addonId)
+        .then(() => {
+          desktopNotifications({
```
kumar303: Hi all, just catching up with this.

I really love the idea of seeing a notification on error but showing a notification every time it reloads seems a bit much to me. I think it would minimize the effect of seeing a notification for failures (i.e. you might start to ignore the notices).

Do you really think we need to show a notice on every reload? If you do, I would request not to make this the default behavior but to allow it with a new option: `--notify-on-reload`. I can see how reload notifications might be useful to trust that the system is running.

saintsebastian: I like this idea! @rpl what do you think?

rpl: @saintsebastian :+1: yeah, I think that I agree with @kumar303 on this, by default we should only show a desktop notification when there is an error. 

(I would be ok to enable the 'desktop notification on success' with an additional cli option, but we can also keep this PR simple and evaluate to introduce the additional cli option and evaluate its correct behavior in a follow up)

kumar303: Here are some other ideas to keep this patch simple so it's easier to land:
* Only show a reload notification in verbose mode
* Remove the reload notification entirely and add it in later once we have a command line option

saintsebastian: @rpl @kumar303 i've removed notification on reload and corresponding test for now, once the code is reviewed and decision is made on how to better handle it, we can continue working on it.

package.json
```diff
@@ -57,6 +57,8 @@
     "minimatch": "3.0.3",
     "mz": "2.6.0",
     "node-firefox-connect": "1.2.0",
+    "node-notifier": "4.6.1",
+    "regenerator-runtime": "0.10.1",
```
rpl: remove this line (regenerator runtime is already included in the dependencies at line 63)

src/cmd/run.js
```diff
@@ -2,6 +2,7 @@
 import type FirefoxProfile from 'firefox-profile';
 import type Watchpack from 'watchpack';
 
+import {desktopNotifications as notifier} from '../util/desktop-notifier';
```
rpl: how about aliasing this as `defaultDesktopNotifications` instead of `notifier`? (like we do for `defaultFirefoxConnector` at line 7)

src/util/desktop-notifier.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+import {NotificationCenter} from 'node-notifier';
+
+const notifier = new NotificationCenter({
+  withFallback: true,
+});
+
+type desktopNotificationsParams = {
+  titleString: string,
```
rpl: let's rename `notifierSource` to `notifier`, `titleString` to `title` and `messageString` to `message`, it will also help to make it line 22 shorter like in:
```
notifier.notify({title, message});
```

src/util/desktop-notifier.js
```diff
@@ -0,0 +1,26 @@
...
+  messageString: string,
+  notifierSource?: NotificationCenter,
```
rpl: this should be `typeof NotificationCenter`.

src/util/desktop-notifier.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+import {NotificationCenter} from 'node-notifier';
+
+const notifier = new NotificationCenter({
```
rpl: let's rename this to `defaultNotifier`

rpl:
It would be nice (even in a follow up of this PR) to include an icon into the desktop notification (e.g. the icon of the reloaded web-extensions)

@kumar303 how it sounds to you?

saintsebastian:
Thanks for the comments! I added icon option to `desktop-notifier.js`so this should be very easy to implement as soon as it is decided what is the best way to represent reload (extension's own icon or a generic [reload](http://fontawesome.io/icon/refresh/)).

kumar303:
> It would be nice (even in a follow up of this PR) to include an icon into the desktop notification

I don't think it's such a big deal but I'm not opposed to it.

src/util/desktop-notifier.js
```diff
@@ -0,0 +1,22 @@
...
+
+const defaultNotifier = new NotificationCenter({
+  withFallback: true,
+});
+
+type desktopNotificationsParams = {
+  title: string,
+  message: string,
+  icon?: string,
+  notifier?: typeof NotificationCenter,
```
kumar303: This shouldn't be part of the main params because it doesn't relate to the notification call itself. You should pass it in as a secondary option so that it won't ever collide with a real notification param (which is unlikely but this follows our pattern elsewhere in web-ext).

Example:
````javascript
export function desktopNotifications(
  {
    title, message, icon,
  }: desktopNotificationsParams,
  {
    notifier = defaultNotifier,
  }: desktopNotificationOptions = {}
): void {
````

**Comment by saintsebastian** [comment]
@kumar303 this is a great suggestion, but I had to employ a kind of work around to test it. Thanks for the review!

File: src/util/desktop-notifier.js [hunk 9]
```diff
@@ -1,22 +1,35 @@
...
 
 export function desktopNotifications(
   {
-    title, message, icon, notifier = defaultNotifier,
-  }: desktopNotificationsParams
+    title, message, icon,
+  }: desktopNotificationsParams,
+  {
+    notifier = defaultNotifier,
+    logger = log,
```
rpl: @kumar303 This solution to spy the logger seems reasonable to me, how it looks to you?

src/util/desktop-notifier.js
```diff
@@ -1,22 +1,35 @@
...
+  }: desktopNotificationsOpts = {}
 ): void {
 
-  notifier.notify({title, message, icon});
+  notifier.notify({title, message, icon}, (err, res) => {
+    if (err) {
+      logger.debug(`notifier error: ${err.message},` +
```
rpl: It would be probably better to capitalize the first letter of this comment (e.g. `Desktop notifier error: ...`)

Also, I'm not sure if the log level is the right one, probably @kumar303 was suggesting to log the error even when web-ext is not running in verbose mode.

 @kumar303 am I right?

src/util/desktop-notifier.js
```diff
@@ -1,22 +1,35 @@
...
+                   ` response: ${res}`);
```
rpl: What is `res` supposed  to contains? 
if it is an object, then this is going to probably log "[Object]", that is not exactly very much informative ;-)

tests/unit/test-util/test.desktop-notifier.js
```diff
@@ -4,26 +4,50 @@ import {assert} from 'chai';
...
+      title: 'web-ext run: title',
+      message: 'message',
+    };
+
     it('is called and creates a message with correct parameters', () => {
       const fakeNotifier = {
         notify: sinon.spy(() => Promise.resolve()),
       };
-
       desktopNotifications(
```
rpl: This could be better formatted as:

```js
desktopNotification(expectedNotification, {
  ...
});
```

(also in the other places where there is a similar function call).

src/util/desktop-notifier.js
```diff
@@ -0,0 +1,35 @@
...
+  {
+    title, message, icon,
+  }: desktopNotificationsParams,
+  {
+    notifier = defaultNotifier,
+    logger = log,
+  }: desktopNotificationsOpts = {}
+): void {
+
+  notifier.notify({title, message, icon}, (err, res) => {
```
kumar303: Logging the error looks good, thanks, but this introduces a synchronization problem for testing. You will need to return a promise instead (just for testing purposes), like:
````javascript
  notifier.notify({title, message, icon}, (err, res) => {
    if (err) {
      logger.debug(`notifier error: ${err.message},` +
                   ` response: ${res}`);
      reject(err);
    } else {
      resolve();
    }
  });
});
````

File: tests/unit/test-util/test.desktop-notifier.js [hunk 14]
```diff
@@ -0,0 +1,53 @@
...
+      const expectedError = new Error('an error');
+      const log = createLogger(__filename);
+      sinon.spy(log, 'debug');
+      const fakeNotifier = {
+        notify: (obj, callback) => {
+          callback(expectedError, 'response');
+        },
+      };
+
+      desktopNotifications(
```
kumar303: After converting this function to a promise, you can do:
````javascript
return desktopNotifications(...)
  .then(makeSureItFails())
  .catch(() => {
    assert.ok(log.debug.called);
  });
````
Without a promise, it's hard to guarantee that this test will run the same way every time (if, say, we refactored some of the `desktopNotifications` code).

File: tests/unit/test-util/test.desktop-notifier.js [hunk 15]
```diff
@@ -0,0 +1,53 @@
...
+    const expectedNotification = {
+      title: 'web-ext run: title',
+      message: 'message',
+    };
+
+    it('is called and creates a message with correct parameters', () => {
+      const fakeNotifier = {
+        notify: sinon.spy(() => Promise.resolve()),
+      };
+      desktopNotifications(
```
kumar303: for consistency, you should probably also use this as a promise everywhere:
````javascript
return desktopNotifications(...)
  .then(() => {
    assert.ok(fakeNotifier.notify.called);
  });
````

File: tests/unit/test-util/test.desktop-notifier.js [hunk 16]
```diff
@@ -0,0 +1,53 @@
...
+describe('util/desktop-notifier', () => {
+  describe('desktopNotifications()', () => {
+    const expectedNotification = {
+      title: 'web-ext run: title',
+      message: 'message',
+    };
+
+    it('is called and creates a message with correct parameters', () => {
+      const fakeNotifier = {
+        notify: sinon.spy(() => Promise.resolve()),
```
kumar303: The signature that you need to mock looks more like this (according to their docs):
````javascript
const fakeNotifier = {
  notify: sinon.spy((options, callback) => callback()),
};
````

File: tests/unit/test-util/test.desktop-notifier.js [hunk 17]
```diff
@@ -0,0 +1,53 @@
...
+      };
+
+      desktopNotifications(
+        expectedNotification, {
+          notifier: fakeNotifier,
+          logger: log,
+        });
+      assert.ok(log.debug.called);
+      assert.equal(log.debug.firstCall.args[0],
+                `notifier error: ${expectedError.message}, response: response`);
```
kumar303: nit: this should either line up with the inside of the opening parenthesis or be indented only two spaces

src/util/desktop-notifier.js
saintsebastian: They are printed in console in the docs, so hopefully they should be readable

saintsebastian:
@kumar303 @rpl thanks for cleaning up this code and also clearing up some things for me in the process

kumar303:
I'm super excited about this! I've stared at Firefox scratching my head more than once, only to discover a syntax error in the console.

PR #676, opened 2016-12-15
feat: web-ext automatically checks for updates 

### LINKED ISSUE #142: Check for updates to self and alert users
Because it's likely to be installed globally and not manually updated (I, and--anecdotally--many other developers, aren't in the habit of updating their globally installed commands on a regular basis), `web-ext` should check for updates to itself every once in awhile. A day or two is probably fine.

It should simply alert the user to the fact that there's a new version and the `npm` command to install it.

shubheksha:
Fixes #142


src/program.js
```diff
@@ -109,6 +110,16 @@ export class Program {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
       await runCommand(argv);
+
+      let pkg = {
```
rpl: `pkg` is never re-assigned again, and so we can use `const` instead of `let`

src/util/updates.js
```diff
@@ -0,0 +1,25 @@
+/* @flow */
+import updateNotifier from 'update-notifier';
```
kumar303: The key to testing this function is to stub out the `updateNotifier` since we can trust it works as documented. The goal of the tests will be to make sure that web-ext is configuring the notifier correctly.

First, import the original package as a default that can be customized:
````javascript
import defaultUpdateNotifier from 'update-notifier';
````
Next, use [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) so that it can be replaced:
````javascript
export function checkForAutomaticUpdates(
  {
    name,
    version,
    updateCheckInterval,
    updateNotifier = defaultUpdateNotifier,
  }: checkForAutomaticUpdatesParams
) {
  // ...
  updateNotifier();
}
````
This allows you to test it with a stub:
````javascript
const updateNotifier = sinon.stub();
checkForAutomaticUpdates({name, version, updateNotifier});

assert.equal(updateNotifier.firstCall.args[0].pkg.name, 'web-ext');
````

File: src/program.js [hunk 3]
```diff
@@ -20,10 +21,15 @@ export class Program {
   yargs: any;
   commands: { [key: string]: Function };
   shouldExitProgram: boolean;
+  checkForUpdates: typeof checkForAutomaticUpdates;
 
   constructor(
     argv: ?Array<string>,
-    {absolutePackageDir = process.cwd()}: {absolutePackageDir?: string} = {}
+    {
+      absolutePackageDir = process.cwd(),
```
kumar303: similarly, this module / class should be set up for dependency injection so that you can replace the update utility with a stub for testing:
````javascript
import {checkForAutomaticUpdates as defaultUpdateChecker} from './util/updates';

class Program {
  constructor(
    argv: ?Array<string>,
    {
      absolutePackageDir?: string,
      checkForAutomaticUpdates = defaultUpdateChecker,
    }
  ) {
    // ...
    this.checkForUpdates = checkForAutomaticUpdates;
  }
}
````

The test can then use a stub:
````javascript
const checkForAutomaticUpdates = sinon.stub();
const program = new Program(['run'], {checkForAutomaticUpdates});
await program.execute();

assert.equal(checkForAutomaticUpdates.firstCall.args[0].version, '1.0.0');
````

File: src/program.js [hunk 4]
```diff
@@ -108,7 +115,14 @@ export class Program {
       if (!runCommand) {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
+
+      this.checkForUpdates ({
+        name: 'web-ext',
+        version: defaultVersionGetter(absolutePackageDir),
```
kumar303: this function already has a `getVersion` parameter which can be replaced by a stub for testing. You can use it like `checkForUpdates({version: getVersion(absolutePackageDir)})`

tests/unit/test-util/test.updates.js
```diff
@@ -0,0 +1,26 @@
...
+
+    checkForAutomaticUpdates({
+      name: 'web-ext',
+      version: '1.0.0',
+      updateCheckInterval: 0,
+      updateNotifier: updateNotifierStub,
+    });
+    assert.equal(updateNotifierStub.firstCall.args[0].pkg.name, 'web-ext');
+    assert.equal(updateNotifierStub.firstCall.args[0].pkg.version, '1.0.0');
+    assert.equal(updateNotifierStub.firstCall.args[0].updateCheckInterval, 0);
```
kumar303: This also needs to assert that `updateNotifierStub.update.called` is `true`

tests/unit/test.program.js
```diff
@@ -6,7 +6,11 @@ import sinon from 'sinon';
 import {assert} from 'chai';
 import {spy} from 'sinon';
 
-import {defaultVersionGetter, main, Program} from '../../src/program';
+import {
+  defaultVersionGetter,
+  main,
+  Program,
```
kumar303: Any reason why you split these into multiple lines? Since the previous statement was less than 80 characters, you don't need to split it.

tests/unit/test.program.js
```diff
@@ -32,7 +36,9 @@ describe('program.Program', () => {
     let thing = spy(() => Promise.resolve());
     let program = new Program(['thing'])
       .command('thing', 'does a thing', thing);
-    return execProgram(program)
+    return execProgram(program, {
+      checkForUpdates: spy(),
```
kumar303: Passing this directly creates unnecessary repetition. I don't see what the problem is with adding it as a default to [execProgram()](https://github.com/shubheksha/web-ext/blob/305f83ffff28cb1bb38aeb91d7636b1bd97559b0/tests/unit/test.program.js#L30). What problem were you running into?

tests/unit/test.program.js
```diff
@@ -242,13 +256,37 @@ describe('program.Program', () => {
       });
   });
 
+  it('checks for updates automatically', () => {
+    const root = path.join(__dirname, '..', '..');
+    const handler = spy();
+    const checkForAutomaticUpdates = sinon.stub();
+    const program = new Program(['run'])
+    .command('run', 'some command', handler);
```
kumar303: This should be indented two spaces to show that it's a continuation of the call up above.

tests/unit/test.program.js
```diff
@@ -242,13 +256,37 @@ describe('program.Program', () => {
...
+    return execProgram(program, {
+      checkForUpdates: checkForAutomaticUpdates,
+    })
+      .then(() => {
+        assert.equal(checkForAutomaticUpdates.firstCall.args[0].name,
+                      'web-ext');
+        let pkgFile = path.join(root, 'package.json');
+        return fs.readFile(pkgFile)
+          .then((pkgData) => {
```
kumar303: You should always chain promises instead of nest them. Otherwise you might as well use callbacks :) Chaining would look like:
````javascript
return execProgram()
  .then(() => {
    assert.equal(...);
    return fs.readFile(pkgFile);
  })
  .then((pkgFile) => {
    assert.equal(...);
  });
````

File: tests/unit/test.program.js [hunk 10]
```diff
@@ -242,13 +256,37 @@ describe('program.Program', () => {
...
+            assert.equal(checkForAutomaticUpdates.firstCall.args[0].version,
+                          JSON.parse(pkgData).version);
```
kumar303: I like this test! Good idea.

tests/unit/test.program.js
shubheksha: It wasn't being passed by default in `execProgram()` and making some of the tests fail. Not sure why.

kumar303: Maybe it wasn't defined in the right part of `execProgram()`? If it's not working, post a broken patch and I can take a look.

src/program.js
```diff
@@ -113,6 +119,12 @@ export class Program {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
       await runCommand(argv);
+
+      checkForUpdates ({
+        version: getVersion(absolutePackageDir),
+        updateNotifier: defaultUpdateChecker,
```
kumar303: why does `updateNotifier` need to be passed? What does it do? I'm confused.

src/program.js
```diff
@@ -24,7 +25,11 @@ export class Program {
 
   constructor(
     argv: ?Array<string>,
-    {absolutePackageDir = process.cwd()}: {absolutePackageDir?: string} = {}
+    {
+      absolutePackageDir = process.cwd(),
+    }: {
+      absolutePackageDir?: string,
```
kumar303: even though there is only one param, I think it's worth moving this to a standalone type object like we do elsewhere:
````javascript
type ProgramOptions = {
  absolutePackageDir?: string,
}
````

File: src/util/updates.js [hunk 13]
```diff
@@ -0,0 +1,20 @@
...
+
+type checkForUpdatesParams = {
+  version: string,
+  updateNotifier?: typeof defaultUpdateNotifier,
+};
+
+export function checkForUpdates({
+  version,
+  updateNotifier = defaultUpdateNotifier,
+}: checkForUpdatesParams
```
kumar303: nit: I think it would be easier to visualize the nesting of these statements if they were indented one more level like:
````javascript
export function checkForUpdates(
  {
    version,
    updateNotifier = defaultUpdateNotifier,
  }: checkForUpdatesParams
) {
  ...
}
````

File: tests/unit/test-util/test.updates.js [hunk 14]
```diff
@@ -0,0 +1,30 @@
...
+import {it, describe} from 'mocha';
+import {assert} from 'chai';
+import sinon from 'sinon';
+
+import {checkForUpdates} from '../../../src/util/updates';
+
+describe('util/updates', () => {
+  describe('checkForUpdates()', () => {
+    it('calls the notifier with the correct parameters', () => {
+      let updateNotifierStub = sinon.spy(() => {
```
kumar303: this should use `const`

src/util/updates.js
```diff
@@ -0,0 +1,20 @@
+/* @flow */
+import defaultUpdateNotifier from 'update-notifier';
+
+type checkForUpdatesParams = {
+  version: string,
+  updateNotifier?: typeof defaultUpdateNotifier,
+};
```
kumar303: This should use the new pipe syntax in Flow that says "only these parameters are allowed" : 
````javascript
type checkForUpdatesParams = {|
  version: string,
  updateNotifier?: typeof defaultUpdateNotifier,
|};
````

Comment (kumar303) [inline src/util/updates.js]: gah, it's not documented yet (or I just can't find where)

File: tests/unit/test-util/test.updates.js [hunk 16]
```diff
@@ -0,0 +1,30 @@
...
+        return {
+          notify: sinon.spy(),
+        };
+      });
+
+      checkForUpdates({
+        name: 'web-ext',
```
kumar303: this call passes a bunch of unnecessary parameters but once you update the Flow type you should get errors about them

tests/unit/test.program.js
```diff
@@ -242,13 +246,37 @@ describe('program.Program', () => {
...
+    const handler = spy();
+    const checkForAutomaticUpdates = sinon.stub();
+    const program = new Program(['run'])
+      .command('run', 'some command', handler);
+    return execProgram(program, {
+      checkForUpdates: checkForAutomaticUpdates,
+    })
+      .then(() => {
+        assert.equal(checkForAutomaticUpdates.firstCall.args[0].name,
+                      'web-ext');
```
kumar303: nit: this should line up with the opening parenthesis

kumar303: actually, you don't need to make this assertion because `program` is no longer responsible for passing the name of the package

tests/unit/test.program.js
```diff
@@ -242,13 +246,37 @@ describe('program.Program', () => {
...
+        return fs.readFile(pkgFile);
+      })
+      .then((pkgData) => {
+        assert.equal(checkForAutomaticUpdates.firstCall.args[0].version,
+                          JSON.parse(pkgData).version);
```
kumar303: nit: this should line up with the opening parenthesis of the `assert.equal()` call

tests/unit/test.program.js
```diff
@@ -242,13 +246,31 @@ describe('program.Program', () => {
...
+    const handler = spy();
+    const checkForAutomaticUpdates = sinon.stub();
+    const program = new Program(['run'])
+      .command('run', 'some command', handler);
+    return execProgram(program, {
+      checkForUpdates: checkForAutomaticUpdates,
+    })
+      .then(() => {
+        assert.equal(checkForAutomaticUpdates.firstCall.args[0].version,
+                          defaultVersionGetter(root));
```
kumar303: I noticed some test failures locally (surprisingly not on Travis though). I'm going to suggest a few things.

* Stub out `getVersion` for all tests that are not covering any version logic
* Use a stub `getVersion` function in your update checker test. I'm suggesting this because the behavior of `defaultVersionGetter` is already covered in other tests.

Here is a patch to illustrate what I mean:
````diff
index 58732b8..a4eea98 100644
@@ -21,6 +21,7 @@ describe('program.Program', () => {
     let absolutePackageDir = path.join(__dirname, '..', '..');
     return program.execute(
       absolutePackageDir, {
+        getVersion: () => 'not-a-real-version',
         checkForUpdates: spy(),
         systemProcess: fakeProcess,
         shouldExitProgram: false,
@@ -251,12 +252,14 @@ describe('program.Program', () => {
     const checkForAutomaticUpdates = sinon.stub();
     const program = new Program(['run'])
       .command('run', 'some command', handler);
+    const getVersion = () => 'some-package-version';
     return execProgram(program, {
+      getVersion,
       checkForUpdates: checkForAutomaticUpdates,
     })
       .then(() => {
         assert.equal(checkForAutomaticUpdates.firstCall.args[0].version,
-                          defaultVersionGetter(root));
+                     'some-package-version');
       });
   });
 });
@@ -266,6 +269,7 @@ describe('program.main', () => {
 
   function execProgram(argv, {projectRoot = '', ...mainOptions}: Object = {}) {
     const runOptions = {
+      getVersion: () => 'not-a-real-version',
       checkForUpdates: spy(),
       shouldExitProgram: false,
       systemProcess: fake(process),
````

File: src/program.js [hunk 20]
```diff
@@ -113,6 +120,11 @@ export class Program {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
       await runCommand(argv);
+
+      checkForUpdates ({
```
kumar303: This needs to start before the command starts, at least it does for the `run` command. I'm not totally sure why, I think it's because of how `run` exits when Firefox quits. I think it's safe to move it up because the updater runs unobtrusively in an async child process.

````diff
index 1f49966..33b34da 100644
--- a/src/program.js
+++ b/src/program.js
@@ -119,12 +119,13 @@ export class Program {
       if (!runCommand) {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
-      await runCommand(argv);
 
       checkForUpdates ({
         version: getVersion(absolutePackageDir),
       });
 
+      await runCommand(argv);
+
     } catch (error) {
       const prefix = cmd ? `${cmd}: ` : '';
       if (!(error instanceof UsageError) || argv.verbose) {
````

It looks nice!

screenshot 2016-12-21 12 25 18

File: package.json [hunk 21]
```diff
@@ -57,12 +57,13 @@
     "minimatch": "3.0.3",
     "mz": "2.6.0",
     "node-firefox-connect": "1.2.0",
-    "regenerator-runtime": "0.10.1",
     "parse-json": "2.2.0",
+    "regenerator-runtime": "0.10.0",
```
shubheksha: I didn't touch package.json at all while resolving the conflicts. There was no conflict in this file that needed to be resolved manually. I get git did it on its own?

kumar303:
@shubheksha oh, I forgot to mention, you need to disable the update check when in development (because it will say web-ext is always out of date). Something like this:
````javascript
if (localEnv === 'production') {
  checkForUpdates({version: getVersion(absolutePackageDir)});
}
````
You can set up `localEnv` similar to how it's set up for [defaultVersionGetter()](https://github.com/shubheksha/web-ext/blob/c12d3692df3ec61454424ce769afdd8785b2d725/src/program.js#L158).

File: tests/unit/test-util/test.updates.js [hunk 22]
```diff
@@ -0,0 +1,28 @@
...
+
+      checkForUpdates({
+        version: '1.0.0',
+        updateNotifier: updateNotifierStub,
+      });
+      assert.equal(updateNotifierStub.called, true);
+      assert.equal(updateNotifierStub.firstCall.args[0].pkg.name, 'web-ext');
+      assert.equal(updateNotifierStub.firstCall.args[0].pkg.version, '1.0.0');
+      assert.equal(updateNotifierStub.firstCall.args[0].updateCheckInterval,
+                  1000 * 60 * 60 * 24 * 7);
```
kumar303: Sorry if I wasn't clear about this style rule. We always align the second argument to the inside of the open parenthesis which allows the reader to easily see that the statement within parentheses continues on the next line:
````javascript
      assert.equal(updateNotifierStub.firstCall.args[0].updateCheckInterval,
                   1000 * 60 * 60 * 24 * 7);
````

We don't have the lint rule for this enabled because it causes some other problems.

File: tests/unit/test.program.js [hunk 23]
```diff
@@ -242,13 +247,34 @@ describe('program.Program', () => {
...
+  it('checks for updates automatically', () => {
+    const handler = spy();
+    const getVersion = () => 'some-package-version';
+    const checkForUpdates = sinon.stub();
+    const program = new Program(['run'])
+      .command('run', 'some command', handler);
+    return execProgram(program, {
+      checkForUpdates,
+      getVersion,
+      localEnv: 'production',
```
kumar303: This needs one more test that does `localEnv: 'development'` and makes sure the update checker doesn't get called, like `assert.equal(checkForUpdates.called, false)`

src/program.js
```diff
@@ -84,8 +90,9 @@ export class Program {
   async execute(
     absolutePackageDir: string,
     {
-      systemProcess = process, logStream = defaultLogStream,
-      getVersion = defaultVersionGetter, shouldExitProgram = true,
+      checkForUpdates = defaultUpdateChecker, systemProcess = process,
+      logStream = defaultLogStream, getVersion = defaultVersionGetter,
+      shouldExitProgram = true, localEnv = 'development',
```
kumar303: This won't work in production :)

We have a [magic build variable](https://github.com/mozilla/web-ext/blob/master/webpack.config.js#L64) which is a bit confusing. You would use it like this:

    localEnv = WEBEXT_BUILD_ENV

This will set `localEnv` to the value of `NODE_ENV` at the time that `web-ext` was built.

Are you testing your patch manually? Automated tests are great but you also need to test it manually. The way I was doing it was temporarily decrementing the version to `1.5.0` in `package.json` (just for testing), running `NODE_ENV=production web-ext build` to get a build, then trying `web-ext run` or whatever to see the update notification.

src/util/updates.js
```diff
@@ -0,0 +1,21 @@
...
+  {
+    version,
+    updateNotifier = defaultUpdateNotifier,
+  }: checkForUpdatesParams
+) {
+  const pkg = {name: 'web-ext', version};
+
+  updateNotifier({
+    pkg,
+    updateCheckInterval: 1000 * 60 * 60 * 24 * 7, // 1 week,
```
kumar303: I think we should make this three days

src/program.js
```diff
@@ -84,8 +90,9 @@ export class Program {
   async execute(
     absolutePackageDir: string,
     {
-      systemProcess = process, logStream = defaultLogStream,
-      getVersion = defaultVersionGetter, shouldExitProgram = true,
+      checkForUpdates = defaultUpdateChecker, systemProcess = process,
+      logStream = defaultLogStream, getVersion = defaultVersionGetter,
+      shouldExitProgram = true, localEnv = WEBEXT_BUILD_ENV,
```
kumar303: Hmm, `localEnv` doesn't really make sense. I'm not sure why we called the last one that. I would suggest naming this `globalEnv` because that's what it is!

tests/unit/test-util/test.updates.js
```diff
@@ -0,0 +1,28 @@
...
+
+      checkForUpdates({
+        version: '1.0.0',
+        updateNotifier: updateNotifierStub,
+      });
+      assert.equal(updateNotifierStub.called, true);
+      assert.equal(updateNotifierStub.firstCall.args[0].pkg.name, 'web-ext');
+      assert.equal(updateNotifierStub.firstCall.args[0].pkg.version, '1.0.0');
+      assert.equal(updateNotifierStub.firstCall.args[0].updateCheckInterval,
+                   1000 * 60 * 60 * 24 * 7);
```
kumar303: I didn't mention this last review but an assertion like this isn't so helpful because when you update the interval, you have to update the test too. I would suggest a looser check, something that just makes sure the option is set and that it's a number:

````javascript
assert.isNumber(updateNotifierStub.firstCall.args[0].updateCheckInterval);
````

File: tests/unit/test.program.js [hunk 28]
```diff
@@ -241,13 +246,50 @@ describe('program.Program', () => {
...
 });
 
 
 describe('program.main', () => {
 
   function execProgram(argv, {projectRoot = '', ...mainOptions}: Object = {}) {
-    const runOptions = {shouldExitProgram: false, systemProcess: fake(process)};
+    const runOptions = {
+      getVersion: () => 'not-a-real-version',
+      checkForUpdates: spy(),
```
kumar303: Since no assertions are made in these tests, you could just assign it as a `sinon.stub()` which does nothing. Doesn't really matter either way though.

shubheksha:
@kumar303, could you review this again, please?

src/program.js
```diff
@@ -138,14 +152,14 @@ declare var WEBEXT_BUILD_ENV: string;
 
 //A defintion of type of argument for defaultVersionGetter
 type versionGetterOptions = {
-  localEnv?: string,
+  globalEnv?: string,
```
kumar303: thanks :)

PR #681, opened 2016-12-16
Added import linting

### LINKED ISSUE #442: Add import linting
We can steal the config from https://github.com/mozilla/addons-frontend/pull/945 to get proper import linting in web-ext. Flow has some nice support for this already but I think the linting rules would provide more granularity.

shubheksha:
Fixes #442

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,17 +9,6 @@
     "flowtype": {
       "onlyFilesWithFlowAnnotation": true
     },
-    "import/ignore": [
-      // Because of CommonJS incompatibility, we can't
-      // check for bad imports in node_modules.
-      "node_modules", 
-    ],
-    "import/resolver": {
-      "node": {
-        // This adds ./src for relative imports.
-        "moduleDirectory": ["node_modules", "src"]
-      }
-    }
   },
   "env": {
     "node": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
   "parser": "babel-eslint",
   "plugins": [
     "async-await",
-    "flowtype"
+    "flowtype",
+    "import"
   ],
   "settings": {
     "flowtype": {
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,22 @@
   "settings": {
     "flowtype": {
       "onlyFilesWithFlowAnnotation": true
+    },
+    "import/ignore": [
+      // Because of CommonJS incompatibility, we can't
+      // check for bad imports in node_modules.
+      "node_modules",
+      // Ignore non-JS imports.
+      "\\.scss$",
+      "\\.jpg$",
+      "\\.mp4$",
+      "\\.webm$"
+    ],
+    "import/resolver": {
+      "node": {
+        // This adds ./src for relative imports.
+        "moduleDirectory": ["node_modules", "src"]
+      }
     }
   },
   "env": {
@@ -118,6 +134,50 @@
 
     "flowtype/define-flow-type": 1,
     "flowtype/use-flow-type": 1,
-    "flowtype/space-after-type-colon": [1, "always"]
+    "flowtype/space-after-type-colon": [1, "always"],
+
+    // This makes sure imported modules exist.
+    "import/no-unresolved": ["error"],
+    // This makes sure imported names exist.
+    "import/named": ["error"],
+    // This will catch accidental default imports when no default is defined.
+    "import/default": ["error"],
+    // This makes sure `*' imports are dereferenced to real exports.
+    "import/namespace": ["error"],
+    // This catches any export mistakes.
+    "import/export": ["error"],
+    // This catches default names that conflict with actual exported names.
+    // For example, this was probably a typo:
+    //   import foo from 'bar';
+    // that should be corrected as:
+    //   import { foo } from 'bar';
+    "import/no-named-as-default": ["error"],
+    // This catches possible typos like trying to access a real export on a
+    // default import.
+    "import/no-named-as-default-member": ["error"],
+    // This prevents exporting a mutable variable.
+    "import/no-mutable-exports": ["error"],
+    // This makes sure package.json defines dev vs. prod dependencies correctly.
+    "import/no-extraneous-dependencies": ["error", {
+      // The following are not allowed to be imported. See .eslintrc in other
+      // directories (like ./test) for where this gets overidden.
+      "devDependencies": false, "optionalDependencies": false, "peerDependencies": false
+    }],
+    // This ensures imports are at the top of the file.
+    "import/imports-first": ["error"],
+    // This catches duplicate exports.
+    "import/no-duplicates": ["error"],
+    // This ensures import statements never provide a file extension in the path.
+    "import/extensions": ["error", "never"],
+    // This ensures imports are organized by type and that groups are separated
+    // by a new line.
+    "import/order": ["error", {
+      "groups": [
+        "builtin", "external", "internal", ["parent", "sibling"], "index"
+      ],
+      "newlines-between": "always"
+    }],
+    // This ensures a new line after all import statements.
+    "import/newline-after-import": ["error"],
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,12 +12,7 @@
     "import/ignore": [
       // Because of CommonJS incompatibility, we can't
       // check for bad imports in node_modules.
-      "node_modules",
-      // Ignore non-JS imports.
-      "\\.scss$",
-      "\\.jpg$",
-      "\\.mp4$",
-      "\\.webm$"
+      "node_modules", 
     ],
     "import/resolver": {
       "node": {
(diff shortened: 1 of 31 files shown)


src/cmd/build.js
```diff
@@ -1,30 +1,31 @@
...
+import streamToPromise from 'stream-to-promise';
 import parseJSON from 'parse-json';
 
 import defaultSourceWatcher from '../watcher';
 import {zipDir} from '../util/zip-dir';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
-import {UsageError} from '../errors';
+import {UsageError} from '../errors'; //eslint-disable-line import/order
```
shubheksha: @kumar303, I spoke to @rpl about this but both of us couldn't figure out while ESLint was complaining at this point. So resorted to this as nothing else seemed to work.

shubheksha: This happens in a couple of other places as well. However, it works in `run.js`. I've been able to figure out why. I tried mimicking everything as is but still it didn't work.

.eslintrc
```diff
@@ -2,11 +2,28 @@
...
   "settings": {
     "flowtype": {
       "onlyFilesWithFlowAnnotation": true
+    },
+    "import/ignore": [
+      // Because of CommonJS incompatibility, we can't
+      // check for bad imports in node_modules.
+      "node_modules",
+      // Ignore non-JS imports.
+      "\\.scss$",
```
rpl: @shubheksha we don't have any of these `.scss`, `.jpg`, `.mp4` or `.webm` imports in our sources, right?
if it is not currently needed, I would remove the lines between 16 and 20.

.eslintrc
```diff
@@ -118,6 +135,50 @@
 
     "flowtype/define-flow-type": 1,
     "flowtype/use-flow-type": 1,
-    "flowtype/space-after-type-colon": [1, "always"]
+    "flowtype/space-after-type-colon": [1, "always"],
+
+    // This makes sure imported modules exist.
+    "import/no-unresolved": ["error"],
```
rpl: @shubheksha in the other rules configured in this .eslintrc file, we currently use `2` in place of "error", we should probably keep it consistent with this new group of rules as well.

(Also, `"rule": ["error"]` can be probably be rewritten to `"rule": 2` for the same reason)

Gruntfile.js
```diff
@@ -9,9 +9,9 @@ module.exports = function(grunt) {
 
 
   // This loads all grunt tasks matching the grunt-*, @*/grunt-* patterns.
-  require('load-grunt-tasks')(grunt);
+  require('load-grunt-tasks')(grunt); //eslint-disable-line import/no-extraneous-dependencies
```
rpl: I would probably prefer to put all the development related files that are not in the tests dir into the .eslintrc file as file glob patterns, instead of ignoring the issues line by line, e.g.:

```json
`
...
"import/no-extraneous-dependencies": [2, {
   "devDependencies": ["Gruntfile.js", "webpack.*.js", "tasks/*.js"]
}],
...
```

but let's see what @kumar303 thinks about it before applying this change.

src/firefox/remote.js
```diff
@@ -16,11 +16,13 @@ export const REMOTE_PORT = 6005;
...
 
 // RemoteFirefox types and implementation
-
+/* eslint-disable import/order, import/imports-first, import/no-extraneous-dependencies */
+//This is related to Flow and will be stripped in production hence won't be treated as a dependency at all
 import type FirefoxClient from 'firefox-client';
 
 export type FirefoxConnectorFn =
   (port?: number) => Promise<FirefoxClient>;
+/* eslint-disable import/order, import/imports-first */
```
rpl: @shubheksha This seems that it should be an `eslint-enable` comment

Also, I'm wondering if `export type FirefoxConnectorFn = ...` needs to be inside the `eslint-disable` comment.

src/firefox/remote.js
```diff
@@ -16,11 +16,13 @@ export const REMOTE_PORT = 6005;
 
 
 // RemoteFirefox types and implementation
-
+/* eslint-disable import/order, import/imports-first, import/no-extraneous-dependencies */
+//This is related to Flow and will be stripped in production hence won't be treated as a dependency at all
```
rpl: when we have regular comments in the same place where we are going to use an eslint-disable block comment, it would be probably better to move that comment outside the eslint-disable, e.g.:

```
// This is related to ...
/* eslint-disable ... */
...
/* eslint-enable ... */
```

and 

```
// import Flow Types
/* eslint-disable ... */
...
/* eslint-enable ... */

File: src/cmd/build.js (hunk 1 above)
Comment (kumar303) [inline src/cmd/build.js]: It's because all imports need to be grouped together. It's not the most helpful message but it's telling you that you forgot to group some imports. Here is a patch that fixes it:

````diff
diff --git a/src/cmd/build.js b/src/cmd/build.js
index 17a71f9..18c4601 100644
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -12,17 +12,13 @@ import {zipDir} from '../util/zip-dir';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
-import {UsageError} from '../errors'; //eslint-disable-line import/order
+import {UsageError} from '../errors';
+import type {OnSourceChangeFn} from '../watcher';
+import type {ExtensionManifest} from '../util/manifest';
 
 
-/* eslint-disable import/order, import/imports-first */
-// Import flow types.
-import type {OnSourceChangeFn} from '../watcher';
-import type {ExtensionManifest} from '../util/manifest';
-/* eslint-disable import/order, import/imports-first */
-
 export function safeFileName(name: string): string {
   return name.toLowerCase().replace(/[^a-z0-9\.-]+/g, '_');
 }
````

File: Gruntfile.js (hunk 4 above)
Comment (kumar303) [inline Gruntfile.js]: yeah, that sounds better to me as well

File: src/cmd/build.js (hunk 1 above)
Comment (kumar303) [inline src/cmd/build.js]: you will run into this with all flow type imports. We kept those separate to make it obvious that they are only for flow but I think it's fine to group them with the other parent/sibling imports.

**Comment by rpl** [comment]
Thanks @shubheksha! this "imports linting" is going to help a lot! 

It is almost done, I added some comments, you can already handle/discuss with me the most important ones and then @kumar303 is going to take a second look.

File: src/cmd/build.js (hunk 1 above)
Comment (shubheksha) [inline src/cmd/build.js]: Oh, so even though the comment disables the imports for types, it doesn't do it for a normal import and hence ESLint complains? I thought disabling ESLint for the type imports should be enough to deal with this. 
However, if we can club module and Flow imports, this won't be an issue at all. I'll go ahead and make changes to all the files in question.

File: tests/unit/test-firefox/test.firefox.js [hunk 7]
```diff
@@ -1,19 +1,21 @@
...
 import * as firefox from '../../../src/firefox';
 import {onlyInstancesOf, UsageError, WebExtError} from '../../../src/errors';
-import {fs} from 'mz';
 import {withTempDir} from '../../../src/util/temp-dir';
 import {TCPConnectError, fixturePath, fake, makeSureItFails} from '../helpers';
 import {basicManifest, manifestWithoutApps} from '../test-util/test.manifest';
-import {defaultFirefoxEnv} from '../../../src/firefox/';
 import {RemoteFirefox} from '../../../src/firefox/remote';
 
+const defaultFirefoxEnv = firefox.defaultFirefoxEnv;
```
shubheksha: I tried that, but it looks like there are no default imports in `firefox.js` hence it fails.

rpl: @shubheksha yeah, you are right, it looks like we have no nice ES6 syntax to merge the two (all named exports, '*', into the 'firefox' alias and the single 'defaultFirefoxEnv' named export with a single import statement :-( )

That's not a big deal, how about just changing the above line to:
```js
const {defaultFirefoxEnv} = firefox;
```

.eslintrc
```diff
@@ -2,11 +2,23 @@
...
+    "import"
   ],
   "settings": {
     "flowtype": {
       "onlyFilesWithFlowAnnotation": true
+    },
+    "import/ignore": [
+      // Because of CommonJS incompatibility, we can't
+      // check for bad imports in node_modules.
+      "node_modules",
```
kumar303: Huh, I forget what the exact issue was here in `addons-frontend`. I just removed it and `web-ext` has no complaints :) Let's remove this entire `import/ignore` block from the config for now until we run into specific problems.

.eslintrc
```diff
@@ -2,11 +2,23 @@
...
+    },
+    "import/ignore": [
+      // Because of CommonJS incompatibility, we can't
+      // check for bad imports in node_modules.
+      "node_modules", 
+    ],
+    "import/resolver": {
+      "node": {
+        // This adds ./src for relative imports.
+        "moduleDirectory": ["node_modules", "src"]
```
kumar303: Since our relative imports all use exact paths (unlike `addons-frontend`), adding `src` here isn't necessary. Also, since `node_modules` is a default, you can just remove this entire `import/resolver` block.

tests/.eslintrc
```diff
@@ -5,5 +5,9 @@
   },
```
kumar303: Huh, I just noticed `mocha: true` in `env`. You can remove that. That rule allowed us to reference `describe()` as a global but we never do that (we import it) so the rule is not necessary.

tests/.eslintrc
```diff
@@ -5,5 +5,9 @@
...
   "rules": {
     "strict": [2, "never"],
```
kumar303: we already had strict in the main `.eslintrc` so you can remove it from here

src/cmd/run.js
```diff
@@ -9,24 +12,18 @@ import {
...
-const log = createLogger(__filename);
-
-
-// Import objects that are only used as Flow types.
-import type FirefoxProfile from 'firefox-profile';
 import type {OnSourceChangeFn} from '../watcher';
-import type Watchpack from 'watchpack';
 import type {
-  FirefoxProcess,
+  FirefoxProcess, //eslint-disable-line import/named
```
kumar303: aww, it doesn't work with the flow linter? That's too bad. This is a good way to workaround it though.

src/firefox/index.js
```diff
@@ -14,26 +15,20 @@ import {getPrefs as defaultPrefGetter} from './preferences';
...
-
-export const defaultFirefoxEnv = {
-  XPCOM_DEBUG_BREAK: 'stack',
-  NS_TRACE_MALLOC_DISABLE_STACKS: '1',
-};
-
-
-// Import flow types.
-
+//Import flow types
```
kumar303: nit: we generally put a space after the slashes for readable comments like `// Import flow types`. We don't have a lint rule for that because when you comment out some code temporarily you get yelled at which can be annoying.

tests/unit/test-cmd/test.sign.js
```diff
@@ -15,7 +16,7 @@ import completeSignCommand, {
   extensionIdFile, getIdFromSourceDir, saveIdToSourceDir,
 } from '../../../src/cmd/sign';
 import {makeSureItFails, fixturePath} from '../helpers';
-
+//Import flow type
```
kumar303: nit: again, space after the slashes

PR #695, opened 2016-12-20
fix: added eslint rule for shorthand property assignment

### LINKED ISSUE #693: Add eslint rule for short hand property assignment
An enhancement

We should add this rule: http://eslint.org/docs/rules/object-shorthand to our eslintrc to use shorthand property assignment wherever applicable

shubheksha:
Fixes #693

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -41,7 +41,7 @@ module.exports = {
     libraryTarget: 'commonjs2',
   },
   module: {
-    preLoaders: preLoaders,
+    preLoaders,
     loaders: [
       {
         exclude: /(node_modules|bower_components)/,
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
     libraryTarget: 'commonjs2',
   },
   module: {
-    preLoaders: preLoaders,
+    preLoaders,
```
kumar303: Does this work? I don't know if the webpack config file is run through babel. It may need to be ignored by eslint for this rule.

shubheksha: Yep, it works! All the tests are passing.

kumar303: I don't know if the tests will cover this. Does `npm run build` still work?

shubheksha: Yes, it does.

kumar303: cool

PR #700, opened 2016-12-21
fix: Added prefer-const to eslint and related changes

### LINKED ISSUE #640: Add lint rule to prefer constants over variables

aniketkudale:
Fixes #640

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -83,8 +83,8 @@ export function defaultReloadStrategy(
   }: ReloadStrategyOptions = {}
 ): void {
   const watcher: Watchpack = (
-      createWatcher({addonId, client, sourceDir, artifactsDir})
-    );
+    createWatcher({addonId, client, sourceDir, artifactsDir})
+  );
 
   firefoxProcess.on('close', () => {
     client.disconnect();
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -182,10 +182,7 @@ export default async function run(
   const requiresRemote = !preInstall;
   let installed = false;
 
-  let runner;
-  let profile;
   let client;
-  let runningFirefox;
   let addonId;
 
   const manifestData = await getValidatedManifest(sourceDir);
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -82,14 +82,15 @@ export function defaultReloadStrategy(
     createWatcher = defaultWatcherCreator,
   }: ReloadStrategyOptions = {}
 ): void {
-  let watcher: Watchpack;
+  const watcher: Watchpack = (
+      createWatcher({addonId, client, sourceDir, artifactsDir})
+    );
 
   firefoxProcess.on('close', () => {
     client.disconnect();
     watcher.close();
   });
 
-  watcher = createWatcher({addonId, client, sourceDir, artifactsDir});
 }
 
 
@@ -187,9 +188,9 @@ export default async function run(
   let runningFirefox;
   let addonId;
 
-  let manifestData = await getValidatedManifest(sourceDir);
+  const manifestData = await getValidatedManifest(sourceDir);
 
-  runner = new ExtensionRunner({
+  const runner = new ExtensionRunner({
     sourceDir,
     firefoxApp,
     firefox,
@@ -198,7 +199,7 @@ export default async function run(
     customPrefs,
   });
 
-  profile = await runner.getProfile();
+  const profile = await runner.getProfile();
 
   if (!preInstall) {
     log.debug('Deferring extension installation until after ' +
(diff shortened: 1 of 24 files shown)


aniketkudale:
Hi @kumar303, sorry for new PR, I messed up in my previous PR so closed that one. Also, the test for commit `6b47da4` was successful on my machine, dont know why it is failing on travis-ci. Once again thanks for taking time for reviewing.

shubheksha:
Hi @aniketkudale, Travis is complaining because the commit message of your second commit doesn't follow format mentioned in the [contributing guide](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages).

aniketkudale:
Ah, thanks. Noted :)

src/cmd/run.js
```diff
@@ -82,14 +82,15 @@ export function defaultReloadStrategy(
     createWatcher = defaultWatcherCreator,
   }: ReloadStrategyOptions = {}
 ): void {
-  let watcher: Watchpack;
+  const watcher: Watchpack = (
+      createWatcher({addonId, client, sourceDir, artifactsDir})
+    );
```
kumar303: This is indented too far. It should look like this:
````javascript
const watcher: Watchpack = (
  createWatcher({addonId, client, sourceDir, artifactsDir})
);
````
This provides a visual indication to the reader that the block is connected to the line opening with `const watcher...`.

**Comment by kumar303** [comment]
In regards to the commit message formatting problem, are you using `git commit --no-verify` ? Because git should not let you commit with a message formatted this way. Be sure all your local dependencies are up to date by running `npm install`.

**Comment by aniketkudale** [comment]
Actually I am using Tortoise SVN tool for commits, so may be this is causing the problem.

**Comment by kumar303** [comment]
Huh, seems like that should work. I see that TortoiseGit has [custom hooks](https://tortoisegit.org/docs/tortoisegit/tgit-dug-settings.html) but I don't see why it would ignore our git hooks.

**Comment by kumar303** [comment]
Let me know if you need help fixing the Flow errors

**Comment by aniketkudale** [comment]
Yes, please! This looks interesting. Neither of cmds, `npm run build` and` npm run lint`, throw any warning on my machine.

**Comment by aniketkudale** [comment]
@kumar303 Will it be ok to use `var`, this doesnt give any warnings/error on `npm run build `and `npm run lint`
```
  for (let pref of cliPrefs) {
    var [key, value] = pref.split('='); // using var here, line 137 (src\firefox\preferences.js)

    if (/[^\w{@}.-]/.test(key)) {
    }

    if (value === `${parseInt(value)}`) {
      value = parseInt(value, 10);
    } else if (value === 'true' || value === 'false') {
      value = (value === 'true');
}
```
```

kumar303:
Hi, thanks for the quick response.

> Neither of cmds, npm run build andnpm run lint, throw any warning on my machine.

You should be using the `npm start` command to develop with. When you get a chance, please read through the development process to get familiar with it: https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md

> Will it be ok to use var?

All variables that do not get modified must be declared with `const`. For any variable that needs to be modified, use [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) because it deals with scoping better than `var`. But, yes, the problem you are running into is that you changed a variable from `let` to `const` but that variable is getting modified so it can't be a constant.

aniketkudale:
oh, so can we write
`const [key value] = pref.split('=');`
as
```
const [key] = pref.split('=');
let [value] = pref.split('=');
```

rpl:
@aniketkudale you should be able to rewrite it as:

```
const prefsAry = pref.split('=');
const [key] = prefsAry;
let [, value] = prefsAry;
```

The comma (`,`) in second line means that you want to skip the first element and assign the second one to value.

Nevertheless, I'm not sure that it is very readable written in this way, I would not be against using the regular syntax in this case, eg.:

```
const prefsAry = pref.split('=');
const key = prefsAry[0];
let value = prefsAry[1];
```

aniketkudale:
Thanks @rpl for reply.
```
const prefsAry = pref.split('=');
const key = prefsAry[0];
let value = prefsAry[1];
```
This looks readable & simple to understand.

aniketkudale:
Thanks a lot @kumar303 for merging my PR.

Issue #131, opened 2016-03-09
Read file-ignore rules from a config for building XPIs


kumar303:
https://github.com/mozilla/web-ext/issues/85 added some simple defaults for ignoring files when building an XPI. Let's look for a config file inside a web extension that the developer can use to declare custom rules.

JPM looked for `.jpmignore` but I think it would be more flexible if we looked for a global `web-ext.config` (or something) and allowed that to set default values for any command line option. We could add an option to `build` like `--ignore-filenames=*.zip,*.foo` so that a config file could declare the same `ignore-filenames` defaults. Yargs [has support for config files](http://yargs.js.org/docs/#methods-configkey-description-parsefn).

See https://github.com/mozilla/web-ext/issues/176 for general config file support.

kumar303:
Let's do the general config file logic first: https://github.com/mozilla/web-ext/issues/176

kumar303:
This seems pretty important. It looks like some people are putting `sign.sh` (which includes API keys) in the root of their project. This will get included in an XPI by default.

lidel:
I have use case that would greatly benefit from adding `--ignore-filenames=*.zip,*.foo` parameter support to both `build` and `lint`.

The add-on I am building relies on "browserified" and "minimized" NPM libraries which need to be included during `build` but produce a lot of warnings when I run `lint`.  
Due to this I often just ignore `lint` output -- too noisy to analyze every time. 

Excluding these 3rd party artifacts from `lint` would improve the situation, as I could focus on removing any warnings from the code I control.

kumar303:
Note to self: when we document this we should link to this cheat sheet for how glob patterns work: https://github.com/isaacs/node-glob#glob-primer

rpl:
@kumar303 additional note: when we document this we should mention that glob patterns (e.g. like `"node_modules/**"`) needs to be quoted to prevent the shell to expand the glob pattern into the entire list of files that verify the glob pattern.

kumar303:
Here is the documentation. @rpl or anyone, let me know if anything is unclear or needs tweaks: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference?document_saved=true#--ignore-files_-i

kumar303:
or @eight04  too! ^

Issue #176, opened 2016-04-06
Set defaults for all options using a config file


kumar303:
If a user has a magically named config file in the current working directory (e.g. `web-ext-config.json`) then `web-ext` should use this to set default option values (and web-ext should log a message about using it). Here are a few other examples of how it should work:

The config can also be specified explicitly:

    web-ext --config=/path/to/web-ext-config.json build ...

The config file should apply to global option values as well as sub-command option values. Here is an example of a config that is the equivalent of `web-ext --verbose sign --api-url-prefix https://addons-dev.allizom.org/api/v3`:

````json
{
  "verbose": true,
  "sign": {
    "api-url-prefix": "https://addons-dev.allizom.org/api/v3"
  }
}
````

The camel case variants of each option should also be supported, such as:
````json
{
  "sign": {
    "apiUrlPrefix": "https://addons-dev.allizom.org/api/v3"
  }
}
````

Some additional rules:
* A command line option will take precedence over any config value
* An environment variable will take precedence over any config value

**Comment by kumar303** [comment]
I have a patch started but it depends on https://github.com/yargs/yargs/issues/464

**Comment by kumar303** [comment]
FWIW, a yaml config might be nicer than JSON because you could put comments in it

**Comment by MarZab** [comment]
I would suggest making a .js file that exports a settings object instead.

How about using a grunt task ie. Gruntfile.js like we can with `jpm`? I feel we are needlessly reinventing the wheel here.

**Comment by kumar303** [comment]
Sure, a `.js` file would be nice because you can add comments. I think yargs might need a patch to support that but maybe not.

**Comment by MarZab** [comment]
Looks like something like this should work:

```
var argv = require('yargs')
  .config('settings', function (configPath) {
    return require(configPath)
  }).argv
```

web-ext.config.json
``` 
module.exports = {
  "sign": {
    "apiUrlPrefix": "https://addons-dev.allizom.org/api/v3"
  }
}
```

kumar303:
oh cool, that looks like it will work. I like using `.js` config files, let's do it.

pdehaan:
You could use something like [**rc**](https://github.com/dominictarr/rc) to use JSON or INI config files.
Or I think ESLint supports config files as:
  - .js (with `module.exports`)
  - .json (with [**strip-json-comments**](https://github.com/sindresorhus/strip-json-comments) to allow for commented JSON)
  - .yml (using [**js-yaml**](https://github.com/nodeca/js-yaml))

[...]

PR #703, opened 2016-12-22
Added support for .xpiignore (at sourceDir) with node-ignore

bluejamesbond:
This pull request is just to see if you are interested. I can refractor and add tests as needed. 

EDIT:
Another module worth considering you think this feature can be adopted - [fsstream-ignore](https://www.npmjs.com/package/fstream-ignore)

Fixes https://github.com/mozilla/web-ext/issues/131


kumar303:
Hi @bluejamesbond , thanks for getting in touch about this!

I would love to help you land this patch but I think it would be better to approach it like this:
* implement it as a command line option, such as `--ignore-files=*.zip,*.xpi`
* stick with `minimatch` unless there is a compelling reason to support the exact `gitignore` matching?

The reason I'm suggesting a command line option instead of a `.web-ext-ignore` file is because we will soon have config parsing support in https://github.com/mozilla/web-ext/issues/176 . After that, the option could be set in a `web-ext-config.js` file like:
````javascript
module.exports = {
  ignoreFiles: [
    '*.zip',
    '*.xpi',
  ]
};
````
Regardless, we definitely need a command line option first because currently the ignore list isn't configurable from the CLI.

**Comment by kumar303** [comment]
Your patch would definitely address https://github.com/mozilla/web-ext/issues/131, thanks for getting it started!

**Comment by Croydon** [comment]
@kumar303 Is the web-ext-config.js file you have in mind a local project file or global? I'm looking for a local .webextignore like .jpmignore

**Comment by kumar303** [comment]
The `web-ext-config.js` could be local (per project) or global in `~/.web-ext-config.js`. Both would be supported.

File: src/cmd/build.js [hunk 1]
```diff
@@ -191,21 +191,47 @@ export default async function build(
...
+    // are defaults required?
+    this._ignore.add(filesToIgnore || [
+      '*.xpi',
+      '*.zip',
+      '.*', // any hidden file
+      'node_modules',
+    ]);
+
+    if (sourceDir) {
+      const ignoreFilePath = path.join(sourceDir, '.xpiignore');
```
wagnerand: Don't want to torpedo this PR, but something like `.webextignore` seems more appropriate to me.

bluejamesbond: I had little free time last month so I couldn't follow up on this PR. It seems the project is opting for a `*.js` file and not using an ignore file I believe.

kumar303:
By the way, I just remembered that we have a newer active patch for this https://github.com/mozilla/web-ext/pull/753 (perhaps the contributor will be able to land it). Thanks again for helping to kick this off though.

bluejamesbond:
@kumar303 My apologies for not catching up. Something other things had come up.

kumar303:
No worries at all! Thanks for your interest in `web-ext`. I'll close this one since https://github.com/mozilla/web-ext/pull/753 is pretty close to landing.

src/cmd/build.js
Croydon: Agreed. In special since .xpi ist not anymore the only supported file ending.

PR #704, opened 2016-12-22
test: Changed changelog linting to check the pull request title

### LINKED ISSUE #701: Make changelog linting more effective

code quality enhancement

When you develop locally or create a PR, you have to follow the [commit message format](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages). If you make a mistake, you'll get a failure locally. If you push a bad commit to your branch, you'll see failures forever unless you rewrite the commit history.

This isn't very effective because ultimately a maintainer will squash and merge the PR to master, choosing a new commit message anyway. There is no linting for this message! As a maintainer myself, I also like to spend time on this final squashed message since it will go into the auto-generated changelog.

We should come up with a more effective linting strategy.

The first action items: 
* remove the local pre-commit hooks
* remove the TravisCI test for changelog linting

Next, we need a new linting approach. Ideas:
* create a bot that watches for new PRs and pings @kumar303 or @rpl in IRC if a title needs fixing. PR titles are used by git to create the squashed commit message.
* create a webextension for maintainers that warns if the text box has an improperly formmated message when doing a squash-on-merge.

rpl:
This pull request contains an attempt to implement a better changelog linting by running conventiona-changelog-lint on the pull request title instead of running it on the the commit messages, especially given that the single commits are going to be squashed when the pull request is merged, while the pull request title is what becomes the new commit message and it wasn't currently linted using conventional-changelog-lint, until it reaches master (which means... too late :-))

Fixes #701


rpl:
@kumar303 This pull request currently contains an "invalid pull request title" on purpose, so that we can test that the related travis build fails, and then we can change the title and force a rebuild to test that it passes when the pull request title follows the conventions correctly.

.travis.yml
```diff
@@ -3,19 +3,14 @@ sudo: false
...
-- git checkout -
 
 before_script:
 # If this command fails and you can't fix it, file an issue and add an exception to .nsprc
 - npm run nsp-check
 
 script:
 - COVERAGE=y NODE_ENV=production npm test
-- npm run changelog-lint
+- npm run travis-pullrequest-title-lint
```
kumar303: How about `github-pr-title-lint` ?

Gruntfile.js
```diff
@@ -51,4 +51,77 @@ module.exports = function(grunt) {
...
+
+        var pullRequestURLPath = '/repos/' + process.env.TRAVIS_REPO_SLUG +
+          /pulls/ + process.env.TRAVIS_PULL_REQUEST + '.json';
+
+        grunt.log.writeln(
+          'Retrieving the pull request title from https://api.github.com' +
+          pullRequestURLPath
+        );
+
+        require('https').get({
```
kumar303: maybe move this require to the top of the file?

rpl:
@kumar303 Here is a link to the expected failure: https://travis-ci.org/mozilla/web-ext/jobs/186182234#L1601-L1634

Gruntfile.js
```diff
@@ -51,4 +51,77 @@ module.exports = function(grunt) {
...
+            body += data;
+          });
+          response.on('error', function(err) {
+            grunt.log.writeln('Failed during pull request title download: ' +
+                              err);
+
+            done(false);
+          });
+          response.on('end', function() {
+            var parsed = JSON.parse(body);
```
kumar303: it would be good to start the promise chain with this wrapped in `new Promise(function(resolve) { ... })` just in case it throws a `SyntaxError`. Otherwise, the grunt task might hang.

kumar303:
It would be nice to get a more readable report of errors, if possible. The standard changelog lint output is already pretty confusing to contributors though. Perhaps this should also include a link to the [formatting guide](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages) ?

rpl:
@kumar303 Here is the new failure messages (formatted as the cli tool does and with a nice message added, which links to the "Writing commit messages" section in the CONTRIBUTING.md docs):

- https://travis-ci.org/mozilla/web-ext/jobs/186214389#L1602-L1618

The name of the task is still `travis-...` because, at least in its current implementation, it is supposed to work correctly only on travis (while locally you have to define the needed environment vars manually to be able to use it).

I opted to move the new task into its own file, it looks like "a good time to move some complex tasks into their own file" can be now after all ;-)

I also changed the task so that it is going to lint the commit message if the branch is composed from a single commit.

Let me know how it looks to you.

Gruntfile.js
```diff
@@ -13,7 +13,7 @@ module.exports = function(grunt) {
 
   var configs = require('load-grunt-configs')(grunt, {
     config: {
-      src: 'tasks/*.js',
+      src: 'tasks/configs/*.js',
```
kumar303: Good idea. The helper comment needs updating too: https://github.com/rpl/web-ext/blob/2e95f3dd3db4df77177c8a459965d61831860425/Gruntfile.js#L8

tasks/travis-pr-title-lint.js
```diff
@@ -0,0 +1,159 @@
...
+          resolve(stdout);
+        }
+      });
+    });
+  }
+
+  function getPullRequestTitle() {
+    return new Promise(function(resolve, reject) {
+      var pullRequestURLPath = `/repos/${ process.env.TRAVIS_REPO_SLUG
+        }${/pulls/ }${process.env.TRAVIS_PULL_REQUEST }.json`;
```
kumar303: Is it ok to use a template string here? I'm surprised this works.

kumar303: nit: I find these easier to read when broken with plus signs (but I know it's not pretty either way):
````javascript
var pullRequestURLPath = '/repos/' +
  `${process.env.TRAVIS_REPO_SLUG}${/pulls/}` +
  `${process.env.TRAVIS_PULL_REQUEST }.json`;
````

Comment (kumar303) [inline tasks/travis-pr-title-lint.js]: is `${/pulls/}` correct?

File: tasks/travis-pr-title-lint.js [hunk 6]
```diff
@@ -0,0 +1,159 @@
...
+          resolve(formatted);
+        } else {
+          reject(formatted);
+        }
+      });
+    });
+  }
+
+  var writeCommitMessagesDocURL =
+        'https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages'; // eslint-disable-line max-len
```
kumar303: nit: we've been doing two spaces for any indent, regardless of the type of block continuation. I know some editors do these breaks with more spaces by default (mine does).

tasks/travis-pr-title-lint.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+        if (!Object.entries) {
+          objectEntries.shim();
+        }
+
+        countGitMergeCommits().then(function(commitsCount) {
+          if (commitsCount === 1) {
+            grunt.log.writeln(
+              'There is only one commit in this pull request, ' +
+                'we are going to check the single commit message...'
```
kumar303: nit: typically we've been indenting breaks like this at the same level since it's not a nested block, it's a continuation of the block on the previous line

tasks/travis-pr-title-lint.js
```diff
@@ -0,0 +1,159 @@
...
+            );
+            return getGitLastCommitMessage().then(lintMessage);
+          } else {
+            grunt.log.writeln(
+              'There is more than one commit in this pull request, ' +
+                'we are going to check the pull request title...'
+            );
+            return getPullRequestTitle().then(lintMessage);
```
kumar303: It looks like `.then(lintMessage)` is the same for both cases so it can be moved down into the promise chain

tasks/travis-pr-title-lint.js
```diff
@@ -0,0 +1,159 @@
...
+          done(true);
+        }).catch(function(err) {
+          var errMessage = err.stack ? err.stack : err;
+          grunt.log.writeln(
+            'Failures during changelog linting the pull request:\n' +
+              errMessage
+          );
+
+          grunt.log.writeln(
+            `${'\nDon\'t panic! If your travis build is failing here, ' +
```
kumar303: I love this message!

tasks/travis-pr-title-lint.js
```diff
@@ -0,0 +1,159 @@
...
+              'please take a look at \n\n' +
+              ' - '}${ writeCommitMessagesDocURL }\n\n` +
+              'and/or mention in a comment one of the mantainers, ' +
+              'we are here to help ;-)'
```
kumar303: I think you could just say: 

Please take a look at [link] and feel free to ask for help from one of the maintainers in a comment; we are here to help ;-)

I'm suggesting that because they may not know what usernames to mention and I think we are pretty good about reading all comments anyway.

tasks/travis-pr-title-lint.js
rpl: ouch, it shouldn't be there, I tried to remove any ES6 syntax to keep it "ES5"-safe :-)
but I forgot a bunch of them (but it looks like template strings are supported in node >= 4).

removed in the updated pull request.

rpl: it shouldn't, and it is very interesting that it worked (and it didn't raise any linting error)
I rewrote all the template strings into strings composition, but I will take another look on this (at least to prevent this kind of weird but working template strings to slip in a patch unnoticed in the future ;-)).

tasks/travis-pr-title-lint.js
rpl: yeah, this will be definitely in a chapter of my book "when your editor auto-indentation feature works against you" :-D

when I moved all the grunt task code into a separate file, the "auto-indentation on copy & paste" feature decided to mess up my manual indentation :-|

tasks/travis-pr-title-lint.js

tasks/travis-pr-title-lint.js
rpl: hehehe, I knew you would like it! 

[Every book should have Don't panic on its cover](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker's_Guide_to_the_Galaxy#Don.27t_Panic) ;-)

tasks/travis-pr-title-lint.js
message tweaked in the last pushed update!

rpl:
@kumar303 The last push build passed, and it shouldn't until we change the pull request title.

do not merge this pull request yet, I'm looking into it

rpl:
@kumar303 I tweaked the grunt task a bit:
- removed more es6 syntaxes (template strings and arrow functions)
- explicitly check and reject on empty message (which should prevent a "wrong" green build like this: https://travis-ci.org/mozilla/web-ext/jobs/186392172#L1117)
- added an explicit "catch & reject" in the "changelog-lint"-related promise chain
- quoted the changelog-lint input (so that it is clear where it starts and where it ends ;-))
- removed git-hooks from the npm dependencies

I've not yet removed the .githooks dir because it seems to me that, on a local dev environment of a contributor where the git hooks are already installed, the raised error message (the one that would be raised if the .githooks dir is missing) can be confusing, and so it is probably better to defer their removal (but in the meantime it is reasonable that we do not auto-install the git hooks anymore).

I played a bit with this branch by rebuilding it on travis multiple times, with and without changing the title,
and it looks to behave correctly, I'm going to do it one more time so that it goes green with a valid pull request title.

Once the travis build is green, it should be ready, let me know how it looks to you.

rpl:
@kumar303 I updated the CONTRIBUTING.md file so that it mentions that the pull requests subject is checked using the same approach used for the commit messages (The new changelog linting task actually checks the commit messages only when there is only one commit in the pull request, but I preferred to keep it simple and don't mention this behavior in the CONTRIBUTING.md change).

rpl:
> Here's one idea about the .githooks. What if you keep the scripts there but make them do nothing? They could also print "This commit hook is no longer needed, you can safely remove it." At the very least, this patch should probably add a comment in each hook script about the change.

:+1: sounds great to me!

I applied it in 63b55a5

PR #706, opened 2016-12-23
feat: Open browser console feature for web-ext run

### LINKED ISSUE #686: web-ext run: open the Browser Console

feature

If you want to develop your extension with the Browser Console open, you have to do this:

- Type: `web-ext run --firefox=nightly`
- Remember that you wanted the Browser Console open
- Manually type ctrl-shift-J when Firefox starts up to open the Browser Console

It would be nice to just do `web-ext run --firefox=nightly --jsconsole`

This should be pretty simple because Firefox already has ` --jsconsole` option we can pass to it.

There are actually a bunch of other options like this that we should expose but that could be done in a separate bug.

aniketkudale:
Fixes #686


aniketkudale:
Hi @kumar303, please review my PR. Thanks

tests/functional/test.cli.run.js
```diff
@@ -9,14 +9,16 @@ import {
 const EXPECTED_MESSAGE = 'Fake Firefox binary executed correctly.';
 
 describe('web-ext run', () => {
-  it('should accept: --no-reload --source-dir SRCDIR --firefox FXPATH',
+  it('should accept: --no-reload' +
+     ' --source-dir SRCDIR --firefox FXPATH --jsconsole',
```
kumar303: Oh yeah, I forgot about this test. I think it mostly only covers the logic of dealing with the Firefox remote debugger but it wouldn't hurt to add in this new option, thanks.

src/cmd/run.js
```diff
@@ -322,6 +322,8 @@ export class ExtensionRunner {
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
     const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefoxBinary: firefox});
+    return firefoxApp.run(
+        profile, {firefoxBinary: firefox, binaryArgs: ['-jsconsole']}
```
kumar303: nit: each block continuation should only be indented two spaces

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +128,18 @@ describe('run', () => {
     });
   });
 
+  it('passes --jsconsole when specified', () => {
```
kumar303: It should just be: `-jsconsole`

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +128,18 @@ describe('run', () => {
...
+    const firefox = '/pretend/path/to/Firefox/firefox-bin';
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+
+    return cmd.run({firefox}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
+      assert.equal(firefoxApp.run.firstCall.args[1].binaryArgs,
+              '-jsconsole');
```
kumar303: when doing a vertical alignment, always line up to the inside of the opening parenthesis, like:
````javascript
      assert.equal(firefoxApp.run.firstCall.args[1].binaryArgs,
                   '-jsconsole');
````

File: tests/unit/test-cmd/test.run.js [hunk 5]
```diff
@@ -36,6 +36,7 @@ describe('run', () => {
       artifactsDir: path.join(sourceDir, 'web-ext-artifacts'),
       sourceDir,
       noReload: true,
+      jsconsole: true,
```
kumar303: This should not be true by default. In fact, you should probably remove it to let the test invoke the default value.

src/program.js
```diff
@@ -298,6 +298,11 @@ Example: $0 --help run.
         type: 'string',
         coerce: coerceCLICustomPreference,
       },
+      'jsconsole': {
+        describe: 'Open web browser console.',
```
kumar303: Let's make it clear that this is a shortcut for something one could do manually. How about this description:

    Open Tools > Web Developer > Browser Console on Firefox startup

src/cmd/run.js
kumar303: You're on the right track but this code will always run Firefox with `-jsconsole`. We only want to run with that arg when the user specifies it as `web-ext run --jsconsole`. 

Because you've already defined `jsconsole` in `program.js`, the [run()](https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L162) function will be invoked like this when someone runs `web-ext run --jsconsole`:

````javascript
run({
  sourceDir: '.', 
  artifactsDir: 'web-ext-artifacts',
  jsconsole: true,
});
````
You need to pass that into `ExtensionRunner.run()` and use it like:
````javascript
const binaryArgs = [];
if (jsconsole) {
  binaryArgs.push('-jsconsole');
}
return firefoxApp.run(
  profile, {firefoxBinary: firefox, binaryArgs}
````

**Comment by kumar303** [comment]
One more thought: I think `--browser-console` would be better than `--jsconsole` since the former differentiates itself from the Web Console (which is for page content).

**Comment by aniketkudale** [comment]
Thanks @kumar303 for quick reply, I will update the changes.

File: src/cmd/run.js [hunk 7]
```diff
@@ -322,6 +322,8 @@ export class ExtensionRunner {
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
```
aniketkudale: Do I need to pass only the parameter into` ExtensionRunner.run()`, like this
`run(profile: FirefoxProfile, jsconsole): Promise<FirefoxProcess> {`

src/cmd/run.js
aniketkudale: Do I need to pass only the `jsconsole: true` parameter into `ExtensionRunner.run()`, like this
`run(profile: FirefoxProfile, jsconsole: true): Promise<FirefoxProcess> {`

src/program.js
aniketkudale: For `type`, when do use `string` and when do we use `boolean`?

src/cmd/run.js
kumar303: yes, that should work

src/program.js
kumar303: this option would be a boolean because it doesn't take a value. It's passed without value like `web-ext run --jsconsole`.

kumar303:
FYI, I may be missing in action next week due to holidays in the US but I'll try to pop in and see how you're doing

src/cmd/run.js
aniketkudale: @kumar303 When I debug this `log.debug(jsconsole)`, I am getting it as undefined.
So this line is not getting executed 
```
if (jsconsole) {
  binaryArgs.push('-jsconsole');
}
```
Do I need to define `jsconsole` anywhere else other than `ExtensionRunner.run()`

aniketkudale: @rpl could you please tell me what I am missing out here?

rpl: @aniketkudale Sure! 

the new "web-ext run" cli option is going to flow through the web-ext sources like:
```
"the new web-ext run cli option" (src/program.js) ->
  "export default async function run" (src/cmd/run.js) ->
    "ExtensionRunner constructor" (src/cmd/run.js) ->
      "ExtensionRunner run method" (src/cmd/run.js)
```

follow some additional details of the changes needed at each of the steps sketched above:

- "the new web-ext run cli option" (src/program.js)
  - the `yargs` type of the option should be a `"boolean"` instead of `"string"` (and, at least in my opinion, it would be probably better to rename it into `--browser-console`, which is how the tool is called in the devtools menu)
- "export default async function run" (src/cmd/run.js) ->
  - the inew option should be added to the `type CmdRunParams` (change needed to pass the flowtype checks), with type `boolean`
  - the new option should be added to the "export default async function run" (change needed to be able to access the new option from the function body), and its default value should be `false`
  - pass the option into the `ExtensionRunner` constructor params
- "ExtensionRunner constructor options" (src/cmd/run.js)
  - the new option should be added to the `type ExtensionRunnerParams` (line 263)
   and to the fields of the `ExtensionRunner` class (line 279), with type `boolean` 
  - the new option should be added to the `ExtensionRunner` constructor params and stored into `this.myVarName` (so that you can use it in the `ExtensionRunner` run method)
 - "`ExtensionRunner` `run` method" (src/cmd/run.js)
   - now you check `this.myVarName` in the `ExtensionRunner` run method (line 331) and
   add `--jsconsole` to the `binaryArgs` if needed (when `this.myVarName` is defined and its value is `true`)

This new cli option will require mostly the same changes needed for #709 (which is still in progress, and I'm going to review asap)

Let me know if the above details are helpful or if something is still missing from the above list.

aniketkudale: @rpl Thanks a lot, this helped.

aniketkudale:
Hi, @kumar303, I have updated the changes, could you please review my PR and please let me know any changes needs to be done. Many thanks.

src/cmd/run.js
```diff
@@ -163,7 +164,7 @@ export default async function run(
   {
     sourceDir, artifactsDir, firefox, firefoxProfile,
     preInstall = false, noReload = false,
-    customPrefs,
+    customPrefs, browserConsole = false,
```
rpl: @aniketkudale do you mind to move the new `browserConsole` property before `customPrefs`?
(and doing the same at line 198)

src/program.js
```diff
@@ -298,6 +298,12 @@ Example: $0 --help run.
         type: 'string',
         coerce: coerceCLICustomPreference,
       },
+      'browser-console': {
+        describe: 'Use --browser-console ' +
```
rpl: The content of `describe` becomes the help message related to the cli option in the `web-ext run --help` output, and so this can be probably just `"Open the DevTools Browser Console."`.

tests/functional/test.cli.run.js
```diff
@@ -9,14 +9,16 @@ import {
 const EXPECTED_MESSAGE = 'Fake Firefox binary executed correctly.';
 
 describe('web-ext run', () => {
-  it('should accept: --no-reload --source-dir SRCDIR --firefox FXPATH',
+  it('should accept: --no-reload' +
```
rpl: @aniketkudale we use the functional test suite only to test that we don't introduce some classes of regressions that are harder to test properly using the unit tests, for this change we can probably use just an additional test in the `tests/unit/test.program.js` file (one additional test similar to this one: https://github.com/mozilla/web-ext/blob/master/tests/unit/test.program.js#L357-L369).

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,18 @@ describe('run', () => {
     });
   });
 
+  it('passes -jsconsole when --browser-console is specified', () => {
+    const firefox = '/pretend/path/to/Firefox/firefox-bin';
```
rpl: this is not needed

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,18 @@ describe('run', () => {
...
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+
+    return cmd.run({firefox} && {browserConsole: true}).then(() => {
```
rpl: this should be just `return cmd.run(`{browserConsole: true}`).then(...)` (and `{firefox} && {browserConsole: true}` is actually exactly equal to just `{browserConsole: true}`, because `{firefox}` is considered [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) and so the expression will be evaluated as the right side of the `... && ...` expression).

aniketkudale:
Hi @rpl thanks for reviewing my PR. I have updated the changes, could you please review the changes, do let me know if any additional tweaks/changes are needed. Many Thanks

aniketkudale:
Hi @rpl / @kumar303 I have updated the changes. Thanks

tests/functional/test.cli.run.js
```diff
@@ -46,4 +46,4 @@ describe('web-ext run', () => {
            }
          });
        }));
-});
+});
```
rpl: Nit, whitespace change (it is probably auto-applied by the editor)

rpl: @aniketkudale :+1: Thanks!

aniketkudale:
@Leif-W, yeah good idea, It would be nice to have an alias for `--browser-console`. I need to discuss this with @kumar303  / @rpl, and come up with a convenient alias name. Thanks

rpl:
@Leif-W @aniketkudale I'm for keeping `--browser-console` has the main name of this cli option, mainly because yargs will turn it into `browserConsole` in the web-ext internals and it will be immediately recognizable without any additional inline comment, but I agree that we can define a shorter alias to make it shorter to type.

I would keep `-c` free, because it will be probably used for the upcoming global `--config` option (which will be also helpful to turn on the `--browser-console` automatically using the user-level config file, e.g. something like `~/.web-ext-rc.js`, if a user wants it enabled by default).

How about `--bc`?
it can be easily added to the current version of this patch by just adding 'bc' to the option aliases, something like:

```diff
diff --git a/src/program.js b/src/program.js
index a46ad9c..5d5d618 100644
--- a/src/program.js
+++ b/src/program.js
@@ -315,6 +315,7 @@ Example: $0 --help run.
         coerce: coerceCLICustomPreference,
       },
       'browser-console': {
+        alias: ['bc'],
         describe: 'Open the DevTools Browser Console.',
         demand: false,
         type: 'boolean',
```

kumar303:
Yeah, adding a `--bc` alias is fine with me

src/program.js
```diff
@@ -314,6 +314,11 @@ Example: $0 --help run.
         type: 'string',
         coerce: coerceCLICustomPreference,
       },
+      'browser-console': {
```
kumar303: as discussed, could you also add a `bc` alias here? : 
````
'browser-console': {
  alias: ['bc'],
  ...
}
````

File: tests/unit/test-cmd/test.run.js [hunk 15]
```diff
@@ -127,6 +127,17 @@ describe('run', () => {
     });
   });
 
+  it('passes -jsconsole when --browser-console is specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+
+    return cmd.run({browserConsole: true}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
```
kumar303: you can use this shortcut: `assert.ok(firefoxApp.run.called)`

Leif-W: I am just an observer but I noticed about 13-14 instances of "assert\.equal\(.*,.*true.*\)", in this file alone.  About ~80 locations in just the web-ext/tests directory.  But assert.ok was used in only 1 location before.  is there likewise an assert.nok or !assert.ok for an assertion of false?  Any benefit to using the shortcut besides shorter line?  Any concern over readability of code?  For example, equals true/false seems clear, where assert.nok or !assert.ok() seems confusing.  If a true must later be changed to a false, or vice-versa.

aniketkudale:
Hi @kumar303 updated the changes, all tests are passing. But, why I am getting warning while running "travis-pr-title-lint" task in travis-ci build?

rpl:
@aniketkudale don't worry about that warning, sometimes the github API is returning an empty message for the PR title, we are still trying to figure out why (it usually passes when the travis job is executed again manually from the travis web ui)

aniketkudale:
Thanks a lot @rpl for quick reply!

rpl:
@aniketkudale :tada: Thanks for this! You have been awesome! :+1:

aniketkudale:
Thanks @kumar / @rpl for the help & merge, you guys are awesome!

Issue #647, opened 2016-11-24
Launch browser at specified page


andymckay:
### Is this a bug or feature request?
Feature

### What is the current behavior?
When you run it starts Firefox with a blank page.

### What is the expected or desired behavior?
Would be awesome if I could define a URL at the command line to open. It could be a page that runs automated tests or just a page that allows me to manually click a few things.

saintsebastian:
@kumar303 @rpl I'd like to work on this

rpl:
@saintsebastian sure! let me know if more context is needed to make this actionable!

gregglind:
my proposal would be a `--url` argument that can be given multiple times, as opposed to a comma separated list.

rpl:
@gregglind :+1: @saintsebastian is working on it on #709 (and it is almost done).

The cli option is going to be called `--start-url`, but (besides the name) it is going to work exactly like you described above (it can be passed multiple times and it is going to open the specified urls into multiple tabs)

gregglind:
1.  Awesome.
2.  Nit:  cli options with internal hyphens... could that just be url?
 (Usual json / python objects with keys with hyphens are annoying.
3.  Would it be worth having a list of all proposed cli stuff somewhere so
it stays coherent?
4.  Ps, I LOVE cli design.

rpl:
@gregglind I would not be strongly against rename it `--url`, the reason why I suggested to name this cli option as `--open-url` or `--start-url` is mostly related to how yargs converts the cli options into javascript properties, because we pass these identifiers around in the web-ext internals and `openUrl`/`startUrl` was more explicit (than `url`) about what the identifier content represents.

Unfortunately it is probably too late to change the general convention on the web-ext cli options, because we have already a number of existent cli options with internal hypens, e.g. `--artifact-dir`, `--source-dir`, `--firefox-profile`, `--no-reload`.

But we can still provide a nicer "single letter alias" for this new cli option, e.g. 
`-u` is currently available

how it sounds to you?

About a list of proposed cli stuff, I agree: keep the cli options consistent is something that we deeply care about and we constantly keep an eye on (we spend so much time on the cli that every small detail matters :-P)

we are currently discussing and enforcing the consistency during/through the code reviews, but I totally agree that we can make it more explicit and keep the convention and the discussions/decisions related to next options planned in a more visible shared place than the single issue/pull request.

My proposal:

- add a new section related to the "cli option naming conventions" into the CONTRIBUTING.md file
[...]

PR #709, opened 2016-12-25
feat: Added an option to launch the browser at a specific url

saintsebastian:
New `web-ext run` argument is added. `--url` just accepts a string and passes it to fx-runner.
As for `binaryArgs` themselves `'--url'` string is really irrelevant and browser would open a page just from a url string just as described [here](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options) even with combinations with other arguments, but just to be safe and to avoid wrong argument attribution I added it. 
There is already a [generic test for binaryArgs](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-firefox/test.firefox.js#L80), since opening browser correctly is delegated to fx-runner, it probably should be tested there. But I will think it over and maybe come up with a better testing strategy. 

Fixes issue #647


src/program.js
```diff
@@ -298,6 +298,12 @@ Example: $0 --help run.
         type: 'string',
         coerce: coerceCLICustomPreference,
       },
+      'url': {
```
rpl: @saintsebastian how about renaming this new CLI option to something like `--open-url` or `--start-url`?

src/cmd/run.js
```diff
@@ -323,7 +328,9 @@ export class ExtensionRunner {
   }
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
-    const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefoxBinary: firefox});
+    const {firefoxApp, firefox, url} = this;
+    const binaryArgsArray = ['--url', url];
```
rpl: it would be better to call it `binaryArgs` and we should add the `url` to the `binaryArgs` only if it is defined.

Also we should keep into account that #706 is going to introduce its own elements in the `binaryArgs` array and we should do it using the same strategy, and as a nice side-effect it will make easier merging/rebasing these two pull request regardless of the landing order, here is where the similar change is applied in #706 https://github.com/aniketkudale/web-ext/blob/4f17f3d11d6602aa31ab00010f58d91284971d5d/src/cmd/run.js#L329-L338)

saintsebastian:
It looks like my commit message is formatted incorrectly. Let me know if I need to add cli.run tests os anything else

src/cmd/run.js
```diff
@@ -276,11 +279,12 @@ export class ExtensionRunner {
   firefoxApp: typeof defaultFirefoxApp;
   firefox: string;
   customPrefs: FirefoxPreferences;
+  startUrl: string;
 
   constructor(
     {
       firefoxApp, sourceDir, manifestData,
-      profilePath, firefox, customPrefs = {},
+      profilePath, firefox, customPrefs = {}, startUrl = '',
```
rpl: @saintsebastian I think that startUrl should not have a default value here.

src/cmd/run.js
```diff
@@ -323,7 +328,12 @@ export class ExtensionRunner {
   }
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
+    const binaryArgs = [];
     const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefoxBinary: firefox});
+    if (this.startUrl) {
+      binaryArgs.push('--url', this.startUrl);
+    }
+    return firefoxApp.run(profile,
```
rpl: If this line is too long we can format it like:
```
return firefoxApp.run(profile, {
  firefoxBinary: firefox, binaryArgs,
});
```

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,19 @@ describe('run', () => {
     });
   });
 
+  it('passes url to Firefox when specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+
+    return cmd.run({startUrl: 'www.example.com'}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
+      assert.equal(firefoxApp.run.firstCall.args[1].binaryArgs[0],
```
rpl: we can probably use [`assert.deepEqual`](http://chaijs.com/api/assert/#method_deepequal) to test `binaryArgs` with a nicer single assertion.

src/cmd/run.js
```diff
@@ -276,11 +279,12 @@ export class ExtensionRunner {
   firefoxApp: typeof defaultFirefoxApp;
   firefox: string;
   customPrefs: FirefoxPreferences;
+  startUrl: ?string = null;
```
rpl: No need to pre-set the `startUrl` property value to `null` here. 

But besides that, Good Job! The type annotation `?string` is definitely the right one, because `startUrl` can be missing (and it will be `undefined`) and by using this annotation we are saying to flow that the `startUrl` property is "maybe a string"!

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,34 @@ describe('run', () => {
...
+      assert.equal(firefoxApp.run.called, true);
+      assert.deepEqual(firefoxApp.run.firstCall.args[1].binaryArgs,
+                   testArgs);
+    });
+  });
+
+  it('passes multiple url parameters to Firefox when specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+    const testArgs = [
```
rpl: nit, let's call this `expectedBinaryArgs`.

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,34 @@ describe('run', () => {
...
+      '--url', 'www.one.com', '--url', 'www.two.com', '--url', 'www.three.com',
+    ];
+
+    return cmd.run({startUrl: [
+      'www.one.com', 'www.two.com', 'www.three.com',
+    ]}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
+      assert.deepEqual(firefoxApp.run.firstCall.args[1].binaryArgs,
+                   testArgs);
```
rpl: nit, align this line to the first letter of the first function parameter on the previous line.

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,34 @@ describe('run', () => {
...
 
+  it('passes single url parameter to Firefox when specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+    const testArgs = ['--url', 'www.example.com'];
+
+    return cmd.run({startUrl: 'www.example.com'}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
+      assert.deepEqual(firefoxApp.run.firstCall.args[1].binaryArgs,
+                   testArgs);
```
rpl: nit, align this line to the first letter of the first function parameter on the previous line.

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,34 @@ describe('run', () => {
     });
   });
 
+  it('passes single url parameter to Firefox when specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+    const testArgs = ['--url', 'www.example.com'];
```
rpl: nit, let's call this `expectedBinaryArgs`.

src/cmd/run.js
```diff
@@ -323,7 +328,16 @@ export class ExtensionRunner {
   }
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
-    const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefoxBinary: firefox});
+    const binaryArgs = [];
+    const {firefoxApp, firefox, startUrl} = this;
+    if (startUrl) {
+      let urls = Array.isArray(startUrl) ? startUrl : [startUrl];
```
rpl: let -> const (this raise an eslint error on travis: https://travis-ci.org/mozilla/web-ext/jobs/188307450#L1217)

src/cmd/run.js
```diff
@@ -323,7 +328,16 @@ export class ExtensionRunner {
...
+      for (let url of urls) {
```
rpl: `for (const url of urls) {...}` 
(This is also detected as a lint error on travis: https://travis-ci.org/mozilla/web-ext/jobs/188307450#L1218)

saintsebastian:
thanks for all your comments @rpl!

src/program.js
```diff
@@ -298,6 +298,12 @@ Example: $0 --help run.
         type: 'string',
         coerce: coerceCLICustomPreference,
       },
+      'start-url': {
+        describe: 'Launch firefox at specified page',
```
rpl: @saintsebastian do you mind to add some alias here, e.g. based on what we discussed with @gregglind in #647:

```
'start-url: {
  alias: ['u', 'url'],
  ...
}
```

so that we can support a shorter cli syntax (helpful especially if the user wants to pass more than one url), while we keep the more descriptive name internally (which helps to make the internals more readable).

saintsebastian:
@gregglind that was a great suggestion, and it is hopefully going to be documented and integrated very soon.
@rpl thanks once again for your help with this feature!

rpl:
@kumar303 This looks like it is ready for you now ;-)

@saintsebastian Thanks for working on it!

src/cmd/run.js
```diff
@@ -323,7 +328,16 @@ export class ExtensionRunner {
...
 
   run(profile: FirefoxProfile): Promise<FirefoxProcess> {
-    const {firefoxApp, firefox} = this;
-    return firefoxApp.run(profile, {firefoxBinary: firefox});
+    const binaryArgs = [];
+    const {firefoxApp, firefox, startUrl} = this;
+    if (startUrl) {
+      const urls = Array.isArray(startUrl) ? startUrl : [startUrl];
+      for (const url of urls) {
+        binaryArgs.push('--url', `${url}`);
```
kumar303: nit: is `${url}` necessary? I think it can just be `url` (without back ticks) because Flow will guarantee that it is a string.

tests/unit/test-cmd/test.run.js
```diff
@@ -127,6 +127,34 @@ describe('run', () => {
     });
   });
 
+  it('passes single url parameter to Firefox when specified', () => {
+    const cmd = prepareRun();
+    const {firefoxApp} = cmd.options;
+    const expectedBinaryArgs = ['--url', 'www.example.com'];
+
+    return cmd.run({startUrl: 'www.example.com'}).then(() => {
+      assert.equal(firefoxApp.run.called, true);
```
kumar303: there is a shortcut for this which is a bit more concise: `assert.ok(firefoxApp.run.called)`

tests/unit/test.program.js
```diff
@@ -326,6 +326,20 @@ describe('program.main', () => {
       });
   });
 
+  it('passes the url of a firefox binary when specified', () => {
+    const fakeCommands = fake(commands, {
+      run: () => Promise.resolve(),
+    });
+    return execProgram(
+      ['run', '--start-url', 'www.example.com'],
```
kumar303: Thanks, this extra test will be helpful. However, I think it will become redundant once we do https://github.com/mozilla/web-ext/issues/377

saintsebastian:
@kumar303 thanks for reviewing the code! I don't feel comfortable deleting the test while there is no substitute yet. But if you think it has to go -- let me know.

rpl:
@saintsebastian :tada: I'm so happy that we got this new feature!  Thank you so much for it!

saintsebastian:
Thanks everyone for your help, this was the feature I felt like I really needed in web-ext.

saintsebastian:
@kumar303 @rpl I believe the new feature should be documented before this issue is closed, should I get on with it?

kumar303:
@saintsebastian ah, looks like this didn't auto-close. I closed the issue since the patch had landed. As for documentation, what we've been doing is looking for [all closed issues that have "needs user docs"](https://github.com/mozilla/web-ext/issues?utf8=%E2%9C%93&q=is%3Aclosed%20label%3A%22needs%20user%20docs%22) but only after the feature actually gets released (which it hasn't yet).

saintsebastian:
@kumar303 thanks for clarification!

PR #710, opened 2016-12-26
test: Changed travis config to run tests with both npm and yarn

rpl:
This PR adds a `yarn.lock` file and it changes the travis config to run the tests using both `yarn` and `npm`.

Pros:
- by including the `yarn.lock` we are able to speed up the tests on nodejs 4 by ~50%
- by using yarn is that the yarn cache can be more safely cached and potentially speed up the builds even more.

Cons: 
- we still have to test that building web-ext using npm works as expected, mostly because npm is the package manager used by default in a classic nodejs environment.
- the `yarn.lock` file needs to be updated every time we update the dependencies (e.g. the greenkeeper pull requests and in general any pull request that adds, removes or changes the package.json dependencies).

Besides being able to get a faster travis build (and a faster failure, if a pull request breaks some tests) on nodejs 4, it would be nice to start test that web-ext can be built using yarn
(e.g. this PR is based on #704 because the git-hooks package postinstall scripts currently seems to break `yarn install`, and #704 is going to get rid of the git-hooks and so it is going to make web-ext buildable using `yarn`). 

I'm opening this PR to start to evaluate if/when we want to introduce the yarn support to our project, and evaluate/discuss how to handle the need to update `yarn.lock` file.

- [ ] add yarn check step
- [ ] add fast_finish


rpl:
@kumar303 The only actual commit that is part of this PR is the last one (76cba5d), the other commits are part of #704

.travis.yml
```diff
@@ -4,6 +4,14 @@ node_js:
 - '4'
 - '6'
 
+# Run on travis with both yarn enabled and disabled.
+env:
+- YARNPKG=true
+- YARNPKG=false
+before_install:
+- if [[ $YARNPKG != "true" ]]; then rm yarn.lock; fi
```
kumar303: doesn't this need an else branch that does `git checkout .` to restore the lock file? I don't know exactly how the matrix build runs but I think it multiplies `env` by each `node_js` version, right? That would mean Node 4 tests would get a yarn.lock file but Node 6 tests would not.

rpl: This is turned into a matrix 2x2, and so we will have 4 builds:

- nodejs 4 without yarn
- nodejs 4 with yarn
- etc.

like in the build associated to this PR: https://travis-ci.org/mozilla/web-ext/builds/188911714

and so we don't need the `else` branch, because we only remove the lock file when yarn is not enabled, so that travis does not "auto-install and enable yarn" as the needed installer and proceed to use the regular npm command as usually

kumar303:
Do we need `fast_finish: true` to get quick feedback? https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

Overall, this approach looks ok but I'm a little hesitant to double the matrix of which tests are run because it means the suite will take longer to complete. I guess the one thing yarn buys us is a fast failure? Perhaps it's necessary to double everything, I'll have to think on it a bit more.

rpl:
@kumar303 it shouldn't increase the time of the Travis builds because they run in parallel (on the contrary we could split the functional test so that the builds became faster)

rpl:
`fast_finish` sounds great to me! I didn't know about it

kumar303:
Oh, they run in parallel? That's nice. I was thinking more about how TravisCI commonly gets overloaded and here we would be throwing more work at it. But let's try it out and see!

As for `fast_finish`, it might already be the default behavior, I'm not sure.

kumar303:
can we merge this?

rpl:
Closing, as this PR is obsolete and has conflicts, and also because I'm really not sure that we want to do this anymore (especially given that we have a package-lock.json file now, I'm not thrilled by the idea of managing two different lock files for the npm dependencies :-))

PR #711, opened 2016-12-26
fix: issue 510 new error type

### LINKED ISSUE #510: web-ext build with a non-existent path to artifacts directory results in an error
When running the following command:

```
web-ext build -a "/non-existant-directory"
```

You get this error:

```
build: Error code: EACCES
```

Instead, you should get this error:

```
build: --artifacts-dir /non-existant-directory is not a directory
```

saintsebastian:
Fixes #510 
This is not exactly what issue is describing, but I think this is understandable for a user. 
`--artifacts-dir /non-existant-directory is not a directory` is only thrown when path leads to a file and not a directory. If access is missing `--artifacts-dir /non-existant-directory lack permissions` is thrown (i am open to better phrasing suggestions). This way `mkdir` is attempted as expected, but user receives clear `UsageError` if it fails

### DIFF
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -18,8 +18,14 @@ export async function prepareArtifactsDir(
     }
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+      await fs.mkdir(artifactsDir, (error) => {
+        if (error) {
+          throw new UsageError(
+            `--artifacts-dir=${artifactsDir} (lack permissions)`);
+        } else {
+          log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        }
+      });
     } else {
       throw error;
     }


src/util/artifacts.js
```diff
@@ -18,8 +18,14 @@ export async function prepareArtifactsDir(
     }
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+      await fs.mkdir(artifactsDir, (error) => {
```
rpl: @saintsebastian I didn't tried to run this PR locally, but there are some additional informations that can be useful to have a picture of what we have to do here (and which part of this first experiment are not going to work exactly as you may expect at a first glance):

- `fs` is not the regular nodejs `fs` module, but the correspondent one wrapped with promises by the `mz` npm dependency (https://www.npmjs.com/package/mz)

- `await` is an ES7 feature, it "suspends" the execution of an `async` function until the passed promise has been resolved/rejected (if the promise is rejected, then an exception is raised, and we can catch it like we are doing on lines 14-19) (you may find this article useful https://pouchdb.com/2015/03/05/taming-the-async-beast-with-es7.html)

- the promise returned by `fs.mkdir` can be rejected for a bunch of different reasons (e.g. there are no more inodes available, or the file already exists but is not a directory, or the filesystem is read-only, or you don't have the permissions, for a more comprehensive list of the possible errors you can look at the list at http://man7.org/linux/man-pages/man2/mkdir.2.html), and so to be able to raise a details error message we should check its error code (e.g. something like at line 20, the underlying mechanisms is not explained in great detail in the nodejs docs but if you search for ".code" in the https://nodejs.org/api/fs.html api docs you can find some useful information and examples) and then raise an `UsageError` with an appropriate message for the one that we are interesting in, and also be sure that we re-raise the original error for the remaining ones (e.g. like we do at line 30) or we are going to "hide" different issues under the wrong error message.

So, long story short, this should be (in pseudo-code) something like:

```
...
try {
  await fs.mkdir(artifactsDir);
} catch (mkdirErr) {
  if (isErrorWithCode('ANERRORCODE', mkdirErr)) {
    throw new UsageError(...);
  } else {
    throw mkdirErr;
  }
}
...
```

or if we want to handle more than one scenario using a `switch/case`, something like:
```
...
} catch (mkdirErr) {
  switch (mkdirErr.code)) {
    case "ERRCODE1": throw new UsageError(...);
    case "ERRCODE2": throw new UsageError(...);
    default: throw mkdirErr;
  }
}
...
```

Let me know if these additional details are helpful or if there is still something missing to make the context of this issue as much detailed as needed.

PR #712, opened 2016-12-28
fix: Logged better messages on errors during artifact dir creation

### LINKED ISSUE #510: web-ext build with a non-existent path to artifacts directory results in an error

saintsebastian:
Fixes #510

@rpl I've restarted this PR because of the wrong branch name with different work. Hope it is not too much troubles. I applied your code and went for if for now. There is still a question of testing this. Is there a good way to simulate this error?

### DIFF
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -1,14 +1,19 @@
 /* @flow */
-import {fs} from 'mz';
+import {fs as defaultFS} from 'mz';
 
 import {UsageError, isErrorWithCode} from '../errors';
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
 
+// type prepareArtifactsDirParamsv = {
+//   artifactsDir: string,
+//   fs?: typeof defaultFS,
+// };
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
+  fs?: typeof defaultFS = defaultFS,
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
@@ -36,7 +41,6 @@ export async function prepareArtifactsDir(
       throw error;
     }
   }
-
 
   return artifactsDir;
 }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -1,5 +1,5 @@
 /* @flow */
-import {fs as defaultFS} from 'mz';
+import {fs} from 'mz';
 import mkdirp from 'mkdirp';
 import promisify from 'es6-promisify';
 
@@ -7,16 +7,13 @@ import {UsageError, isErrorWithCode} from '../errors';
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-const promisifiedMkdirp = promisify(mkdirp);
+const asyncMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
   artifactsDir: string,
-  fs?: typeof defaultFS = defaultFS,
-  mkdir?: typeof promisifiedMkdirp = promisifiedMkdirp,
 ): Promise<string> {
   try {
-    const stats = await defaultFS.stat(artifactsDir);
-    // Check that path is a directory.
+    const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
       throw new UsageError(
         `"${artifactsDir}" exists and it is not a directory.`);
@@ -31,18 +28,18 @@ export async function prepareArtifactsDir(
       }
     }
   } catch (error) {
-    // Handle errors when the artifactsDir cannot be accessed.
     if (isErrorWithCode('EACCES', error)) {
+      // Handle errors when the artifactsDir cannot be accessed.
       throw new UsageError(
         `Cannot access "${artifactsDir}", user lacks permissions.`);
-    // Handle errors when the artifactsDir doesn't exists yet.
     } else if (isErrorWithCode('ENOENT', error)) {
+      // Handle errors when the artifactsDir doesn't exists yet.
       try {
         log.debug(`Creating artifacts directory: ${artifactsDir}`);
-        await mkdir(artifactsDir);
+        await asyncMkdirp(artifactsDir);
       } catch (mkdirErr) {
-        // Handle errors when the artifactsDir cannot be created for lack of permissions.
         if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
           throw new UsageError(
             `Cannot create "${artifactsDir}", user lacks permissions.`);
         } else {
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -18,8 +18,14 @@ export async function prepareArtifactsDir(
     }
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+      await fs.mkdir(artifactsDir, (error) => {
+        if (error) {
+          throw new UsageError(
+            `--artifacts-dir=${artifactsDir} (lack permissions)`);
+        } else {
+          log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        }
+      });
     } else {
       throw error;
     }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -1,42 +1,54 @@
 /* @flow */
 import {fs as defaultFS} from 'mz';
+import mkdirp from 'mkdirp';
+import promisify from 'es6-promisify';
 
 import {UsageError, isErrorWithCode} from '../errors';
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-
-// type prepareArtifactsDirParamsv = {
-//   artifactsDir: string,
-//   fs?: typeof defaultFS,
-// };
+const promisifiedMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
   artifactsDir: string,
   fs?: typeof defaultFS = defaultFS,
+  mkdir?: typeof promisifiedMkdirp = promisifiedMkdirp,
 ): Promise<string> {
   try {
-    const stats = await fs.stat(artifactsDir);
+    const stats = await defaultFS.stat(artifactsDir);
+    // Check that path is a directory.
     if (!stats.isDirectory()) {
       throw new UsageError(
-        `--artifacts-dir=${artifactsDir} (this value is not a directory)`);
+        `"${artifactsDir}" exists and it is not a directory.`);
+    }
+    // Check that we can write in the artifacts dir.
+    try {
+      await fs.access(artifactsDir, fs.W_OK);
+    } catch (accessErr) {
+      if (isErrorWithCode('EACCES', accessErr)) {
+        throw new UsageError(
+          `"${artifactsDir}" exists but the user lacks permissions on it.`);
+      }
     }
   } catch (error) {
-    if (isErrorWithCode('ENOENT', error)) {
+    // Handle errors when the artifactsDir cannot be accessed.
+    if (isErrorWithCode('EACCES', error)) {
+      throw new UsageError(
+        `Cannot access "${artifactsDir}", user lacks permissions.`);
+    // Handle errors when the artifactsDir doesn't exists yet.
+    } else if (isErrorWithCode('ENOENT', error)) {
       try {
         log.debug(`Creating artifacts directory: ${artifactsDir}`);
-        await fs.mkdir(artifactsDir);
+        await mkdir(artifactsDir);
       } catch (mkdirErr) {
+        // Handle errors when the artifactsDir cannot be created for lack of permissions.
         if (isErrorWithCode('EACCES', mkdirErr)) {
           throw new UsageError(
-            `--artifacts-dir=${artifactsDir} (lack permissions)`);
+            `Cannot create "${artifactsDir}", user lacks permissions.`);
         } else {
           throw mkdirErr;
         }
       }
-    } else if (isErrorWithCode('EACCES', error)) {
-      throw new UsageError(
-        `--artifacts-dir=${artifactsDir} (lack permissions)`);
     } else {
       throw error;
     }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -29,10 +29,14 @@ export async function prepareArtifactsDir(
           throw mkdirErr;
         }
       }
+    } else if (isErrorWithCode('EACCES', error)) {
+      throw new UsageError(
+        `--artifacts-dir=${artifactsDir} (lack permissions)`);
     } else {
       throw error;
     }
   }
+
 
   return artifactsDir;
 }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -18,14 +18,17 @@ export async function prepareArtifactsDir(
     }
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
-      await fs.mkdir(artifactsDir, (error) => {
-        if (error) {
+      try {
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await fs.mkdir(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
           throw new UsageError(
             `--artifacts-dir=${artifactsDir} (lack permissions)`);
         } else {
-          log.debug(`Creating artifacts directory: ${artifactsDir}`);
+          throw mkdirErr;
         }
-      });
+      }
     } else {
       throw error;
     }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -35,7 +35,7 @@ export async function prepareArtifactsDir(
       // Handle errors when the artifactsDir cannot be accessed.
       throw new UsageError(
         `Cannot access --artifacts-dir="${artifactsDir}" because the user ` +
-        'lacks permissions.');
+        `lacks permissions: ${error}`);
     } else if (isErrorWithCode('ENOENT', error)) {
       // Create the artifact dir if it doesn't exist yet.
       try {
@@ -46,7 +46,7 @@ export async function prepareArtifactsDir(
           // Handle errors when the artifactsDir cannot be created for lack of permissions.
           throw new UsageError(
             `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
-            'user lacks permissions.');
+            `user lacks permissions: ${mkdirErr}`);
         } else {
           throw mkdirErr;
         }
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -16,7 +16,7 @@ export async function prepareArtifactsDir(
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
       throw new UsageError(
-        `"${artifactsDir}" exists and it is not a directory.`);
+        `--artifacts-dir="${artifactsDir}" exists but it is not a directory.`);
     }
     // If the artifactsDir already exists, check that we have the write permissions on it.
     try {
@@ -24,7 +24,8 @@ export async function prepareArtifactsDir(
     } catch (accessErr) {
       if (isErrorWithCode('EACCES', accessErr)) {
         throw new UsageError(
-          `"${artifactsDir}" exists but the user lacks permissions on it.`);
+          `--artifacts-dir="${artifactsDir}" exists but the user lacks ` +
+          'permissions on it.');
       } else {
         throw accessErr;
       }
@@ -33,7 +34,8 @@ export async function prepareArtifactsDir(
     if (isErrorWithCode('EACCES', error)) {
       // Handle errors when the artifactsDir cannot be accessed.
       throw new UsageError(
-        `Cannot access "${artifactsDir}", user lacks permissions.`);
+        `Cannot access --artifacts-dir="${artifactsDir}" because the user ` +
+        'lacks permissions.');
     } else if (isErrorWithCode('ENOENT', error)) {
       // Create the artifact dir if it doesn't exist yet.
       try {
@@ -43,7 +45,8 @@ export async function prepareArtifactsDir(
         if (isErrorWithCode('EACCES', mkdirErr)) {
           // Handle errors when the artifactsDir cannot be created for lack of permissions.
           throw new UsageError(
-            `Cannot create "${artifactsDir}", user lacks permissions.`);
+            `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
+            'user lacks permissions.');
         }
       }
     } else {
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -15,7 +15,7 @@ type PrepareArtifactsDirOptions = {
 
 export async function prepareArtifactsDir(
   artifactsDir: string,
-  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions,
+  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions = {},
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -7,10 +7,11 @@ import {UsageError, isErrorWithCode} from '../errors';
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-const asyncMkdirp = promisify(mkdirp);
+const defaultAsyncMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
+  asyncMkdirp?: typeof defaultAsyncMkdirp = defaultAsyncMkdirp,
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -47,6 +47,8 @@ export async function prepareArtifactsDir(
           throw new UsageError(
             `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
             'user lacks permissions.');
+        } else {
+          throw mkdirErr;
         }
       }
     } else {
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -10,7 +10,7 @@ const log = createLogger(__filename);
 const asyncMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
-  artifactsDir: string,
+  artifactsDir: string
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -9,9 +9,13 @@ import {createLogger} from './logger';
 const log = createLogger(__filename);
 const defaultAsyncMkdirp = promisify(mkdirp);
 
+type PrepareArtifactsDirOptions = {
+  asyncMkdirp?: typeof defaultAsyncMkdirp,
+}
+
 export async function prepareArtifactsDir(
   artifactsDir: string,
-  asyncMkdirp?: typeof defaultAsyncMkdirp = defaultAsyncMkdirp,
+  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions,
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
--- a/src/util/artifacts.js
+++ b/src/util/artifacts.js
@@ -18,13 +18,15 @@ export async function prepareArtifactsDir(
       throw new UsageError(
         `"${artifactsDir}" exists and it is not a directory.`);
     }
-    // Check that we can write in the artifacts dir.
+    // If the artifactsDir already exists, check that we have the write permissions on it.
     try {
       await fs.access(artifactsDir, fs.W_OK);
     } catch (accessErr) {
       if (isErrorWithCode('EACCES', accessErr)) {
         throw new UsageError(
           `"${artifactsDir}" exists but the user lacks permissions on it.`);
+      } else {
+        throw accessErr;
       }
     }
   } catch (error) {
@@ -33,7 +35,7 @@ export async function prepareArtifactsDir(
       throw new UsageError(
         `Cannot access "${artifactsDir}", user lacks permissions.`);
     } else if (isErrorWithCode('ENOENT', error)) {
-      // Handle errors when the artifactsDir doesn't exists yet.
+      // Create the artifact dir if it doesn't exist yet.
       try {
         log.debug(`Creating artifacts directory: ${artifactsDir}`);
         await asyncMkdirp(artifactsDir);
@@ -42,8 +44,6 @@ export async function prepareArtifactsDir(
           // Handle errors when the artifactsDir cannot be created for lack of permissions.
           throw new UsageError(
             `Cannot create "${artifactsDir}", user lacks permissions.`);
-        } else {
-          throw mkdirErr;
         }
       }
     } else {
--- a/tests/unit/test-util/test.artifacts.js
+++ b/tests/unit/test-util/test.artifacts.js
@@ -9,7 +9,7 @@ import {fs} from 'mz';
 import {onlyInstancesOf, UsageError} from '../../../src/errors';
 import {withTempDir} from '../../../src/util/temp-dir';
 import {prepareArtifactsDir} from '../../../src/util/artifacts';
-import {makeSureItFails} from '../helpers';
+import {makeSureItFails, ErrorWithCode} from '../helpers';
 
 
 describe('prepareArtifactsDir', () => {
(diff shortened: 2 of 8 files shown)


saintsebastian:
I found a [fs-mock module](https://www.npmjs.com/package/fs-mock) which can simulate those errors and will test how it works.

rpl:
@saintsebastian unfortunately the `"fs-mock"` module doesn't seem to be maintained anymore, and it is not enough on its own, because we need to override the real one to make it work as you expect.

By the way, in other tests we tests similar scenarios using a temporary directory that is automatically removed after the test is completed, e.g. like this test case https://github.com/mozilla/web-ext/blob/master/tests/unit/test.watcher.js#L15

The helper is called `withTempDir` and it is defined here: https://github.com/mozilla/web-ext/blob/master/src/util/temp-dir.js#L25

and so I would suggest you to create a new test case that:

- uses the `withTempDir` helper to create a temporary directory that is automatically removed when the tests is done
- in the callback passed to the `withTempDir`, you prepare the environment that you need to test the issue (e.g. you create a regular file with the same name of the requested artifact dir path)
- and then you run the component that we are testing and make the needed assertions on the results

If this is not enough to test the needed scenarios we can evaluate other options (e.g. using something like [rewrite](https://github.com/jhnns/rewire) to override the external module that we want to mock).

saintsebastian:
@rpl the temporary directory generated by `withTempDir` is accessible for writing and therefore doesn't tigger the error. Am I missing something?

rpl:
@saintsebastian can you create a subdir that is not accessible for writing inside the directory that withTempDir creates and use it as the artifact dir that should raise the error?

saintsebastian:
@rpl many thanks for helping me with this one and finding such simple and elegant solutions

saintsebastian:
I added one more test and reused an `errorWithCode` class. Since it is used in a few files maybe it should be exportable?

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +56,38 @@ describe('prepareArtifactsDir', () => {
...
+      class ErrorWithCode extends Error {
+        code: string;
+        constructor() {
+          super('some error');
+          this.code = 'EACCES';
+        }
+      }
+
+      const tmpPath = path.join(tmpDir.path(), 'build');
+      fs.mkdir = sinon.spy(() => Promise.reject(new ErrorWithCode()));
```
rpl: This is risky, if we change this module function and the spy is actually accessed by other modules, we are going to potentially affect other tests, and it is going to potentially generate errors that can be very hard to be debugged (because a fake installed by a test can generate a failure in another test executed after it).

If we want to do this kind of tests (and it is actually a good idea, even better than creating the actually files and dirs on the filesystem), we should use the pattern that we already use for the other similar scenarios:

- the tested component (e.g. the `prepareArtifactsDir` is this case) will take an additional parameter which contains the dependencies (the `fs` module in this case), with the related properties defaulted to the default implementation (the `fs` property imported from the `"mz"` npm module in this case)
- in the real code, the additional parameter is not passed (and the defaults are used)
- in the test code, we pass the additional parameter to inject the fake implementation that behaves as needed by the test case

(e.g. basically the same pattern that you are using for #674)

saintsebastian: @rpl It is indeed not the safest way, thanks for pointing it out. I wonder if it make sense to rewrite the first test accordingly?

src/util/artifacts.js
```diff
@@ -18,8 +23,20 @@ export async function prepareArtifactsDir(
     }
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+      try {
```
rpl: I tested this function a bit locally, especially to:
- verify that the main scenarios are covered
- that the error message raised is nice and clear
- that all these try/catch blocks are needed

I think that it is pretty near to what we want it to be, but I'd like to propose some further tweaks on it, e.g.:
- tweak the error message to be more readable
- reorder the `if/if else/else` inside the `try/catch`, so that the `EACCES` raised from `fs.stat` is the first one, then the `ENOENT` (also raised from fs.stat) and finally `EACCES` error that can be raised from the `fs.mkdir`
- put an inline comment in this `isErrorWithCode` if branches (so that we know which scenario anyone of them is going to handle, e.g.

```
  try {
    const stats = await fs.stat(artifactsDir);
    if (!stats.isDirectory()) {
      throw new UsageError(
        `"${artifactsDir}" exists and it is not a directory.`);
    }
  } catch (error) {
    if (isErrorWithCode('EACCES', error)) {
      // Handle errors when the artifactsDir cannot be accessed.
      throw new UsageError(
        `"Cannot access ${artifactsDir}", user lacks permissions.`);
    } else if (isErrorWithCode('ENOENT', error)) {
      // Handle errors when the artifactsDir doesn't exists yet.
      try {
        log.debug(`Creating artifacts directory: ${artifactsDir}`);
        await fs.mkdir(artifactsDir);
      } catch (mkdirErr) {
        if (isErrorWithCode('EACCES', mkdirErr)) {
          // Handle errors when the artifactsDir cannot be created for lack of permissions.
          throw new UsageError(
            `Cannot create "${artifactsDir}", user lacks permissions.`);
        } else {
          throw mkdirErr;
        }
      }
    } else {
      throw error;
    }
  }
```

also, it would be nice to use something like [node-mkdirp](https://github.com/substack/node-mkdirp) instead of `fs.mkdir` so that we create the parent dirs if they don't exists yet (NOTE: `mkdirp` uses the usual nodejs callback style, and so we need to wrap it into a promise to be able to use it with `await`)

Finally, if I'm not wrong, `fs.stat` is going to raise an error if we cannot access the artifacts dir, but if the artifacts dir already exists and we cannot create any file in it (for lack of permissions), stat is not going to complain, as a workaround we can probably use `fs.access` to check that we can actually create a file in it, e.g. something like:

```
  try {
    const stats = await fs.stat(artifactsDir);
    if (!stats.isDirectory()) {
      throw new UsageError(
        `"${artifactsDir}" exists and it is not a directory.`);
    }

    // Check that we can write in the artifacts dir.
    try {
      await fs.access(path.join(artifactsDir, "someRandomFileNameHere"), fs.W_OK);
    } catch (error) {
      if (isErrorWithCode('EACCES', error)) {
        throw new UsageError(
          `"${artifactsDir}" exists but the user lacks permissions on it.`);
      }
    }
  } catch (error) {
```

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +56,43 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws error when checking the directory if lacks writing permissions',
+    () => withTempDir(
+    (tmpDir) => {
+      const tmpPath = path.join(tmpDir.path(), 'build');
+      return fs.mkdir(tmpPath, '0622').then(() => {
```
rpl: @kumar303 to test this change we have two possible strategies (which are already used in the project):

- recreate the scenario on the real filesystem like in this test (in a temp dir that is automatically removed after the test)
- inject the dependency (`fs` in this case) and use sinon to fake all the error scenarios without recreating anything on disk

I would be ok (e.g. test everything using sinon) to use just one of the two approaches, because using both is probably going to be confusing.

The scenarios to be tested are:

- the artifact dir exists but it is not a directory
- the artifact dir exists but it is a file
- the artifact dir exists but we lack the permissions to write files in it
- the artifact dir do not exist, but we lack the permission to create it

which one of the two approach do you prefer?

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +56,43 @@ describe('prepareArtifactsDir', () => {
...
+            assert.match(error.message, /lack permissions/);
+          }));
+      });
+    }
+  ));
+
+  it('throws error when creating a folder if lacks writing permissions',
+    () => withTempDir(
+    (tmpDir) => {
+      class ErrorWithCode extends Error {
```
rpl: If we decide to use this approach, it would be nice to move this definition of `ErrorWithCode` in a file shared with the other test file that is already using a similar class.

src/util/artifacts.js
saintsebastian: This is a great improvement on the error messages. Also do we really need `await mkdirp` outside of existing project conventions?

saintsebastian:
Thanks for review @rpl! I've added additional test in a mocking style, since I think it is more suitable, but this can be resolved when you and @kumar303 decide otherwise.

src/util/artifacts.js
```diff
@@ -1,14 +1,19 @@
 /* @flow */
-import {fs} from 'mz';
+import {fs as defaultFS} from 'mz';
```
kumar303: We currently use `mz/fs` everywhere else simply as `fs`. I would like to keep it consistent.

tests/unit/test-util/test.artifacts.js
kumar303: I prefer to see tests with real files using the temp dir (like the existing ones). However, in cases like lacking write-permissions it might be too hard to set up a test for that with real files. I think it would be fine to use a mock for that but only as a last resort.

rpl: As per discussion with @kumar303, we would both prefer to be able to do not inject the `fs` module as a dependency, and to be able to do this we need to test the above scenarios without any sinon spy.

I gave it a try and it looks like (by this test as an example), we can create all the needed tests using the `tmpDir` strategy:

-  "throws an UsageError when directory exists but lacks writing permissions": can be tested by creating the artifactsDir with permissions 555 (it lacks the write permission), something like:
``` 
      const artifactsDir = path.join(tmpDir.path(), 'dir-nowrite');
      return fs.mkdir(artifactsDir, '555').then(() => {
         ...
```

- "throws an UsageError if lacks permissions to create the directory": by creating a parent dir with permissions 555,
```
       const parentDir = path.join(tmpDir.path(), 'dir-nowrite');
       const artifactsDir = path.join(parentDir, 'artifacts');
       return fs.mkdir(parentDir, '555').then(() => {
```

- "throws an UsageError when it lacks permissions to stat the directory": by creating the parent dir with permissions 444 (it lacks the execute permission and nodejs cannot even stat the artifact dir path)
```
       const parentDir = path.join(tmpDir.path(), 'dir-nowrite');
       const artifactsDir = path.join(parentDir, 'artifacts');
       return fs.mkdir(parentDir, '444').then(() => {
```

- "creates the artifacts dir successfully if the parent dir doesn't exist": this should be pretty simple, it should be mostly like the fist test form this file but with an additional directory level (e.g. something like `path.join(tmpDir.path(), 'build', 'subdir')` 

NOTE: on asserting the error message various, try to be specific enough to do not catch one of the other error message by mistake (e.g. `/user lacks permissions/` is not specific enough, and you can use `.*`, e.g. `/myprefix.*mysuffix/`, to match a more specific error message with a regular expression)

src/util/artifacts.js
```diff
@@ -1,25 +1,54 @@
...
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-
+const promisifiedMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
+  fs?: typeof defaultFS = defaultFS,
```
rpl: Let's remove the injected `fs` and `mdir` dependency and let's try to use the tmpDir based testing approach describe in the related review comment.

(it will make this change smaller and the tests without sinon will make the change even safer)

tests/unit/test.errors.js
```diff
@@ -5,7 +5,7 @@ import {assert} from 'chai';
 import {
   onlyErrorsWithCode, isErrorWithCode, onlyInstancesOf,
 } from '../../src/errors';
-import {makeSureItFails} from './helpers';
+import {makeSureItFails, ErrorWithCode} from './helpers';
```
rpl: If we do not use sinon for the new tests, these changes related to `ErrorWithCode` are not needed anymore, but I'm not strongly against them.

saintsebastian: since `ErrorWithCode` are used in two separate test files and seem potentially very useful, I'd leave them as it is

src/util/artifacts.js
```diff
@@ -1,25 +1,51 @@
...
+        `"${artifactsDir}" exists and it is not a directory.`);
+    }
+    // Check that we can write in the artifacts dir.
+    try {
+      await fs.access(artifactsDir, fs.W_OK);
+    } catch (accessErr) {
+      if (isErrorWithCode('EACCES', accessErr)) {
+        throw new UsageError(
+          `"${artifactsDir}" exists but the user lacks permissions on it.`);
+      }
```
rpl: I think that we also need to throw the `accessErr` if it is not an 'EACCES' error.

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,59 @@ describe('prepareArtifactsDir', () => {
...
+          }));
+      });
+    }
+  ));
+
+  it('throws error when directory exists but lacks writing permissions',
+    () => withTempDir(
+    (tmpDir) => {
+      const artifactsDir = path.join(tmpDir.path(), 'dir-nowrite');
+      return fs.mkdir(artifactsDir, '555').then(() => {
```
rpl: `0555` (to match the syntax used in the test cause above)

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,59 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws error when checking the directory if lacks writing permissions',
+    () => withTempDir(
+    (tmpDir) => {
+      const tmpPath = path.join(tmpDir.path(), 'build');
+      return fs.mkdir(tmpPath, '0622').then(() => {
```
rpl: @saintsebastian like we discussed previously, the tests that are creating the directory with the custom permissions are not going to work on Windows (and I tried directly, by rebasing temporarily this PR on #766, and it looks that it fails because the custom permissions have no effect on the Windows system: https://ci.appveyor.com/project/rpl/web-ext/build/1.0.15/job/ro3024btux5tnm01#L1513)

Do you mind to:
- open a new issue to "investigate and re-enable this tests on windows"  
- prepare these tests to be skipped (using something like `if (process.platform == "win32") { this.skip(); return }`, 
- add an inline comment near `this.skip()` which reference the newly created issue

NOTE: to be able to use `this.skip()`, the test case needs to be in a regular function (`function () { ... }`) instead of an arrow function (`() => { ... }`), because the arrow function rebind `this` and so it will not provide the `skip` method.

src/util/artifacts.js
```diff
@@ -1,25 +1,51 @@
...
 ): Promise<string> {
   try {
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
       throw new UsageError(
-        `--artifacts-dir=${artifactsDir} (this value is not a directory)`);
+        `"${artifactsDir}" exists and it is not a directory.`);
+    }
+    // Check that we can write in the artifacts dir.
+    try {
```
rpl: I would really like to not have so many `try/catch` nested here, but I tried it rewritten without the nested `try/catch` and it didn't look better (to be fair the number of nested `try/catch` is not higher than it was before this change).

src/util/artifacts.js
```diff
@@ -1,25 +1,51 @@
...
+      // Handle errors when the artifactsDir doesn't exists yet.
+      try {
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await asyncMkdirp(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
+          throw new UsageError(
+            `Cannot create "${artifactsDir}", user lacks permissions.`);
+        } else {
```
rpl: no need for the `else` block here (the only other scenario is that we throw the UsageError above.

tests/unit/test-util/test.artifacts.js
saintsebastian: thanks for a clarifying (about arrow functions and in general). It is surprising that `win32` is true even for 64-bit systems

src/util/artifacts.js
```diff
@@ -1,25 +1,51 @@
...
 import {UsageError, isErrorWithCode} from '../errors';
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-
+const asyncMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
```
rpl: we should remove this trailing comma from the `prepareArtifactsDir` params

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,71 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws an UsageError when it lacks permissions to stat the directory',
+    () => withTempDir(
+    function(tmpDir) {
+      if (process.platform === 'win32') {
```
rpl: @saintsebastian can you add an inline comment here (and in the following two other skipped tests) which explains why they are skipped and link the related issue?
something like: 
```// Skipped on windows because the mkdir permissions do not have any effect (web-ext#ISSUENUMBER)```

File: src/util/artifacts.js (hunk 13 above)
Comment (saintsebastian) [inline src/util/artifacts.js]: How did I miss this one! Thanks for reminding me

**Comment by saintsebastian** [comment]
Thanks @rpl for you patience with all these tests

File: src/util/artifacts.js [hunk 15]
```diff
@@ -14,12 +16,36 @@ export async function prepareArtifactsDir(
     const stats = await fs.stat(artifactsDir);
     if (!stats.isDirectory()) {
       throw new UsageError(
-        `--artifacts-dir=${artifactsDir} (this value is not a directory)`);
+        `"${artifactsDir}" exists and it is not a directory.`);
```
kumar303: Any reason why you dropped the prefix? I think the prefix will make it unambiguous that the error is talking about the value passed to `--artifacts-dir`, like:

    `--artifacts-dir="${artifactsDir}" exists and it is not a directory.`

Also, just a nit, but using a two-word conjunction will help to convey that its existence is ok but its file state is not ok:

    `--artifacts-dir="${artifactsDir}" exists but it is not a directory.`

src/util/artifacts.js
```diff
@@ -14,12 +16,36 @@ export async function prepareArtifactsDir(
...
+    }
+    // If the artifactsDir already exists, check that we have the write permissions on it.
+    try {
+      await fs.access(artifactsDir, fs.W_OK);
+    } catch (accessErr) {
+      if (isErrorWithCode('EACCES', accessErr)) {
+        throw new UsageError(
+          `"${artifactsDir}" exists but the user lacks permissions on it.`);
```
kumar303: I think a prefix would be helpful here too. I'd also suggest saying specifically that the file is not writable:

    `--artifacts-dir="${artifactsDir}" exists but the user lacks permissions to write in it.`

src/util/artifacts.js
```diff
@@ -14,12 +16,36 @@ export async function prepareArtifactsDir(
...
+      }
     }
   } catch (error) {
-    if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+    if (isErrorWithCode('EACCES', error)) {
+      // Handle errors when the artifactsDir cannot be accessed.
+      throw new UsageError(
+        `Cannot access "${artifactsDir}", user lacks permissions.`);
```
kumar303: The prefix will help provide context. I also think the reason could be made more specific:

    `Cannot access --artifacts-dir="${artifactsDir}" because the user lacks permissions.`

src/util/artifacts.js
```diff
@@ -14,12 +16,36 @@ export async function prepareArtifactsDir(
...
+    } else if (isErrorWithCode('ENOENT', error)) {
+      // Create the artifact dir if it doesn't exist yet.
+      try {
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await asyncMkdirp(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
+          throw new UsageError(
+            `Cannot create "${artifactsDir}", user lacks permissions.`);
```
kumar303: Same suggestion here:

    `Cannot create --artifacts-dir="${artifactsDir}" because the user lacks permissions.`

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws an UsageError when it lacks permissions to stat the directory',
+    () => withTempDir(
+    function(tmpDir) {
```
kumar303: nit: I'd like to see this block (along with those of similar test cases) indented two spaces to show that it's a continuation of the call started above.

When we do https://github.com/mozilla/web-ext/issues/755 we can get rid of `withTempDir` ! It's annoying, sorry.

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
...
+      const artifactsDir = path.join(parentDir, 'artifacts');
+      return fs.mkdir(parentDir, '0555').then(() => {
+        return prepareArtifactsDir(artifactsDir)
+          .then(makeSureItFails())
+          .catch(onlyInstancesOf(UsageError, (error) => {
+            assert.match(error.message, /Cannot create.*lacks permissions/);
+          }));
+      });
+    }
+ ));
```
kumar303: indentation looks off

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
...
+
+  it('creates the artifacts dir successfully if the parent dir does not exist',
```
kumar303: ooh, nice test

saintsebastian:
@kumar303 thanks for the review

src/util/artifacts.js
```diff
@@ -14,12 +16,39 @@ export async function prepareArtifactsDir(
...
+      try {
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await asyncMkdirp(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
+          throw new UsageError(
+            `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
+            'user lacks permissions.');
+        }
```
kumar303: I just noticed something: shouldn't this have an `else` block that throws the error? Otherwise it will be ignored. Is that intentional? If it is, please at least add a `log.debug()` statement showing what the ignored error was.

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws an UsageError when it lacks permissions to stat the directory',
+    () => withTempDir(
+      function(tmpDir) {
```
kumar303: Hmm, I don't think this is quite right. I believe that in order to use `this.skip()` you'll need to make two changes:
* Pass a `function` to `it()` so that mocha will bind `this` to the test case instance
* Use arrow functions everywhere else in the test case so that `this` won't be bound to an instance

Example:
````js
it('throws an UsageError when it lacks permissions to stat the directory',
  function() {
    return withTempDir(
      (tmpDir) => {
        if (process.platform === 'win32') {
          // Now `this` will be bound to the mocha test case instance:
          this.skip();
          return;
        }
      }
    );
  }
);
````

More info: https://mochajs.org/#inclusive-tests

File: tests/unit/test-util/test.artifacts.js [hunk 24]
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
...
+            .then(makeSureItFails())
+            .catch(onlyInstancesOf(UsageError, (error) => {
+              assert.match(error.message, /Cannot access.*lacks permissions/);
+            }));
+        });
+      }
+  ));
+
+  it('throws error when directory exists but lacks writing permissions',
+    () => withTempDir(
```
kumar303: This has the same problem as above. It needs to begin with `function` so that `this` gets bound to the test instance.

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,74 @@ describe('prepareArtifactsDir', () => {
...
+            .then(makeSureItFails())
+            .catch(onlyInstancesOf(UsageError, (error) => {
+              assert.match(error.message, /exists.*lacks permissions/);
+            }));
+        });
+      }
+  ));
+
+  it('throws error when creating a folder if lacks writing permissions',
+    () => withTempDir(
```
kumar303: This has the same problem as above. It needs to begin with `function` so that `this` gets bound to the test instance.

src/util/artifacts.js
saintsebastian: that bit was taken out recently (https://github.com/mozilla/web-ext/pull/712/files/de75bcf35b02ac882d9f3de244fcaafbc6420037#r97771085) but i prefer having an error back just for user's information

src/util/artifacts.js
```diff
@@ -14,12 +16,41 @@ export async function prepareArtifactsDir(
...
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await asyncMkdirp(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
+          throw new UsageError(
+            `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
+            'user lacks permissions.');
+        } else {
+          throw mkdirErr;
```
kumar303: Yes, this seems better to me. A plausible error would be ENOSPC, No space left on device.

This needs test coverage. Surprisingly, the report says you are at 100% but when I commented out the `throw` statement, all your tests still pass. At least one should fail. You should be able to add a test that makes `asyncMkdirp` throw an error having the code `ENOSPC` to cover it.

src/util/artifacts.js
```diff
@@ -14,12 +16,41 @@ export async function prepareArtifactsDir(
...
+      // Create the artifact dir if it doesn't exist yet.
+      try {
+        log.debug(`Creating artifacts directory: ${artifactsDir}`);
+        await asyncMkdirp(artifactsDir);
+      } catch (mkdirErr) {
+        if (isErrorWithCode('EACCES', mkdirErr)) {
+          // Handle errors when the artifactsDir cannot be created for lack of permissions.
+          throw new UsageError(
+            `Cannot create --artifacts-dir="${artifactsDir}" because the ` +
+            'user lacks permissions.');
```
kumar303: This has the same issue as above. You can fix it like:
````js
`Cannot create --artifacts-dir="${artifactsDir}" because the ` +
`user lacks permissions: ${mkdirErr}`
````

File: src/util/artifacts.js [hunk 28]
```diff
@@ -14,12 +16,41 @@ export async function prepareArtifactsDir(
...
     }
   } catch (error) {
-    if (isErrorWithCode('ENOENT', error)) {
-      log.debug(`Creating artifacts directory: ${artifactsDir}`);
-      await fs.mkdir(artifactsDir);
+    if (isErrorWithCode('EACCES', error)) {
+      // Handle errors when the artifactsDir cannot be accessed.
+      throw new UsageError(
+        `Cannot access --artifacts-dir="${artifactsDir}" because the user ` +
+        'lacks permissions.');
```
kumar303: I think this should include the original error to help the user pinpoint exactly what the access problem is. You can do it like:
````js
`Cannot access --artifacts-dir="${artifactsDir}" because the user ` +
`lacks permissions: ${error}`
````
Otherwise the permission error may not be completely obvious: the user may not own own the file, the read bit may not be set, the parent dir may not have an executable bit, etc.

File: tests/unit/test-util/test.artifacts.js [hunk 29]
```diff
@@ -55,4 +55,83 @@ describe('prepareArtifactsDir', () => {
...
+      );
+    }
+  );
+
+  it('throws error when directory exists but lacks writing permissions',
+    function() {
+      return withTempDir(
+        (tmpDir) => {
+          if (process.platform === 'win32') {
+            // Now `this` will be bound to the mocha test case instance:
```
kumar303: and this one

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,83 @@ describe('prepareArtifactsDir', () => {
...
+      );
+    }
+  );
+
+  it('throws error when creating a folder if lacks writing permissions',
+    function() {
+      return withTempDir(
+        (tmpDir) => {
+          if (process.platform === 'win32') {
+            // Now `this` will be bound to the mocha test case instance:
```
kumar303: and this one too

tests/unit/test-util/test.artifacts.js
```diff
@@ -55,4 +55,83 @@ describe('prepareArtifactsDir', () => {
     }
   ));
 
+  it('throws an UsageError when it lacks permissions to stat the directory',
+    function() {
+      return withTempDir(
+        (tmpDir) => {
+          if (process.platform === 'win32') {
+            // Now `this` will be bound to the mocha test case instance:
```
saintsebastian: Oh sorry for this sloppiness!

src/util/artifacts.js
saintsebastian: @kumar303 what's the best way to do it? Sinon would do the job, but those test have been rewritten to avoid using it and rely on permission codes instead.

rpl: @saintsebastian being able to reproduce in a test case the `ENOSPC` scenario is not as "easy" as it was with the "lack of permissions" scenarios, and so it is reasonable to use the the usual "dependency injection strategy", this time restricted to this single parameter and this single test.

saintsebastian: @rpl thanks for the quick answer!

src/util/artifacts.js
```diff
@@ -1,25 +1,57 @@
...
 import {createLogger} from './logger';
 
 const log = createLogger(__filename);
-
+const defaultAsyncMkdirp = promisify(mkdirp);
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
+  asyncMkdirp?: typeof defaultAsyncMkdirp = defaultAsyncMkdirp,
```
kumar303: To follow the convention we use everywhere else in web-ext, I'd like optional parameters to always be object properties. This solves a couple issues. First, it makes the calling code easier to read because you don't have to remember what all the positional arguments are. Secondly, it makes it easier to add additional options to this function in the future. The new signature would look like:
````js
type PrepareArtifactsDirOptions = {
  asyncMkdirp?: typeof defaultAsyncMkdirp,
}

export async function prepareArtifactsDir(
  artifactsDir: string,
  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions,
): Promise<string> {
  ...
}
````

File: tests/unit/test-util/test.artifacts.js [hunk 33]
```diff
@@ -55,4 +57,95 @@ describe('prepareArtifactsDir', () => {
...
+    (tmpDir) => {
+      const fakeAsyncMkdirp = sinon.spy(() =>
+        Promise.reject(new ErrorWithCode('ENOSPC', 'an error')));
+      const tmpPath = path.join(tmpDir.path(), 'build', 'subdir');
+      return prepareArtifactsDir(tmpPath, fakeAsyncMkdirp)
+        .then(makeSureItFails())
+        .catch((error) => {
+          assert.ok(fakeAsyncMkdirp.called);
+          assert.equal(error.message, 'an error');
+        });
```
kumar303: There is a pattern I like a little bit better for this because it will throw the error from `makeSureItFails()` rather than pass it to the assertion in `catch()`. Both approaches will work but I'd like us to use this pattern for future code:
````js
return prepareArtifactsDir(tmpPath, fakeAsyncMkdirp)
  .then(makeSureItFails(), (error) => {
    assert.ok(fakeAsyncMkdirp.called);
    assert.equal(error.message, 'an error');
  });
````

Comment (saintsebastian) [inline tests/unit/test-util/test.artifacts.js]: @kumar303 the other tests deviate from this structure (`catch(onlyInstancesOf(UsageError, (error) => {}`). Should they be fixed as well?

Comment (kumar303) [inline tests/unit/test-util/test.artifacts.js]: We should fix them eventually but don't worry about it for now. Let's keep this patch focused.

File: src/util/artifacts.js [hunk 34]
```diff
@@ -1,25 +1,61 @@
...
+const defaultAsyncMkdirp = promisify(mkdirp);
 
+type PrepareArtifactsDirOptions = {
+  asyncMkdirp?: typeof defaultAsyncMkdirp,
+}
 
 export async function prepareArtifactsDir(
-  artifactsDir: string
+  artifactsDir: string,
+  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions,
```
kumar303: This looks good but in my example I forgot to say that this should be optional. You can tell that it needs to be optional because it has a default value. The definition to make it optional would look like this. The syntax is a bit weird but what it's doing is initializing the options parameter to an empty object.
````js
export async function prepareArtifactsDir(
  artifactsDir: string,
  {asyncMkdirp = defaultAsyncMkdirp}: PrepareArtifactsDirOptions = {},
): Promise<string> {
  ...
}
````
After that, all the tests can call it like before:
````js
return prepareArtifactsDir(tmpPath)
  ..
````

saintsebastian: thanks for the review(s)!

PR #726, opened 2017-01-03
refactor: changed named functions params to have Flow types

### LINKED ISSUE #696: Make sure all named function parameters have Flow types

code quality enhancement

The code base still has a lot of untyped object parameters. They look something like this:
````javascript
function doTheThing({
  paramOne,
  paramTwo,
}: Object = {}) {
  ...
}
````

If you use a wrong parameter value when calling a function like this, Flow should tell you but it won't because `Object` is a loose type that allows any value.

We should change all instances of this to use typed objects like:

````javascript
type doTheThingParams = {
  paramOne: string,
  paramTwo: number,
}

function doTheThing({
  paramOne,
  paramTwo,
}: doTheThingParams = {}) {
  ...
}
````

### PR DESCRIPTION
**Pull request by shubheksha**
Fixes #696 
Can't figure out a way to use Flow types for rest params yet.

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -27,31 +27,31 @@ export function safeFileName(name: string): string {
 
 // defaultPackageCreator types and implementation.
 
-export type ExtensionBuildResult = {
+export type ExtensionBuildResult = {|
   extensionPath: string,
-};
+|};
 
-export type PackageCreatorParams = {
+export type PackageCreatorParams = {|
   manifestData?: ExtensionManifest,
   sourceDir: string,
   fileFilter: FileFilter,
   artifactsDir: string,
   showReadyMessage: boolean
-};
+|};
 
-export type LocalizedNameParams = {
+export type LocalizedNameParams = {|
   messageFile: string,
   manifestData: ExtensionManifest,
-}
+|}
 
 // This defines the _locales/messages.json type. See:
 // https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales
-type LocalizedMessageData = {
+type LocalizedMessageData = {|
   [messageName: string]: {
     description: string,
     message: string,
   },
-}
+|}
 
 export async function getDefaultLocalizedName(
   {messageFile, manifestData}: LocalizedNameParams
@@ -135,19 +135,19 @@ async function defaultPackageCreator({
 
 // Build command types and implementation.
 
-export type BuildCmdParams = {
+export type BuildCmdParams = {|
   sourceDir: string,
   artifactsDir: string,
   asNeeded?: boolean,
-};
+|};
 
-export type BuildCmdOptions = {
+export type BuildCmdOptions = {|
   manifestData?: ExtensionManifest,
   fileFilter?: FileFilter,
   onSourceChange?: OnSourceChangeFn,
   packageCreator?: PackageCreatorFn,
   showReadyMessage?: boolean
-};
+|};
 
 export default async function build(
   {sourceDir, artifactsDir, asNeeded = false}: BuildCmdParams,
@@ -189,9 +189,9 @@ export default async function build(
 
 // FileFilter types and implementation.
 
-export type FileFilterOptions = {
+export type FileFilterOptions = {|
   filesToIgnore?: Array<string>,
-};
+|};
 
 /*
  * Allows or ignores files when creating a ZIP archive.
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -47,10 +47,10 @@ export type LocalizedNameParams = {|
 // This defines the _locales/messages.json type. See:
 // https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales
 type LocalizedMessageData = {|
-  [messageName: string]: {
+  messageName: {|
     description: string,
     message: string,
-  },
+  |},
 |}
 
 export async function getDefaultLocalizedName(
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -47,7 +47,7 @@ export type LocalizedNameParams = {|
 // This defines the _locales/messages.json type. See:
 // https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales
 type LocalizedMessageData = {|
-  messageName: {|
+  [messageName: string]: {|
     description: string,
     message: string,
   |},
--- a/src/program.js
+++ b/src/program.js
@@ -188,10 +188,10 @@ export function defaultVersionGetter(
 // TODO: add pipes to Flow type after https://github.com/facebook/flow/issues/2405 is fixed
 
 type MainParams = {
-  getVersion: Function,
-  commands: Object,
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

**Comment by kumar303** [comment]
You should be able to catch rest params just like regular variables. Something like this is working for me:
````diff
diff --git a/tests/unit/test-firefox/test.firefox.js b/tests/unit/test-firefox/test.firefox.js
index 73c55e8..d298cfa 100644
--- a/tests/unit/test-firefox/test.firefox.js
+++ b/tests/unit/test-firefox/test.firefox.js
@@ -47,7 +47,12 @@ describe('firefox', () => {
       }));
     }
 
-    function runFirefox({profile = fakeProfile, ...args}: Object = {}) {
+    type RunFirefoxParams = {
+      profile?: FirefoxProfile,
+      args?: Array<any>,
+    }
+
+    function runFirefox({profile = fakeProfile, ...args}: RunFirefoxParams = {}) {
       return firefox.run(profile, {
         fxRunner: createFakeFxRunner(),
         findRemotePort: () => Promise.resolve(6000),

````
Although using something more specific than `any` would be better.

Also, the rest param is just a convenience for the function. I would suggest explicitly defining each supported type to avoid surprises, like this:
````diff
diff --git a/tests/unit/test-firefox/test.firefox.js b/tests/unit/test-firefox/test.firefox.js
index 73c55e8..d9d5ae4 100644
--- a/tests/unit/test-firefox/test.firefox.js
+++ b/tests/unit/test-firefox/test.firefox.js
@@ -47,7 +47,12 @@ describe('firefox', () => {
       }));
     }
 
-    function runFirefox({profile = fakeProfile, ...args}: Object = {}) {
+    type RunFirefoxParams = {
+      profile?: FirefoxProfile,
+      firefoxBinary?: string,
+    }
+
+    function runFirefox({profile = fakeProfile, ...args}: RunFirefoxParams = {}) {
       return firefox.run(profile, {
         fxRunner: createFakeFxRunner(),
         findRemotePort: () => Promise.resolve(6000),

````

File: tests/unit/helpers.js [hunk 1]
```diff
@@ -182,10 +182,17 @@ export function fake<T>(original: Object, methods: Object = {}): T {
  * Returns a fake Firefox client as would be returned by
  * connect() of 'node-firefox-connect'
  */
+
+type FakeFirefoxClientParams = {
+  requestResult?: Object,
+  requestError?: Object | void,
```
kumar303: why do you need `void`?

tests/unit/helpers.js
```diff
@@ -182,10 +182,17 @@ export function fake<T>(original: Object, methods: Object = {}): T {
...
+  makeRequestResult?: Object,
+  makeRequestError?: Object | void,
```
kumar303: why do you need `void` here? I think these were just left over from before you removed null (which was the right move)

tests/unit/test-cmd/test.lint.js
```diff
@@ -7,9 +7,13 @@ import defaultLintCommand from '../../../src/cmd/lint';
 import {FileFilter} from '../../../src/cmd/build';
 import {fake, makeSureItFails} from '../helpers';
 
+type setUpParams = {
+  createLinter?: Function,
+  fileFilter?: Function,
+}
```
kumar303: nit: this should have a blank line after it, separating this block from the call to `describe()`

tests/unit/test-cmd/test.sign.js
```diff
@@ -55,9 +55,14 @@ describe('sign', () => {
   /*
    * Run the sign command with stubs for all dependencies.
    */
+  type SignExtraParams = {
```
kumar303: the pattern we use is to call something like this `SignOptions`.

tests/unit/helpers.js
```diff
@@ -182,10 +182,17 @@ export function fake<T>(original: Object, methods: Object = {}): T {
  * Returns a fake Firefox client as would be returned by
  * connect() of 'node-firefox-connect'
  */
+
+type FakeFirefoxClientParams = {
```
kumar303: Please use pipes for this and all type definitions. This will make sure that no *unknown* object keys are allowed. Example:
````javascript
type FakeFirefoxClientParams = {|
  requestResult?: Object,
|}
````

**Comment by kumar303** [comment]
These also need parameter types:

* https://github.com/mozilla/web-ext/blob/master/src/program.js#L96
* https://github.com/mozilla/web-ext/blob/master/src/program.js#L179

**Comment by shubheksha** [comment]
@kumar303, could you review this again? Thanks!

File: tests/unit/test-firefox/test.firefox.js [hunk 6]
```diff
@@ -47,7 +48,19 @@ describe('firefox', () => {
       }));
     }
 
-    function runFirefox({profile = fakeProfile, ...args}: Object = {}) {
+    // TODO: add pipes to Flow type after https://github.com/facebook/flow/issues/2405 is fixed
+
+    type RunFirefoxOptions = {
+      profile?: typeof FirefoxProfile,
+      args?: Array<any>,
```
kumar303: I think `args` can be deleted because that's actually a property name that gets passed to the function. The `...args` syntax will spread the object into specific properties (none of which are named `args`). 

Also, I think it's ok in this case *not* to define each specific property because it's just passing through properties to the `firefox.run()` call. In other words, if one of the tests calls `runFirefox()` with invalid properties, the `FirefoxRunOptions` type will catch them. 

Thus, you can edit the TODO comment and just say something like "This object should accept dynamic properties since those are passed to firefox.run()"

shubheksha:
@kumar303, I addressed all the feedback, could you review this again? Thanks!

src/cmd/build.js
```diff
@@ -27,31 +27,31 @@ export function safeFileName(name: string): string {
...
   manifestData: ExtensionManifest,
-}
+|}
 
 // This defines the _locales/messages.json type. See:
 // https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales
-type LocalizedMessageData = {
-  [messageName: string]: {
+type LocalizedMessageData = {|
+  messageName: {|
```
kumar303: oh, sorry! This should be `[messageName: string]`. I was confused for some reason. The message file can have many keys of this type: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales

src/program.js
```diff
@@ -174,13 +185,21 @@ export function defaultVersionGetter(
   }
 }
 
+// TODO: add pipes to Flow type after https://github.com/facebook/flow/issues/2405 is fixed
+
+type MainParams = {
+  getVersion: Function,
```
kumar303: this should be optional: `getVersion?:` since it has a default value in the function signature

src/program.js
```diff
@@ -174,13 +185,21 @@ export function defaultVersionGetter(
...
+  commands: Object,
```
kumar303: this should be optional: `commands?:` since it has a default value in the function signature

src/program.js
```diff
@@ -174,13 +185,21 @@ export function defaultVersionGetter(
...
+  argv: Array<any>,
+  runOptions: Object,
```
kumar303: this should be optional: `runOptions?:` since it has a default value in the function signature

PR #727, opened 2017-01-03
fix: Support custom prefs values that contains '=' chars

rpl:
This PR contains two small fixes which cover the following corner cases:

- custom prefs values which contains '=' chars
- custom prefs which do not specify a value

### DIFF
--- a/src/firefox/preferences.js
+++ b/src/firefox/preferences.js
@@ -136,8 +136,16 @@ export function coerceCLICustomPreference(
 
   for (const pref of cliPrefs) {
     const prefsAry = pref.split('=');
+
+    if (prefsAry.length < 2) {
+      throw new UsageError(
+        `Incomplete custom preference: "${pref}". ` +
+        'Syntax expected: "prefname=prefvalue".'
+      );
+    }
+
     const key = prefsAry[0];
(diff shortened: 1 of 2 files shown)


src/firefox/preferences.js
```diff
@@ -136,8 +136,13 @@ export function coerceCLICustomPreference(
 
   for (const pref of cliPrefs) {
     const prefsAry = pref.split('=');
+
+    if (prefsAry.length < 2) {
+      throw new UsageError(`Incomplete custom preference: "${pref}"`);
```
kumar303: This might be more helpful with an example, like "Incomplete custom preference (expected key=value): ..."

tests/unit/test-firefox/test.preferences.js
```diff
@@ -68,6 +68,13 @@ describe('firefox/preferences', () => {
       assert.equal(prefs['valid.preference'], '4.55');
     });
 
+    it('support string values with "=" chars', () => {
```
kumar303: nit: typo: `it('supports...`

src/firefox/preferences.js
rpl: :+1:

PR #731, opened 2017-01-03
chore: Fixed the intermittent empty pull request title failures

rpl:
This PR should finally fix the intermittency of the "pull request title" linting ;-)

### DIFF
--- a/tasks/travis-pr-title-lint.js
+++ b/tasks/travis-pr-title-lint.js
@@ -72,25 +72,64 @@ module.exports = function(grunt) {
 
   function getPullRequestTitle() {
     return new Promise(function(resolve, reject) {
-      var pullRequestURLPath = '/repos/' +
-        process.env.TRAVIS_REPO_SLUG + '/pulls/' +
-        process.env.TRAVIS_PULL_REQUEST + '.json';
+      var pullRequestURLPath = '/' +
+        process.env.TRAVIS_REPO_SLUG + '/pull/' +
+        process.env.TRAVIS_PULL_REQUEST;
 
       grunt.log.writeln(
-        'Retrieving the pull request title from https://api.github.com' +
+        'Retrieving the pull request title from https://github.com' +
         pullRequestURLPath
       );
 
-      https.get({
-        host: 'api.github.com',
+      var req = https.get({
+        host: 'github.com',
         path: pullRequestURLPath,
         headers: {
-          'User-Agent': 'mozilla web-ext grunt tasks',
+          'User-Agent': 'mozilla web-ext grunt tasks'
         },
       }, function(response) {
         var body = '';
         response.on('data', function(data) {
-          body += data;
+          try {
+            body += data;
+
+            // Once we get the closing title tag, we can read
+            // the pull request title and
+            if (body.includes('</title>')) {
+              response.removeAllListeners('data');
+              response.emit('end');
+
+              var titleStart = body.indexOf('<title>');
+              var titleEnd = body.indexOf('</title>');
+
+              // NOTE: page slice is going to be something like:
+              // "<title> PR title by author · Pull Request #NUM · mozilla/web-ext · GitHub"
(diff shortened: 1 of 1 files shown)


rpl:
@kumar303 I'm not exactly "proud" of it, but this seems to do the trick ;-)

tasks/travis-pr-title-lint.js
```diff
@@ -72,25 +72,69 @@ module.exports = function(grunt) {
...
+                  .replace('<title>', '')
+                  .split(' · ');
+
+              // Check that we have really got the title of a real pull request.
+              var expectedPart1 = 'Pull Request #' +
+                process.env.TRAVIS_PULL_REQUEST;
+
+              if (pageTitleParts[1] === expectedPart1) {
+                // NOTE: the extracted title still contains the
+                // "by author" part, but it should not affect the
```
rpl: @kumar303 actually this will affect the "changelog linting" if the resulting string  "pull request title + by author" exceeds the 72 character limit :-(

tasks/travis-pr-title-lint.js
```diff
@@ -72,25 +72,68 @@ module.exports = function(grunt) {
...
+              var expectedPart1 = 'Pull Request #' +
+                process.env.TRAVIS_PULL_REQUEST;
+
+              if (pageTitleParts[1] === expectedPart1) {
+                // Remove the "by author" part.
+                var prTitleEnd = pageTitleParts[0].lastIndexOf(' by ');
+                resolve(pageTitleParts[0].slice(0, prTitleEnd));
+              } else {
+                if (process.env.VERBOSE === 'true') {
+                  console.log('DEBUG getPullRequestTitle:', body);
```
kumar303: good idea on checking the title. However, I feel like this is very likely to break one day. How about doing a console log even if it's not in verbose mode? That will save us the hassle of having to re-run it in verbose mode. In every day usage it won't log anything so there's no added noise.

tasks/travis-pr-title-lint.js
```diff
@@ -72,25 +72,68 @@ module.exports = function(grunt) {
...
         pullRequestURLPath
       );
 
-      https.get({
-        host: 'api.github.com',
+      var req = https.get({
+        host: 'github.com',
         path: pullRequestURLPath,
         headers: {
           'User-Agent': 'mozilla web-ext grunt tasks',
```
kumar303: I feel like this user-agent should be named more like "your API is annoying" :)

tasks/travis-pr-title-lint.js
```diff
@@ -72,25 +72,68 @@ module.exports = function(grunt) {
...
+                  console.log('DEBUG getPullRequestTitle:', body);
+                }
+                reject(new Error('Unable to retrieve the pull request title'));
+              }
+
+              req.abort();
+            }
+          } catch (err) {
+            reject(err);
+            req.abort();
```
kumar303: should it call abort() before rejecting? I'm just wondering if the error gets thrown that the abort() may get lost on the stack (but maybe not).

tasks/travis-pr-title-lint.js
rpl: :+1: I like it :-P

tasks/travis-pr-title-lint.js

tasks/travis-pr-title-lint.js
rpl: I'm calling about explicitly only to force nodejs to close the connection when we are sure that it is not needed anymore, I'm aborting it after the rejection on purpose because I'm afraid that we can reject from the `response.on('error', ...)` and we want the caught exception instead.

kumar303: ok, sounds good

rpl:
@kumar303 Travis has completed the build on the last tweaks, hopefully this is the last time I change this grunt task for a while :-)

I'm going to merge it now!

PR #735, opened 2017-01-04
chore: after upgrading yargs, fix deprecated use of demand()

kumar303:
`demand()` changed after yargs 6.6.0 https://github.com/yargs/yargs/pull/740

Test bustage was in https://travis-ci.org/mozilla/web-ext/builds/189002823 (but this mysteriously didn't fail in the PR that actually upgraded yargs, hmm)

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -59,7 +59,11 @@ export class Program {
         return;
       }
       return yargs
-        .demand(0, 0, 'This command does not take any arguments')
+        // Make sure the user does not add any extra commands. For example,
+        // this would be a mistake because lint does not accept arguments:
+        // web-ext lint ./src/path/to/file.js
+        .demandCommand(0, 0, undefined,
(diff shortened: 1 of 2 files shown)


kumar303:
@rpl this fixes the failing test and I discovered that we were missing a test for the other use of `demand()`. Could you take another look?

rpl:
@kumar303 it is even more weird because it did fail as expected here:

- https://travis-ci.org/mozilla/web-ext/jobs/188997915#L1528

but the grunt task exited with 0:

- https://travis-ci.org/mozilla/web-ext/jobs/188997915#L1580

can you file a separate issue to investigate how it is happening? 

it is weird (and scary ;-)) and I'd like to take a deeper look into it to prevent similar regressions to happen again in the future.

kumar303:
oh, good spotting. I filed https://github.com/mozilla/web-ext/issues/736

PR #744, opened 2017-01-09
feat: parse explicit config file to set global options

### LINKED ISSUE #728: Parse an explicit config file to set global options
This is a stepping stone toward implementing all of https://github.com/mozilla/web-ext/issues/176

For this step, let's do the following:
* Add a new global `--config, -c` option that accepts a file path
* Parse the config file with `require()` as if it were a NodeJS module
* The configured value for any option can be overridden by a specific command line value

See https://github.com/mozilla/web-ext/issues/738 for handling default values

shubheksha:
Fixes #728


src/config.js
```diff
@@ -0,0 +1,61 @@
...
+import path from 'path';
+
+import requireUncached from 'require-uncached';
+
+import {createLogger} from './util/logger';
+import {UsageError} from './errors';
+
+const log = createLogger(__filename);
+
+type ApplyConfigToArgvParams = {
```
kumar303: use pipes for this object type

src/config.js
```diff
@@ -0,0 +1,61 @@
...
+  argv: Object,
+  configObject?: Object,
+  configFileName?: string,
+  defaultValues?: Object,
+};
+
+type ParseConfigParams = {
```
kumar303: use pipes for this object type

src/program.js
```diff
@@ -121,7 +129,8 @@ export class Program {
     argv.customPrefs = argv.pref;
 
     const runCommand = this.commands[cmd];
-
+    const configFileName = argv.config;
+    const newArgv = applyConfigToArgv(argv, configFileName, this.defaultValues);
```
kumar303: Hmm, how did Flow let you do this? It should be `applyConfigToArgv({argv, configFileName, defaultValues})`.

You will need to set up a class variable for `applyConfigToArgv` on `Program` so that it can be replaced with a stub for testing.

src/config.js
```diff
@@ -0,0 +1,61 @@
...
+  for (const option in configObject) {
+    if (!argv.hasOwnProperty(option) || defaultValues[option]) {
+      argv[option] = configObject[option];
+    }
+  }
+  return argv;
+}
+
+export function parseConfig({
+  argv,
```
shubheksha: This small change is making ~25 completely unrelated tests fail. I've addressed all the other nits and feedback, will take some time to get to the bottom of this one.

shubheksha:
I addressed all the feedback except the one I commented on and adding the two tests you mentioned.

tests/unit/test.config.js
```diff
@@ -0,0 +1,162 @@
+import path from 'path';
+
+import {assert} from 'chai';
+import {describe, it} from 'mocha';
+import sinon from 'sinon';
+import {fs} from 'mz';
+
+//import {fake} from './helpers';
```
kumar303: this looks like it can be removed

tests/unit/test.config.js
```diff
@@ -0,0 +1,162 @@
...
+}) {
+  const program = new Program(userCmd);
+
+  if (globalOpt) {
+    program.setGlobalOptions(globalOpt);
+  }
+  if (commandOpt) {
+    program.command(command, commandDesc, commandExecutor, commandOpt);
+  }
+  return program;
```
kumar303: it looks like you can return an object with commonly needed properties for convenience:
````js
return {
  argv: program.yargs.exitProcess(false).argv,
  defaultValues: program.defaultValues,
};
````

And use them in tests like:
````js
const {argv, defaultValues} = makeArgv(...);
````

File: tests/unit/test.config.js [hunk 7]
```diff
@@ -0,0 +1,162 @@
...
+        },
+      });
+      const argv = program.yargs.exitProcess(false).argv;
+      const configObject = {
+        sourceDir: '/configured/source/dir',
+      };
+      const defaultValues = {
+        sourceDir: 'default/source/dir',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
```
kumar303: this should use `program.defaultValues`, not a custom `defaultValues` object.

src/config.js
```diff
@@ -0,0 +1,62 @@
...
+
+import {createLogger} from './util/logger';
+import {UsageError} from './errors';
+
+const log = createLogger(__filename);
+
+type ApplyConfigToArgvParams = {|
+  argv: Object,
+  configObject: Object,
+  defaultValues?: Object,
```
kumar303: This should not be optional. I only see a few cases in your tests where it's not passed but you can update those calls to pass in the object. This function needs to operate on `defaultValues` for it to do its job.

tests/unit/test.config.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+  });
+
+  describe('loadJSConfigFile', () => {
+    it('throws an error if the config file does not exist', () => {
+      return withTempDir (
+        (tmpDir) => {
+          assert.throws(() => {
+            loadJSConfigFile(tmpDir.path());
+          }, UsageError);
```
kumar303: this should assert a little bit of the error message to make sure that the right `UsageError` instance was the one thrown. For example, if you wanted to match the string "foo is undefined" then you'd pass a regular expression like this:
````js
assert.throws(() => loadJSConfigFile(tmpDir.path()), UsageError,
  /foo is undefined/);
````

File: tests/unit/test.config.js [hunk 10]
```diff
@@ -0,0 +1,162 @@
...
+          },
+        },
+      });
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'config.js');
+          const configFileName = 'config.js';
+          argv.sourceDir = tmpDir.path();
+          fs.writeFileSync(configFilePath,
+          `module.exports = {
```
kumar303: this block should be indented two spaces to show that it's a continuation of the function call

src/program.js
```diff
@@ -121,7 +129,9 @@ export class Program {
     argv.customPrefs = argv.pref;
 
     const runCommand = this.commands[cmd];
-
+    const configFileName = argv.config;
+    const configObject = parseConfig(argv, configFileName);
+    const newArgv = applyConfigToArgv(argv, configObject, this.defaultValues);
```
kumar303: As we discussed, remove all of this from the patch. We add it back to `program.js` in another patch. By removing it, it will allow you to focus just on the units themselves and not have to deal with failures in `test.program.js`.

kumar303: I filed this to add this code back into `program.js` https://github.com/mozilla/web-ext/issues/775

src/program.js
```diff
@@ -246,6 +254,13 @@ Example: $0 --help run.
       describe: 'Show verbose output',
       type: 'boolean',
     },
+    'config': {
+      alias: 'c',
+      describe: 'Supply a custom configuration file',
+      type: 'string',
+      requiresArg: true,
+      demand: false,
+    },
```
kumar303: You can also remove this code for now and do it later when you work on https://github.com/mozilla/web-ext/issues/775

tests/unit/test.config.js
```diff
@@ -0,0 +1,162 @@
+import path from 'path';
```
kumar303: The top of this file needs the comment `/* @flow */` so that Flow checks it

src/config.js
```diff
@@ -0,0 +1,62 @@
...
+type ParseConfigParams = {|
+  sourceDir: string,
+  configFileName: string,
+|};
+
+export function applyConfigToArgv({
+  argv,
+  configObject,
+  defaultValues = {},
+}: ApplyConfigToArgvParams): Object {
```
kumar303: This function should not mutate `argv` because that can create subtle bugs. You can work around this by first copying all of its properties and working with that variable:
````js
const newArgv = {...argv};
  // ...
}
// ...
return newArgv;
````

File: src/config.js [hunk 15]
```diff
@@ -0,0 +1,62 @@
...
+  for (const option in configObject) {
+    if (argv[option] && defaultValues[option]) {
+      const wasValueSetOnCLI = argv[option] !== defaultValues[option];
+      if (wasValueSetOnCLI) {
```
kumar303: Add some debug logging here so we can detect when this happens (since we know there are subtle edge cases when it will be wrong). I suggest:
````js
log.debug(`Favoring CLI ${option}=${argv[option]} over configuration: ${option}=${configObject[option]}`);
````

File: src/config.js [hunk 16]
```diff
@@ -0,0 +1,62 @@
...
+|};
+
+export function applyConfigToArgv({
+  argv,
+  configObject,
+  defaultValues = {},
+}: ApplyConfigToArgvParams): Object {
+  for (const option in configObject) {
+    if (argv[option] && defaultValues[option]) {
+      const wasValueSetOnCLI = argv[option] !== defaultValues[option];
```
kumar303: If you change this to the following then you won't need to wrap it in the if statement up above:
````js
const wasValueSetOnCLI = argv[option] && argv[option] !== defaultValues[option];
````

File: src/config.js [hunk 17]
```diff
@@ -0,0 +1,62 @@
...
+      if (wasValueSetOnCLI) {
+        continue;
+      }
+    }
+    if (!argv.hasOwnProperty(option) || defaultValues[option]) {
```
kumar303: A debug statement would be helpful here too:
````js
log.debug(`Favoring configuration: ${option}=${configObject[option]} over CLI: ${option}=${argv[option]}`);
````

File: src/config.js [hunk 18]
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function loadJSConfigFile(filePath: string) {
+  log.debug(`Loading JS config file: ${filePath}`);
+  try {
+    return requireUncached(filePath);
+  } catch (error) {
+    log.debug('Handling error:', error);
+    throw new UsageError
+        (`Cannot read config file: ${filePath}\nError: ${error.message}`);
```
kumar303: nit: To keep style consistent in web-ext, preserve the opening parenthesis here, like:
````js
throw new UsageError(
  `Cannot read config file: ${filePath}\nError: ${error.message}`);
````

File: src/program.js [hunk 19]
```diff
@@ -121,7 +128,6 @@ export class Program {
     argv.customPrefs = argv.pref;
 
     const runCommand = this.commands[cmd];
-
```
kumar303: can you put these lines back just to make the diff easier to read?

tests/unit/test.config.js
```diff
@@ -0,0 +1,185 @@
...
```
kumar303: The top of this file needs the comment `/* @flow */` so that Flow checks it

tests/unit/test.config.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+  return {
+    argv: program.yargs.exitProcess(false).argv,
+    defaultValues: program.defaultValues,
+  };
+}
+
+describe('config', () => {
+  describe('applyConfigToArgv', () => {
+    it('overrides the default value with a configured value', () => {
```
kumar303: This test isn't doing what it says because the option is not configured with a default value. You already added 'preserves configured value over default' so I think you can delete this one.

tests/unit/test.config.js
```diff
@@ -0,0 +1,185 @@
...
+            demand: false,
+            default: 'default/value/option/definition',
+          },
+        },
+      });
+      const configObject = {
+        foo: '/configured/foo',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.sourceDir, argv.sourceDir);
```
kumar303: I would rather see `argv.sourceDir` replaced with `'default/value/option/definition'` to be sure it's getting the default.

tests/unit/test.config.js
```diff
@@ -0,0 +1,178 @@
...
+  });
+
+  describe('parseConfig', () => {
+    it('parses the configuration file correctly', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'config.js');
+          const configFileName = 'config.js';
+          fs.writeFileSync(configFilePath,
+          `module.exports = {
```
kumar303: nit: call continuations should always be indented two spaces

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "vsicons.presets.angular": false
```
kumar303: Is this really necessary? Why does it assume web-ext is written in Angular?

shubheksha: Ugh, this is my vscode settings being committed by mistake. I'll delete the directory.

src/config.js
```diff
@@ -0,0 +1,64 @@
...
+  return newArgv;
+}
+
+export function parseConfig({
+  sourceDir,
+  configFileName,
+}: ParseConfigParams): Object {
+  let configObject = {};
+  if (configFileName) {
+    const configFilePath = path.join(`/${sourceDir}`, configFileName);
```
shubheksha: If I remove this, there will be no default place to look. Even the existing test will fail. How will I pass `tmpDir.path()` to `parseConfig()`?

.vscode/settings.json
kumar303: You can add `.vscode` to [.gitignore](https://github.com/mozilla/addons-frontend/blob/master/.gitignore) to prevent it from happening.

tests/unit/test.config.js
```diff
@@ -0,0 +1,187 @@
...
+import {Program} from '../../src/program';
+import {
+  applyConfigToArgv,
+  loadJSConfigFile,
+} from '../../src/config';
+import {withTempDir} from '../../src/util/temp-dir';
+import {UsageError} from '../../src/errors';
+
+type MakeArgvParams = {|
+  userCmd: Array<string>,
```
kumar303: this should be marked as optional since the implementation initializes it to a default value

tests/unit/test.config.js
```diff
@@ -0,0 +1,187 @@
...
+      assert.strictEqual(newArgv.sourceDir, cmdLineSrcDir);
+    });
+  });
+
+  describe('loadJSConfigFile', () => {
+    it('throws an error if the config file does not exist', () => {
+      return withTempDir (
+        (tmpDir) => {
+          assert.throws(() => {
+            loadJSConfigFile(tmpDir.path());
```
kumar303: this is producing another kind of error ("Cannot find module"), not an error about a config file that doesn't exist. You should explicitly pass it a path that doesn't exist, like:
````js
loadJSConfigFile(path.join(tmpDir.path(), 'non-existant-config.js'));
````

File: tests/unit/test.config.js [hunk 28]
```diff
@@ -0,0 +1,187 @@
...
+        });
+    });
+
+    it('throws an error if the config file has syntax errors', () => {
+      return withTempDir (
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'config.js');
+          fs.writeFileSync(configFilePath,
+            `module.exports = {
+                sourceDir 'path/to/fake/source/dir',
```
kumar303: since this is only missing one character it's hard to read this code and immediately see that there is a syntax error in it. Can you either rewrite it to look like a more obvious syntax error or put a comment at the top explaining what the syntax error is?

src/config.js
```diff
@@ -0,0 +1,47 @@
...
+
+export function loadJSConfigFile(filePath: string): Object {
+  log.debug(`Loading JS config file: ${filePath}`);
+  try {
+    return requireUncached(filePath);
+  } catch (error) {
+    log.debug('Handling error:', error);
+    throw new UsageError(
+        `Cannot read config file: ${filePath}\nError: ${error.message}`);
+  }
```
kumar303: I think it would be nice to add a debug warning about empty configs here since we don't have a way to detect when a user forgets to type `module.exports = ...`. Something like:
````js
if (Object.keys(configObject).length === 0) {
  log.debug(
    `Config file ${filePath} did not define any options. ` +
    `Did you set module.exports = {...}?`);
}
````
I'm suggesting `log.debug()` not `log.warn()` because there could be legitimate cases where the config won't have any keys, such as when working with a template for a new project.

Because of code coverage, you will need to add a test that tries to load an empty config but that test won't need to make any assertions. It will just be a test to make sure no errors are thrown.

File: src/program.js [hunk 30]
```diff
@@ -90,7 +92,15 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
+    this.defaultValues = {};
     Object.keys(options).forEach((key) => {
+      const camelCasedKey = camelCase(key);
+      if (options[key].type === 'boolean') {
+        this.defaultValues[camelCasedKey] = 'defaultValueExists';
```
kumar303: This should be set to the actual default value, which for booleans would be:
````js
this.defaultValues[camelCasedKey] = false;
````

File: src/config.js [hunk 31]
```diff
@@ -0,0 +1,53 @@
...
+  let configObject;
+  try {
+    configObject = requireUncached(filePath);
+  } catch (error) {
+    log.debug('Handling error:', error);
+    throw new UsageError(
+        `Cannot read config file: ${filePath}\nError: ${error.message}`);
+  }
+  if (Object.keys(configObject).length === 0) {
+    log.debug('Config file ${filePath} did not define any options. ' +
```
kumar303: this needs to be a template string

tests/unit/test.config.js
```diff
@@ -0,0 +1,249 @@
...
+  userCmd?: Array<string>,
+  command?: string,
+  commandDesc?: string,
+  commandExecutor?: Function,
+  commandOpt?: Object,
+  globalOpt?: Object,
+|}
+
+function makeArgv({
+  userCmd = [],
```
kumar303: To make this a bit more useful, you could set the default to:
````js
userCmd = ['fakecommand'],
````

File: tests/unit/test.config.js [hunk 33]
```diff
@@ -0,0 +1,249 @@
...
+      const configObject = {
+        sourceDir: '/configured/source/dir',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.sourceDir, cmdLineSrcDir);
+    });
+
+    it('preserves configured value over default', () => {
+      const {argv, defaultValues} = makeArgv({
+        userCmd: ['fakecommand'],
```
kumar303: You can now remove this since the default will set it correctly

tests/unit/test.config.js
```diff
@@ -0,0 +1,249 @@
...
+      const configObject = {
+        sourceDir: '/configured/source/dir',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.sourceDir, cmdLineSrcDir);
+    });
+
+    it('preserves default value of option if not in config', () => {
+      const {argv, defaultValues} = makeArgv({
+        userCmd: ['fakecommand'],
```
kumar303: You can now remove this since the default will set it correctly

tests/unit/test.config.js
```diff
@@ -0,0 +1,249 @@
...
+      const configObject = {
+        foo: '/configured/foo',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.sourceDir, cmdLineSrcDir);
+    });
+
+    it('uses a configured boolean value over an implicit default', () => {
+      const {argv, defaultValues} = makeArgv({
+        userCmd: ['fakecommand'],
```
kumar303: You can now remove this since the default will set it correctly

tests/unit/test.config.js
```diff
@@ -0,0 +1,249 @@
...
+      const configObject = {
+        overwriteFiles: true,
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.overwriteFiles, true);
+    });
+
+    it('uses a configured boolean value over an explicit default', () => {
+      const {argv, defaultValues} = makeArgv({
+        userCmd: ['fakecommand'],
```
kumar303: You can now remove this since the default will set it correctly

src/config.js
```diff
@@ -0,0 +1,56 @@
...
+
+export function applyConfigToArgv({
+  argv,
+  configObject,
+  defaultValues = {},
+}: ApplyConfigToArgvParams): Object {
+  const newArgv = {...argv};
+  for (const option in configObject) {
+    const wasValueSetOnCLI = typeof(newArgv[option]) !== 'undefined' &&
+      (argv[option] !== defaultValues[option]);
```
kumar303: For consistency, use either `argv` or `newArgv` in this entire statement. Either one is fine. Since you're only reading the old values, I think `argv` would make a bit more sense.

src/config.js
```diff
@@ -0,0 +1,56 @@
...
+|};
+
+export function applyConfigToArgv({
+  argv,
+  configObject,
+  defaultValues = {},
+}: ApplyConfigToArgvParams): Object {
+  const newArgv = {...argv};
+  for (const option in configObject) {
+    const wasValueSetOnCLI = typeof(newArgv[option]) !== 'undefined' &&
```
kumar303: When someone reads this code, it won't be immediately clear what's going on since we both know it's a hack. For situations like this, you should add a comment above it explaining in plain English why you are assuming the value was set on the command line.

src/config.js
```diff
@@ -0,0 +1,56 @@
...
+      (argv[option] !== defaultValues[option]);
+    if (wasValueSetOnCLI) {
+      log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
+        `configuration: ${option}=${configObject[option]}`);
+      continue;
+    }
+    if (!newArgv.hasOwnProperty(option) || defaultValues[option] ||
```
kumar303: There are a couple things I'd like to see changed here:

* Check for an `undefined` value instead of using `hasOwnProperty` because that will work even in a future case where yargs uses prototype inheritance to build up an `argv` object (it's unlikely but possible)
* Remove the second part of the if statement because it's redundant now that your `wasValueSetOnCLI` statement is working

Here is a suggested patch:
````diff
index 3735ae6..4a92a7e 100644
--- a/src/config.js
+++ b/src/config.js
@@ -26,14 +26,15 @@ export function applyConfigToArgv({
         `configuration: ${option}=${configObject[option]}`);
       continue;
     }
-    if (!newArgv.hasOwnProperty(option) || defaultValues[option] ||
-      typeof(newArgv[option]) === 'boolean' ||
-      typeof(newArgv[option]) === 'number') {
-
-      newArgv[option] = configObject[option];
-      log.debug(`Favoring configuration: ${option}=${configObject[option]} ` +
-        `over CLI: ${option}=${argv[option]}`);
+    if (typeof(argv[option]) === 'undefined') {
+      log.debug(`Ignoring config option: ${option}=${configObject[option]} ` +
+        'because this is an unknown option');
+      continue;
     }
+
+    newArgv[option] = configObject[option];
+    log.debug(`Favoring configuration: ${option}=${configObject[option]} ` +
+      `over CLI: ${option}=${argv[option]}`);
   }
   return newArgv;
 }
@@ -53,4 +54,4 @@ export function loadJSConfigFile(filePath: string): Object {
       'Did you set module.exports = {...}?');
   }
   return configObject;
-}
\ No newline at end of file
+}

````

File: src/config.js [hunk 40]
```diff
@@ -0,0 +1,58 @@
...
+    // not the same as that on the argv object as there is a very rare chance
+    // this happening
+    const wasValueSetOnCLI = typeof(argv[option]) !== 'undefined' &&
+      (argv[option] !== defaultValues[option]);
+    if (wasValueSetOnCLI) {
+      log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
+        `configuration: ${option}=${configObject[option]}`);
+      continue;
+    }
+    // We want to include options of types string, boolean and number
```
kumar303: This comment isn't entirely true since it will include more than these types listed (like Object types). You can probably just remove the comment since the debug message below will explain what's happening.

src/config.js
```diff
@@ -0,0 +1,58 @@
...
+    if (argv[option] === undefined) {
+      log.debug(`Favoring configuration: ${option}=${configObject[option]} ` +
+        `over CLI: ${option}=${argv[option]}`);
```
kumar303: This debug message isn't correct. It should say:
````js
log.debug(`Ignoring configuration: ${option}=${configObject[option]} ` +
  'because this is an unknown option');
````

File: src/program.js [hunk 42]
```diff
@@ -90,7 +92,16 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
+    this.defaultValues = {};
     Object.keys(options).forEach((key) => {
+      const camelCasedKey = camelCase(key);
+      if (options[key].type === 'boolean') {
+        this.defaultValues[camelCasedKey] = false;
+      }
+      // We want to include options of types string, boolean and number
```
kumar303: This comment also isn't totally true because it can include object types

src/config.js
```diff
@@ -0,0 +1,58 @@
...
+    // this happening
+    const wasValueSetOnCLI = typeof(argv[option]) !== 'undefined' &&
+      (argv[option] !== defaultValues[option]);
+    if (wasValueSetOnCLI) {
+      log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
+        `configuration: ${option}=${configObject[option]}`);
+      continue;
+    }
+    // We want to include options of types string, boolean and number
+    if (argv[option] === undefined) {
```
kumar303: Sorry, I was wrong earlier about the prototype inheritance problem when using `hasOwnProperty()`. That method was the correct one to use because it's definitely possible for a *defined* option to be given a value of undefined (this is one of the biggest head scratchers in the JavaScript language but it is what it is).

Here is a patch for a test that fails with the way you have it here along with a fix:

````diff
index f33206b..ee0f0d0 100644
--- a/src/config.js
+++ b/src/config.js
@@ -30,7 +30,7 @@ export function applyConfigToArgv({
       continue;
     }
     // We want to include options of types string, boolean and number
-    if (argv[option] === undefined) {
+    if (!argv.hasOwnProperty(option)) {
       log.debug(`Favoring configuration: ${option}=${configObject[option]} ` +
         `over CLI: ${option}=${argv[option]}`);
       continue;
@@ -55,4 +55,4 @@ export function loadJSConfigFile(filePath: string): Object {
       'Did you set module.exports = {...}?');
   }
   return configObject;
-}
\ No newline at end of file
+}
diff --git a/tests/unit/test.config.js b/tests/unit/test.config.js
index fae0bd8..5753c37 100644
--- a/tests/unit/test.config.js
+++ b/tests/unit/test.config.js
@@ -210,6 +210,22 @@ describe('config', () => {
       assert.strictEqual(newArgv.sourceDir, cmdLineSrcDir);
     });
 
+    it('uses configured value even when option defaults to undefined', () => {
+      const {argv, defaultValues} = makeArgv({
+        globalOpt: {
+          'source-dir': {
+            type: 'string',
+            default: undefined,
+          },
+        },
+      });
+      const configObject = {
+        sourceDir: '/configured/directory',
+      };
+      const newArgv = applyConfigToArgv({argv, configObject, defaultValues});
+      assert.strictEqual(newArgv.sourceDir, '/configured/directory');
+    });
+
     it('uses a configured number value over a falsey default', () => {
       const {argv, defaultValues} = makeArgv({
         userCmd: ['fakecommand'],
````

File: src/config.js [hunk 44]
```diff
@@ -0,0 +1,57 @@
...
+    newArgv[option] = configObject[option];
+  }
+  return newArgv;
+}
+
+export function loadJSConfigFile(filePath: string): Object {
+  log.debug(`Loading JS config file: ${filePath}`);
+  let configObject;
+  try {
+    configObject = requireUncached(filePath);
```
kumar303: I think this got lost in the churn. I had requested this change:
````js
const resolvedFilePath = path.resolve(filePath);
log.debug(
  `Loading JS config file: "${filePath}" (resolved to "${resolvedFilePath}")`);
let configObject;
try {
  configObject = requireUncached(resolvedFilePath);
}
...
````
Be sure to change all `filePath` variables to `resolvedFilePath`.

Without this change, a relative path might get mistaken for a node module. It's a little too hard to test this scenario so don't worry about adding a test.

kumar303:
Something else touched `package.json` so you'll need to resolve the conflict. Just be sure to use the right version numbers of each package when resolving it.

shubheksha:
I resolved the conflict and incorporated the other change as well. Ready for another review.

PR #746, opened 2017-01-11
feat: Added reload from the terminal by pressing a key

### LINKED ISSUE #705: Trigger an extension reload from the CLI with keyboard input

feature request

When using `web-ext run`, the extension will be automatically reloaded when you edit the source. 

It would be nice to trigger a reload manually from the CLI. You can make a new blank line in one of your files and save it but that's a hacky workaround. 

I found myself needing such a feature because I had erased some logs and I wanted to reload my background script to see the startup logging again. I don't know if people would need it much but it should be easy to implement. We could just listen for `Enter` or `R` or whatever key and do the reload.

saintsebastian:
Fixes #705

Would love to here ideas about testing, my idea is [mocking stdin](https://github.com/caitp/node-mock-stdin), but I am open to suggestions.


src/cmd/run.js
```diff
@@ -256,6 +273,19 @@ export default async function run(
         );
       }
 
+      readline.emitKeypressEvents(process.stdin);
+      if (process.stdin.isTTY) {
+        log.info('Press R to reload');
+        (process.stdin: any).setRawMode(true);
```
rpl: This `(process.stdin: any).setRawMode(...)` looks like something that is hiding a flow validation error, am I right?

we should try to fix the flow validation error properly (and not by using the `any` type workaround which is too broad), e.g. it seems that the issue could be that flow is not able to know if process.stdin is going to be a tty.ReadStream (https://nodejs.org/api/tty.html#tty_class_tty_readstream), and so my guess is that something like this should make flow happy:

```
if (process.stdin instanceof tty.ReadStream) {
  process.stdin.setRawMode(true);
}
```

src/cmd/run.js
```diff
@@ -256,6 +273,19 @@ export default async function run(
         );
       }
 
+      readline.emitKeypressEvents(process.stdin);
```
rpl: This would be probably better moved inside the `if (process.stdin.isTTY) { ... }` below.

saintsebastian:
@rpl the issue you've mentioned may cause troubles, is there some way (automated testing, virtual machine) i can use to make sure this feature actually works outside of my system?

rpl:
> @rpl the issue you've mentioned may cause troubles, is there some way (automated testing, virtual machine) i can use to make sure this feature actually works outside of my system?

@saintsebastian I just tried on a Windows7 system and it worked correctly (with the interactive reload and exit on Ctrl-C as on OSX and Linux) with both cmd and ConEmu.

@saintsebastian currently there are conflicts with master (related to PR #753 landed recently), do you mind to merge the updated master into your branch to resolve the conflicts? 
(the final "squash and merge" will squash all the commit from this PR into a single commit and so the merge commit will not be visible in the master history)

saintsebastian:
@rpl please, take a look at this feature after the merging

src/cmd/run.js
```diff
@@ -44,16 +47,18 @@ export type WatcherCreatorParams = {|
   desktopNotifications?: typeof defaultDesktopNotifications,
   ignoreFiles?: Array<string>,
   createFileFilter?: FileFilterCreatorFn,
+  addonReload?: typeof defaultAddonReload,
 |};
 
+
```
rpl: extra empty line

src/cmd/run.js
```diff
@@ -44,16 +47,18 @@ export type WatcherCreatorParams = {|
   desktopNotifications?: typeof defaultDesktopNotifications,
```
rpl: This is not anymore a parameter of the `defaultWatchCreator` utility method, we should remove it from this type as well.

src/cmd/run.js
```diff
@@ -63,23 +68,48 @@ export function defaultWatcherCreator(
     sourceDir,
     artifactsDir,
     onChange: () => {
```
rpl: It looks that this can now be rewritten as a one line:

```
...
onChange: () => addonReload({addonId, client}),
...
```

(when the arrow function is composed by a single expression, we can omit the curly brackets and the arrow function will return the value of that single expression without the need to explicitly return it).

src/cmd/run.js
```diff
@@ -44,16 +47,18 @@ export type WatcherCreatorParams = {|
...
   createFileFilter?: FileFilterCreatorFn,
+  addonReload?: typeof defaultAddonReload,
 |};
 
+
 export type WatcherCreatorFn = (params: WatcherCreatorParams) => Watchpack;
 
 export function defaultWatcherCreator(
   {
     addonId, client, sourceDir, artifactsDir, ignoreFiles,
```
rpl: it looks like `addonId` and `client` are also unused in defaultWatcherCreator now (they are parameters of the `addonReload`)

src/cmd/run.js
```diff
@@ -63,23 +68,48 @@ export function defaultWatcherCreator(
...
 
+export type ReloadParams = {
+  addonId: string,
+  client: RemoteFirefox,
+  desktopNotifications?: typeof defaultDesktopNotifications,
+};
+
+export function defaultAddonReload(
+  {
+  addonId, client, desktopNotifications = defaultDesktopNotifications,
```
rpl: It looks like that the indentation needs some tweaks here, e.g.:
```
export function defaultAddonReload(
  {
    addonId, client, 
    desktopNotifications = defaultDesktopNotifications,
  }: ReloadParams
): Promise<void> {
  ...
}
```

src/cmd/run.js
```diff
@@ -63,23 +68,48 @@ export function defaultWatcherCreator(
...
+export type ExitParams = {
+  logger?: typeof log,
+  killer?: typeof process.kill,
+};
+
+export async function defaultExitProgram(
+  {logger = log, killer = process.kill}: ExitParams = {},
+): Promise<void> {
+  logger.info('Exiting web-ext on user request');
+  killer(process.pid, 'SIGINT');
```
rpl: @saintsebastian is this needed to be sure that web-ext cleanups everything when the user exits using Ctrl-C?

src/cmd/run.js
saintsebastian: looks like my merge wasn't as clean as I thought! thanks for finding this

src/cmd/run.js
saintsebastian: @rpl I am in no way an expert on node, however, after you pointed out that `process.exit()` wasn't the best strategy, I looked for some answers and it seems, that sending `'SIGINT'` allows graceful shutdown and it also allows us to `process.on('SIGINT', function () {...}` in case we want to clean up something which is not taken care of. I am open to any suggestion and better sources of course!

src/cmd/run.js
saintsebastian: @rpl what's the best way to handle this in your opinion? Have a second set of types, so the function of both is clear like in `defaultWatcherCreator ({...WatcherCreatorParams}, {...ReloadParams})` or pass them like `defaultWatcherCreator (addend, client, {...ReloadParams})` ?

rpl: @saintsebastian actually I think that we can leave it like it is right now 

(initially I was thinking that we could pass an `addonReload` utility method that do not need to receive the `addonId` and the `client` from the `defaultWatcherCreator`, e.g. by capture them into a closure, which is absolutely possible, but after re-reading it again I changed my mind and I think that we can leave to `defaultWatcherCreator` the responsability of providing `addonId` and `client` to the `addonReload` function)

src/cmd/run.js
rpl: @saintsebastian I did some testing on Windows (and some additional reading about the "how to cleanly exit a nodejs app") and it seems that the current implementation (the one based on `SIGINT`) has a couple of issues:

- web-ext does not exit anymore when the Firefox instance started by web-ext is closed (this is probably related to the new listener attached to the `process.stdin`, and it happens on every platform, I directly tried only on Linux and Windows, but given the kind of issue OSX has probably the same issue)

- on Windows, web-ext does not remove the "temporarely created profile" anymore (it does on Linux, and probably also on OSX)

The above "Windows specific issue" seems to be related to the `"SIGINT"` strategy, which probably doesn't behave correctly on Window (e.g. it exits the process buy the `"SIGINT"` event is probably missing, and the `"firefox-profile"` npm dependency subscribe a listener to the `"exit"` and `"SIGINT"` events on the `process` object to be able to cleanup the "temporarily generated profile")

I did some further experimentations on Windows and it seems that if web-ext exits using the `process.exit` method, the `"exit"` event is correctly received on Windows (as well as on Linux and OSX).

Unfortunately, on Windows if the Firefox instance is still running, `"firefox-profile"` fails to remove the generate temporary profile dir (because of an `"EBUSY: resource busy or locked, unlink 'C:\Users\..."` error).

The following is the only workaround that I found to fix both the above issues:

- define a global module var (e.g. `let userExitRequest = false;`) to be able to recognize when a Firefox instance is exiting because of an explicit user request

- at line 312, when Ctrl-C shortcut has been received, set to true the defined global module var (e.g. `userExitRequest = true;`) to recognize that we are killing the running Firefox instance because of a user request, and then kill the Firefox instance using `runningFirefox.kill()`

- at line 143, when the "close" event has been emitted by `firefoxProcess`, check if the user has requested to exit and in that case use `process.exit(0)` to exit it (with a "success" process exit value), on the contrary if the Firefox instance has been closed for unknown reasons we should probably use `process.exit(1)` (Unfortunately it seems that we need to force the nodejs process in both the scenarios, or web-ext will never exit if it is running in "interactive mode" and the Firefox instance closes itself without any user interventions).

By killing the running Firefox instance, and exiting the nodejs process once it has been closed, it seems that Windows is able to unlink the temporary generated profile as expected.

Note: the main problem with using `process.exit` to exit a nodejs application seems to be related to the flushing of the stdout/stderr data, that is asynchronous.

kumar303: Yeah, I think using a variable to flag user-initiated requests sounds like a good strategy.

saintsebastian:
@rpl Unfortunately this is still not ready. Biggest question is how to test new exit and flow type error. 
`307:             runningFirefox.kill();
                                ^^^^ property `kill`. Property not found in
307:             runningFirefox.kill();
                 ^^^^^^^^^^^^^^ FirefoxProcess`
My next commit is going to delegate exit to utility function, so it can be tested better.

rpl:
Hi @saintsebastian the kill method is not recognized but it is there (https://nodejs.org/api/child_process.html#child_process_child_kill_signal)

we should be able to fix the flow error by declaring the kill method in this flow type definition:

https://github.com/mozilla/web-ext/blob/master/src/firefox/index.js#L96

Let me know how it goes.

src/cmd/run.js
```diff
@@ -312,14 +310,18 @@ export default async function run(
         log.info('Press R to reload (and Ctrl-C to quit)');
         if (process.stdin instanceof tty.ReadStream) {
           process.stdin.setRawMode(true);
+          process.stdin.on('keypress', (str, key) => {
+            if (key.ctrl && key.name === 'c') {
+              if (process.stdin instanceof tty.ReadStream) {
```
saintsebastian: This is really awkward, but flow forgets about the previous check.

src/cmd/run.js
```diff
@@ -99,16 +97,19 @@ export function defaultAddonReload(
...
+  {
+    firefox, stdin, logger = log,
+  }: ExitParams = {},
 ): Promise<void> {
-  logger.info('Exiting web-ext on user request');
-  userExitRequest = true;
-  runningFirefox.kill();
+  log.info('\nExiting web-ext on user request');
+  firefox.kill();
+  stdin.pause();
```
saintsebastian: Since it is not a `process.exit()` anymore I decided to place it here instead. Without `firefox.kill();` pause wouldn't work

src/cmd/run.js
rpl: @saintsebastian that's weird, but I think that there is a better workaround:
we can extract the `process.stdin` property into its own `const` first and then used it in the code that follows, that should convince flow that `stdin` is not going to change its type, something like:

```js
      const {stdin} = process;
      if (stdin.isTTY && stdin instanceof tty.ReadStream) {
        readline.emitKeypressEvents(process.stdin);
        log.info('Press R to reload (and Ctrl-C to quit)');
        stdin.setRawMode(true);
        stdin.on('keypress', (str, key) => {
          if (key.ctrl && key.name === 'c') {
            exitProgram({firefox: runningFirefox, stdin});
          } else if (key.name === 'r' && addonId) {
            addonReload({addonId, client});
          }
        });
      }
```

How that sounds to you?

tests/unit/test-cmd/test.run.js
```diff
@@ -221,6 +225,35 @@ describe('run', () => {
     });
   });
 
+  it('can reload when user presses R in shell console', () => {
+    const cmd = prepareRun();
+    const {addonReload} = cmd.options;
+
+    return cmd.run({noReload: false})
+      .then(() => {
+        process.stdin.emit('keypress', 'r', {name: 'r', ctrl: false});
```
rpl: Given that we are using `mock-stdin` elsewhere, it would be better to use it also to test it here.

saintsebastian:
in unexpected turn of events travis tests failed, yet appveyor tests passed!

rpl:
@saintsebastian ouch, it seems to fail because of a nodejs crash (related to an assertion, apparently related to libuv), I'm going to take a look at it asap, I'd like to check if it fails with the same crash locally on my Linux system (where it would be more easily debuggable).

rpl:
Hi @saintsebastian 
I did some debugging of this crash: https://travis-ci.org/mozilla/web-ext/jobs/208220672#L1554-L1557
(basically the reason why the tests are failing on travis).

Unfortunately I've not being able to reproduce the crash locally, but after some digging (and running the test on travis a bunch of times ;-)) I've been able to identify how the crash happens:

The crash happens in the `tests/unit/test.program.js` test file, where we create and inject a `fakeProcess`, that is created using the `fake` test helper. This `fakeProcess` is the one that generates the crash.

To double-check my theory I made the following changes to the `tests/unit/test.program.js` and `tests/units/helpers.js` modules:

- In `tests/unit/helpers.js`, I changed fake to optionally support an array of skipped properties, that are not going to be added to the created `fake object`, and I created a new helper function named `createFakeProcess` that creates a fake process and skips the `stdio` property (and the `EventEmitter` which raises a deprecation warning when accessed by the `fake` method)

```diff
diff --git a/tests/unit/helpers.js b/tests/unit/helpers.js
index ae49bde..3649795 100644
--- a/tests/unit/helpers.js
+++ b/tests/unit/helpers.js
@@ -149,7 +149,9 @@ export function makeSureItFails(): Function {
  *
  */
  // $FLOW_IGNORE: fake can return any kind of object and fake a defined set of methods for testing.
-export function fake<T>(original: Object, methods: Object = {}): T {
+export function fake<T>(
+  original: Object, methods: Object = {}, skipProperties: Array<string> = []
+): T {
   var stub = {};

   // Provide stubs for all original members:
@@ -162,7 +164,8 @@ export function fake<T>(original: Object, methods: Object = {}): T {

   var proto = Object.getPrototypeOf(original);
   for (const key of props) {
-    if (!original.hasOwnProperty(key) && !proto.hasOwnProperty(key)) {
+    if (skipProperties.indexOf(key) >= 0 ||
+        (!original.hasOwnProperty(key) && !proto.hasOwnProperty(key))) {
       continue;
     }
     const definition = original[key] || proto[key];
@@ -190,6 +193,9 @@ export function fake<T>(original: Object, methods: Object = {}): T {
   return stub;
 }

+export function createFakeProcess() {
+  return fake(process, {}, ['EventEmitter', 'stdin']);
+}
```

- then in `tests/unit/test.program.js` I changed every line that creates the `fakeProcess` object to use the new `createFakeProcess` helper:

```diff
diff --git a/tests/unit/test.program.js b/tests/unit/test.program.js
index 6d432be..599586f 100644
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -10,14 +10,19 @@ import {assert} from 'chai';
 import {defaultVersionGetter, main, Program} from '../../src/program';
 import commands from '../../src/cmd';
 import {onlyInstancesOf, UsageError} from '../../src/errors';
-import {fake, makeSureItFails, ErrorWithCode} from './helpers';
+import {
+  createFakeProcess,
+  fake,
+  makeSureItFails,
+  ErrorWithCode,
+} from './helpers';
 import {ConsoleStream} from '../../src/util/logger';


 describe('program.Program', () => {

   function execProgram(program, options = {}) {
-    const fakeProcess = fake(process);
+    const fakeProcess = createFakeProcess();
     const absolutePackageDir = path.join(__dirname, '..', '..');
     return program.execute(
       absolutePackageDir, {
@@ -58,7 +63,7 @@ describe('program.Program', () => {
   });

   it('exits 1 on a thrown error', () => {
-    const fakeProcess = fake(process);
+    const fakeProcess = createFakeProcess();
     const program = new Program(['cmd'])
       .command('cmd', 'some command', () => {
         throw new Error('this is an error from a command handler');
@@ -280,7 +285,7 @@ describe('program.main', () => {
       getVersion: () => 'not-a-real-version',
       checkForUpdates: spy(),
       shouldExitProgram: false,
-      systemProcess: fake(process),
+      systemProcess: createFakeProcess,
     };
     return main(projectRoot, {argv, runOptions, ...mainOptions});
   }
```

The above changes seems to be enough to prevent the tests from crashing nodejs.

package.json
```diff
@@ -73,6 +73,7 @@
     "tmp": "0.0.30",
     "watchpack": "1.3.0",
     "update-notifier": "2.0.0",
+    "watchpack": "1.2.1",
```
rpl: this line should be removed (watchpack is already included in the dependencies at line 74)

src/cmd/run.js
```diff
@@ -62,24 +65,36 @@ export function defaultWatcherCreator(
...
-          log.error('\n');
-          log.error(error.stack);
-          desktopNotifications({
-            title: 'web-ext run: error occurred',
-            message: error.message,
-          });
-          throw error;
-        });
-    },
+    onChange: () => addonReload({addonId, client}),
```
rpl: If I'm not wrong, these `addonReload/ ReloadParams/ defaultAddonReload` can be reverted: in the current version of this change, the `addonReload` helper seems to be no different from the previous version of this `onChange` callback.

Reverting these changes should help to greatly simplify this change.

src/cmd/run.js
```diff
@@ -303,7 +339,7 @@ export type ExtensionRunnerParams = {|
   startUrl?: string | Array<string>,
 |};
 
-export class ExtensionRunner {
+export class DefaultExtensionRunner {
```
rpl: can we leave this named as `ExtensionRunner` instead of renaming it to `DefaultExtensionRunner`?

tests/unit/test-cmd/test.run.js
```diff
@@ -51,6 +61,8 @@ describe('run', () => {
       reloadStrategy: sinon.spy(() => {
         log.debug('fake: reloadStrategy()');
       }),
+      addonReload: sinon.spy(),
+      ExtensionRunner: DefaultExtensionRunner,
```
rpl: it shouldn't be necessary to pass `ExtensionRunner` explicitly if we are using its default value.

tests/unit/test-cmd/test.run.js
```diff
@@ -403,6 +434,7 @@ describe('run', () => {
       const args = {
         addonId: 'some-addon@test-suite',
         client,
+        // $FLOW_IGNORE: fake can return any kind of object and fake a defined set of methods for testing.
```
rpl: I think that we can remove this flow suppress comment, by explicitly define the kill property on the `StubChildProcess`, e.g.:

```diff
diff --git a/tests/unit/test-cmd/test.run.js b/tests/unit/test-cmd/test.run.js
index f4dc88d..6898664 100644
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -424,6 +424,7 @@ describe('run', () => {
     class StubChildProcess extends EventEmitter {
       stderr = new EventEmitter();
       stdout = new EventEmitter();
+      kill = sinon.spy(() => {});
     }

     function prepare() {
@@ -434,7 +435,6 @@ describe('run', () => {
       const args = {
         addonId: 'some-addon@test-suite',
         client,
-        // $FLOW_IGNORE: fake can return any kind of object and fake a defined set of methods for testing.
         firefoxProcess: new StubChildProcess(),
         profile: {},
         sourceDir: '/path/to/extension/source',
```

tests/unit/test-cmd/test.run.js
```diff
@@ -347,42 +409,11 @@ describe('run', () => {
...
-          assert.equal(config.client.reloadAddon.called, true);
-          const reloadArgs = config.client.reloadAddon.firstCall.args;
+          assert.equal(config.addonReload.called, true);
+          const reloadArgs = config.addonReload.firstCall.args;
           assert.ok(config.addonId);
-          assert.equal(reloadArgs[0], config.addonId);
-        });
-    });
-
-    it('notifies user on error from source change handler', () => {
```
rpl: If I'm not wrong, once we revert the changes related to the `addonReload` helper, these changes to the tests are not needed anymore and they can also be reverted (and the resulting change should be shorter and simpler)

src/cmd/run.js
```diff
@@ -272,6 +293,21 @@ export default async function run(
         );
       }
 
+      if (stdin.isTTY && stdin instanceof tty.ReadStream) {
+        readline.emitKeypressEvents(stdin);
+        log.info('Press R to reload (and Ctrl-C to quit)');
+        stdin.setRawMode(true);
+        stdin.on('keypress', (str, key) => {
```
rpl: I've been thinking a bit about this, and I'm wondering if this can be rewritten into an `async function` and a `while` loop (mostly like we are planning to do in the 'web-ext create' PR), e.g. something like:

```js
if (stdin.isTTY && stdin instanceof tty.ReadStream) {
  readline.emitKeypressEvents(stdin);
  stdin.setRawMode(true);

  // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
  // before the callback attached to the Promise returned by this function (and it allows the `run` function
  // to not be stuck in the while loop).
  Promise.resolve().then(async function() {
    log.info('Press R to reload (and Ctrl-C to quit)');

    let userExit = false;

    while (!userExit) {
      const keyPressed = await new Promise((resolve) => {
        stdin.once('keypress', (str, key) => resolve(key));
      });

      if (keyPressed.ctrl && keyPressed.name === 'c') {
        userExit = true;
      } else if (keyPressed.name === 'r' && addonId) {
        await addonReload({addonId, client});
      }
    }

    log.info('\nExiting web-ext on user request');
    runningFirefox.kill();
    stdin.pause();
  });
}
```

The above version has also the following advantages:
- no listener to the stdin can be leaked
- if the user press 'r' multiple times in fast sequence, it will not generate a queue of addon reloads request (because the keypress listener is not subscribed until the next loop, after the addon has been restarted).

src/cmd/run.js
rpl: Actually, I was wrong, I was forgetting that the `addReload` utility function is needed by the keypress handler.

@saintsebastian As per discussion over email, you can ignore this comment and only handle the rest of the review comments.

saintsebastian:
@rpl I found your note in the comment for run.js very helpful and I left it untouched. I believe it may be necessary since the existing solution is a result of almost 2 months of cooperation and experiments in different PR as well.

rpl:
@kumar303 this PR is now ready for a final look:
This change has presented some interesting challenges (e.g. we are introducing user interaction through the stdin stream for the first time, we had to carefully handle any risk of cross-"operating systems" compatibility issues, the `fake` test helper presented an interesting nodejs crash on travis, which has been currently fixed by tweaking the test helper, and more :-)) and so it is one of those changes where I'd like to get one additional round of review from you before landing it.

@saintsebastian Thank you so much for all the passion, hard work (and patience :-)) that you have put in it during the past reviews on this PR! :+1:

saintsebastian:
@rpl this was lots of fun! thanks for sticking with me all along!

src/cmd/run.js
```diff
@@ -272,6 +293,36 @@ export default async function run(
...
+        readline.emitKeypressEvents(stdin);
+        stdin.setRawMode(true);
+
+        // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
+        // before the callback attached to the Promise returned by this function (and it allows the `run` function
+        // to not be stuck in the while loop).
+        Promise.resolve().then(async function() {
+          log.info('Press R to reload (and Ctrl-C to quit)');
+
+          let userExit = false;
```
kumar303: To expand on my comment above about moving this logic to `defaultReloadStrategy`, you're going to need to handle the case where Firefox quits. In other words, try these steps:

* Execute `web-ext run`
* Let Firefox open up
* Go to File > Quit in the Firefox application to close it

After this, `web-ext` should stop running the while loop to check `stdin` for keyboard input but currently it does not. If you move your logic to `defaultReloadStrategy` then you can hook something into the already existing close event:

````js
  firefoxProcess.on('close', () => {
    client.disconnect();
    watcher.close();
    // stop polling stdin...
  });
````

**Comment by saintsebastian** [comment]
@kumar303 hey Kumar! i think this is ready for further review

File: src/cmd/run.js [hunk 20]
```diff
@@ -104,18 +121,49 @@ export function defaultReloadStrategy(
...
+      let userExit = false;
+
+      while (!userExit) {
+        const keyPressed = await new Promise((resolve) => {
+          stdin.once('keypress', (str, key) => resolve(key));
+        });
+
+        if (keyPressed.ctrl && keyPressed.name === 'c') {
+          userExit = true;
+        } else if (keyPressed.name === 'r' && addonId) {
```
kumar303: I think it would be nice to log a debug message here so that we know the reload was triggered by a key press

tests/unit/test-cmd/test.run.js
```diff
@@ -449,6 +431,90 @@ describe('run', () => {
...
+          stdin.emit('keypress', 'r', {name: 'r', ctrl: false});
+        })
+        .then(() => {
+          assert.ok(stdin.setRawMode.called);
+          assert.ok(addonReload.called);
+          assert.equal(addonReload.firstCall.args[0].addonId,
+            tempInstallResult.addon.id);
+        });
+    });
+
```
kumar303: nit: we usually only separate functions by one space (but I guess we don't have an eslint rule for it)

tests/unit/test.program.js
```diff
@@ -280,7 +285,7 @@ describe('program.main', () => {
       getVersion: () => 'not-a-real-version',
       checkForUpdates: spy(),
       shouldExitProgram: false,
-      systemProcess: fake(process),
+      systemProcess: createFakeProcess,
```
kumar303: I think was supposed to be `createFakeProcess()`, right?

src/cmd/run.js
```diff
@@ -104,18 +121,49 @@ export function defaultReloadStrategy(
...
+
+    // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
+    // before the callback attached to the Promise returned by this function (and it allows the `run` function
+    // to not be stuck in the while loop).
+    Promise.resolve().then(async function() {
+      log.info('Press R to reload (and Ctrl-C to quit)');
+
+      let userExit = false;
+
+      while (!userExit) {
```
saintsebastian: @kumar thanks for catching this, indeed defaults prevent the loop from starting, so it only happens in two tests, in one it doesn't exit since it is waiting for possible input after reload in another it exits on ctrl+c. Hope I understood your correctly and implemented this the way you imagined

saintsebastian:
@kumar303 thanks for catching all these things!

src/cmd/run.js
```diff
@@ -104,18 +121,50 @@ export function defaultReloadStrategy(
...
+        const keyPressed = await new Promise((resolve) => {
+          stdin.once('keypress', (str, key) => resolve(key));
+        });
+
+        if (keyPressed.ctrl && keyPressed.name === 'c') {
+          userExit = true;
+        } else if (keyPressed.name === 'r' && addonId) {
+          log.info('\nReloading extension on user request');
```
kumar303: This should be `log.debug()` because we only want it to be printed in `--verbose` mode. Since we don't reset the cursor in verbose mode, you won't need a `\n` (new line) in the message.

src/cmd/run.js
```diff
@@ -104,18 +121,50 @@ export function defaultReloadStrategy(
...
     watcher.close();
+    stdin.pause();
   });
 
+  if (stdin.isTTY && stdin instanceof tty.ReadStream) {
+    readline.emitKeypressEvents(stdin);
+    stdin.setRawMode(true);
+
+    // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
+    // before the callback attached to the Promise returned by this function (and it allows the `run` function
```
kumar303: I think it's no longer accurate to say this is "executed before the callback attached to the Promise returned by this function..." (since the function no longer returns a promise, right?) -- could you update the comment? It's important to get this comment right because I had to read it several times to understand what this code is doing :)

tests/unit/test-cmd/test.run.js
```diff
@@ -410,14 +386,16 @@ describe('run', () => {
         ignoreFiles: ['first/file', 'second/file'],
       };
       const options = {
+        addonReload: sinon.spy(() => Promise.resolve()),
         createWatcher: sinon.spy(() => watcher),
+        stdin: fakeStdin,
```
kumar303: this could just be: `stdin: new stream.Readable(),`

saintsebastian:
@kumar303 thanks for such a fast review! Hopefully most of the work is done and I can replicate what I've learnt here in other PRs.

PR #747, opened 2017-01-11
docs: adding new cli option

saintsebastian:
Added information on adding new cli options and some general CLI info to Contributing Guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,20 +141,21 @@ To add a command option locate the relevant command description (i.e. `run`).
 Here is an example (note the trailing comma):
 
 ````
-'new-option': {
-  describe: 'A wonderful new option with useful functionality.',
-  alias: ['alias'],
+'file-path': {
+  describe: 'An absolute file path.',
+  alias: ['fp'],
   demand: false,
   requiresArg: true,
-  type: 'boolean',
+  type: 'string',
   coerce: nameOfCoerceFunction,
 },
 ````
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -127,6 +127,38 @@ so as not to disturb any real `addons.mozilla.org` data.
...
+
+````
+'new-option': {
+  describe: 'A wonderful new option with useful functionality.',
+  alias: ['alias'],
+  demand: false,
+  requiresArg: true,
+  type: 'boolean',
+  coerce: nameOfCoerceFunction,
+},
```
kumar303: This looks good but it's a little bit abstract which will make it hard for someone to understand. If it were me, I'd be wondering what is `new-option` and what is `alias`? I would suggest trying to make it a bit more realistic, like:
````
'file-path': {
  describe: 'An absolute file path.',
  alias: ['fp'],
  demand: false,
  requiresArg: true,
  coerce: nameOfCoerceFunction,
},
````
That way you can refer to it as `--file-path=./path/to/file` which looks more like a realistic option. You can talk about `--fp` as the alias which makes a bit more sense. I also removed the `boolean` type because `requiresArg: true` contradicts that.

saintsebastian:
@kumar303 Thanks! I tried to recreate how I've learned doing that, will try add better examples next time.

rpl:
@saintsebastian :+1: Thanks a lot!

Issue #739, opened 2017-01-05
Improve the CONTRIBUTING.md file by including more informations about the tools and conventions used


rpl:
### Is this a bug or feature request?

it is a **doc** **feature** to better **fix** **bugs** :-)

This is basically a **tracking** issue, it is unlikely that we can add everything in a single pull request that completely close this issue.

### What is the current behavior?

There are a number of "recurrent patterns" and "often needed information" that could be useful to include in the CONTRIBUTING.md file, so that they are easier to find (and reference in the single PRs). 

### What is the expected or desired behavior?

The CONTRIBUTING.md file should include more information related to the coding style and common patterns used in the web-ext internals, eg.:

Most important docs:

- [x] flowtypes:
  - why flow? what is it used for? link to the official docs
  - how to recognize the Flow vs. ES6 syntaxes? how to read the errors?
  - common conventions currently used in the codebase (e.g. flowtype names)
- [x] command line command / long and short options conventions
  - what we use to parse the command line? (yargs) link to the docs 
  - how the command line is parsed and how the parsed options "flow" into the internals?
  - how to add a command, how to add an option?
  - conventions used for the command and options (how to propose new and check that the
    new options are consistent with the conventions)

Nice to have:

- [ ] tests:
  - what is tested as a unit? when we need a new functional test?
  - what are the tools used and what they do?(mocha, chai, sinon) link to the offial docs
  - common conventions currently used in the codebase (e.g. dependencies injection and test fakes)
- [ ] coding styles conventions:
  - what is a linting error? how to fix it?
  - code formatting requirements/tips that do not generate linting errors (e.g. vertical alignment,
    formatting objects, how to break long lines nicely)
- [ ] more (suggestions are welcome in the comments, so that we can discuss them and grow this list accordingly)

kumar303:
These all look important but it will take some time to write all of this up in the docs. Without it, we'll just have to keep offering feedback in PR reviews which is okay because that's usually only needed for new contributors. 

If I had to choose, the most important thing on this list is the docs for how to add a new command and how that flows through the program.

rpl:
@kumar303 :+1: list splitted and reordered accordingly

shubheksha:
I'll try to work on this one thing at a time. 😄

rpl:
@shubheksha :heart:

we can simply start by creating one of these new sections and by including into it the most common/useful information that have been added to the past pull requests on that topic, during the reviews we will refine it together

@kumar303 how that sounds to you?

[...]

PR #749, opened 2017-01-12
docs: Added more docs related to the flow errors

rpl:
This PR adds some "flowtype" docs to the CONTRIBUTING.md file (as part of #739).

Add more info in the intro to flow:

- [x] small introduction to how to learn more about Flow
- [x] link to the official Flow docs

include some helpful example: 

- [x] missing annotation (the simplest one, used to introduce "how to fix a flow error" on the simplest scenario)
- [x] "How to read Flow errors related to type inconsistencies" (used to explain how to read and evaluate the kind of flow error message that is usually harder to read)
- [x] Common "Flow types definitions" conventions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,89 @@ the test suite without lint checks:
 This project relies on [Flow] to ensure functions and
 classes are used correctly. Run all Flow checks with `npm run flow-check`.
 
+The first steps to learn how to fix flow errors are:
+
+- learn how to read the flow annotations
+- learn how to write new or change the existent type definitions
+- learn to read the flow errors and know some of the more common errors
+
+To learn more about the syntax used to add the flow annotations and how to
+write/change the type definitions, you should take a look at the official
+[Flow docs](https://flowtype.org/docs/getting-started.html)
+
+Follows some additional information related to common flow errors and how
+to read them.
+
+### Missing annotation
+
+This is a pretty common flow error and it is usually the simplest to fix.
+
+It means that the new code added in the sources doesn't define the types
+of the functions and methods parameters, e.g. on the following snippet:
+
+```js
+export default async function getValidatedManifest(sourceDir) {
+  ...
+}
+```
+
+flow is going to raise the error:
+
+```
+src/util/manifest.js:32
+ 32:   sourceDir
+       ^^^^^^^^^ parameter `sourceDir`. Missing annotation
+```
+
+which is fixed by annotate the function correctly, e.g.:
+
+```js
+export default async function getValidatedManifest(
+  sourceDir: string
+): Promise<ExtensionManifest> {
+  ...
+}
+```
+
+### How to read Flow errors related to type inconsistencies
+
+Some of the flow errors are going to contains references to the two sides
+of the flowtype errors:
+
+```
+tests/unit/test-cmd/test.build.js:193
+193:         manifestData: basicManifest,
+                           ^^^^^^^^^^^^^ property `applications`. Property not found in
+ 24: export type ExtensionManifest = {|
+                                     ^ object type. See: src/util/manifest.js:24
+
+```
+
+- The first part points to the offending code (where the type violation has been found)
+- The second part points to the offended type annotation (where the type has been defined)
+
+When flow raises this kind of errors (e.g. it is pretty common during a refactoring),
+we have to evaluate which one of the two sides is wrong.
+
+As an example, by reading the above error is not immediately clear which part should be fixed,
+To be sure about which is the proper fix, we have to look at the code near to both the lines
+and evaluate the actual reason (that can be different from time to time), e.g.:
+
+- it is possible that we wrote some of property names wrong (in the code or in the type definitions)
+- or the defined type is supposed to contains a new property and it is not yet in the related type definitions
+
+### Flow type definitions conventions
+
+In the `web-ext` sources we are currently using the following conventions, that should be preserved
+when we change or add flow types definitions:
+
+- the type names should be CamelCased (e.g. `ExtensionManifest`)
+- the types used to annotate functions or methods exported from the module should be exported
+  (`export type ExtensionManifest = ...`)
+- object types should be exact object types (e.g. `{| ... |}`), because flow will be able to raise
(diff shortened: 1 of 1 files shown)


rpl:
@kumar303 here is a proposal of an expanded version of the existent section related to flow, I started by including some of the informations that can be helpful to a new contributor.

I'm obviously open to new suggestion ;-)

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
 This project relies on [Flow] to ensure functions and
 classes are used correctly. Run all Flow checks with `npm run flow-check`.
 
+The first steps to learn how to fix flow errors are:
+
+- learn how to read the flow annotations
+- learn how to write new or change the existent type definitions
```
kumar303: typo: learn how to write new type definitions or change the existing definitions

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+- learn to read the flow errors and know some of the more common errors
+
+To learn more about the syntax used to add the flow annotations and how to
+write/change the type definitions, you should take a look at the official
+[Flow docs](https://flowtype.org/docs/getting-started.html)
+
+Follows some additional information related to common flow errors and how
```
kumar303: typo: The following is additional information...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+
+flow is going to raise the error:
+
+```
+src/util/manifest.js:32
+ 32:   sourceDir
+       ^^^^^^^^^ parameter `sourceDir`. Missing annotation
+```
+
+which is fixed by annotate the function correctly, e.g.:
```
kumar303: typo: which is fixed by annotating...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+export default async function getValidatedManifest(
+  sourceDir: string
+): Promise<ExtensionManifest> {
+  ...
+}
+```
+
+### How to read Flow errors related to type inconsistencies
+
+Some of the flow errors are going to contains references to the two sides
```
kumar303: typo: are going to contain...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+tests/unit/test-cmd/test.build.js:193
+193:         manifestData: basicManifest,
+                           ^^^^^^^^^^^^^ property `applications`. Property not found in
+ 24: export type ExtensionManifest = {|
+                                     ^ object type. See: src/util/manifest.js:24
+
+```
+
+- The first part points to the offending code (where the type violation has been found)
+- The second part points to the offended type annotation (where the type has been defined)
```
kumar303: It's rare to use 'offended' in this context so I think it would be a little confusing to native English readers. I had to think on it for a minute because it may be technically correct, just awkward. I would suggest replacing 'offended' with 'violated'.

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+
+When flow raises this kind of errors (e.g. it is pretty common during a refactoring),
```
kumar303: typo: raises this kind of error...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+we have to evaluate which one of the two sides is wrong.
+
+As an example, by reading the above error is not immediately clear which part should be fixed,
```
kumar303: typo: by reading the above error it is not...

kumar303: This should be the end of the sentence: ...which part should be fixed.

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+To be sure about which is the proper fix, we have to look at the code near to both the lines
+and evaluate the actual reason (that can be different from time to time), e.g.:
```
kumar303: nit: remove '(that can be different from time to time)' to make the sentence more concise

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+
+- it is possible that we wrote some of property names wrong (in the code or in the type definitions)
```
kumar303: typo: we wrote some of the property...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+- or the defined type is supposed to contains a new property and it is not yet in the related type definitions
```
kumar303: typo: type is supposed to contain...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+
+### Flow type definitions conventions
```
kumar303: nit: reads slightly better to me as: Flow type conventions

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+
+In the `web-ext` sources we are currently using the following conventions, that should be preserved
```
kumar303: I think you meant to split this sentence into two clauses? An easy fix would be:  ...we are currently using the following conventions; they should be preserved...

CONTRIBUTING.md
```diff
@@ -85,6 +85,89 @@ the test suite without lint checks:
...
+when we change or add flow types definitions:
+
+- the type names should be CamelCased (e.g. `ExtensionManifest`)
+- the types used to annotate functions or methods exported from the module should be exported
+  (`export type ExtensionManifest = ...`)
```
kumar303: we only need to export types that are shared by other modules, which is worth clarifying

kumar303: related to exporting/importing, let's mention that we prefer to write all type definitions in code close to the function they annotate and export/import them when necessary. We do not use "include" files. Maybe this deserves a separate bullet point or maybe you can weave it into that same bullet point.

Issue #239, opened 2016-05-03
Defer temp dir teardown until the process exits


kumar303:
There is a pattern internal to web-ext that is becoming cumbersome and hard to deal with. It looks like:
````javascript
import {withTempDir} from '../../src/util/temp-dir';
withTempDir(
  (tmpDir) => Promise.resolve(tmpDir.path()))
  .then((tmpPath) => {
    // Do something with tmpPath.
  })
});
````
Running a promise inside of a callback is necessary because the temp dir is torn down after the promise succeeds or fails. We can just make the temp dir clean up after itself when the process exits. However, there may still be some cases where we want immediate cleanup, such as a long running session where a file watcher will need to create many temp dirs.

kumar303:
Hmm, I'm not so sure that deferring the temp dir teardown until the process exits would be so good for [auto-reloading](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Automatic_extension_reloading) performance. That's something to keep an eye on.

PR #751, opened 2017-01-15
test: Windows compatibilities

eight04:
This PR makes `npm test` be able to pass under Windows 7, Node 7.0.0. However `functional/test.cli.run.js` still failed since web-ext uses `fx-runner` to spawn the browser, which is a fake .js file. Any idea how to fix this?

The original `stream-to-promise` doesn't ensure that the file descriptor is closed. This sometimes makes `fs.rmdir` throws during cleaning the tempdir when testing.

A small test script:
```javascript
var streamToPromise = require("stream-to-promise"),
	fs = require("fs");

function streamToPromise2(stream) {
	return new Promise(resolve => {
		stream.on("close", resolve);
	});
}
	
function test(toPromise, fn) {
	return new Promise(resolve => {
		var i = 0;
		(function loop() {
			var stream = fs.createWriteStream(fn),
				pending = toPromise(stream);
				
			stream.end("test");
			
			pending.then(() => {
				fs.unlinkSync(fn);
				var files = fs.readdirSync(".");
				if (files.includes(fn)) {
					console.log("fd is still opened");
					resolve();
					return;
				}
				if (i < 1000) {
					i++;
					loop();
				} else {
					console.log("all tests are passed");
					resolve();
				}
			}).catch(err => {
				console.log(err);
				resolve();
			});
		})();
	});
}

test(streamToPromise, "test.txt")
	.then(() => test(streamToPromise2, "test2.txt"));
```

### DIFF
--- a/src/util/stream-to-promise.js
+++ b/src/util/stream-to-promise.js
@@ -1,9 +1,0 @@
-/* @flow */
-
-export async function streamToPromise(
-  stream: stream$Readable | stream$Writable
-) {
-  return new Promise((resolve) => {
-    stream.on('close', resolve);
-  });
-}
--- a/src/util/stream-to-promise.js
+++ b/src/util/stream-to-promise.js
@@ -1,0 +1,9 @@
+/* @flow */
+
+export async function streamToPromise(
+  stream: stream$Readable | stream$Writable
+) {
+  return new Promise((resolve) => {
+    stream.on('close', resolve);
+  });
+}
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -91,7 +91,8 @@ export function execWebExt(
 ): RunningWebExt {
 
   const spawnedProcess = spawn(
-      process.execPath, [webExt, ...argv], spawnOptions);
+    process.execPath, [webExt, ...argv], spawnOptions
+  );
 
   const waitForExit = new Promise((resolve) => {
     let errorData = '';
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -90,6 +90,12 @@ export type RunningCommand = {|
 export function execCommand(
   execPath: string, argv: Array<string>, spawnOptions: child_process$spawnOpts,
 ): RunningCommand {
+  // bin/web-ext can't be directly spawned on Windows
+  if (execPath === webExt && process.platform === 'win32') {
+    argv.unshift(execPath);
+    execPath = process.execPath;
+  }
+
   const spawnedProcess = spawn(execPath, argv, spawnOptions);
   const waitForExit = new Promise((resolve) => {
     let errorData = '';
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -72,31 +72,27 @@ export function reportCommandErrors(obj: Object, msg: ?string) {
   throw error;
 }
 
-// execCommand helper
+// execWebExt helper
 
-export type RunCommandResult = {|
+export type RunWebExtResult = {|
   exitCode: number,
   stderr: string,
   stdout: string,
 |};
 
-export type RunningCommand = {|
-  execPath: string,
+export type RunningWebExt = {|
   argv: Array<string>,
-  waitForExit: Promise<RunCommandResult>,
+  waitForExit: Promise<RunWebExtResult>,
   spawnedProcess: ChildProcess,
 |};
 
-export function execCommand(
-  execPath: string, argv: Array<string>, spawnOptions: child_process$spawnOpts,
-): RunningCommand {
-  // bin/web-ext can't be directly spawned on Windows
-  if (execPath === webExt && process.platform === 'win32') {
-    argv.unshift(execPath);
-    execPath = process.execPath;
-  }
+export function execWebExt(
+  argv: Array<string>, spawnOptions: child_process$spawnOpts,
+): RunningWebExt {
 
-  const spawnedProcess = spawn(execPath, argv, spawnOptions);
+  const spawnedProcess = spawn(
+      process.execPath, [webExt, ...argv], spawnOptions);
+
   const waitForExit = new Promise((resolve) => {
     let errorData = '';
     let outputData = '';
@@ -113,5 +109,5 @@ export function execCommand(
     });
   });
 
-  return {execPath, argv, waitForExit, spawnedProcess};
+  return {argv, waitForExit, spawnedProcess};
 }
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -74,7 +74,7 @@ export function reportCommandErrors(obj: Object, msg: ?string) {
 
 // execWebExt helper
 
-export type RunWebExtResult = {|
+export type WebExtResult = {|
   exitCode: number,
   stderr: string,
   stdout: string,
@@ -82,7 +82,7 @@ export type RunWebExtResult = {|
 
 export type RunningWebExt = {|
   argv: Array<string>,
-  waitForExit: Promise<RunWebExtResult>,
+  waitForExit: Promise<WebExtResult>,
   spawnedProcess: ChildProcess,
 |};
 
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -17,7 +17,9 @@ export const webExt = path.join(projectDir, 'bin', 'web-ext');
 export const fixturesDir = path.join(functionalTestsDir, '..', 'fixtures');
 export const addonPath = path.join(fixturesDir, 'minimal-web-ext');
 export const fakeFirefoxPath = path.join(
-  functionalTestsDir, 'fake-firefox-binary.js'
+  functionalTestsDir,
+  process.platform === 'win32' ?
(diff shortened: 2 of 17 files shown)


eight04:
I guess we should make badManifest a separate file. Maybe place in `fixures/badManifest.json`.

kumar303:
Thanks for the patch. Before I take a closer look, I'm updating the pull request title to match our [format](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages) (we use this for squash and merge which eventually auto-generates us a changelog too). This is why the test suite failed on this one and your other pull request.

tests/unit/test-cmd/test.build.js
```diff
@@ -90,7 +92,12 @@ describe('build', () => {
     );
   });
 
-  it('checks locale file for malformed json', () => {
+  // FIXME: Failed on Node 7.
+  // https://github.com/sindresorhus/parse-json/issues/6
+  it('checks locale file for malformed json', function() {
+    if (semver.gte(process.versions.node, '7.0.0')) {
+      return this.skip();
```
kumar303: We try to stick to [LTS](https://github.com/nodejs/LTS) pretty closely and Node 7 isn't there yet. I'd rather let this fail for now since we don't officially support Node 7.

tests/functional/common.js
```diff
@@ -90,6 +90,12 @@ export type RunningCommand = {|
 export function execCommand(
   execPath: string, argv: Array<string>, spawnOptions: child_process$spawnOpts,
 ): RunningCommand {
+  // bin/web-ext can't be directly spawned on Windows
+  if (execPath === webExt && process.platform === 'win32') {
+    argv.unshift(execPath);
+    execPath = process.execPath;
+  }
+
   const spawnedProcess = spawn(execPath, argv, spawnOptions);
```
kumar303: if you change this to:

    const spawnedProcess = spawn(process.execPath, webExt, argv, spawnOptions);

won't it work for all platforms?

Every call to it begins with `execCommand(webExt...)` so we could maybe just rename it to `execWebExt()`.

tests/unit/test-cmd/test.build.js
```diff
@@ -44,6 +45,7 @@ describe('build', () => {
             fileNames.sort();
             assert.deepEqual(fileNames,
                              ['background-script.js', 'manifest.json']);
+            return zipFile.close();
```
kumar303: Ugh, all of these would be way easier with try/catch/finally. We need https://github.com/mozilla/web-ext/issues/755 first though. This was a nasty bug catch, thanks.

tests/unit/test-util/test.manifest.js
```diff
@@ -60,23 +61,30 @@ describe('util/manifest', () => {
...
-          .then(makeSureItFails())
-          .catch(onlyInstancesOf(InvalidManifest, (error) => {
-            assert.match(error.message, /Error parsing manifest\.json at /);
-            assert.include(error.message, 'Unexpected token \' \' at 2:49');
-            assert.include(error.message, manifestFile);
-          }));
+    // FIXME: Failed on Node 7.
+    // https://github.com/sindresorhus/parse-json/issues/6
+    it('reports an error for invalid manifest JSON', function() {
+      if (semver.gte(process.versions.node, '7.0.0')) {
```
kumar303: Again, I think we should let it fail for Node 7 for now

tests/unit/test.watcher.js
```diff
@@ -34,7 +34,7 @@ describe('watcher', () => {
           });
         })
         .then((watcher) => {
-          return fs.utimes(someFile, Date.now(), Date.now())
+          return fs.utimes(someFile, Date.now() / 1000, Date.now() / 1000)
```
kumar303: whoops, good catch

kumar303:
> I guess we should make badManifest a separate file. Maybe place at fixures/badManifest.json.

Why? The nice thing about defining it at the top of the test function is that you can see it easily while reading the test without having to open a separate file.

tests/unit/test-cmd/test.build.js
eight04: OK, let me revert 3139e5d.

eight04:
Because the test needs to be rewritten each time the indent changed. Anyway, 51fea52 should fix that.

eight04:
Oops, I forgot to change the title.

src/util/stream-to-promise.js
```diff
@@ -0,0 +1,9 @@
+/* @flow */
+
+export async function streamToPromise(
```
kumar303: Why did you have to replace [stream-to-promise](https://github.com/eight04/web-ext/blob/e58888bae779b65985ca5951e003d3b420506fce/package.json#L66)? Is it broken on Windows? Can you [patch it upstream](https://github.com/bendrucker/stream-to-promise) instead? If you really need to replace it (I hope not) then the patch needs some changes:
* All the various error callbacks need to result in promise rejections
* I'd like to see some debug logging
* We'll need unit test cases for all possible stream to promise scenarios
* `stream-to-promise` should be removed from `package.json`

tests/functional/common.js
```diff
@@ -72,25 +72,27 @@ export function reportCommandErrors(obj: Object, msg: ?string) {
...
-export function execCommand(
-  execPath: string, argv: Array<string>, spawnOptions: child_process$spawnOpts,
-): RunningCommand {
-  const spawnedProcess = spawn(execPath, argv, spawnOptions);
+export function execWebExt(
+  argv: Array<string>, spawnOptions: child_process$spawnOpts,
+): RunningWebExt {
+
+  const spawnedProcess = spawn(
+      process.execPath, [webExt, ...argv], spawnOptions);
```
kumar303: nit: all block continuations should only be indented by two spaces. Sorry, we don't have a lint rule for it because it's too complicated :clown_face:

tests/functional/common.js
```diff
@@ -72,25 +72,27 @@ export function reportCommandErrors(obj: Object, msg: ?string) {
   throw error;
 }
 
-// execCommand helper
+// execWebExt helper
 
-export type RunCommandResult = {|
+export type RunWebExtResult = {|
```
kumar303: Maybe `WebExtResult` is fine?

tests/functional/test.cli.run.js
```diff
@@ -1,14 +1,21 @@
...
-  withTempAddonDir, execCommand, reportCommandErrors,
+  addonPath, fakeFirefoxPath,
+  withTempAddonDir, execWebExt, reportCommandErrors,
 } from './common';
 
 const EXPECTED_MESSAGE = 'Fake Firefox binary executed correctly.';
 
 describe('web-ext run', () => {
+  before(function() {
+    // you can't spawn .js file as an executable on Windows
```
kumar303: would it be possible to put a `fake-firefox-binary.bat` file in the same directory that simply executes `fake-firefox-binary.js`? If it's a hassle, we could leave it as a skip and file an issue to fix it later.

src/util/stream-to-promise.js
eight04: [stream-to-promise](https://github.com/bendrucker/stream-to-promise) depends on [end-of-stream](https://github.com/mafintosh/end-of-stream), which [calls the callback on `finish` event](https://github.com/mafintosh/end-of-stream/blob/master/index.js#L64).

[From Node.js document](https://nodejs.org/api/stream.html#stream_event_finish),

> The 'finish' event is emitted after the stream.end() method has been called, and all data has been flushed to the underlying system.

If we need to wait for the release of the file descriptor, we should listen to [`close` event](https://nodejs.org/api/stream.html#stream_event_close).

> The 'close' event is emitted when the stream and any of its underlying resources (a file descriptor, for example) have been closed.

So the problem is: when the promise is resolved, the file is still open, which make tmp module fail to do the cleanup task and leave some junk data in the file system.

Related issue: https://github.com/mozilla/web-ext/issues/239

kumar303: Oh wow, good catch. I would suggest filing a bug upstream when you get a chance.

For `web-ext` purposes, we need the error handling and test coverage I mentioned here. However, if that feels like a big undertaking maybe you can just add `skip()` to the relating tests and file a `web-ext` issue to add our own `streamToPromise`.

eight04: If we are just using stream-to-promise to wait for file closing, would it be better to switch to something like [event-to-promise](https://www.npmjs.com/package/event-to-promise)? So that we can explicitly specify what we are waiting, i.e. a `close` event.

kumar303: Yes! I didn't know about it. Does switching to that lib get it working?

kumar303:
Looks like you'll have to update your master and merge to fix the package.json conflict.

caitmuenster:
Thanks, @eight04, and welcome onboard! Your contribution has been added to our [Recognition Wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#January_2017).

PR #753, opened 2017-01-15
feat: Added --ignore-files option and build now ignores artifacts folder

### LINKED ISSUE #131: Read file-ignore rules from a config for building XPIs

### LINKED ISSUE #186: `web-ext build` should ignore the `artifacts-dir`.
When I build my web-extension, it includes an empty directory (`web-ext-artifacts/`) where it put the previous build of the XPI. 
Since we know that's a place for the artifacts, we should probably not add it to the XPI.  (Alternatively, maybe we should skip all empty directories, in case people put stuff in their artifact directory, or build their XPIs into the root directory.)

``` sh
🌒  🕞  bwinton@15:22/reviewboard-collapse/ (master) $ web-ext build
Building web extension from /Users/bwinton/Programming/addons/WebExtensions/reviewboard-collapse
Your web extension is ready: /Users/bwinton/Programming/addons/WebExtensions/reviewboard-collapse/web-ext-artifacts/reviewboard_collapser-1.0.xpi
🌒  🕞  bwinton@15:23/reviewboard-collapse/ (master) $ unzip -l web-ext-artifacts/reviewboard_collapser-1.0.xpi
Archive:  web-ext-artifacts/reviewboard_collapser-1.0.xpi
  Length     Date   Time    Name
 --------    ----   ----    ----
        0  04-08-16 15:23   web-ext-artifacts/
    16725  04-08-16 15:23   LICENSE
     6056  04-08-16 15:23   icon.png
       97  04-08-16 15:23   README.md
     1638  04-08-16 15:23   content.js
      451  04-08-16 15:23   manifest.json
 --------                   -------
    24967                   6 files
🌒  🕞  bwinton@15:23/reviewboard-collapse/ (master) $
```

eight04:
Fixes #131, Fixes #186


eight04:
Seems that there are only 2 tests using `filesToIgnore` option. Can we replace it completely with `ignoreFiles`?
https://github.com/mozilla/web-ext/search?utf8=%E2%9C%93&q=filesToIgnore

kumar303:
> Seems that there are only 2 tests using filesToIgnore option. Can we replace it completely with ignoreFiles?

I think so. It is only used by `build` to set some defaults but it currently relies on minimatch patterns for nested directories.

src/program.js
```diff
@@ -246,6 +246,12 @@ Example: $0 --help run.
       describe: 'Show verbose output',
       type: 'boolean',
     },
+    'ignore-files': {
+      alias: 'i',
+      describe: 'Files to ignore.',
```
kumar303: This should be more explicit and should show an example of ignoring multiple files. It should say that the file must be an absolute path. I think an example of multiple paths would look like this, no? `--ignore-files /path/to/first.js /path/to/second.js`

tests/unit/test-cmd/test.build.js
```diff
@@ -48,6 +53,50 @@ describe('build', () => {
     );
   });
 
+  it('ignore artifacts dir if included by source dir when zipping', () => {
```
kumar303: rejoice! Since you moved all of the ignoring logic, you no longer need such a heavyweight test. Here's what you can do:
* you can delete all of the fixture directories (with the manifest, etc)
* you can simply assert that build is configuring the file watcher correctly with its value of `ignoreFiles`. Here are [some examples](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-cmd/test.build.js#L249) of how we do things like that.

src/util/file-filter.js
```diff
@@ -0,0 +1,87 @@
+/* @flow */
```
kumar303: Good call on moving this to its own file. You also need to move the tests out of [test.build.js](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-cmd/test.build.js#L308). The new `ignoreFiles` needs test coverage too.

src/cmd/run.js
```diff
@@ -257,14 +262,21 @@ export default async function run(
       }
 
       log.info('The extension will reload if any source file changes');
+      const createWatcher = (...args) => {
```
eight04: I have to solve #751 to do the "run" test.

src/cmd/run.js
```diff
@@ -257,14 +262,21 @@ export default async function run(
...
+        const fileFilter = new FileFilter({
+          sourceDir, artifactsDir, ignoreFiles,
+        });
+        args[0].shouldWatchFile = (file) => fileFilter.wantFile(file);
+        return defaultWatcherCreator(...args);
+      };
```
eight04: You're right. I should make `reloadStrategy` accept `ignoreFiles` and build `shouldWatchFile` inside `reloadStrategy`.

Also maybe we should remove `FileFilter` from the watcher so it doesn't depend on `FileFilter` anymore. Make `shouldWatchFile` default to `() => true`.

src/program.js
eight04: It can be relative path. [`FileFilter` resolves it with `sourceDir`](https://github.com/eight04/web-ext/blob/cae516e433e66ce6489fb9a1a65f8603f887daea/src/util/file-filter.js#L56).

src/util/file-filter.js
eight04: I found some bugs when writing tests for FileFilter.
```
    it('ignore artifactsDir and its content', () => {
      const filter = new FileFilter({
        artifactsDir: 'artifacts',
      });
      assert.equal(filter.wantFile('artifacts'), false);
      assert.equal(filter.wantFile('artifacts/some.js'), false); // assertion error
    });
```
It might be fine for `build.js`, since the `zipDir` will skip the content if the parent dir is skipped.

But it cause problem for `watcher.js`. Make the content inside `artifactsDir` becomes wanted files. This bug applies to `.git` and `node_modules` folders as well.

eight04: I'm not sure how should we fix it. Just explicitly ignore all content inside the folder?
```diff
    filesToIgnore = [
      '**/*.xpi',
      '**/*.zip',
      '**/.*', // any hidden file
+     '**/.*/**/*', // any content in the hidden folder
      '**/node_modules',
+     '**/node_modules/**/*',
    ],
```

src/cmd/run.js
kumar303: ah, yes of course! OK, let's try to land https://github.com/mozilla/web-ext/pull/751 first so you can merge it into this branch when it hits master.

src/cmd/run.js
kumar303: Yeah, there is probably a bit too much complexity here.

src/program.js
```diff
@@ -246,6 +246,13 @@ Example: $0 --help run.
       describe: 'Show verbose output',
       type: 'boolean',
     },
+    'ignore-files': {
+      alias: 'i',
+      describe: 'A list of file patterns to ignore. (example: ' +
+                '--ignore-files=path/to/first.js path/to/second.js **/*.log)',
```
wagnerand: Does passing several arguments work (without quotes), if you use `=`?

eight04: It does on my machine. I never look into yargs's doc though.

eight04:
I get stuck in writing tests for lint.

Currently we have a nice test for `FileFilter` to check if it correctly ignores `artifactsDir` and `ignoreFiles`. But how do I test if `lint()` has correctly passed `artifactsDir` and `ignoreFiles` to the file filter?

In `test.build.js`, we mock `packageCreator()` and check if the argument `fileFilter` is correctly ignoring `artifactsDir` passed to `build()` (452e382). Should I do this to other commands (i.e. lint) too?

kumar303:
btw, you can pull from master and merge it into your branch to get your test fixes \o/ https://github.com/mozilla/web-ext/pull/751

kumar303:
> how do I test if lint() has correctly passed artifactsDir and ignoreFiles to the file filter?

The way we do this is replace `FileFilter` with a mock and simply check that the mock class instance constructed with `artifactsDir`, `ignoreFiles`, etc.

This will probably require some adjustment to the `lint()` signature. Maybe something like this:

````diff
diff --git a/src/cmd/lint.js b/src/cmd/lint.js
index 75a7af4..77e550a 100644
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -2,7 +2,7 @@
 import {createInstance as defaultLinterCreator} from 'addons-linter';
 
-import {FileFilter} from './build';
+import {FileFilter as DefaultFileFilter} from './build';
 
 
@@ -60,10 +60,11 @@ export default function lint(
   }: LintCmdParams,
   {
     createLinter = defaultLinterCreator,
-    fileFilter = new FileFilter(),
+    FileFiler = DefaultFileFilter,
   }: LintCmdOptions = {}
 ): Promise<void> {
   log.debug(`Running addons-linter on ${sourceDir}`);
+  const fileFilter = new FileFilter({ artifactsDir, sourceDir, });
   const linter = createLinter({
     config: {
       logLevel: verbose ? 'debug' : 'fatal',
````

**Comment by eight04** [comment]
I think all commands are tested now.

File: src/util/file-filter.js [hunk 7]
```diff
@@ -0,0 +1,109 @@
...
+
+export type FileFilterOptions = {|
+  filesToIgnore?: Array<string>,
+  ignoreFiles?: Array<string>,
+  sourceDir?: string,
+  artifactsDir?: string,
+|};
+
+/*
+ * Allows or ignores files when creating a ZIP archive.
```
kumar303: Probably a good time to update this to: Allows or ignores files

src/util/file-filter.js
```diff
@@ -0,0 +1,109 @@
...
+      this.filesToIgnore.push(this.resolve(file));
+      // If file is a directory, ignore its content too.
+      if (/([/\\]\.{0,2})$/.test(file) || isDir) {
+        this.filesToIgnore.push(this.resolve(path.join(file, '**/*')));
+      }
+    }
+  }
+
+  /*
+   * Returns true if the file is wanted for the ZIP archive.
```
kumar303: can be updated to: Returns true if the file is wanted

src/util/file-filter.js
```diff
@@ -0,0 +1,109 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+}
+
+// a helper function to make mocking easier
+
+export type FileFilterCreatorFn = (params: FileFilterOptions) => FileFilter;
```
kumar303: I think you can just move this underneath `createFileFilter` and define it like:
````js
export type FileFilterCreatorFn = typeof createFileFilter;
````

File: tests/unit/test-cmd/test.build.js [hunk 10]
```diff
@@ -49,6 +49,25 @@ describe('build', () => {
...
+    const params = {
+      sourceDir: '/src',
+      artifactsDir: 'artifacts',
+      ignoreFiles: ['**/*.log'],
+    };
+    return build(params, {packageCreator, createFileFilter}).then(() => {
+      // ensure sourceDir, artifactsDir, ignoreFiles is used
+      assert(createFileFilter.calledWithMatch(params));
+      // ensure packageCreator received correct fileFilter
+      assert(packageCreator.calledWithMatch({fileFilter}));
```
kumar303: yep, you got it. This is the general pattern. What it does is allows us to only test the file filter in *one* place and bind all other uses of it as dependency contracts.

src/util/file-filter.js
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   *  Insert more files into filesToIgnore array.
+   */
+  addToIgnoreList(files: Array<string>, isDir: boolean = false) {
+    for (const file of files) {
+      this.filesToIgnore.push(this.resolve(file));
+      // If file is a directory, ignore its content too.
+      if (/([/\\]\.{0,2})$/.test(file) || isDir) {
+        this.filesToIgnore.push(this.resolve(path.join(file, '**/*')));
```
kumar303: I think it makes `FileFilter` less powerful when we insert a new pattern dynamically like this. What if someone doesn't want to match all subdirectories in their custom ignore rules? I don't understand why the original patterns can't just be updated. For example, can't we just call it like this to solve the problem? I.E. change the default patterns to include subdirectory matching:
````js
export class FileFilter {
 constructor({
   filesToIgnore = [
     '**/*.xpi',
     '**/*.zip',
     '**/.*/**/*', // any hidden file and folder
     '**/node_modules/**/*',
   ],
  }) {
    ...
  }
}
````
Or do we need to double them up? I would be fine with this too:

````js
export class FileFilter {
 constructor({
   filesToIgnore = [
     '**/*.xpi',
     '**/*.zip',
     '**/.*', // any hidden file
     '**/.*/**/*', // any hidden folder and its contents
     '**/node_modules/',
     '**/node_modules/**/*', // all sub-directory contents
   ],
  }) {
    ...
  }
}
````

File: src/util/file-filter.js [hunk 12]
```diff
@@ -0,0 +1,109 @@
...
+import path from 'path';
+
+import minimatch from 'minimatch';
+
+import {createLogger} from './logger';
+
+const log = createLogger(__filename);
+
+// Use this function to mimic path.resolve without resolving to absolute path.
+export const normalizeResolve = (file: string): string => {
```
kumar303: I don't fully understand why we need this function. Is it just for convenience so that someone can use an ignore path of `src/package/../data/*.json` instead of `src/data/*.json`? And the trailing slash protection is also for convenience? I feel like we shouldn't impose too many constraints on the `ignoreFiles` logic since that's up to the developer to use for customization.

eight04: https://github.com/eight04/web-ext/blob/71e3da42208cbca3fc98296c2e16c2a18f82d6fb/src/util/file-filter.js#L65
I want to make sure the result returned from `FileFilter.resolve` is consistent. Currently `fileFilter.resolve` has two behaviors. 

1. If `sourceDir` is set, all relative paths will be converted into absolute path by `path.resolve(sourceDir, file)`, which can represent as `path.join -> path.normalize -> trim trailing slash`.
2. Otherwise, just do `path.normalize -> trim trailing slash` to the path.

This make `sourceDir` optional and `fileFilter` can be run in "relative mode", which is the original behavior in `test.build.js`.

src/util/file-filter.js
eight04: It is a simpler fix, but then the users also need to pass two patterns to ignore a folder. If that is acceptable, I think we should document somewhere that "Ignoring a folder will not ignore its content".

src/util/file-filter.js
kumar303: ok, that makes sense. Consistency is good.

src/util/file-filter.js
kumar303: Yeah, I think it's acceptable. We can point this out in the documentation. I just don't want to limit what you can do with the minimatch patterns. The user should be able to configure them how they want.

src/util/file-filter.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /*
+   * Returns true if the file is wanted for the ZIP archive.
+   *
+   * This is called by zipdir as wantFile(path, stat) for each
+   * file in the folder that is being archived.
+   */
+  wantFile(path: string): boolean {
+    path = this.resolve(path);
```
kumar303: I've been thinking more about this and I don't think it's right to always assume `sourceDir` is the root. Can't we use [minimatch(..., {matchBase: true})](https://github.com/isaacs/minimatch#matchbase) to solve the problem?

eight04: I'm not sure what is the problem described here. `matchBase: true` changes the behavior of match pattern that doesn't contain slashes, make `*.js` behaves like `**/*.js`.

We normalize the path here is to make sure it can match normalized patterns, so `fileFilter.wantFile("path/to/file")` works correctly on Windows. Normally the path is already normalized since `zip-dir`, `watchpack`, and `addons-linter` always pass back full, normalized path, so this change only affect our `test.file-filter.js`.

kumar303: ah, whoops, `matchBase` is not right then. I was thinking more about how `sourceDir` gets prepended to all patterns.

tests/unit/test-cmd/test.build.js
```diff
@@ -49,6 +49,25 @@ describe('build', () => {
...
+    const fileFilter = {wantFile: () => true};
+    const createFileFilter = sinon.spy(() => fileFilter);
+    const params = {
+      sourceDir: '/src',
+      artifactsDir: 'artifacts',
+      ignoreFiles: ['**/*.log'],
+    };
+    return build(params, {packageCreator, createFileFilter}).then(() => {
+      // ensure sourceDir, artifactsDir, ignoreFiles is used
+      assert(createFileFilter.calledWithMatch(params));
```
kumar303: This error would be easier to debug if it reported the incorrect params. I suggest this instead:
````js
assert.ok(createFileFilter.called);
assert.deepEqual(createFileFilter.firstCall.args[0], params);
````

File: tests/unit/test-cmd/test.build.js (hunk 10 above)
Comment (kumar303) [inline tests/unit/test-cmd/test.build.js]: This error would be easier to debug if it reported the incorrect params. I suggest this instead:
````js
assert.ok(packageCreator.called);
assert.deepEqual(packageCreator.firstCall.args[0], params);
````

File: tests/unit/test-cmd/test.lint.js [hunk 15]
```diff
@@ -127,20 +126,22 @@ describe('lint', () => {
...
-        assert.equal(createLinter.called, true);
-        const config = createLinter.firstCall.args[0].config;
-        assert.isFunction(config.shouldScanFile);
-
-        // Simulate how the linter will use this callback.
-        config.shouldScanFile('manifest.json');
-        assert.equal(fileFilter.wantFile.called, true);
-        assert.equal(fileFilter.wantFile.firstCall.args[0], 'manifest.json');
-      });
+  it('ensure linter build fileFilter and shouldScanFile correctly', () => {
```
kumar303: This test spec is hard for me to read. How about changing it to: "configures and passes a fileFilter to the linter"

tests/unit/test-cmd/test.run.js
```diff
@@ -299,6 +300,21 @@ describe('run', () => {
...
+    it('configure fileFilter, shouldWatchFile', () => {
+      const fileFilter = {wantFile: sinon.spy()};
+      const createFileFilter = sinon.spy(() => fileFilter);
+      const {config, createWatcher} = prepare();
+      createWatcher({createFileFilter});
+      assert(createFileFilter.calledWithMatch({
+        sourceDir: config.sourceDir,
+        artifactsDir: config.artifactsDir,
+        ignoreFiles: config.ignoreFiles,
+      }));
```
kumar303: I'd like to see this assertion made more readable too. Like:
````js
assert.ok(createFileFilter.called);
assert.deepEqual(createFileFilter.firstCall.args[0], { sourceDir: ... , artifactsDir: ... });
````

File: tests/unit/test-cmd/test.run.js [hunk 17]
```diff
@@ -299,6 +300,21 @@ describe('run', () => {
...
+      const {shouldWatchFile} = config.onSourceChange.firstCall.args[0];
+      shouldWatchFile('path/to/file');
+      assert(fileFilter.wantFile.calledWith('path/to/file'));
```
kumar303: This should be made more readable too:
````js
assert.ok(fileFilter.wantFile.called);
assert.equal(fileFilter.wantFile.firstCall.args[0], 'path/to/file');
````

File: tests/unit/test-cmd/test.lint.js [hunk 18]
```diff
@@ -127,20 +126,22 @@ describe('lint', () => {
...
+    const fileFilter = {wantFile: sinon.spy(() => true)};
+    const createFileFilter = sinon.spy(() => fileFilter);
+    const {lint, createLinter} = setUp({createFileFilter});
+    const params = {
+      sourceDir: '.',
+      artifactsDir: 'artifacts',
+      ignoreFiles: ['file1', '**/file2'],
+    };
+    return lint(params).then(() => {
+      assert(createFileFilter.calledWithMatch(params));
```
kumar303: this should change to the `assert.deepEqual()` pattern too

tests/unit/test-cmd/test.lint.js
```diff
@@ -127,20 +126,22 @@ describe('lint', () => {
...
+      assert(createLinter.called);
+      const {shouldScanFile} = createLinter.firstCall.args[0].config;
+      shouldScanFile('path/to/file');
+      assert(fileFilter.wantFile.calledWith('path/to/file'));
```
kumar303: this should change to the `assert.equal(fileFilter.wantFile.firstCall.args[0], ...)` pattern too

tests/unit/test-util/test.file-filter.js
```diff
@@ -0,0 +1,114 @@
...
+      });
+      assert.equal(filter.wantFile('/src/artifacts'), false);
+      assert.equal(filter.wantFile('/src/ignore-dir'), false);
+      assert.equal(filter.wantFile('/src/ignore-dir/some.css'), true);
+      assert.equal(filter.wantFile('/src/some.js'), false);
+      assert.equal(filter.wantFile('/src/some.log'), false);
+      assert.equal(filter.wantFile('/src/other/some.js'), true);
+      assert.equal(filter.wantFile('/src/other/some.log'), false);
+      assert.equal(filter.wantFile('/src/ignore/dir/content'), true);
+      assert.equal(filter.wantFile('/src/ignore/dir/content/file.js'), false);
```
kumar303: Can you add one more to assert how absolute paths might not match some patterns?

````diff
diff --git a/tests/unit/test-util/test.file-filter.js b/tests/unit/test-util/test.file-filter.js
index 45468dd..0146aad 100644
--- a/tests/unit/test-util/test.file-filter.js
+++ b/tests/unit/test-util/test.file-filter.js
@@ -88,6 +88,8 @@ describe('util/file-filter', () => {
       assert.equal(filter.wantFile('/src/other/some.log'), false);
       assert.equal(filter.wantFile('/src/ignore/dir/content'), true);
       assert.equal(filter.wantFile('/src/ignore/dir/content/file.js'), false);
+      // This file is not ignored because it's not relative to /src:
+      assert.equal(filter.wantFile('/some.js'), true);
     });
 
   });
````

File: src/util/file-filter.js [hunk 21]
```diff
@@ -0,0 +1,110 @@
...
+   *  Insert more files into filesToIgnore array.
+   */
+  addToIgnoreList(files: Array<string>) {
+    for (const file of files) {
+      this.filesToIgnore.push(this.resolve(file));
+    }
+  }
+
+  /*
+   * Returns true if the file is wanted.
```
kumar303: I think we should add another sentence to this comment (if you agree that it is correct):

If `path` does not start with a slash, it will be treated as a path relative to `sourceDir` when matching it against all configured ignore-patterns.

src/util/file-filter.js
```diff
@@ -0,0 +1,110 @@
...
+        path.join(artifactsDir, '**/*'),
+      ]);
+    }
+  }
+
+  /**
+   *  Resolve relative path to absolute path if sourceDir is setted.
+   */
+  resolve(file: string): string {
+    if (this.sourceDir) {
```
eight04: Prepending only happens if `sourceDir` is set and `file` is relative. Should we check whether the file is relative to log the info?

src/util/file-filter.js
```diff
@@ -0,0 +1,110 @@
...
+    filesToIgnore = [
+      '**/*.xpi',
+      '**/*.zip',
+      '**/.*', // any hidden file and folder
+      '**/.*/**/*', // and the content inside hidden folder
+      '**/node_modules',
+      '**/node_modules/**/*',
+    ],
+    ignoreFiles = [],
+    sourceDir = '',
```
eight04: `sourceDir?: string;` cause parsing error in eslint. Is `sourceDir: string | typeof undefined;` correct?

eight04:
How do I fix this error?

BTW, what does the vertical bar mean in type definition? I can't find them in flow's document.
```
exports type ... = {|
	...
|};
```

kumar303:
> How do I fix this error?

Huh, looks like a Flow bug. Flow is smart enough to see that you checked the type of the variable in an if statement but it seems to *forget* the if statement after one line of code. I moved the `resolve()` call right under the if statement and it worked. However, you can see that the second line still needs protection against undefined values (which is a Flow bug). Here's my patch:
````diff
diff --git a/src/util/file-filter.js b/src/util/file-filter.js
index 23a941d..13d046a 100644
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -68,11 +68,13 @@ export class FileFilter {
    *  Resolve relative path to absolute path if sourceDir is setted.
    */
   resolve(file: string): string {
-    if (typeof this.sourceDir === 'string') {
+    if (this.sourceDir !== undefined) {
+      const resolvedPath = path.resolve(this.sourceDir, file);
       log.debug(
-        `Adding sourceDir ${this.sourceDir} to the beginning of file ${file}`
+        `Resolved path ${file} with sourceDir ${this.sourceDir || ''} ` +
+        `to ${resolvedPath}`
       );
-      return path.resolve(this.sourceDir, file);
+      return resolvedPath;
     }
     return normalizeResolve(file);
   }

````

I also made the debug statement more verbose to help detect the situation you mentioned above when the path is absolute (which is when it won't be pre-pended with `sourceDir`).

**Comment by kumar303** [comment]
> BTW, what does the vertical bar mean in type definition?

Sigh, I can't find it in the docs either. It's a very important feature. If you have a function like this:

````js
type Params = {|
  size: 'small' | 'large',
|}

function render({size}: Params) {
}
````
And you call it like this:
````js
render({size: 'small', color: 'blue'});
````
then Flow will raise an error because `color` is not a valid parameter. If you do not use the pipes when defining `Params`, Flow will not raise an error. I think this was a pretty major flaw in Flow's original implementation but they couldn't fix it without new syntax because of backwards compatibility. All object types should always use pipes (except there are a few open bugs).

File: src/util/file-filter.js [hunk 24]
```diff
@@ -0,0 +1,117 @@
...
+  sourceDir?: string,
+  artifactsDir?: string,
+|};
+
+/*
+ * Allows or ignores files.
+ */
+export class FileFilter {
+  filesToIgnore: Array<string>;
+  sourceDir: string | typeof undefined;
```
kumar303: I guess this works. I suggested the wrong thing originally but I was thinking of:
````js
sourceDir: ?string;
````
They are pretty much the same though. The one with the question marks allows it to be `null` which is the only difference.

File: src/util/file-filter.js (hunk 3 above)
Comment (kumar303) [inline src/util/file-filter.js]: These issues are both resolved in the patch now.

File: src/util/file-filter.js (hunk 22 above)
Comment (kumar303) [inline src/util/file-filter.js]: Thanks, the debug statement now shows if the path is relative or not.

File: src/util/file-filter.js [hunk 25]
```diff
@@ -0,0 +1,119 @@
...
+
+/*
+ * Allows or ignores files.
+ */
+export class FileFilter {
+  filesToIgnore: Array<string>;
+  sourceDir: ?string;
+
+  constructor({
+    filesToIgnore = [
```
rpl: The `FileFilter` now takes two parameters with two very similar names but they have different meaning, it would be nice to make it more clear by make the names a bit more descriptive (e.g. `filesToIgnore` renamed to `defaultFilesToIgnore`, and `ignoreFiles` renamed to `customFilesToIgnore`?)

eight04: If we change `ignoreFiles` to `customFilesToIgnore`, every calls to `createFileFilter` would also need to be changed to `createFileFilter({sourceDir, artifactsDir, customFilesToIgnore: ignoreFiles})`.

How about to change `filesToIgnore` to `defaultIgnorePatterns` and keep `ignoreFiles` so `createFileFilter` could be called nicely as `createFileFilter({sourceDir, artifactsDir, ignoreFiles})`?

tests/unit/test-util/test.file-filter.js
```diff
@@ -0,0 +1,125 @@
...
+      'path/to/dir/.', 'path/to/dir/..',
+    ];
+
+    it('mimic path.resolve', () => {
+      const src = '/src/';
+
+      paths.forEach((file) => {
+        assert.equal(
+          path.resolve(src, file),
+          path.resolve(src) + path.sep + normalizeResolve(file)
```
eight04: I guess it's OK? This test is to make sure:
```
normalizeResolve(relativePath) == stripRoot(path.resolve(root, relativePath), root)
```

src/program.js
```diff
@@ -246,6 +246,14 @@ Example: $0 --help run.
       describe: 'Show verbose output',
       type: 'boolean',
     },
+    'ignore-files': {
```
eight04: It seems that setting `requiresArg` would force yargs to validate `ignoreFiles` is not a empty list. Since the `ignoreFiles` defaults to an empty list, this make yargs always argue `Missing argument value: ignore-files`.

src/util/file-filter.js
rpl: @eight04 yeah, I agree that not being able to use the ES6 syntax shortcut is a pity.

What I'd like to achieve is to make it clear that the FileFilter constructor property currently named `filesToIgnore` is not overridden by the new `ignoreFiles` property, but extended.

How about rename `filesToIgnore` to `baseIgnoredPatterns` (or `commonIgnoredPatterns`) and leave `ignoreFiles` unmodified?

src/program.js
rpl: This comment is supposed to be part of the last review comments but it has not been submitted by github, and so I added it here as a separate comment.

@eight04 yargs can be very annoying sometimes :-)

Anyway, it looks like that we can use a combination of yargs option config properties to make it work as expected:

```js
    'ignore-files': {
      alias: 'i',
      describe: '...',
      demand: false,
      requiresArg: true,
      type: 'array',
    },
```

Basically it seems that we have to:
- omit the empty array currently set as its default value
- add a `demand: false` config to allow the --ignore-files option to be undefined (which is going to mean "that the option has not been specified") 
- add the `requiresArg: true` which force "--ignore-files" to have a value when specified

I briefly tried the above config locally and it seems to work as expected, 
let me know if it doesn't for you.

kumar303:
Thanks again for all your work on this and especially for all the quick turnarounds.

caitmuenster:
Nice work! 👍

PR #765, opened 2017-01-20
fix: Fixed loading changelog lint config

### LINKED ISSUE #767: Don't fail the build when a commit message is too long

bug

If the commit message is too long, the changelog linter will fail the build

I know that the changelog generator truncates commit messages but that's okay, it shouldn't cause a build failure. We hand edit the changelog after auto-generation anyway.

This is a problem mainly in greenkeeper pull requests because commit messages can get pretty long and it's out of our control. See https://github.com/mozilla/web-ext/pull/763

rpl:
Fixes #767

This PR introduces a small change in the `travis-pr-title-lint` grunt task that fixes the `conventional-changelog-lint` config loading.

Based on the failure from this Travis build https://travis-ci.org/mozilla/web-ext/jobs/193576176, we noticed that the configuration file `'.conventional-changelog-lintrc'` was not loading correctly, by looking at the `conventional-changelog-lint`, sources it looks like the reason was related to the usage of `getConfiguration` with the wrong parameters (which unfortunately doesn't generate any complains or promise rejection, it just return a default configuration based on the selected preset).

### DIFF
--- a/tasks/travis-pr-title-lint.js
+++ b/tasks/travis-pr-title-lint.js
@@ -154,7 +154,7 @@ module.exports = function(grunt) {
 
       Promise.all([
         getPreset('angular'),
-        getConfiguration('.conventional-changelog-lintrc'),
+        getConfiguration(),
       ]).then(function(results) {
         var preset = results[0];
         var configuration = results[1];
--- a/tasks/travis-pr-title-lint.js
+++ b/tasks/travis-pr-title-lint.js
@@ -1,9 +1,11 @@
 /* eslint prefer-template: 0, no-console: 0 */
 
 var exec = require('child_process').exec;
+var path = require('path');
 var https = require('https');
 
 var changelogLintPkg = require('conventional-changelog-lint');
+var fs = require('mz').fs;
 var objectValues = require('object.values');
 var objectEntries = require('object.entries');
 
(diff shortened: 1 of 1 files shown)


rpl:
@kumar303 This PR should fix the regression described by #767

tasks/travis-pr-title-lint.js
```diff
@@ -154,7 +154,7 @@ module.exports = function(grunt) {
 
       Promise.all([
         getPreset('angular'),
-        getConfiguration('.conventional-changelog-lintrc'),
```
kumar303: Oh man, I can't believe they don't raise an error for this. How about adding some annoying code to check the file since they don't raise an error? Something like:
````javascript
var config = path.join(__dirname, '..', '.conventional-changelog-lintrc');
var configStat = file.statSync(config);
if (!configStat || !configStat.isFile()) {
  throw new Error(...);
}
````

**Comment by rpl** [comment]
@kumar303 I've just updated this PR with the proposed tweaks (I opted to use 'mz/fs' which provides the promisified version of the fs module, and I added both an fs.stat and an fs.access check, and so it is going to check both that the file exists, that is readable and that is a regular file).

File: tasks/travis-pr-title-lint.js [hunk 2]
```diff
@@ -152,12 +154,30 @@ module.exports = function(grunt) {
...
+        })
+        .catch(function(err) {
+          throw new Error(`Unable to load changelog lint config: ${err.stack}`);
+        });
+
       Promise.all([
+        checkLintConfig,
         getPreset('angular'),
-        getConfiguration('.conventional-changelog-lintrc'),
+        getConfiguration(),
```
kumar303: hmm, why do you no longer need to pass in the file? Is it because we can assume it will be auto-detected? I thought it would be: `getConfiguration(config)`

rpl: yep, unfortunately the getConfiguration helper doesn't take the filename, but a series of parameter that will affect the final config file name :-)

Code References:
- https://github.com/marionebl/conventional-changelog-lint/blob/875a77d66cf09f8106fc348e85b8a6ddcbcfa55d/source/cli.js#L103-L109
- https://github.com/marionebl/conventional-changelog-lint/blob/master/source/library/get-configuration.js#L11-L17

kumar303: wow, that's confusing!

PR #766, opened 2017-01-20
test: Run unit and function tests on Window through AppVeyor

### LINKED ISSUE #761: Run tests in Windows as part of continuous integration
Now that we have Windows support via https://github.com/mozilla/web-ext/pull/751 let's not regress! I think we can use https://ci.appveyor.com/ which @rpl has been experimenting with.

rpl:
This PR aims to run the unit and functional tests on Windows on every change (fixes #761), so that we can keep an eye on any future Windows specific regressions.

All the test unit and functional tests passes, besides the test case related to git-rev-sync, which fails because it is not able to find the git repository (I'm not currently sure if it is a limitation of the git-rev-sync, or something different in the AppVeyor build environment), and so I've added a small helper called `it_skipOnWin32` which skips on AppVeyor the tests that do not currently work on Windows.

Also flow-bin seems to have some issues while running on AppVeyor, and so I opted to create a new grunt task which only run the unit and functional tests on AppVeyor (especially given that the eslint linting, flow checks and changelog linting is already happening on Travis and it would not add much running them again on AppVeyor).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -30,19 +30,23 @@ module.exports = function(grunt) {
     'webpack:functional_tests',
   ]);
 
-  grunt.registerTask('test', [
-    'lint',
-    'flowbin:check',
-    'build-tests',
-    'mochaTest:unit',
-    'mochaTest:functional',
-  ]);
+  grunt.registerTask('test', 'run linting and test suites', function() {
+    var tasks = [
+      'lint',
+      'flowbin:check',
+      'build-tests',
+      'mochaTest:unit',
+      'mochaTest:functional',
+    ];
 
-  grunt.registerTask('test-appveyor', [
-    'build-tests',
-    'mochaTest:unit',
-    'mochaTest:functional',
-  ]);
+    // TODO: enable the flowbin:check task on AppVeyor (mozilla/web-ext#773)
+    if (process.env.APPVEYOR) {
+      tasks = tasks.filter((t) => t !== 'flowbin:check');
+      grunt.log.writeln('flowbin:check task skipped because of $APPVEYOR');
+    }
+
+    grunt.task.run(tasks);
+  });
 
   grunt.registerTask('develop', [
     'flowbin:start',
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,6 +38,12 @@ module.exports = function(grunt) {
     'mochaTest:functional',
   ]);
 
+  grunt.registerTask('test-appveyor', [
+    'build-tests',
+    'mochaTest:unit',
+    'mochaTest:functional',
+  ]);
+
   grunt.registerTask('develop', [
     'flowbin:start',
     'watch:develop',
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,3 @@
-environment:
-  matrix:
-    - nodejs_version: '6'
-    - nodejs_version: '4'
-
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
@@ -15,9 +10,15 @@ matrix:
 build: off
 shallow_clone: true
 
+# Keep the following configs in sync with .travis.yml
+environment:
+  matrix:
+    - nodejs_version: '6'
+    - nodejs_version: '4'
+
 test_script:
   - node --version
   - npm --version
   - set COVERAGE=y
   - set NODE_ENV=production
-  - npm run test-appveyor
+  - npm run test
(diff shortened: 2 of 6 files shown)


appveyor.yml
```diff
@@ -0,0 +1,23 @@
+environment:
+  matrix:
```
kumar303: Ugh, I can think of many things that can get out of sync now that we need to maintain two CI config files. Most of the things will result in a build failure which is fine but some things will not. Changing Node versions is one of them. I suggest:

* Add a comment to the version list in `.travis.yml` saying: "Keep this in sync with appveyor.yml"
* Add another comment above the version list here saying "Keep this in sync with .travis.yml"

Gruntfile.js
```diff
@@ -38,6 +38,12 @@ module.exports = function(grunt) {
     'mochaTest:functional',
   ]);
 
+  grunt.registerTask('test-appveyor', [
```
kumar303: I would rather not introduce another execution path through the test suite unless we have to. It creates a maintenance burden. 

Can we just create a wrapper around the flowbin task that skips the test? It could just look for an appveyor env var and skip [like this one](https://github.com/rpl/web-ext/blob/3c7a2c7cae90a589b7d9ad6dbc9fe1f82ef60c84/Gruntfile.js#L53).

I don't mind running eslint on Windows because that's what devs will be doing locally. It would be good to have coverage that it works on Windows.

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+  - set PATH=%APPDATA%\npm;%PATH%
+  - npm install
+
+matrix:
+  fast_finish: true
+
+build: off
+shallow_clone: true
+
+test_script:
```
kumar303: The test script and env vars also needs to keep in sync with `.travis.yml`. Can add comments in both places asking to keep them in sync with one another?

tests/unit/test.program.js
```diff
@@ -434,11 +434,10 @@ describe('program.defaultVersionGetter', () => {
       });
   });
 
-  it('returns git commit information in development', () => {
+  it_skipOnWin32('returns git commit information in development', () => {
```
kumar303: I'm pretty sure the `git-rev-sync` issue is not a Windows problem but an Appveyor one. I don't recall any mention of `git-rev-sync` problems in https://github.com/mozilla/web-ext/pull/751 .

Thus, could you make this skip happen only when a custom Appveyor env var is present? 

Also, I find the `it` wrapper hard to read. If it were done with an if statement inline it's only two extra lines. That would be tolerable to me. Something like:

````javascript
it('returns git commit information in development', () => {
  if (runningOnAppveyor()) {
    return this.skip();
  }
});
````
or maybe even without a function:

````javascript
it('returns git commit information in development', () => {
  if (process.env.APPVEYOR_CI) {
    return this.skip();
  }
});
````

**Comment by rpl** [comment]
@kumar303 I've just updated this PR with the suggested tweaks

This is the last build on appveyor:
- https://ci.appveyor.com/project/rpl/web-ext/build/1.0.9

NOTE: we have to register this repo to AppVeyor.
(I've currently registered only my fork to be able to test the AppVeyor service).

File: tests/unit/test.program.js [hunk 5]
```diff
@@ -434,11 +434,15 @@ describe('program.defaultVersionGetter', () => {
       });
   });
 
-  it('returns git commit information in development', () => {
+  it('returns git commit information in development', function() {
+    if (process.env.APPVEYOR) {
+      console.log('Test skipped because of $APPVEYOR'); // eslint-disable-line no-console
```
kumar303: I'm not so sure we need a console statement here. Do we?

Gruntfile.js
```diff
@@ -30,13 +30,22 @@ module.exports = function(grunt) {
...
+    var tasks = [
+      'lint',
+      'flowbin:check',
+      'build-tests',
+      'mochaTest:unit',
+      'mochaTest:functional',
+    ];
+
+    if (process.env.APPVEYOR) {
+      tasks = tasks.filter((t) => t !== 'flowbin:check');
```
kumar303: Cool. Could you also file a bug in web-ext (or maybe in Appveyor if that's easy?) so we don't forget to fix this. I guess it's low priority but since flowbin:check seems to work on Windows AFAIK we should investigate.

tests/unit/test.program.js
rpl: :+1: an inline comment which reference the related github issue will be enough ;-)

Gruntfile.js
rpl: sure, I opened #773 and #774 (and added a reference to them in an inline comment near the skipped grunt task and the skipped testcase)

PR #768, opened 2017-01-20
chore: turn on no-shadow lint rule

kumar303:
One of our contributors got tripped up on overshadowed variables (https://github.com/mozilla/web-ext/pull/552#discussion_r82895201) and Flow might even have some bugs around it.


kumar303:
@rpl what do you think? Hopefully not too annoying?

kumar303:
@rpl will you have some time soon to take a look? No rush.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -38,14 +38,11 @@ describe('firefox', () => {
     };
 
     function createFakeFxRunner(firefoxOverrides = {}) {
-      const firefox = {
+      const process = {
```
rpl: How about calling it `firefoxProcess` or `fxProcess` instead? 
`process` is going to shadow the `process` identifier as defined by nodejs 
(that's a pity that the eslint `no-shadow` rule doesn't seem to have any config available for customized disallowed identifier)

kumar303:
Good call on the built-ins, thanks. That exposed a few more problems. I think https://github.com/mozilla/web-ext/pull/744/ is close to merge so I'm going to wait on that before landing this one just in case more fixes are needed.

PR #771, opened 2017-01-23
feat: web-ext create simple version

### LINKED ISSUE #540: web-ext new: create the directory layout for an extension
We should add a `web-ext new` command that automatically generate a `manifest.json` file and other common files to help the developer get started on a new WebExtension.

You could take a look at how [ember new](https://ember-cli.com/user-guide/#create-a-new-project) works for ideas.

The [jpm init](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/jpm#jpm_init) command is also similar but I like the general behavior of `ember new` better.

saintsebastian:
fixes #540 

I came up with a couple of question about this issue:
1.  behaviour of `web-ext new`, at the moment name of the current directory is used to generate the manifest as the name of the addon, maybe user would rather expect `web-ext new name` which would create a `name` folder and manifest file inside of it
2. the current structure and content of default `manifest.json` may not be perfect, I'd really like to know your opinion on what elements are necessary


src/cmd/index.js
```diff
@@ -3,5 +3,6 @@ import build from './build';
 import lint from './lint';
 import run from './run';
 import sign from './sign';
+import newCommand from './new';
 
-export default {build, lint, run, sign};
+export default {build, lint, run, sign, newCommand};
```
saintsebastian: I agree `web-ext init` is a better option, I'll proceed with this changes. Unless @kumar303 disagrees.

kumar303: I think renaming the command is the best solution since it will allow us to match the name of the function to the command exactly.

Here are some ideas along with how they might be mnemonically associated in the developer's mind:

* `web-ext create`
  * "Create a new web extension"
* `web-ext start`
  * "Start developing a new web extension"
  * One downside is that `npm start` is a convention which usually applies more to starting a server or a test suite, depending on the project.
* `web-ext init`
  * "Initialize a new web extension"
  * I'm not totally opposed to this, it just seems like an unnecessary abbreviation to me when we have options for shorter words.

Any of these are fine by me so pick your favorite :) I would probably choose `create` myself.

saintsebastian:
@rpl To avoid creating a file in working directory I wrote test to `chdir` into temporary folder before and after test, which may not be the best solution. Also I am still figuring out how to test [usage error](https://github.com/saintsebastian/web-ext/blob/647c0a1fb0a856aad68957fde6cf3cd70b41e99b/src/cmd/create.js#L56) because flow really doesn't like `process.stdin.isTTY` method.

tests/unit/test-cmd/test.create.js
```diff
@@ -0,0 +1,118 @@
...
+
+describe('create', () => {
+
+  afterEach(() => {
+    process.chdir(homeDir);
+  });
+
+  it('creates files including manifest with correct name', () => withTempDir(
+    (tmpDir) => {
+      process.chdir(tmpDir.path());
```
rpl: Like discussed, we should change the current work dir during the tests (and it should fix the failures of these tests on windows) here, in the afterEach and elsewhere in this test file.

package.json
```diff
@@ -113,6 +113,7 @@
     "load-grunt-tasks": "3.5.2",
     "mocha": "3.2.0",
     "mocha-multi": "0.10.0",
+    "mock-stdin": "0.3.1",
```
rpl: We can probably reuse the mock that we created for #746 and do not include this as a devDependencies.

src/cmd/create.js
```diff
@@ -0,0 +1,133 @@
...
+import mkdirp from 'mkdirp';
+import promisify from 'es6-promisify';
+
+import {createLogger} from '../util/logger';
+import {UsageError, isErrorWithCode} from '../errors';
+
+const log = createLogger(__filename);
+const defaultAsyncMkdirp = promisify(mkdirp);
+
+export type CreateParams = {|
```
rpl: we should split these properties into two parameters and flow types like we do for the other commands:
- the first one should be named `SomethingParams` (e.g. `CreateCmdParams`) and it should only contains what is supposed to be in the regular parameters of the command (e.g. `dirPath` in this case)
- the second one should be named `SomethingOptions` (e.g. `CreateCmdOptions`) and it should only contains the optional injected dependencies (`stdin?` in this case), which are used mostly to override the dependencies in the unit tests

src/cmd/create.js
```diff
@@ -0,0 +1,133 @@
...
+  stdin?: stream$Readable,
+|};
+
+export default async function create(
+  {
+    dirPath,
+    stdin = process.stdin,
+  }: CreateParams
+): Promise<void> {
+  const targetPath = path.join(process.cwd(), dirPath);
```
rpl: I'm thinking that instead of using `process.cwd()` directly here, it would be better to turn it into a parameter (which defaults to process.cwd() if missing) and/or into an injected dependency (e.g. something like `function getBaseDir(baseDir) { return baseDir || process.cwd(); }`).

This way, the command can provide much more control to the caller if needed, which should be helpful both in the unit tests and while using web-ext as a library.

rpl:
Closing this PR, we have moved this feature into its own [create-webextension](https://www.npmjs.com/package/create-webextension) npm package and [github repo](http://github.com/rpl/create-webextension) for now.

PR #782, opened 2017-01-30
feat: run extension in a custom profile

### LINKED ISSUE #123: Run an extension in a linked profile for persistence
Calling [run --firefox-profile=myprofile](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run) will copy a custom profile into a temporary profile. It will not affect the original profile. This new feature is to run the extension directly in the custom profile so that any profile changes will be saved. 

Example use case: if you're developing an extension that has a splash screen on install, you may not want to see that screen every time you launch Firefox to test it.

We should probably introduce a new option that can be called alongside the existing ones like this:
````bash
web-ext run --firefox-profile=myprofile --keep-profile-changes
````

### PR DESCRIPTION
**Pull request by saintsebastian**
Fixes #123 

Added new function and tests

### CODE REVIEW

**Comment by saintsebastian** [comment]
Important note: `--custom-prefs` are currently applied, but that need not to stay this way. Some users actually may like this side-effect.

**Comment by saintsebastian** [comment]
@rpl thanks for great suggestions!

File: tests/unit/test-firefox/test.firefox.js [hunk 1]
```diff
@@ -297,6 +297,46 @@ describe('firefox', () => {
 
   });
 
+  describe('useProfile', () => {
+
+    function withBaseProfile(callback) {
```
rpl: @saintsebastian `withBaseProfile` this utility function is identical to the one already defined by the `copyProfile` tests (at line 172 of the same file).

We should move this helper into the head of this module, so that it can be shared by both `copyProfile` and `useProfile` without duplicating its definition.

saintsebastian:
@rpl thanks for the review! curiously, windows test have failed.

rpl:
@saintsebastian that's pretty weird because on appveyor it has failed only on node 4 during the functional tests because of a `ECONNRESET` exception during the web-ext sign test:
- https://ci.appveyor.com/project/kumar303/web-ext/build/1.0.113/job/73adn7y5sfpur03h#L1086

@kumar303 can you try to re-run the job on appveyor from your user? it looks like a potential intermittent

saintsebastian:
@rpl i triggered tests again just in case

rpl:
@kumar303 this PR is now ready for your final review!

rpl:
As briefly discussed with @kumar303 over IRC, let's merge this PR!

🎉 Thanks @saintsebastian for making the new "--keep-profile-changes" option real!

saintsebastian:
@rpl my pleasure 👩🏽‍💻!

PR #790, opened 2017-02-04
fix: Ignore artifactsDir only if it is a sub directory of sourceDir

### LINKED ISSUE #789: web-ext producing empty .zip files in v1.8.0
We're using the `web-ext` tool to build Snooze Tabs. [As of v1.8.0 we're getting an empty .zip file](https://github.com/bwinton/SnoozeTabs/issues/139).

With v1.7.0, things work as expected (487k zip file):
```
➜  tmp git clone https://github.com/bwinton/SnoozeTabs.git && cd SnoozeTabs
➜  SnoozeTabs git:(master) npm install && npm install web-ext@1.7.0 && npm run package && ls -al *.zip

# (...snip build process...)

Building web extension from /Users/lmorchard/devel/SnoozeTabs/dist
Your web extension is ready: snoozetabs-1.0.9.zip
-rw-r--r--  1 lmorchard  staff  487689 Feb  3 12:28 snoozetabs-1.0.9.zip
```

With v1.8.0, things do not work as expected (22 byte empty zip file):
```bash
➜  SnoozeTabs git:(master) rm *.zip && npm install web-ext@1.8.0 && npm run package && ls -al *.zip

# (...snip build process...)

Building web extension from /Users/lmorchard/devel/SnoozeTabs/dist
Your web extension is ready: snoozetabs-1.0.9.zip
-rw-r--r--  1 lmorchard  staff  22 Feb  3 12:25 snoozetabs-1.0.9.zip
```

Additional version info:
````
➜  SnoozeTabs git:(master) node --version && npm --version && web-ext --version
v6.2.0
3.8.9
1.8.0
````

### PR DESCRIPTION
**Pull request by eight04**
Fixes #789

I tried to drop "relative mode" to make FileFilter simpler:

1. `sourceDir` defaults to `.`
2. `FileFilter` would only match files inside `sourceDir`. (prefix all patterns with `sourceDir`)
3. `sourceDir` and `artifactsDir` are resolved with `cwd`.
4. `wantFile(path)` is resolved with `cwd`.

The change in (2) makes `proxyFileChanges` unable to filter out `baseIgnoredPatterns` since it doesn't know `sourceDir`, and I just removed that test.

### DIFF
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -96,7 +96,7 @@ export class FileFilter {
    * file in the folder that is being archived.
    */
   wantFile(filePath: string): boolean {
-    const resolvedPath = path.resolve(filePath);
+    const resolvedPath = this.resolve(filePath);
     for (const test of this.filesToIgnore) {
       if (minimatch(resolvedPath, test)) {
         log.debug(
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -8,7 +8,7 @@ import {createLogger} from './logger';
 const log = createLogger(__filename);
 
 // check if target is a sub directory of src
-export const isSubDir = (src: string, target: string): boolean => {
+export const isSubPath = (src: string, target: string): boolean => {
   const relate = path.relative(src, target);
   // same dir
   if (!relate) {
@@ -58,7 +58,7 @@ export class FileFilter {
     if (ignoreFiles) {
       this.addToIgnoreList(ignoreFiles);
     }
-    if (artifactsDir && isSubDir(sourceDir, artifactsDir)) {
+    if (artifactsDir && isSubPath(sourceDir, artifactsDir)) {
       artifactsDir = path.resolve(artifactsDir);
       log.debug(
         `Ignoring artifacts directory "${artifactsDir}" ` +
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -22,7 +22,7 @@ export const isSubDir = (src: string, target: string): boolean => {
 export type FileFilterOptions = {|
   baseIgnoredPatterns?: Array<string>,
   ignoreFiles?: Array<string>,
-  sourceDir?: string,
+  sourceDir: string,
   artifactsDir?: string,
 |};
 
@@ -43,7 +43,7 @@ export class FileFilter {
       '**/node_modules/**/*',
     ],
     ignoreFiles = [],
-    sourceDir = '.',
+    sourceDir,
     artifactsDir,
   }: FileFilterOptions = {}) {
     sourceDir = path.resolve(sourceDir);
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -92,7 +92,7 @@ export class FileFilter {
    * file in the folder that is being archived.
    */
   wantFile(filePath: string): boolean {
-    const resolvedPath = this.resolve(filePath);
+    const resolvedPath = path.resolve(filePath);
     for (const test of this.filesToIgnore) {
       if (minimatch(resolvedPath, test)) {
         log.debug(
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -69,11 +69,15 @@ export class FileFilter {
   }
 
   /**
-   *  Resolve relative path to absolute path if sourceDir is setted.
+   *  Resolve relative path to absolute path with sourceDir.
    */
   resolve(file: string): string {
-    log.debug(`Resolved path ${file} with sourceDir ${this.sourceDir}`);
-    return path.resolve(this.sourceDir, file);
+    const resolvedPath = path.resolve(this.sourceDir, file);
+    log.debug(
+      `Resolved path ${file} with sourceDir ${this.sourceDir} ` +
+      `to ${resolvedPath}`
+    );
+    return resolvedPath;
   }
 
   /**
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -14,7 +14,10 @@ export const isSubDir = (src: string, target: string): boolean => {
   if (!relate) {
     return false;
   }
-  return relate[0] !== '.';
+  if (relate === '..') {
+    return false;
+  }
+  return !relate.startsWith(`..${path.sep}`);
 };
 
 // FileFilter types and implementation.
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -69,19 +69,16 @@ export class FileFilter {
   }
 
   /**
-   *  Resolve relative path to absolute path if sourceDir is setted.
-   */
-  resolve(file: string): string {
-    log.debug(`Resolved path ${file} with sourceDir ${this.sourceDir}`);
-    return path.resolve(this.sourceDir, file);
-  }
-
-  /**
    *  Insert more files into filesToIgnore array.
    */
   addToIgnoreList(files: Array<string>) {
     for (const file of files) {
-      this.filesToIgnore.push(this.resolve(file));
+      const resolvedFile = path.resolve(this.sourceDir, file);
+      log.debug(
+        `Resolved path ${file} with sourceDir ${this.sourceDir} ` +
+        `to ${resolvedFile}`
+      );
+      this.filesToIgnore.push(resolvedFile);
     }
   }
 
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -71,7 +71,7 @@ export class FileFilter {
   /**
    *  Resolve relative path to absolute path with sourceDir.
    */
-  resolve(file: string): string {
+  resolveWithSourceDir(file: string): string {
     const resolvedPath = path.resolve(this.sourceDir, file);
     log.debug(
       `Resolved path ${file} with sourceDir ${this.sourceDir} ` +
@@ -85,7 +85,7 @@ export class FileFilter {
    */
   addToIgnoreList(files: Array<string>) {
     for (const file of files) {
-      this.filesToIgnore.push(this.resolve(file));
+      this.filesToIgnore.push(this.resolveWithSourceDir(file));
     }
   }
 
@@ -100,7 +100,7 @@ export class FileFilter {
    * file in the folder that is being archived.
    */
   wantFile(filePath: string): boolean {
-    const resolvedPath = this.resolve(filePath);
+    const resolvedPath = this.resolveWithSourceDir(filePath);
     for (const test of this.filesToIgnore) {
       if (minimatch(resolvedPath, test)) {
         log.debug(
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -57,6 +57,10 @@ export class FileFilter {
     }
     if (artifactsDir && isSubDir(sourceDir, artifactsDir)) {
       artifactsDir = path.resolve(artifactsDir);
+      log.debug(
+        `Ignoring artifacts directory "${artifactsDir}" ` +
+        'and all its subdirectories'
+      );
       this.addToIgnoreList([
         artifactsDir,
         path.join(artifactsDir, '**', '*'),
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -69,16 +69,19 @@ export class FileFilter {
   }
 
   /**
+   *  Resolve relative path to absolute path if sourceDir is setted.
+   */
+  resolve(file: string): string {
+    log.debug(`Resolved path ${file} with sourceDir ${this.sourceDir}`);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/watcher.js [hunk 1]
```diff
@@ -66,8 +65,7 @@ export function proxyFileChanges(
   {artifactsDir, onChange, filePath, shouldWatchFile}: ProxyFileChangesParams
 ): void {
   if (!shouldWatchFile) {
```
kumar303: I say just remove this and make `shouldWatchFile` mandatory. It's not needed for anything except the tests (which can be fixed up). I tried this change locally so I think it will work:
````diff
diff --git a/src/watcher.js b/src/watcher.js
index 83ff86f..c42bf04 100644
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -19,7 +19,7 @@ export type OnSourceChangeParams = {|
   sourceDir: string,
   artifactsDir: string,
   onChange: OnChangeFn,
-  shouldWatchFile?: ShouldWatchFn,
+  shouldWatchFile: ShouldWatchFn,
 |};
 
 // NOTE: this fix an issue with flow and default exports (which currently
@@ -59,7 +59,7 @@ export type ProxyFileChangesParams = {|
   artifactsDir: string,
   onChange: OnChangeFn,
   filePath: string,
-  shouldWatchFile?: ShouldWatchFn,
+  shouldWatchFile: ShouldWatchFn,
 |};
 
 export function proxyFileChanges(
````

File: src/util/file-filter.js [hunk 2]
```diff
@@ -45,9 +43,10 @@ export class FileFilter {
       '**/node_modules/**/*',
     ],
     ignoreFiles = [],
-    sourceDir,
+    sourceDir = '.',
```
eight04: There are some tests using `FileFilter` without `sourceDir`. This would force them to explicitly use `sourceDir: '.'`. Maybe it is a good thing?

tests/unit/test-util/test.file-filter.js
```diff
@@ -103,22 +113,14 @@ describe('util/file-filter', () => {
...
-        );
-      });
+  describe('isSubDir', () => {
+    it('test if target is a sub directory of src', () => {
+      assert.equal(isSubDir('dist', '.'), false);
+      assert.equal(isSubDir('.', 'artifacts'), true);
+      assert.equal(isSubDir('.', '.'), false);
+      assert.equal(isSubDir('/src/dist', '/src'), false);
+      assert.equal(isSubDir('/src', '/src/artifacts'), true);
+      assert.equal(isSubDir('/src', '/src'), false);
```
kumar303: Can you add one more test to check that two completely unrelated paths return the right value? This is passing but I think it will be good to cover it.
````diff
diff --git a/tests/unit/test-util/test.file-filter.js b/tests/unit/test-util/test.file-filter.js
index a939c5d..0ee8e41 100644
--- a/tests/unit/test-util/test.file-filter.js
+++ b/tests/unit/test-util/test.file-filter.js
@@ -121,6 +121,7 @@ describe('util/file-filter', () => {
       assert.equal(isSubDir('/src/dist', '/src'), false);
       assert.equal(isSubDir('/src', '/src/artifacts'), true);
       assert.equal(isSubDir('/src', '/src'), false);
+      assert.equal(isSubDir('/firstroot', '/secondroot'), false);
     });
   });
 

````

File: src/util/file-filter.js [hunk 4]
```diff
@@ -99,7 +96,7 @@ export class FileFilter {
    * file in the folder that is being archived.
    */
   wantFile(filePath: string): boolean {
-    const resolvedPath = this.resolve(filePath);
+    const resolvedPath = path.resolve(filePath);
```
kumar303: Can you explain why this change is needed? If I put it back to `this.resolve()` then all tests still pass.

tests/unit/test-util/test.file-filter.js
```diff
@@ -1,15 +1,13 @@
...
 import {assert} from 'chai';
 
-import {FileFilter, normalizeResolve} from '../../../src/util/file-filter';
+import {FileFilter, isSubDir} from '../../../src/util/file-filter';
 
 describe('util/file-filter', () => {
 
   describe('default', () => {
-    const defaultFilter = new FileFilter();
+    const defaultFilter = new FileFilter({sourceDir: '.'});
```
kumar303: In situations where the tests rarely need to change one of the parameters, I think it's helpful to create a helper function:
````diff
diff --git a/tests/unit/test-util/test.file-filter.js b/tests/unit/test-util/test.file-filter.js
index a939c5d..34a5d4f 100644
--- a/tests/unit/test-util/test.file-filter.js
+++ b/tests/unit/test-util/test.file-filter.js
@@ -6,6 +6,11 @@ import {FileFilter, isSubDir} from '../../../src/util/file-filter';
 
 describe('util/file-filter', () => {
 
+  const newFileFilter = (params = {}) => new FileFilter({
+    sourceDir: '.',
+    ...params,
+  });
+
   describe('default', () => {
     const defaultFilter = new FileFilter({sourceDir: '.'});
 
````

File: tests/unit/test-util/test.file-filter.js [hunk 6]
```diff
@@ -74,12 +74,22 @@ describe('util/file-filter', () => {
...
       assert.equal(filter.wantFile('artifacts'), false);
       assert.equal(filter.wantFile('artifacts/some.js'), false);
     });
 
+    it('does not ignore an artifactsDir outside of sourceDir', () => {
+      const filter = new FileFilter({
+        artifactsDir: '.',
+        sourceDir: 'dist',
+      });
+      assert.equal(filter.wantFile('file'), true);
```
kumar303: it is slightly confusing to me that this is `true`. I guess it's because `file` is assumed to be relative to `sourceDir` and not relative to `.`?

src/util/file-filter.js
eight04: I want to remove some "concatenation magic". I'll put the details in next comment.

tests/unit/test-util/test.file-filter.js
eight04: Here's the detail. Currently,

1. All ignored patterns are resolved with `sourceDir`.
2. Resolve other paths (e.g. `sourceDir`, `artifactsDir`, `path` in `wantFile`) with `cwd`.
3. Since all ignored patterns are resolved with `sourceDir`, everything outside of `sourceDir` won't be ignored.

```JavaScript
const fileFilter = new FileFilter({
	sourceDir: 'dist/',
	ignoreFiles: ['some-script.js']
});
assert.equal(fileFilter.wantFile('dist/some-script.js'), false); // ignored by ignoreFiles
assert.equal(fileFilter.wantFile('some-script.js'), true);	// outside of sourceDir
```

I'd like to discuss some problems here:

**Is resolving ignored patterns with `sourceDir` make sense?** Or should we resolve ignored patterns with `cwd`? Which would look like:
```diff
const fileFilter = new FileFilter({
	sourceDir: 'dist/',
	ignoreFiles: ['some-script.js']
});
-assert.equal(fileFilter.wantFile('dist/some-script.js'), false); // ignored by ignoreFiles
+assert.equal(fileFilter.wantFile('dist/some-script.js'), true); // some-script.js is resolved with cwd instead of dist
-assert.equal(fileFilter.wantFile('some-script.js'), true);	// outside of sourceDir
+assert.equal(fileFilter.wantFile('some-script.js'), false);	// ignored by ignoreFiles
```

Second, **since we make `sourceDir` mandatory now, should we ignore all files outside of `sourceDir`?**
```diff
const fileFilter = new FileFilter({
	sourceDir: 'dist/',
	ignoreFiles: ['some-script.js']
});
assert.equal(fileFilter.wantFile('dist/some-script.js'), false); // ignored by ignoreFiles
-assert.equal(fileFilter.wantFile('some-script.js'), true);	// outside of sourceDir
+assert.equal(fileFilter.wantFile('some-script.js'), false);	// outside of sourceDir
```

src/util/file-filter.js
```diff
@@ -68,15 +72,8 @@ export class FileFilter {
    *  Resolve relative path to absolute path if sourceDir is setted.
    */
   resolve(file: string): string {
```
eight04: I guess we can completely drop this method since it is only used by `FileFilter.addToIgnoreList`.

kumar303:
Thanks for the detailed explanations. I think we should ignore everything in relation to `sourceDir`. The purpose of `FileFilter` is, in all cases, to work *within* the WebExtension's source directory. For every command -- `run`, `lint`, and `build` -- there is no need to think of anything besides what's inside the source directory.

I suggest:
* Continue to ignore all incoming patterns (`ignoreFiles`, `baseIgnoredPatterns`, `artifactsDir`) in relation to `sourceDir`
* Go back to resolving all paths in `wantFile` with `sourceDir` before applying the ignore patterns

Does that sound right? We can't really rely on `cwd` because none of the commands expect to change into the `sourceDir` while running.

eight04:
I think `artifactsDir` should be resolved with `cwd` (`.`) or we will break #789:
```
~/myaddon$ web-ext build -s dist -a .

sourceDir -> '~/myaddon/dist'
artifactsDir -> '~/myaddon'
```

For `wantFile`, if we can make sure that all paths sent to `wantFile` is absolute (how do `watchpack`, `zipdir` work?), maybe we should just change all `wantFile` calls in `test.file-filter.js` to use absolute path, and add some validation to `wantFile` simply throwing errors for relative paths.

kumar303:
> I think artifactsDir should be resolved with cwd (.) 

Ah, yes, true.

> make wantFile simply throw errors for relative paths?

I thought about that too. I'm not sure exactly how watchpack and zipdir are doing it. I think zipdir might use relative paths but it would need some investigation.

src/util/file-filter.js
```diff
@@ -7,24 +7,22 @@ import {createLogger} from './logger';
...
-    return file.slice(0, -1);
+// check if target is a sub directory of src
+export const isSubDir = (src: string, target: string): boolean => {
+  const relate = path.relative(src, target);
+  // same dir
+  if (!relate) {
+    return false;
   }
-  return file;
+  return relate[0] !== '.';
```
rpl: @kumar303 @eight04 If I'm reading it correctly, this check assumes that the `relate` path produced by `path.relative` is going to contain a `"."` as its first character only if the path is not a subdir (e.g. "../artifact"), which can be a wrong assumption if the file name starts with a `"."` (e.g. "web-ext build -s . -a .artifact")

In that case the artifact dir ".artifact" is not going to be recognized as a subdir anymore, but it is currently excluded by the default filter list (which contains "**/.*"), nevertheless if we are going to allow the customization of the base ignored patterns, the artifact dir content can be included by mistake in the zip file.

How about making this check more explicit? e.g. `return !relate.startsWith("../");`

Nit, This helper do not check if target is a dir, to make it more clear we could name it `isSubPath`.

src/util/file-filter.js
```diff
@@ -7,24 +7,22 @@ import {createLogger} from './logger';
...
-
-  // trim trailing slash
-  if (path.parse(file).base && file.endsWith(path.sep)) {
-    return file.slice(0, -1);
+// check if target is a sub directory of src
+export const isSubDir = (src: string, target: string): boolean => {
+  const relate = path.relative(src, target);
+  // same dir
+  if (!relate) {
+    return false;
```
rpl: @kumar303 @eight04 we should probably raise a UsageError if the artifact dir and the source dir are the same, but I'm not sure that it is nice to raise it from this utility funtion. (it would be probably nicer to raise the UsageError in the FileFilter constructor to make it more explicit).

eight04: I think this doesn't belong to `FileFilter`. Maybe some kind of "validation phase" after parsing cli arguments?

rpl: :+1: yep, I agree, I think that it is reasonable to check this condition in src/program.js (e.g. near here https://github.com/mozilla/web-ext/blob/master/src/program.js#L127 , or here where we raise other UsageError instances https://github.com/mozilla/web-ext/blob/master/src/program.js#L141-L152)

kumar303:
@rpl aha, great catches, thanks. In regards to the problem of `artifactsDir` being set to `sourceDir`, I think we handle that case fine by *not* ignoring all of the `sourceDir` contents (see https://github.com/mozilla/web-ext/pull/790/files#diff-e42a3bd68a308cc596e1b6d76d610728R127). Since ignoring files is the main responsibility of FileFilter, I think it will be ok for a user to set `artifactsDir` equal to `sourceDir`.

Do you think the new changes address all of your requests ok?

kumar303:
@eight04 let me echo our thanks once again as this was blocking some extension developers from upgrading web-ext -- your speedy help on this is greatly appreciated. I'll cut a new bugfix release ASAP.

PR #792, opened 2017-02-06
fix: Protect config files from declaring invalid keys

### LINKED ISSUE #772: Protect config files from declaring invalid keys
This is part of implementing https://github.com/mozilla/web-ext/issues/176

Let's say you have a config with a typo in it:
````js
module.exports = {
  sourceDirectory: './src',
};
````
This should raise an error like "sourceDirectory is not a valid config key."

This would also naturally raise an error for non-camel cased options, like:
````js
module.exports = {
  'source-dir': './src',
};
````
-> "source-dir is not a valid config key."

### PR DESCRIPTION
**Pull request by shubheksha**
Fixes #772

### CODE REVIEW

File: src/config.js [hunk 1]
```diff
@@ -10,7 +10,7 @@ const log = createLogger(__filename);
 
 type ApplyConfigToArgvParams = {|
   argv: Object,
-  configObject: Object,
+  configObject?: Object,
```
shubheksha: @kumar303, this was an unintended error, I'll fix this in the next commit.

kumar303: ok (in reply to https://github.com/mozilla/web-ext/pull/792/files#r99653520)

src/config.js
```diff
@@ -34,7 +34,8 @@ export function applyConfigToArgv({
     if (!argv.hasOwnProperty(option)) {
       log.debug(`Ignoring configuration: ${option}=${configObject[option]} ` +
         'because this is an unknown option');
-      continue;
+      throw new UsageError(`The option ${option} is invalid.` +
+        ' Please fix your config and try again.');
```
kumar303: This error message should have the path to the config so that the user can find out which config they need to fix. This means `applyConfigToArgv()` will need to accept a filename argument.

tests/unit/test.config.js
```diff
@@ -140,8 +143,10 @@ describe('config', () => {
       const configObject = {
         foo: '/configured/foo',
```
kumar303: The same comment applies to this test. Change `foo` to a valid option.

tests/unit/test.config.js
```diff
@@ -206,8 +211,10 @@ describe('config', () => {
       const configObject = {
         foo: true,
```
kumar303: The same comment applies to this test. Change `foo` to a valid option.

tests/unit/test.config.js
```diff
@@ -259,6 +266,23 @@ describe('config', () => {
...
+            demand: false,
+          },
+        },
+      });
+      const configObject = {
+        foo: 'fake/value/',
+      };
+      assert.throws(() => {
+        applyConfigToArgv({argv, configObject, defaultValues});
+      }, UsageError, /Please fix your config and try again/);
```
kumar303: This regex should make sure that the option name was specified in the error message

tests/unit/test.config.js
```diff
@@ -120,8 +120,11 @@ describe('config', () => {
       const configObject = {
         foo: '/configured/foo',
```
kumar303: This test should not throw an error because it's testing that default values are preserved. You can fix it by changing `foo` to some kind of valid option. However, you will have to define a new option in addition to `source-dir` to do that.

src/config.js
```diff
@@ -23,18 +25,24 @@ export function applyConfigToArgv({
...
     const wasValueSetOnCLI = typeof(argv[option]) !== 'undefined' &&
       (argv[option] !== defaultValues[option]);
     if (wasValueSetOnCLI) {
       log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
         `configuration: ${option}=${configObject[option]}`);
       continue;
     }
-    if (!argv.hasOwnProperty(option)) {
+    if (!argv.hasOwnProperty(option) &&
+      !argv.hasOwnProperty(decamelize(option, '-'))) {
```
kumar303: can't this just be `if (!argv.hasOwnProperty(decamelize(option, '-'))) { ...` ?

src/config.js
```diff
@@ -23,18 +25,24 @@ export function applyConfigToArgv({
...
       log.debug(`Ignoring configuration: ${option}=${configObject[option]} ` +
         'because this is an unknown option');
-      continue;
+      throw new UsageError(`The option ${option} is invalid.` +
+        ' Please fix your config and try again.');
```
kumar303: This error message should have the path to the config so that the user can find out which config they need to fix. This means applyConfigToArgv() will need to accept a filename argument.

src/config.js
```diff
@@ -23,18 +25,24 @@ export function applyConfigToArgv({
...
       (argv[option] !== defaultValues[option]);
     if (wasValueSetOnCLI) {
       log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
         `configuration: ${option}=${configObject[option]}`);
       continue;
     }
-    if (!argv.hasOwnProperty(option)) {
+    if (!argv.hasOwnProperty(option) &&
+      !argv.hasOwnProperty(decamelize(option, '-'))) {
       log.debug(`Ignoring configuration: ${option}=${configObject[option]} ` +
```
kumar303: this debug statement isn't necessary anymore because the developer will see an informative error about it.

src/config.js
```diff
@@ -23,18 +25,24 @@ export function applyConfigToArgv({
   for (const option in configObject) {
     // we assume the value was set on the CLI if the default value is
     // not the same as that on the argv object as there is a very rare chance
-    // this happening
+    // of this happening
+    if (camelCase(option) !== option) {
+      throw new UsageError(`Please use camel case to specify ${option} ` +
```
kumar303: since you have the function already you can just tell them what to do (which will help them fix it faster):
````js
throw new UsageError(
  `The config option "${option}" must be specified in camel case: "${camelCase(option)}"`
);
````

File: src/config.js (hunk 8 above)
Comment (kumar303) [inline src/config.js]: Rather than say "please fix [thing]," I think it's more useful to state exactly what the error was as a matter of fact. I would suggest something more like:
````js
throw new UsageError(
  `The config file at ${configFilename} specified an unknown option: "${option}"`
);
````

File: src/config.js [hunk 11]
```diff
@@ -12,29 +14,34 @@ type ApplyConfigToArgvParams = {|
   argv: Object,
   configObject: Object,
   defaultValues: Object,
+  configFileName?: string,
```
kumar303: This should be a required argument. Why did you want it optional? For the tests?

tests/unit/test.config.js
```diff
@@ -259,6 +270,43 @@ describe('config', () => {
...
+            type: 'string',
+            demand: false,
+          },
+        },
+      });
+      const configObject = {
+        'source-dir': 'fake/value/',
+      };
+      assert.throws(() => {
+        applyConfigToArgv({argv, configObject, defaultValues});
```
kumar303: you could make a shorter helper function just for testing to make it easier to omit defaults. Something like this:
````js
const applyConf = (params) => applyConfigToArgv({
  configFilename: 'some/path/to/config.js',
  ...params,
});
````

File: tests/unit/test.config.js [hunk 13]
```diff
@@ -259,6 +270,43 @@ describe('config', () => {
...
+        globalOpt: {
+          'source-dir': {
+            type: 'string',
+            demand: false,
+          },
+        },
+      });
+      const configFileName = 'fake/path/to/config';
+      const configObject = {
+        artifacts: 'fake/value/',
```
kumar303: Since the check for invalid options relies on the original hyphenated key, I think it would be good to make this `artifactsDir` to cover that part of the logic.

src/config.js
shubheksha: Yes, if I don't make this optional, I'll have to add a file path to all the tests.

shubheksha:
@kumar303, could you review this again? Thanks!

PR #795, opened 2017-02-08
feat: add support for reading sub options from the config

### LINKED ISSUE #729: Apply config file keys to sub commands
This is a stepping stone toward implementing all of https://github.com/mozilla/web-ext/issues/176

After https://github.com/mozilla/web-ext/issues/728, let's allow a config file to define both global options and sub-command options like this:
````javascript
module.exports = {
  sourceDir: './extension',
  sign: {
    apiKey: process.env.MY_API_KEY,
    apiSecret: process.env.MY_API_SECRET,
  },
  run: {
    firefoxBinary: 'nightly',
  },
};
````

Any sub-command option should be named after the command (like `sign`, `run`, `lint`, etc).

### PR DESCRIPTION
**Pull request by shubheksha**
Fixes #729

### CODE REVIEW

File: src/config.js [hunk 1]
```diff
@@ -17,6 +17,35 @@ type ApplyConfigToArgvParams = {|
...
+  mapCommandToSubOpts,
+  configObject,
+  defaultValues,
+  configFileName,
+}: ApplySubOptionsToArgvParams): Object {
+  // cannot use spread operator because Flow complains. See: https://github.com/facebook/flow/pull/3381/
+  const newArgv = JSON.parse(JSON.stringify(argv));
+  if (mapCommandToSubOpts[commandExecuted]) {
+    for (const opt in mapCommandToSubOpts) {
+      newArgv[opt] = mapCommandToSubOpts[opt];
```
kumar303: `mapCommandToSubOpts` [looks like](https://github.com/mozilla/web-ext/pull/795/files#diff-9589e3a0cc72e26a4814bc9ef05aa2d4R74) it will only contain option definitions. I don't think you should be assigning those to `newArgv`. You need to assign values from `configObject` only.

src/config.js
```diff
@@ -17,6 +17,35 @@ type ApplyConfigToArgvParams = {|
...
+export function applySubOptionsToArgv({
+  commandExecuted,
+  argv,
+  mapCommandToSubOpts,
+  configObject,
+  defaultValues,
+  configFileName,
+}: ApplySubOptionsToArgvParams): Object {
+  // cannot use spread operator because Flow complains. See: https://github.com/facebook/flow/pull/3381/
+  const newArgv = JSON.parse(JSON.stringify(argv));
```
shubheksha: I tried removing the pipes, it doesn't work.

kumar303: actually, there's another problem. Read the Flow error more closely, it explains exactly what's happening (be sure you are looking at the right line numbers).

src/config.js
kumar303: To be clear, I think you meant to assign each option in `configObject[commandExecuted]` to `newArgv`, right?

shubheksha:
The issues with test coverage are mainly due to duplication across the functions applying options and sub options to `argv`. It should be fixed after refactoring.

shubheksha:
@kumar303, I tried to make the function recursive, but that introduced some mutation as well. I'm not sure if this is acceptable while using recursion or not. Could you review this again?

src/config.js
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
-    if (wasValueSetOnCLI) {
-      log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
-        `configuration: ${option}=${configObject[option]}`);
+
+    validateOption(option);
+    if (option === commandExecuted) {
+      configObject = configObject[option];
+      newArgv = applyConfigToArgv({
+        argv,
+        configObject,
```
kumar303: you shouldn't need to redefine this variable since it's not used anywhere besides here as a param. Just keep `configObject` defined as a constant and adjust the param to the following:
````js
newArgv = applyConfigToArgv({
  ...
});
````

File: src/config.js [hunk 4]
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
+        subCommandDefaultValues,
```
kumar303: if you change this to `defaultValues: subCommandDefaultValues` then you can get rid of the if statement below that sets `defaultValues = subCommandDefaultValues`.

src/config.js
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
     }
+
+    checkOptionPresence(argv, option, configFileName);
+
     newArgv[option] = configObject[option];
   }
   return newArgv;
 }
 
+function checkOptionPresence(argv: Object, option: string,
```
kumar303: this no longer needs to be called in multiple places so I don't think it deserves its own function

src/config.js
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
+  configFileName: string) {
+  if (!argv.hasOwnProperty(decamelize(option, '-'))) {
+    throw new UsageError(`The config file at ${configFileName} specified ` +
+      `an unknown option: "${option}"`);
+  }
+
+}
+
+function validateOption(option: string) {
```
kumar303: same here: this doesn't need to be given its own function anymore

src/config.js
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
+  }
+}
+
+function wasValueSetOnCLI(optionName: string, optionValue: string,
+  defaultValue: string, configOption: string) {
+  // we assume the value was set on the CLI if the default value is
+  // not the same as that on the argv object as there is a very rare chance
+  // of this happening
+  const setOnCLI = typeof(optionValue) !== 'undefined' &&
+  (optionValue !== defaultValue);
```
kumar303: nit: always indent two spaces when you're continuing a statement on a new line.

src/config.js
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
-    if (!argv.hasOwnProperty(decamelize(option, '-'))) {
-      throw new UsageError(`The config file at ${configFileName} specified ` +
-        `an unknown option: "${option}"`);
+
+    if (defaultValues[option] === undefined && subCommandDefaultValues) {
+      defaultValues = subCommandDefaultValues;
+    }
+
+    if (wasValueSetOnCLI(option, argv[option], defaultValues[option],
+      configObject[option])) {
```
kumar303: nit: it would be easier for me to read all function arguments on one line, like this:
````js
if (wasValueSetOnCLI(
  option, argv[option], defaultValues[option], configObject[option])) {
````

File: src/config.js [hunk 9]
```diff
@@ -13,41 +13,82 @@ const log = createLogger(__filename);
...
+}
+
+function validateOption(option: string) {
+  if (camelCase(option) !== option) {
+    throw new UsageError(`The config option "${option}" must be ` +
+      `specified in camel case: "${camelCase(option)}"`);
+  }
+}
+
+function wasValueSetOnCLI(optionName: string, optionValue: string,
```
kumar303: I know this has a bit more code to it but I also don't think it deserves its own function (since it's not shared)

src/program.js
```diff
@@ -175,6 +176,23 @@ export class Program {
 // A global variable generated by DefinePlugin, generated in webpack.config.js
 declare var WEBEXT_BUILD_ENV: string;
 
+function setDefaultValues(
+  options: Object): Object {
```
kumar303: nit: this looks like it no longer needs to be broken into a new line

src/config.js
```diff
@@ -14,40 +14,62 @@ type ApplyConfigToArgvParams = {|
...
+  let newArgv = {...argv};
   for (const option in configObject) {
+
+    validateOption(option);
+    if (option === commandExecuted) {
+      newArgv = applyConfigToArgv({
+        argv,
+        configObject: configObject[option],
+        defaultValues: subCommandDefaultValues,
+        subCommandDefaultValues,
```
shubheksha: @kumar303, here I am being forced to pass `subCommandDefaultValues` again because if I make it optional, Flow starts complaining that it can be undefined on the previous line. Is there a way to get around this?

shubheksha:
@kumar303, I addressed all the feedback, this could use another review!

src/config.js
```diff
@@ -14,40 +14,62 @@ type ApplyConfigToArgvParams = {|
...
       throw new UsageError(`The config file at ${configFileName} specified ` +
         `an unknown option: "${option}"`);
     }
+
     newArgv[option] = configObject[option];
   }
   return newArgv;
 }
 
+function validateOption(option: string) {
```
kumar303: This doesn't need to be in its own function since the code isn't shared. It's only called in one place.

src/program.js
```diff
@@ -67,6 +70,11 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
+    this.mapCommandToSubOpts = {};
+    this.mapCommandToSubOpts[name] = commandOptions;
+    this.commandExecuted = name;
+    this.subCommandDefaultValues = setDefaultValues(commandOptions);
```
kumar303: `this.subCommandDefaultValues` is no different from [this.defaultValues](https://github.com/shubheksha/web-ext/blob/854caca3615b512d75d1d88ee850ad9fd9c92a74/src/program.js#L103) except that it has a nested key (the command name). Try to see if you can use `this.defaultValues` for everything -- I think you can.

src/config.js
shubheksha: I left this as is because I it'll have more code needed for validation coming in the next patch.

kumar303: I suggest waiting until you do the next patch to see if it needs its own function. It will be easy to move if needed but I don't think it will need its own function.

kumar303: by the way, try to avoid what's referred to as [premature optimization](https://shreevatsa.wordpress.com/2008/05/16/premature-optimization-is-the-root-of-all-evil/). It's very tempting and I have to fight it a lot. Only optimize what you need to optimize now.

shubheksha: Alright, I'll remove it for now.

src/program.js
```diff
@@ -67,6 +69,13 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
+    this.mapCommandToSubOpts = {};
+    this.mapCommandToSubOpts[name] = commandOptions;
```
kumar303: this doesn't look like it's used anywhere. I think you can delete `this.mapCommandToSubOpts`.

src/program.js
```diff
@@ -67,6 +69,13 @@ export class Program {
...
+    this.commandExecuted = name;
+    const subCommandDefaultValues = setDefaultValues(commandOptions);
+    this.defaultValues = {...this.defaultValues,
+      ...subCommandDefaultValues};
```
kumar303: While technically unlikely, a sub-command option default named `sourceDir` would overwrite the global value named `sourceDir`. This would be confusing and I don't think it's necessary. You should be able to do this, no?

````js
this.defaultValues[name] = setDefaultValues(commandOptions);
````

File: src/program.js [hunk 16]
```diff
@@ -67,6 +69,13 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
+    this.mapCommandToSubOpts = {};
+    this.mapCommandToSubOpts[name] = commandOptions;
+    this.commandExecuted = name;
```
kumar303: This is unused

src/program.js
```diff
@@ -92,15 +101,8 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
-    this.defaultValues = {};
+    this.defaultValues = setDefaultValues(options);
```
kumar303: Since `this.defaultValues` is mutated in several places, you'll need to be a little bit more careful about setting it.

First, initialize it in the `constructor()` like this:
````js
this.defaultValues = {};
````

Next, make sure you are preserving the old values on this line:
````js
this.defaultValues = {...this.defaultValues, setDefaultValues(options)};
````

File: src/program.js [hunk 18]
```diff
@@ -175,6 +177,22 @@ export class Program {
 // A global variable generated by DefinePlugin, generated in webpack.config.js
 declare var WEBEXT_BUILD_ENV: string;
 
+function setDefaultValues(options: Object): Object {
+  const defaultValues = {};
+  if (options !== undefined && options !== null) {
```
kumar303: why is this check necessary? The Flow definition of `options` says that it can't be undefined or null.

src/program.js
```diff
@@ -67,6 +68,11 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
+    if (!this.defaultValues) {
+      this.defaultValues = {};
```
kumar303: Do this in the [constructor()](https://github.com/shubheksha/web-ext/blob/6761b989e41f55a22ce6789f966dd92392729b76/src/program.js#L43) instead.

src/program.js
```diff
@@ -92,15 +98,12 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
-    this.defaultValues = {};
+    const defaultValuesSet = setDefaultValues(options);
+    if (!this.defaultValues) {
+      this.defaultValues = {};
```
kumar303: You can delete this when the initialization is moved to the constructor

src/program.js
```diff
@@ -67,6 +68,11 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
+    if (!this.defaultValues) {
+      this.defaultValues = {};
+    }
+    this.defaultValues[name] = setDefaultValues(commandOptions);
+    this.commandExecuted = name;
```
kumar303: This is unused, it can be deleted.

src/program.js
```diff
@@ -92,15 +98,12 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
-    this.defaultValues = {};
+    const defaultValuesSet = setDefaultValues(options);
+    if (!this.defaultValues) {
+      this.defaultValues = {};
+    }
+    this.defaultValues = {...this.defaultValues, ...defaultValuesSet};
```
kumar303: No need for a new variable here:
````js
this.defaultValues = {...this.defaultValues, ...setDefaultValues(options)};
````

File: src/program.js [hunk 23]
```diff
@@ -38,6 +38,7 @@ export class Program {
   commands: { [key: string]: Function };
   shouldExitProgram: boolean;
   defaultValues: Object;
+  commandExecuted: string;
```
kumar303: this can be deleted

shubheksha:
Done! Hopefully it's ready to land now :smile:

PR #817, opened 2017-02-23
feat: linting on run command

### LINKED ISSUE #384: Check for lint while running an extension
It might be nice to do this:

```
web-ext run --lint
```

This would check an extension for [lint](https://github.com/mozilla/web-ext/issues/204), run the extension in Firefox, watch for source changes, lint again when the source changes, reload the extension, and repeat. It would add a little bit of slowness but it might be useful.

saintsebastian:
At the moment I am still perplexed by how to test `shouldScanFiles`. Would welcome any suggestions!

Fixes #384


saintsebastian:
@rpl thanks, this is definitely improves a structure of those changes

saintsebastian:
@rpl Hey Luca! Thanks for your suggestions! I addressed you review and the first big thing (reusing code and rewriting test accordingly). However I am not yet done with the second (blocking listing errors). I agree listing on run should be informational only and not fatal, however I am not so sure about reload linting. Let me know what you think!

saintsebastian:
@rpl thanks again for bearing with me on this issue. As linter implemented at the moment, the first linter fail on run doesn't affect the user. I think it is a good idea to talk error handling in general before we move on. What do you think? 
1. first fail on run doesn't affect the build, just creates message in console
2. later fails also are printed in console (since notification means error in application, listing error doesn't seem important enough for notifications)
3. notification is shown when the linting fails on the watched file (since user specifically asked for that)

I will resolve the conflict in the meanwhile, and revisit this again

saintsebastian:
@rpl here is finally a merged version of this PR. Btw i got to see you changes and they really improved how fast i was navigating around the code

src/cmd/lint.js
```diff
@@ -15,29 +9,6 @@ const log = createLogger(__filename);
 
 export type LinterOutputType = 'text' | 'json';
```
rpl: This `LinterOutputType` type is used in the `LinterCreatorParams`, and so it is supposed to be moved into the new `"src/util/linter"` module (also I'm wondering why flow didn't complain about it)

src/cmd/run.js
```diff
@@ -69,6 +73,10 @@ export default async function run(
 
   const manifestData = await getValidatedManifest(sourceDir);
 
+  if (lint) {
+    await linter({sourceDir, artifactsDir, ignoreFiles});
```
rpl: could we pass the lint and linter parameters to the `MultiExtensionRunner` instance and leave it to handle both the initial linting and the linting on file changes?

saintsebastian:
@rpl thanks for looking at this PR! let me know if anything else needs fixing

tests/unit/test-cmd/test.lint.js
```diff
@@ -3,157 +3,39 @@ import {it, describe} from 'mocha';
...
-
-  it('runs as a binary', () => {
-    const {lint, createLinter} = setUp();
-    return lint().then(() => {
-      const args = createLinter.firstCall.args[0];
-      assert.equal(args.runAsBinary, true);
-    });
-  });
-
-  it('sets runAsBinary according shouldExitProgram option', () => {
```
rpl: I cannot find this test case anymore, it should probably be one of the tests of the new linter module.

tests/unit/test-cmd/test.lint.js
```diff
@@ -3,157 +3,39 @@ import {it, describe} from 'mocha';
...
-
-  it('passes sourceDir to the linter', () => {
-    const {lint, createLinter} = setUp();
-    return lint({sourceDir: '/some/path'}).then(() => {
-      const config = createLinter.firstCall.args[0].config;
-      assert.equal(config._[0], '/some/path');
-    });
-  });
 
-  it('passes warningsAsErrors to the linter', () => {
```
rpl: I cannot find this test case anymore, it should probably be one of the tests of the new linter module.

tests/unit/test-cmd/test.lint.js
```diff
@@ -3,157 +3,39 @@ import {it, describe} from 'mocha';
...
-    const {lint, createLinter} = setUp();
-    return lint({warningsAsErrors: true}).then(() => {
-      const config = createLinter.firstCall.args[0].config;
-      assert.equal(config.warningsAsErrors, true);
-    });
-  });
-
-  it('passes warningsAsErrors undefined to the linter', () => {
```
rpl: same as above.

tests/unit/test-cmd/test.lint.js
```diff
@@ -3,157 +3,39 @@ import {it, describe} from 'mocha';
...
-  it('configures the linter when not verbose', () => {
-    const {lint, createLinter} = setUp();
-    return lint({verbose: false}).then(() => {
-      const config = createLinter.firstCall.args[0].config;
-      assert.equal(config.logLevel, 'fatal');
-      assert.equal(config.stack, false);
-    });
-  });
-
-  it('passes through linter configuration', () => {
```
rpl: I cannot find this test case anymore, it should probably be one of the tests of the new linter module.

tests/unit/test-cmd/test.run.js
```diff
@@ -29,6 +29,7 @@ describe('run', () => {
       artifactsDir: path.join(sourceDir, 'web-ext-artifacts'),
       sourceDir,
       noReload: true,
+      lint: false,
```
rpl: We should probably add a new test case in this file to ensure that the CLI option --lint is passed to the extension runner.

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -368,6 +405,24 @@ describe('util/extension-runners', () => {
       );
     });
 
+    it('lints changed files', () => {
```
rpl: we should also add some test cases for the failures scenarios (e.g. linting failure when the first linting has been executed while the extension runner is starting, linting failures on source change) and ensure that it is going to behave like we want (e.g. the linting failure on the first linting could be a fatal error, and exit the web-ext run command, while a linting failure on change could send a desktop notification and it does not reload the changed extension, this sounds like a reasonable behavior to me)

tests/unit/test-cmd/test.lint.js
saintsebastian: @rpl correct, i only test for default config. thanks for pointing this out

tests/unit/test-cmd/test.lint.js
saintsebastian: i think this is a lint cmd test (since i pass it to linter as `runAsBinary`). I'll add it again

tests/unit/test-extension-runners/test.extension-runners.js
saintsebastian: @rpl i've just noticed that there are no tests for `MultiExtensionRunner` in test.run.js. before i continue, should l i write these? just one basic to test all passed param (including `desktopNotifications` which is [defined but never used](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-cmd/test.run.js#L49)) would already be something

saintsebastian:
@rpl this commit partially addresses this review, I'll add tests for linting failure scenario and overall test for `MultiExtensionRunner`. Let me know if you have better ideas

saintsebastian:
@rpl i think we still haven't decided how to handle linting errors. Maybe you and @kumar303 have some ideas. I think in SF we talked about how desktop notification usually means fatal error, so linting failure is probably not big enough deal for that. Also I don't know if error of first linting on run should be fatal or not. The tests are for how the situation currently is (they are pretty useless), but rewriting them shouldn't be hard once decision is made. We can test the linting results for errors and decide what to do with them

tests/unit/test-extension-runners/test.extension-runners.js
rpl: @saintsebastian oh yeah, that is definitely missing, we should have some tests similar to the one that inject a fake FirefoxDesktopExtensionRunner into the cmd run and checks that it has been created with the expected parameters.

Thanks for spotting the missing test!

rpl:
> @rpl i think we still haven't decided how to handle linting errors. Maybe you and @kumar303 have some ideas. I think in SF we talked about how desktop notification usually means fatal error, so linting failure is probably not big enough deal for that. Also I don't know if error of first linting on run should be fatal or not. The tests are for how the situation currently is (they are pretty useless), but rewriting them shouldn't be hard once decision is made. We can test the linting results for errors and decide what to do with them

Yeah, that's true.

k, I vote for the following behaviors:

- none of the linting failures is fatal
- a desktop notification when lint fails on the source changes would be nice, assuming that the user have to turn on the linting manually (using --lint in web-ext run), I think that it is reasonable to want to know if there is a linting failure without switching to the console (or maybe this behavior can be part of an optional parameter which make `web-ext run` to consider the linting issues as errors, something like `web-ext run --lint --fail-on-lint` or similar, which can be part of a follow up of this change).

kumar303:
Yeah, sounds good to me. I agree that with `--lint` there should be a desktop notification for all lint errors. It should not abort the `run` command though because the developer has a chance to fix it and wait for auto-reload to test for lint again.

saintsebastian:
@rpl @kumar303 thanks for your ideas!
My future plans for this PR are:
1. MultiExtensionRunner injection tests for run
2. To add parameter `--fail-on-lint`
Let me know if anything else is needed.

saintsebastian:
This PR needs extensive rebasing, are you still interested in this feature? Functionally after rebase it is almost ready (missing `--fail-on-lint` parameter). Let me know and i'll wrap it up

rpl:
> This PR needs extensive rebasing, are you still interested in this feature? Functionally after rebase it is almost ready (missing --fail-on-lint parameter). Let me know and i'll wrap it up

@saintsebastian sure, this feature still makes sense (sorry for the conflicts, I've probably introduced a number of them when we landed the android support PR :-))

saintsebastian:
@rpl Hi! This PR has been rebased, I believe everything is preserved as it should be. I honestly think that it's pretty much ready, but let me know if i miss something :)

src/cmd/run.js
```diff
@@ -189,6 +191,11 @@ export default async function run(
   const extensionRunner = new MultiExtensionRunner({
     desktopNotifications,
     runners,
+    sourceDir,
```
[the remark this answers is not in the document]
saintsebastian: @rpl I think there is no avoiding passing certain thing to `MultiExtensionRunner`, but i agree, moving actual linting to run is a good idea. About `ReloadSrategy`, it's possible, only caveat being that watcher can pass linter a specific file path, and inside `reloadExtensionBySourceDir` the entire folder will be linted.

rpl:
Closing as "state: incomplete", the issue is still open and valid, unfortunately this PR has been branched out quite some time ago and it is unlikely to be resumed from this old branch.

PR #830, opened 2017-03-03
feat: Added web-ext run --experiments to install additional webextensions experiments/addons

### LINKED ISSUE #741: Make working with experiments easier

Feature

So you want to write an experiment. That's an old style legacy add-on:

https://webextensions-experiments.readthedocs.io/en/latest/

You then write an WebExtension that uses the experiment. But web-ext run won't work out of the box, you need to make a custom profile that has the legacy add-on installed. Then use that with --profile with your web-ext run.

Then you go back and fiddle with the legacy add-on now you have to reinstall the legacy add-on in the profile and start again.

Is there a way we can make this a little easier?

rpl:
This PR goal is to make working with experiments easier and faster (which should fix #741) by providing an additional command line option for "web-ext run" (that is "--experiments" in the current version of this patch) which:

- install the experiments (one or more than one) before installing the extension (which is needed because the extension will depends on the experiments and a "your extension is not compatible with the application version" error is going to be raised if we try to install the extension before the experiments it depends on)

- watch the experiments sources and reload the experiment addon when they change (which also already has the very nice side-effect of reloading the addon that depends from the reloaded experiment)

- raise an error if we try to use it with --pre-install (mainly because with --pre-install the experiment cannot be reloaded, well... it can be disabled and re-enabled to achieve the same effect, but I'm not sure there is a lot of value on supporting the experiments with this mode)

- raise an error if the experiment addon id is missing during the experiments installation

Example command line:
```
web-ext run -s ./extension --experiments ./experiment ../../another-experiment -f /path/to/nightly-or-devedition/bin/firefox
```

Note: the "-- experiments" option take an array of paths to the experiments, but they do not need to really be webextension experiment addons, they can also be just regular addons, the only requirement is that they should be able to install as temporarily installed addons.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -1,4 +1,7 @@
 /* @flow */
+
+import path from 'path';
+
 import type FirefoxProfile from 'firefox-profile';
 import type Watchpack from 'watchpack';
 
@@ -11,6 +14,7 @@ import {
   isErrorWithCode,
   RemoteTempInstallNotSupported,
   WebExtError,
+  UsageError,
 } from '../errors';
 import {createLogger} from '../util/logger';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
@@ -83,8 +87,17 @@ export function defaultWatcherCreator(
 
 // defaultReloadStrategy types and implementation.
 
+export type InstalledExperiment = {|
+  addonId: string,
+|};
+
+export type InstalledExperiments = {
+  [experimentPath: string]: InstalledExperiment,
+};
+
 export type ReloadStrategyParams = {|
   addonId: string,
+  installedExperiments: InstalledExperiments,
   firefoxProcess: FirefoxProcess,
   client: RemoteFirefox,
   profile: FirefoxProfile,
@@ -100,13 +113,28 @@ export type ReloadStrategyOptions = {|
 
 export function defaultReloadStrategy(
   {
-    addonId, firefoxProcess, client, profile,
+    addonId, installedExperiments = {}, firefoxProcess, client, profile,
     sourceDir, artifactsDir, ignoreFiles,
   }: ReloadStrategyParams,
   {
     createWatcher = defaultWatcherCreator,
   }: ReloadStrategyOptions = {}
 ): void {
+  const experimentWatchers: Array<Watchpack> = [];
+
+  for (const experimentDir of Object.keys(installedExperiments)) {
+    const installedExperiment = installedExperiments[experimentDir];
+
+    const experimentWatcher: Watchpack = createWatcher({
+      addonId: installedExperiment.addonId,
+      artifactsDir,
+      client,
+      ignoreFiles,
+      sourceDir: experimentDir,
+    });
+    experimentWatchers.push(experimentWatcher);
+  }
+
   const watcher: Watchpack = (
     createWatcher({addonId, client, sourceDir, artifactsDir, ignoreFiles})
   );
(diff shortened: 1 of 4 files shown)


rpl:
Hey @kumar303 
during this week I was working on a couple of WebExtensions Experiments and I suddenly found myself with a patch for #741 :-) (well... I was tired of manually reload the experiment addon of every change in the sources and I thought to tweak web-ext to make it able to install, watch and reload the experiment for me, as it already does with the main addon).

Yesterday I added a bunch of new test cases to it (and cleaned a bit the initial hack accordingly) to push it here for an initial feedback, and so that I can rebase it more easily on top of the upcoming changes related to the open pull requests from @saintsebastian.

kumar303:
Cool! I can take a look shortly. Can you fix the conflicts?

src/cmd/run.js
```diff
@@ -280,6 +328,7 @@ export default async function run(
         sourceDir,
         artifactsDir,
         addonId,
+        installedExperiments,
```
kumar303: Did you consider making this more abstract in a way that will support both experiments and regular extensions? Maybe a structure like this could accommodate both:
````js
type Extension = {
  addonId: string,
  sourceDir: string,
};

const extensions: Array<Extension> = [];

// Add experiments:
const experiments = await runner.installExperiments(client);
experiments.forEach(
  (experiment) => extensions.push({
    addonId: experiment.addonId, 
    sourceDir: experiment.sourceDir,
  })
);


// Add the extension under test:
const installResult: FirefoxRDPResponseAddon = 
  await runner.installAsTemporaryAddon(client);
extensions.push({addonId: installResult.addon.id, sourceDir});

// Now all functions should be able to work with the array of extension data.
````

File: src/program.js [hunk 2]
```diff
@@ -388,6 +388,13 @@ Example: $0 --help run.
         demand: false,
         type: 'boolean',
       },
+      'experiments': {
+        describe: 'Install one or more WebExtensions Experiments ' +
+                  'alongside the WebExtension Addon.',
```
kumar303: I think you could just say:

    Install one or more WebExtension Experiments alongside the main WebExtension.

tests/unit/test-cmd/test.run.js
```diff
@@ -282,6 +290,82 @@ describe('run', () => {
...
+    assert.equal(firefoxClient.installTemporaryAddon.callCount, 3);
+    assert.equal(firefoxClient.installTemporaryAddon.firstCall.args[0],
+                 path.resolve(experiments[0]));
+    assert.equal(firefoxClient.installTemporaryAddon.secondCall.args[0],
+                 path.resolve(experiments[1]));
+    assert.equal(firefoxClient.installTemporaryAddon.lastCall.args[0],
+                 cmd.argv.sourceDir);
+  });
+
+  it('raise an error on missing experiment addonId', async () => {
```
kumar303: it throws an error...

tests/unit/test-cmd/test.run.js
```diff
@@ -282,6 +290,82 @@ describe('run', () => {
...
+
+  it('raises an error message on --experiments and --pre-install', () => {
+    const experiments = [
+      fixturePath('fake-experiment'),
+      fixturePath('fake-experiment2'),
+    ];
+
+    const cmd = prepareRun(null, experiments);
+
+    assert.isRejected(cmd.run({preInstall: true}), UsageError);
```
kumar303: Since there are many possible UsageErrors I think it would help to add a regex to match a bit of the error message

rpl:
I'm closing this PR because the experiments can now be embedded in an extension, and so a special behavior for the experiments doesn't seem to be really needed anymore (it could be more reasonable to optionally allow "web-ext run" to install more than a single extension on the same Firefox instance).

PR #842, opened 2017-03-07
fix: add validation for option values of incorrect types

### LINKED ISSUE #798: Throw an error if a config file defines an option with the wrong type
Every option is defined with a type. Each possible type is [documented here](http://yargs.js.org/docs/#methods-optionkey-opt). If a config file defines a value in the wrong type, let's throw an error. Otherwise, confusing and unpredictable things might happen when the command gets executed.

This is part of implementing https://github.com/mozilla/web-ext/issues/176

shubheksha:
Fixes #798


shubheksha:
@kumar303, could you tell me if the approach is correct or am I missing something here?

I'll add tests once you've reviewed it.
Thanks!

tests/unit/test.config.js
```diff
@@ -70,12 +71,15 @@ describe('config', () => {
       const configObject = {
         sourceDir: '/configured/source/dir',
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({argv,
+        configObject,
+        defaultValues,
+        optionTypes});
```
kumar303: This adds a lot of verbosity. The `argv`, `defaultValues` and `optionTypes` variables are not actually needed for this test (and many others) so you could make it less verbose by passing them with a spread operator:
````js
const params = makeArgv();
...
const newArgv = applyConf({...params, configObject});
````

File: tests/unit/test.config.js [hunk 2]
```diff
@@ -460,10 +545,39 @@ describe('config', () => {
         argv,
         configObject,
         defaultValues,
+        optionTypes,
         commandExecuted,
       });
       assert.strictEqual(newArgv.apiKey, cmdApiKey);
     });
+
+    it('throws an error when the type of sub option is invalid', () => {
```
kumar303: This needs another test that configures a value for a command that is not being executed. This would highlight the error you had where the `optionTypes` are not keyed by command. For example, set up a test where the user executes `run` but specifies an invalid type in a config file for `sign.apiKey`.

kumar303:
Hi @shubheksha , would you have some time to address the change requests soon?

shubheksha:
hey @kumar303, I'll be travelling for the first half of May, so can't get to it soon, but I'll wrap it up as soon as I find time :)

kumar303:
No problem, thanks for the update. There's no rush on it. I just don't want it to bitrot too much and it seems close to landing.

src/program.js
```diff
@@ -63,14 +63,16 @@ export class Program {
...
+    this.options = {};
   }
 
   command(
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
-    this.defaultValues[name] = setDefaultValues(commandOptions);
+    this.options = Object.assign(this.options, commandOptions);
+    this.mainCommandsList.push(name);
```
kumar303: Consider this instead:

```js
this.options[name] = commandOptions;
```
Then you won't need `mainCommandsList` at all.

shubheksha: That would create issues with having the list of options using `Object.keys()` since they won't be the top-level options anymore.

shubheksha:
@kumar303, I used another approach to fix this: I used a local variable & added the property to `options` as an array instead of using it as a key.

src/program.js
```diff
@@ -63,14 +62,19 @@ export class Program {
...
 
   command(
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
-    this.defaultValues[name] = setDefaultValues(commandOptions);
+    const mainCommandsList = [];
+    mainCommandsList.push(name);
+    this.options = Object.assign(this.options, commandOptions);
+    this.options.mainCommandsList = mainCommandsList;
```
kumar303: I still think you can get it working with just this:

```js
this.options[name] = commandOptions;
```
All the data is there. Try to see if you can get it working.

shubheksha:
I've no clue why it didn't work the first time around!

kumar303:
Two more tests and some fixes:

```diff
diff --git a/src/config.js b/src/config.js
index 4386468..a7d0d69 100644
--- a/src/config.js
+++ b/src/config.js
@@ -33,7 +33,9 @@ export function applyConfigToArgv({
         `specified in camel case: "${camelCase(option)}"`);
     }
 
-    if (option === commandExecuted || Object.keys(options).includes(option)) {
+    if (typeof options[option] === 'object' &&
+        typeof configObject[option] === 'object') {
+      // Descend into the nested configuration for a sub-command.
       newArgv = applyConfigToArgv({
         argv,
         configObject: configObject[option],
@@ -115,4 +117,4 @@ export function loadJSConfigFile(filePath: string): Object {
       'Did you set module.exports = {...}?');
   }
   return configObject;
-}
\ No newline at end of file
+}
diff --git a/src/program.js b/src/program.js
index 3594a27..34226fe 100644
--- a/src/program.js
+++ b/src/program.js
@@ -2,6 +2,7 @@
 import path from 'path';
 import {readFileSync} from 'fs';
 
+import camelCase from 'camelcase';
 import git from 'git-rev-sync';
 import yargs from 'yargs';
 
@@ -69,7 +70,7 @@ export class Program {
     name: string, description: string, executor: Function,
     commandOptions: Object = {}
   ): Program {
-    this.options[name] = commandOptions;
+    this.options[camelCase(name)] = commandOptions;
 
     this.yargs.command(name, description, (yargsForCmd) => {
       if (!commandOptions) {
diff --git a/tests/unit/test.config.js b/tests/unit/test.config.js
index 711d6e1..4b47d27 100644
--- a/tests/unit/test.config.js
+++ b/tests/unit/test.config.js
@@ -702,6 +702,52 @@ describe('config', () => {
         ' (expected type: "string")');
     });
 
+    it('handles camel case sub-commands', () => {
+      const params = makeArgv({
+        userCmd: ['sign-extension'],
+        command: 'sign-extension',
+        commandOpt: {
+          'api-url': {
+            requiresArg: true,
+            type: 'string',
+            default: 'pretend-default-value-of-apiKey',
+          },
+        },
+      });
+      const configObject = {
+        signExtension: {
+          apiUrl: 2,
+        },
+      };
+      assert.throws(
+        () => applyConf({...params, configObject}),
+        UsageError,
+        'UsageError: The config file at some/path/to/config.js ' +
+        'specified the type of "apiUrl" incorrectly'
+      );
+    });
+
+    it('throws an error when a global option type is invalid', () => {
+      const params = makeArgv({
+        globalOpt: {
+          retries: {
+            type: 'number',
+            default: 1,
+          },
+        },
+      });
+      const configObject = {
+        retries: 'invalid-value',
+      };
+      assert.throws(
+        () => applyConf({...params, configObject}),
+        UsageError,
+        'UsageError: The config file at some/path/to/config.js ' +
+        'specified the type of "retries" incorrectly as ' +
+        '"string" (expected type: "number")'
+      );
+    });
+
   });
 
   describe('loadJSConfigFile', () => {
```

src/program.js
```diff
@@ -96,7 +97,7 @@ export class Program {
     // This is a convenience for setting global options.
     // An option is only global (i.e. available to all sub commands)
     // with the `global` flag so this makes sure every option has it.
-    this.defaultValues = {...this.defaultValues, ...setDefaultValues(options)};
+    this.options = Object.assign(this.options, options);
```
kumar303: This should use ES6 object spreads

tests/unit/test.config.js
```diff
@@ -70,12 +69,13 @@ describe('config', () => {
       const configObject = {
         sourceDir: '/configured/source/dir',
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -88,13 +88,14 @@ describe('config', () => {
...
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -108,12 +109,13 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -130,13 +132,14 @@ describe('config', () => {
       const configObject = {
         artifactsDir: '/configured/artifacts/dir',
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -154,12 +157,13 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -171,12 +175,13 @@ describe('config', () => {
       const configObject = {
         overwriteFiles: true,
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -187,12 +192,13 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -203,13 +209,14 @@ describe('config', () => {
       const configObject = {
         overwriteFiles: false,
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -223,12 +230,13 @@ describe('config', () => {
       const configObject = {
         verbose: true,
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -240,12 +248,13 @@ describe('config', () => {
       const configObject = {
         numberOfRetries: 1,
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -257,12 +266,13 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -273,12 +283,13 @@ describe('config', () => {
       const configObject = {
         sourceDir: '/configured/directory',
       };
-      const newArgv = applyConf({argv, configObject, defaultValues});
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -290,34 +301,91 @@ describe('config', () => {
         'source-dir': 'fake/value/',
       };
       assert.throws(() => {
-        applyConf({argv, configObject, defaultValues});
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -290,34 +301,91 @@ describe('config', () => {
...
       const configObject = {
-        artifactsDir: 'fake/artifacts/dir',
+        randomDir: 'fake/artifacts/dir',
       };
       assert.throws(() => {
-        applyConf({argv, configObject, defaultValues, configFileName});
-      }, UsageError, 'UsageError: The config file at fake/path/to/config ' +
-        'specified an unknown option: "artifactsDir"');
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -290,34 +301,91 @@ describe('config', () => {
...
+            demand: false,
+          },
+        },
+      });
+      const configObject = {
+        sourceDir: {randomKey: 'randomValue'},
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -290,34 +301,91 @@ describe('config', () => {
...
+          verbose: {
+            type: 'count',
+          },
+        },
+      });
+      const configObject = {
+        verbose: 2,
+      };
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -334,18 +402,14 @@ describe('config', () => {
...
         },
       };
-      const newArgv = applyConf({
-        argv,
-        configObject,
-        defaultValues,
-        commandExecuted,
-      });
+      const newArgv = applyConf({...params,
+        configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -362,18 +426,14 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -389,17 +449,13 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -422,18 +478,14 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+        },
+      });
+      const configObject = {
+        sign: {
+          'api-url': 2,
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+        },
+      });
+      const configObject = {
+        sign: {
+          randomOption: 'random-value',
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+        },
+      });
+      const configObject = {
+        sign: {
+          apiUrl: 2,
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+      });
+      const configObject = {
+        sign: {
+          apiUrl: 2,
+          apiKey: 'fake-api-key',
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
```
kumar303: This should be on one line

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+      }, UsageError,
+        'UsageError: Option: apiUrl was defined without a type.');
+    });
+
+    it('throws an error when the type of an unrelated sub option is invalid',
+    () => {
```
kumar303: It would be nicer to shorten this test description so that you don't have to break the line. It's a bit hard to read when scanning tests.

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+      const program = new Program(['run']);
+
+      const commandDesc = 'this is a fake command';
+      const commandExecutor = sinon.stub();
+
+      program.command('run', commandDesc, commandExecutor,
```
kumar303: You can replace `commandExecutor` here and below with `sinon.stub()`; it's shorter ;)

kumar303: You can replace `commandDesc` here are below with `'fake command'`. It's not important enough to make into a separate variable.

tests/unit/test.config.js
```diff
@@ -456,14 +508,244 @@ describe('config', () => {
...
+        },
+      };
+      const params = {
+        argv: program.yargs.exitProcess(false).argv,
+        options: program.options,
+      };
+
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This should be on one line

src/config.js
```diff
@@ -33,27 +31,62 @@ export function applyConfigToArgv({
...
+        options: options[option],
         configFileName});
       continue;
     }
 
+    const decamelizedOptName = decamelize(option, '-');
+    let expectedType;
+    if (options[decamelizedOptName]) {
+      if (options[decamelizedOptName].type === undefined) {
+        throw new UsageError
```
shubheksha: This is something that can be fixed in the config by the developer, right?

src/config.js
```diff
@@ -33,27 +31,62 @@ export function applyConfigToArgv({
...
+        configObject: configObject[option],
+        options: options[option],
         configFileName});
       continue;
     }
 
+    const decamelizedOptName = decamelize(option, '-');
+    let expectedType;
+    if (options[decamelizedOptName]) {
+      if (options[decamelizedOptName].type === undefined) {
```
shubheksha: This doesn't work for some reason. It keeps trying to read `type` even if the object is `undefined`. I'm not sure why.

shubheksha: Basically, the `&&` isn't exhibiting short circuiting.

shubheksha:
@kumar303, I addressed all the feedback except for the comments I left about the `&&` and the `UsageError` vs the `WebExtError`.

src/config.js
```diff
@@ -33,27 +31,61 @@ export function applyConfigToArgv({
...
-        defaultValues: defaultValues[commandExecuted],
+        configObject: configObject[option],
+        options: options[option],
         configFileName});
       continue;
     }
 
+    const decamelizedOptName = decamelize(option, '-');
+    let expectedType;
+    if (options[decamelizedOptName]) {
```
kumar303: Some of these if statements are duplications of other ones. I think this is why you couldn't simplify them before. Here is a way to re-organize it a bit to allow for simplification:

```diff
diff --git a/src/config.js b/src/config.js
index f1f194b..e5c1598 100644
--- a/src/config.js
+++ b/src/config.js
@@ -43,17 +43,19 @@ export function applyConfigToArgv({
     }
 
     const decamelizedOptName = decamelize(option, '-');
-    let expectedType;
-    if (options[decamelizedOptName]) {
-      if (options[decamelizedOptName].type === undefined) {
-        throw new UsageError(
-          `Option: ${option} was defined without a type.`);
-      } else {
-        expectedType = options[decamelizedOptName].type ===
-          'count' ? 'number' : options[decamelizedOptName].type;
-      }
+
+    if (typeof options[decamelizedOptName] !== 'object') {
+      throw new UsageError(`The config file at ${configFileName} specified ` +
+        `an unknown option: "${option}"`);
+    }
+    if (options[decamelizedOptName].type === undefined) {
+      throw new WebExtError(
+        `Option: ${option} was defined without a type.`);
     }
 
+    const expectedType = options[decamelizedOptName].type ===
+      'count' ? 'number' : options[decamelizedOptName].type;
+
     if (configObject[option]) {
       const optionType = typeof(configObject[option]);
       if (optionType !== expectedType &&
@@ -85,11 +87,6 @@ export function applyConfigToArgv({
       continue;
     }
 
-    if (typeof options[decamelizedOptName] !== 'object') {
-      throw new UsageError(`The config file at ${configFileName} specified ` +
-        `an unknown option: "${option}"`);
-    }
-
     newArgv[option] = configObject[option];
   }
   return newArgv;
@@ -114,4 +111,4 @@ export function loadJSConfigFile(filePath: string): Object {
       'Did you set module.exports = {...}?');
   }
   return configObject;
-}
\ No newline at end of file
+}
```

tests/unit/test.config.js
```diff
@@ -456,14 +488,233 @@ describe('config', () => {
...
+        },
+      });
+      const configObject = {
+        sign: {
+          'api-url': 2,
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params,
+          configObject});
```
kumar303: This one still needs fixing

shubheksha:
@kumar303, done!

src/config.js
```diff
@@ -33,31 +31,62 @@ export function applyConfigToArgv({
...
+    }
+    if (options[decamelizedOptName].type === undefined) {
+      throw new WebExtError(
+        `Option: ${option} was defined without a type.`);
+    }
+
+    const expectedType = options[decamelizedOptName].type ===
+      'count' ? 'number' : options[decamelizedOptName].type;
+
+    if (configObject[option]) {
```
kumar303: This if statement is no longer necessary since the recursive function is only triggered for object types now

src/config.js
```diff
@@ -33,31 +31,62 @@ export function applyConfigToArgv({
...
+      const optionType = typeof(configObject[option]);
+      if (optionType !== expectedType &&
+        expectedType !== undefined) {
```
kumar303: You can remove `expectedType !== undefined` because an error is already thrown for that case up above

src/config.js
```diff
@@ -33,31 +31,62 @@ export function applyConfigToArgv({
...
+    if (options[decamelizedOptName].type === undefined) {
+      throw new WebExtError(
+        `Option: ${option} was defined without a type.`);
+    }
+
+    const expectedType = options[decamelizedOptName].type ===
+      'count' ? 'number' : options[decamelizedOptName].type;
+
+    if (configObject[option]) {
+      const optionType = typeof(configObject[option]);
```
kumar303: To keep the code consistent with the rest of web-ext, make this:

```js
const optionType = typeof configObject[option];
```

src/config.js
```diff
@@ -33,31 +31,62 @@ export function applyConfigToArgv({
...
+        expectedType !== undefined) {
+        throw new UsageError(`The config file at ${configFileName} specified ` +
+          `the type of "${option}" incorrectly as "${optionType}"` +
+          ` (expected type: "${expectedType}")`);
+      }
+    }
+
+    let defaultValue;
+    if (options[decamelizedOptName]) {
+      if (options[decamelizedOptName].default !== undefined) {
```
kumar303: The order of if statements here are correct now but it needs test coverage. In other words, if I switch them the other way around, the tests should fail but they do not:

```js
      if (expectedType === 'boolean') {
        defaultValue = false;
      } else if (options[decamelizedOptName].default !== undefined) {
        defaultValue = options[decamelizedOptName].default;
      }
```

tests/unit/test.config.js
```diff
@@ -290,34 +289,86 @@ describe('config', () => {
...
+          verbose: {
+            type: 'count',
+          },
+        },
+      });
+      const configObject = {
+        verbose: 2,
+      };
+      const newArgv = applyConf({...params, configObject});
+      assert.strictEqual(newArgv.verbose, 0);
```
kumar303: Why does `newArgv.verbose` not equal 2? Shouldn't the configured value win?

tests/unit/test.config.js
```diff
@@ -456,14 +487,232 @@ describe('config', () => {
...
+          apiKey: 'fake-api-key',
+        },
+      };
+      assert.throws(() => {
+        applyConf({...params, configObject});
+      }, UsageError, 'UsageError: The config file at some/path/to/config.js ' +
+        'specified the type of "apiUrl" incorrectly');
+    });
+
+    it('throws an error when the type of option in config is missing', () => {
```
kumar303: This test spec shouldn't use the word config because that's misleading. This test is simulating a case where someone had programmed web-ext itself incorrectly (with a missing option type).

tests/unit/test.config.js
```diff
@@ -456,14 +487,232 @@ describe('config', () => {
...
+            default: 'pretend-default-value-of-apiKey',
+          },
+        });
+
+      const configObject = {
+        sign: {
+          apiUrl: 2,
+        },
+      };
+      const params = {
```
kumar303: There's no need to make a variable for this. Just pass them directly:

```js
applyConf({
  argv: program.yargs.exitProcess(false).argv, 
  options: program.options, 
  configObject,
});
```

src/config.js
shubheksha: I don't understand the need for switching these. You mentioned that `true` values will never be honored but I don't see how because the condition is `!== undefined`. Could you elaborate a little?

kumar303: All I mean is that this logic needs test coverage. I can tell right now that it doesn't have test coverage. As an exercise, you can check it yourself. Change the code to this (which is wrong):

```js
      if (expectedType === 'boolean') {
        defaultValue = false;
      } else if (options[decamelizedOptName].default !== undefined) {
        defaultValue = options[decamelizedOptName].default;
      }
```
At least one test should fail but they all pass. Once you write a failing test, change the code back and it should pass:

```js 
      if (options[decamelizedOptName].default !== undefined) {
        defaultValue = options[decamelizedOptName].default;
      } else if (expectedType === 'boolean') {
        defaultValue = false;
      }
```

kumar303: It just needs a test that defines a `boolean` option with a default value of true

tests/unit/test.config.js
shubheksha: I dug into this. This test works fine if a default value is explicitly specified in the option definition, but otherwise it triggers the `wasValueSetOnCli` condition to be true, assigning a default value of 0 (because the type is `count`, I guess) & ignores the configured value.

src/config.js
shubheksha: Got it, makes sense, thanks!

tests/unit/test.config.js
kumar303: ok, that sounds like a bug. It should use the configured value in this case. You will probably need to make a special exception for `count` types.

kumar303:
Sorry, I somehow missed that this patch was ready to go! Thanks for all your work on this.

PR #846, opened 2017-03-08
fix: Allowed --ignore-files option within sign command

### LINKED ISSUE #843: Option `--ignore-files` doesn't seem to affect `web-ext sign`
Hi,

A bug.

I'm trying to package and sign an extension, which is built with webpack.
I've got source directories in the tree that I don't want to include into the packaged extension.

With the `web-ext build` command, I'm using the global option `--ignore-file` to ignore those source directories, and it works as expected (i.e. `web-ext build --ignore-files ./src`).
But with `web-ext sign`, this option doesn't seem to be taken into account.

This makes the validation step fail, because it tries to validate source files written in ES6, and it keeps finding syntax errors in them.

I would expect `web-ext sign` to understand the `--ignore-files` options, as it is a global one.

```
node: v6.7.0
npm: 3.10.3
web-ext: 1.8.1
```

### Notes

If I understand the code correctly, it seems that in the `sign` command, the `--ignore-files` is not passed to the `build` sub-command (https://github.com/mozilla/web-ext/blob/master/src/cmd/sign.js#L69).

I'll be glad to contribute fixing this, but I wanted to check first if this was a documentation issue or one in the code.

felixgirault:
This is meant to fix #843.

Also worth noting, some unrelated tests failed on my computer, due to i18n : 

```
1) program.Program throws an error about unknown options:
     AssertionError: expected 'Argument inconnu: nope' to match /Unknown argument: nope/
      at Function.assert.match (node_modules/chai/lib/chai/interface/assert.js:883:32)
      at dist/webpack:/tests/unit/test.program.js:226:16
      at runMicrotasksCallback (internal/process/next_tick.js:58:5)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)

  2) program.Program throws an error about unknown sub-command options:
     AssertionError: expected 'Argument inconnu: nope' to match /Unknown argument: nope/
      at Function.assert.match (node_modules/chai/lib/chai/interface/assert.js:883:32)
      at dist/webpack:/tests/unit/test.program.js:237:16
      at runMicrotasksCallback (internal/process/next_tick.js:58:5)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

This is because an exception is thrown in french : `Argument inconnu: nope` instead of `Unknown argument: nope`.

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -24,6 +24,7 @@ export type SignParams = {|
   verbose?: boolean,
   sourceDir: string,
   artifactsDir: string,
+  ignoreFiles?: Array<string>,
   apiKey: string,
   apiSecret: string,
   apiUrlPrefix: string,
@@ -45,8 +46,8 @@ export type SignResult = {|
 
 export default function sign(
   {
-    verbose, sourceDir, artifactsDir, apiKey, apiSecret,
-    apiUrlPrefix, apiProxy, id, timeout,
+    verbose, sourceDir, artifactsDir, ignoreFiles = [],
+    apiKey, apiSecret, apiUrlPrefix, apiProxy, id, timeout,
   }: SignParams,
   {
     build = defaultBuilder, signAddon = defaultAddonSigner,
@@ -66,7 +67,7 @@ export default function sign(
       }
 
       const [buildResult, idFromSourceDir] = await Promise.all([
-        build({sourceDir, artifactsDir: tmpDir.path()},
+        build({sourceDir, ignoreFiles, artifactsDir: tmpDir.path()},
               {manifestData, showReadyMessage: false}),
         getIdFromSourceDir(sourceDir),
       ]);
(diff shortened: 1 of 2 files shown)


kumar303:
> Also worth noting, some unrelated tests failed on my computer, due to i18n

d'oh, could you file a separate bug for that? Good catch.

kumar303:
Here is some general info about getting started with the test suite https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md Thanks for taking the time to submit a patch!

felixgirault:
Thank you for the hint about the test implementation, it should be ok this time :)

felixgirault:
Also, I just saw your message about the other problem, I open an issue here https://github.com/mozilla/web-ext/issues/852

caitmuenster:
Thanks, @felixgirault! Your contribution has been added to the [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#March_2017). 

If you would like to create a profile on [mozillians.org](https://mozillians.org), I would be happy to vouch for you!

felixgirault:
Glad it helped, thank you for the quick responses ;)

PR #849, opened 2017-03-08
fix: Changed web-ext build to suggest submitting the generated zip file to AMO (Fixes #847)

### LINKED ISSUE #847: web-ext build shouldn't say "Your web extension is ready" if manifest.json doesn't have applications.gecko.id key
related to https://bugzilla.mozilla.org/show_bug.cgi?id=1344674

a kind of UX bug

Steps:
1. create `test` directory
2. enter `test` directory
3. create `manifest.json` with the following content
```
{
  "manifest_version": 2,
  "name": "test",
  "version": "1.0",
  "description": "test"
}
```
4. run `web-ext build`
5. open Nightly 55.0a1 (2017-03-08) (64-bit)
6. open about:addons
7. drag and drop generated .zip file onto "Extensions" pane

Result:

`web-ext build` prints the following message:
```
Building web extension from /tmp/test
Your web extension is ready: /tmp/test/web-ext-artifacts/test-1.0.zip
```

and then Firefox says:
```
undefined
This add-on couldn't be installed because it appears to be corrupt.
```

`web-ext build` shouldn't say "Your web extension is ready" for this case.

Firefox is complaining the add-on is corrupt because `manifest.json` doesn't have `applications.gecko.id` key.

It's hard to figure out why Firefox is saying it's corrupt, because there's no issue while debugging with temporary add-on, and `applications.gecko.id` doesn't exist in almost all examples.

`web-ext build` should check the existence of `applications.gecko.id` key before saying "Your web extension is ready",
and if there's no `applications.gecko.id` key, it should suggest adding it, or just say "Your web extension is ready for submitting to addons.mozilla.org",
to prevent developers from trying to install the zip to Firefox and bumps into the unhelpful error message.

revision c3ff1c3ee7eb81aaf62522b00e0e7a310bfec02a

* **Firefox version**: 55.0a1 (2017-03-08) (64-bit)
* **Your OS and version**: macOS Sierra 10.12.3

````
````

```
v7.5.0
4.1.2
master-c3ff1c3ee7eb81aaf62522b00e0e7a310bfec02a
```

### PR DESCRIPTION
**Pull request by arai-a**
this fixes https://github.com/mozilla/web-ext/issues/847

Now `web-ext build` prints the following message:

```
Building web extension from /tmp/test
Your web extension is ready for being submitted on addons.mozilla.org: /tmp/test/web-ext-artifacts/test-1.0.zip
```

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -131,7 +131,8 @@ async function defaultPackageCreator({
   await eventToPromise(stream, 'close');
 
   if (showReadyMessage) {
-    log.info(`Your web extension is ready: ${extensionPath}`);
+    log.info('Your web extension is ready for being submitted on ' +
+             `addons.mozilla.org: ${extensionPath}`);
   }
   return {extensionPath};
 }

### CODE REVIEW

File: src/cmd/build.js [hunk 1]
```diff
@@ -131,7 +131,8 @@ async function defaultPackageCreator({
   await eventToPromise(stream, 'close');
 
   if (showReadyMessage) {
-    log.info(`Your web extension is ready: ${extensionPath}`);
+    log.info(`Your web extension is ready for being submitted on ` +
```
rpl: This line currently raises an eslint error (https://travis-ci.org/mozilla/web-ext/jobs/209043848#L1227), the reason is that the first part of the string doesn't contain any interpolated value and the eslint rule enforces a single quote string in this case.

arai-a: thanks!
fixed

src/cmd/build.js
```diff
@@ -131,7 +131,8 @@ async function defaultPackageCreator({
   await eventToPromise(stream, 'close');
 
   if (showReadyMessage) {
-    log.info(`Your web extension is ready: ${extensionPath}`);
```
kumar303: I think the old message was correct. Here is why:
* A built `.zip` runs fine in Chrome
* A built `.zip` can be submitted for signing by Mozilla through the [API](http://addons-server.readthedocs.io/en/latest/topics/api/signing.html).
* Submitting the `.zip` to addons.mozilla.org is only one of many things you can do
* The bug you noticed where Firefox thinks the extension is corrupt is a [bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1224591#c6). In other words, the error message is terrible but no one is convinced enough to fix (it should be fixed IMO).

arai-a: in that case, how do you think about adding a note that says the built .zip doesn't work on Firefox without signing?

kumar303: Adding a note where? In the docs? I saw the [warning you added](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Packaging_your_extension) to the docs already. It might scare some people into using an ID when they don't have to but I guess it's helpful.

arai-a: I meant, in the output of `web-ext build`, like this (I'm not sure what would the good wording for it is tho):

```
Building web extension from /tmp/test
Your web extension is ready: /tmp/test/web-ext-artifacts/test-1.0.zip
NOTE: the generated zip file may not work on Firefox without signing it.
```

about the documentation, I'll change it to suggest signing first.

kumar303: A generated zip should work fine in Firefox without signing it (as long as signing is disabled and bugs are fixed!) so I'm hesitant to add a note like this. There are a lot of use cases for the `build` command which I started listing at the top of this thread. 

Maybe it would just be more helpful to fix the bug in Firefox that initially caused confusion? It might help to comment on that bug to try and convince them why it's important.

arai-a: so, the expected behavior here is that:
  * "web-ext build" says "it's ready"
  *  Firefox accepts the .zip file even if there's no signature nor id, as long as signature check is disabled
right?

arai-a:
sorry for closing.

src/cmd/build.js
kumar303: I think Firefox should accept the .zip without an ID as long as signature checks are disabled but this doc is suggesting otherwise: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/WebExtensions_and_the_Add-on_ID#Basic_workflow_with_no_Add-on_ID

PR #853, opened 2017-03-09
test: Fix test failure in non-English locales

### LINKED ISSUE #852: Tests are failing due to i18n

A bug.

Some tests in the test suite are checking the message of an exception.
This message seems to be localized, but the test doesn't take that into account.

```
  1) program.Program throws an error about unknown options:
     AssertionError: expected 'Argument inconnu: nope' to match /Unknown argument: nope/
      at Function.assert.match (node_modules/chai/lib/chai/interface/assert.js:883:32)
      at dist/webpack:/tests/unit/test.program.js:226:16
      at runMicrotasksCallback (internal/process/next_tick.js:58:5)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)

  2) program.Program throws an error about unknown sub-command options:
     AssertionError: expected 'Argument inconnu: nope' to match /Unknown argument: nope/
      at Function.assert.match (node_modules/chai/lib/chai/interface/assert.js:883:32)
      at dist/webpack:/tests/unit/test.program.js:237:16
      at runMicrotasksCallback (internal/process/next_tick.js:58:5)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

The exception is thrown in french (`Argument inconnu: nope`) instead of in english (`Unknown argument: nope`).

The tests should not fail if they are run on a non-english machine.

````
node: v6.7.0
npm: 3.10.3
web-ext: 1.8.1
````

kumar303:
Fixes #852

### DIFF
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -221,9 +221,9 @@ describe('program.Program', () => {
     return execProgram(new Program(['--nope']))
       .then(makeSureItFails())
       .catch((error) => {
-        // It's a bit weird that yargs calls this an argument rather
-        // than an option but, hey, it's an error.
-        assert.match(error.message, /Unknown argument: nope/);
+        // Make sure that the option name is in the error message.
+        // Be careful not to rely on any text from yargs since it's localized.
+        assert.match(error.message, /nope/);
       });
   });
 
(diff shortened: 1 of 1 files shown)


kumar303:
I think adding a note sounds good but I'm not sure we need to slow down our builds (with yet another matrix) just to test it. I think it's fine for developers to file bugs when they notice problems :)

PR #868, opened 2017-03-17
feat: Added Firefox for Android support and --android and --android-device CLI options

### LINKED ISSUE #737: Support `web-ext run` on Firefox for Android (fennec)

feature

`web-ext run` will let you test an extension on Firefox desktop.

It would be nice to type `web-ext run --remote-firefox=X.X.X.X:9001` (see https://github.com/mozilla/web-ext/issues/521) on your desktop work station to test the extension in Firefox for Android on your phone.

This depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1305143

rpl:
This PR contains an initial prototype of "web-ext run" support on Firefox for Android and it should fix #737 by providing mostly the same "addon developer experience" and features supported by "web-ext run" on Firefox Desktop (e.g. run Firefox for Android in a temporary Firefox profile, watch the sources and then rebuild and reinstall the addon on changes, show a desktop notification when the build or install is failing with errors).

The current implementation already passes the eslint and flow type checks, and it doesn't break any of the existent tests, and I'm going to create some additional tests for the android support module asap.

In the meantime, you can watch the following demo screencast and/or build web-ext from this branch if you want to use it immediately and provide some preliminary feedback, that will be very much appreciated :-)

https://youtu.be/C_R1C2s3AWQ

Testing a WebExtension on Android is never been so easy! 
\o/ Enjoy!

## TODO List

- [x] refactor this branch on top of #962
- [x] waiting for #962 to be ready to land
- [x] add new tests to cover the new extension runner and the new code in cmd/run
- [x] choose a random tcp port to be used for the "remote debugging adb forwarding"
- [x] ensure that the unit tests are as fast as possible
- [x] speed up the android extension runner as much as possible (e.g. by running some
  of the setup steps in parallel)

## Followups List

- #985 - ensure that when 'web-ext run' is configured to run the extension on multiple targets (e.g. Firefox Desktop and Firefox Android), the logged messages (and errors) makes clear which of the target the message (or error) is coming from


rpl:
The travis build is currently failing because of `nsp check` detecting a vulnerability in a library included by a dependency of adbkit (the nodejs library used by this PR to interact with adb):

https://travis-ci.org/rpl/web-ext/jobs/212115440#L1234-L1237

I've reported this issue to the upstream in openstf/adbkit#67

kumar303:
This is very cool! I only took a quick look so here are some initial thoughts.

This is the first time we will be introducing a new `run` target. In theory, one should be able to say "run my add-on in both desktop Firefox and Firefox for Android" and have it install/run in both places simultaneously. This would let you take advantage of the auto-reloader while testing cross-platform functionality. Such a use case would also apply to running an add-on in both Firefox and Chrome. Should we try and work towards this goal in this patch? It could come later too. 

To achieve multiple targets like this, you'd probably want to fold your code into the already existing abstraction around running and watching/reloading. That code is a bit hairy though.

The command line option should probably signify "firefox" in some way since Chrome for Android may also support add-ons in the future. So maybe `web-ext run --firefox-mobile` or `--firefox-android` ? We may also want to consider handling multiple targets like `web-ext run --target firefox firefox-android chrome` (i.e. array type option).

shatur:
Hey!
Out of curiosity, I tried to use this really cool feature. But I am unable to load my web-extension on fennec on a real device(tried it on multiple devices). Then I tried it on emulator device and it worked fine. So, @rpl have you tried it on a real device or this issue is on my end.

I am really sorry if this is not the right place to ask this question.

rpl:
@shatur Thanks for trying this PR, I didn't tried it on a real device yet (I was trying to ensure that it will not be destructive first ;-)), did the issue show any visible error (e.g. adb logcat)?

Thanks again for your help.

rpl:
Hey @kumar303, thanks for this first round of comments, I like the idea of preparing the CLI options to be able to turn "web-ext run" into a multi-target command.

Also, my initial intention was to add the android support into the cmd/run.js module, but I had the immediate feeling that it needs a refactoring first and so I though to defer the refactoring a bit by starting with creating a new module (but preventing any duplication of shared feature between the two) and, so that we could take a look at the different workflows/needs of the two targets in this PR and then refactor the two modules accordingly (possibly before landing this feature).

I'll take a look at both asap.

(Another reason why I deferred the refactoring is that we have a couple of pending PRs from contributors that would be affected by a refactoring on the "./cmd/run.js" module, and so I opted to wait a bit, so that I can rebase my change on top of them and start the refactoring from there).

shatur:
Hey @rpl!
Did a little more digging, and I think the problem I am facing is due to my rooted device. I tried this PR on 4 devices, out of which two were rooted devices. This PR works really fine on non-rooted device:-). But on rooted device it gives me this error https://pastebin.mozilla.org/8983546. I will update you if I find something new!

rpl:
@shatur my apologies for not being able to come back to this sooner (it turned out that it would have been better to refactor the `"cmd/run"` module before introducing these changes, and so, now that we have merged the refactoring PR in #962, I started to look actively into this again).

I suspect that the issue could be something related to something that is installed only on the rooted devices and that can prevent it from working correctly (I did most of the testing while working on this on a rooted emulator and it didn't have any issue, and so I think that the reason is not that the devices are "rooted" but something that is installed only on the "rooted" devices used to test it).

Another possible issue can be related to the specific Android versions (the emulator I used was an Android 4.3 AOSP), I'll try to put together a bit more variegated testing environment before landing it.

also, thank you a lot for testing it on some non-rooted devices, that was definitely an important scenario, that I didn't have tested yet this branch on.

src/cmd/run.js
```diff
@@ -69,32 +87,83 @@ export default async function run(
...
+      firefoxClient,
+      adb: defaultADB,
+      desktopNotifications: defaultDesktopNotifications,
+      buildSourceDir: (extensionSourceDir: string) => {
+        return buildExtension({
+          sourceDir: extensionSourceDir,
+          artifactsDir,
+          ignoreFiles,
+          asNeeded: false,
+          // TODO: choose a different artifactsDir for additional safety?
```
rpl: @kumar303 tons of unit tests (and also some "code removal") later... I think that PR is almost ready for you.

Unfortunately even after the "code removal" phase, it is still a quite big (most of it is composed by the new extension runner file and its related test file).
On the bright side, the changes in this PR are enough to be able to also use the feature for real and being sure that the unit tests are testing actual real scenarios
(I'm still tempted to add some additional test case :-P).

This line is the only remaining one from my current todo list related to this PR that I haven't worked on yet, we can probably defer it to a follow up PR (and prevent this one from growing further).

I'll take another look at the final diff of this PR tomorrow, and then we can start to review it.

rpl:
Hey @kumar303,
this is now ready for your review, while re-looking at the full diff I thought that a "high level view" of the `FirefoxAndroidExtensionRunner.prototype.run` method (which contains most of the steps needed to prepare the android device to run the extension) could be helpful as a "map" during the review, and so I've put together a smal (sort of) UML activity diagram of it:

FirefoxAndroidExtensionRunner.prototype.run - UML activity diagram

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+      selectedAdbDevice,
+      params: {
+        adb,
+        firefoxApk,
+      },
+    } = this;
+    // Discovery and select a Firefox for Android version.
+    const pmList = await adbClient.shell(selectedAdbDevice.id, [
+      'pm', 'list', 'packages',
+    ]).then(adb.util.readAll);
```
rpl: @kumar303 I'm thinking that we should turn this `adbClient.shell(...).then(adb.util.readAll)` pattern into a reusable helper, I'm afraid that forgetting to drain the output from an `adbClient.shell` command can lead to bugs hard to debug (e.g. if the output of a previous `adbClient.shell` command can become part of the output read from another unrelated `adbClient.shell` command which reads the output using `adb.util.readAll`).

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+
+    await adbClient.forward(selectedAdbDevice.id, `tcp:${tcpPort}`,
+                            `localfilesystem:${this.selectedRDPSocketFile}`);
+    this.selectedTCPPort = tcpPort;
+  }
+
+  chooseLocalTcpPort(): Promise<number> {
+    return new Promise((resolve) => {
+      const srv = net.createServer();
+      // $FLOW_FIXME: flow has his own opinions on this method signature.
```
rpl: I hate this `$FLOW_FIXME`, I'll take a look if I can find a better way to handle this (e.g. explicitly define the additional signature or use one of the signatures that flow already recognize as valid)

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+    // Log the choosen tcp port at info level (useful to the user to be able
+    // to connect the Firefox DevTools to the Firefox for Android instance).
+    log.info(`Android Remote Debugging TCP Port selected: ${tcpPort}`);
+
+    await adbClient.forward(selectedAdbDevice.id, `tcp:${tcpPort}`,
+                            `localfilesystem:${this.selectedRDPSocketFile}`);
+    this.selectedTCPPort = tcpPort;
+  }
+
+  chooseLocalTcpPort(): Promise<number> {
```
rpl: This is also something that I think we can refactor and move into a separate module, so that we can reuse in other places where it could be useful to easily choose a free tcp port.

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+        throw new WebExtError(
+          'Timeout while waiting for the Android Firefox Debugger Socket'
+        );
+      }
+
+      await new Promise((resolve) => {
+        setTimeout(resolve, unixSocketDiscoveryRetryTime);
+      });
+
+      log.info(`Waiting ${selectedFirefoxApk} to be ready... ` +
```
kumar303: This should be `Waiting for ${selectedFirefoxApk} to be ready...`

kumar303: I thought it was a bit annoying to see this message printed repeatedly so I made a little loading indicator if you want to use it. It might be hard to test though. Also, I think the modulo math is off by one maybe.

```diff
diff --git a/src/extension-runners/firefox-android.js b/src/extension-runners/firefox-android.js
index fae5b61..dfbd7d3 100644
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -540,29 +540,50 @@ export class FirefoxAndroidExtensionRunner {
       stdin.on('keypress', handleCtrlC);
     }
 
-    while (!userExit && androidUnixSockets.length === 0) {
-      if (Date.now() - startedDiscoveryTime > unixSocketDiscoveryMaxTime) {
-        throw new WebExtError(
-          'Timeout while waiting for the Android Firefox Debugger Socket'
-        );
-      }
 
-      await new Promise((resolve) => {
-        setTimeout(resolve, unixSocketDiscoveryRetryTime);
-      });
+    const loadingPositions = ['---', ' / ', ' | ', ' \ '];
+    log.info(`Waiting ${selectedFirefoxApk} to be ready... `);
+    let loadingPos = 0;
+
+    const loader = setInterval(() => {
+      // Show a little loading indicator.
+      process.stdout.write(
+        `\r${loadingPositions[loadingPos % (loadingPositions.length - 1)]}`);
+      loadingPos += 1;
+    }, 200);
 
-      log.info(`Waiting ${selectedFirefoxApk} to be ready... ` +
-               '(it can take a while on an ARM emulator)');
-      androidUnixSockets = await adbClient.shell(selectedAdbDevice.id, [
-        'cat', '/proc/net/unix',
-      ]).then(adb.util.readAll);
+    function clearLoader() {
+      clearInterval(loader);
+    }
 
-      androidUnixSockets = androidUnixSockets.toString()
-        .split('\n').filter((line) => {
-          return line.trim().endsWith('firefox-debugger-socket');
+    while (!userExit && androidUnixSockets.length === 0) {
+      try {
+        if (Date.now() - startedDiscoveryTime > unixSocketDiscoveryMaxTime) {
+          throw new WebExtError(
+            'Timeout while waiting for the Android Firefox Debugger Socket'
+          );
+        }
+
+        await new Promise((resolve) => {
+          setTimeout(resolve, unixSocketDiscoveryRetryTime);
         });
+
+        androidUnixSockets = await adbClient.shell(selectedAdbDevice.id, [
+          'cat', '/proc/net/unix',
+        ]).then(adb.util.readAll);
+
+        androidUnixSockets = androidUnixSockets.toString()
+          .split('\n').filter((line) => {
+            return line.trim().endsWith('firefox-debugger-socket');
+          });
+      } catch (error) {
+        clearLoader();
+        throw error;
+      }
     }
 
+    clearLoader();
+
     if (stdin.isTTY && stdin instanceof tty.ReadStream) {
       stdin.removeListener('keypress', handleCtrlC);
     }
```

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+    });
+
+    if (foundDevices.length === 0) {
+      const devicesMsg = JSON.stringify(devices);
+      throw new UsageError(
+        `Android Device not found: ${adbDevice} in ${devicesMsg}`);
+    }
+
+    this.selectedAdbDevice = foundDevices[0];
+    log.info(`Selected adb device: ${this.selectedAdbDevice.id}`);
```
kumar303: I think this could be a debug message

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+  async adbForceStopSelectedPackage() {
+    const {
+      adbClient,
+      selectedAdbDevice,
+      selectedFirefoxApk,
+      params: {
+        adb,
+      },
+    } = this;
+    log.info(`Stop any existent instance of ${selectedFirefoxApk}...`);
```
kumar303: Stopping any existing instances of...

kumar303: This should be a debug message

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+    const {
+      adbClient,
+      selectedAdbDevice,
+      selectedFirefoxApk,
+      params: {
+        firefoxApp,
+      },
+    } = this;
+    // Create the preferences file and the Fennec temporary profile.
+    log.info(`Preparing a temporary profile for ${selectedFirefoxApk}...`);
```
kumar303: This should be a debug message

src/extension-runners/firefox-android.js
```diff
@@ -0,0 +1,653 @@
...
+    const {
+      adbClient,
+      selectedFirefoxApk,
+      selectedAdbDevice,
+    } = this;
+
+    const deviceProfileDir = this.getDeviceProfileDir();
+
+    log.info(
+      `Starting ${selectedFirefoxApk} on the profile ${deviceProfileDir}...`
```
kumar303: I think it's not so important to tell the user what profile it is so this could be split into two messages:
```js
log.info(`Starting ${selectedFirefoxApk}...`); 
log.debug(`Using profile ${deviceProfileDir}`);
```

src/cmd/run.js
```diff
@@ -69,32 +87,83 @@ export default async function run(
...
+      adbPort,
+      adbBinary,
+
+      // Injected dependencies.
+      firefoxApp,
+      firefoxClient,
+      adb: defaultADB,
+      desktopNotifications: defaultDesktopNotifications,
+      buildSourceDir: (extensionSourceDir: string) => {
+        return buildExtension({
```
kumar303: This call should add `showReadyMessage: false` so that it doesn't log an info message

kumar303:
I was getting a timeout when trying to run [this extension](https://github.com/mdn/webextensions-examples/tree/master/context-menu-demo) on my Pixel. I saw this in the log:

```
SQLiteDatabase: Failed to open database '/sdcard/web-ext-artifacts-1499372416336/profile/browser.db
'.
```

... but maybe there are other errors? Here are some logs:

My output from `web-ext run --target firefox-android -v --android-device FA6AX0301736 --firefox-apk org.mozilla.fennec_aurora`: [android-web-ext-verbose-log.txt](https://github.com/mozilla/web-ext/files/1129348/android-web-ext-verbose-log.txt)

And an associated `adb logcat` (from a different session): 
[android-web-ext-adb-log-fennec.txt](https://github.com/mozilla/web-ext/files/1129351/android-web-ext-adb-log-fennec.txt)

It timed out on Firefox release too. Here is the `adb logcat` from that: 
[android-web-ext-adb-log-release.txt](https://github.com/mozilla/web-ext/files/1129352/android-web-ext-adb-log-release.txt)

rpl:
@kumar303 I've been finally able to dig into this and find which are the issues on Android versions >= Lollypop.

Follow some additional details about the issues and the related applied fixes.

> I was getting a timeout when trying to run [this extension](https://github.com/mdn/webextensions-examples/tree/master/context-menu-demo) on my Pixel. I saw this in the log:
> ```
> SQLiteDatabase: Failed to open database '/sdcard/web-ext-artifacts-1499372416336/profile/browser.db
> '.
> ```

This error is due to the `android.permission.READ_EXTERNAL_STORAGE` and `android.permission.WRITE_EXTERNAL_STORAGE` not granted to the Firefox apk, which is required starting from Android Lollypop (android sdk >= 21) to be able to access (and modify) the files of the temporarily created Firefox profile (which we absolutely need, e.g. because of the custom preferences that we set on that profile).

In the updated version of this PR, the Firefox for Android ExtensionRunner detects the android sdk version and if it is `>= 21`, it checks the granted permissions and raises an helpful error message if the required permissions are not already granted (we could grant them automatically, but I'm a bit worried of granting this permission automatically on a device, it is a one time change and so it is probably better to leave it to the user).

After solving this issue, I noticed that the remote debugging unix socket file discovery is failing on Android Lollypop, probably due to a different formatting on the output lines, and so I tweaked it so that it is able to extract the needed unix socket file path across the android versions.

tests/unit/test-util/test.adb.js
```diff
@@ -0,0 +1,887 @@
...
+} from '../../../src/errors';
+import ADBUtils from '../../../src/util/adb';
+
+const fakeADBPackageList = `
+package:org.mozilla.fennec
+package:org.mozilla.firefox
+package:org.some.other.software
+`;
+
+const fakeSocketFilePrefix = (
```
kumar303: Could you add a comment about this? Did you copy it from a real file?

tests/unit/test-util/test.adb.js
```diff
@@ -0,0 +1,887 @@
...
+           adbClient: {
+             shell: sinon.spy(() => Promise.resolve('')),
+           },
+           adbkitUtil: {
+             readAll: sinon.spy(() => {
+               return Promise.resolve(
+                 new Buffer(`
+                android.permission.READ_EXTERNAL_STORAGE: granted=true
+                android.permission.WRITE_EXTERNAL_STORAGE: granted=true
+              `)
```
kumar303: this indent looks off

src/extension-runners/index.js
```diff
@@ -31,6 +31,7 @@ export type MultiExtensionRunnerParams = {|
 
 // Export everything exported by the firefox-desktop runner.
 export * from './firefox-desktop';
+export * from './firefox-android';
```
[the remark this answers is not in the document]
rpl: `export * from '...'` only exports what the module has exported explicitly and not what they have imported inside the module it.

Nevertheless I'm totally ok to change it if we don't like it because it is not explicit enough.

tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -0,0 +1,799 @@
...
+    it('resolves to an array of WebExtError if the extension is not reloadable',
+       async () => {
+         const {params} = prepareSelectedDeviceAndAPKParams();
+
+         const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+         await runnerInstance.run();
+
+         await runnerInstance.reloadExtensionBySourceDir(
+           '/non-existent/source-dir'
+         ).then((results) => {
```
kumar303: Since you're using `await`, this call to `then()` could be removed.

tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -0,0 +1,799 @@
...
+               () => Promise.reject(Error('Reload failure'))
+             ),
+           },
+         });
+
+         const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+         await runnerInstance.run();
+
+         await runnerInstance.reloadAllExtensions()
+           .then((results) => {
```
kumar303: It looks like you can remove this call to `then()` as well

tests/unit/test-util/test.adb.js
```diff
@@ -0,0 +1,896 @@
...
+
+// NOTE: fake 'pm dump <APK>' output related to the granted permissions for an
+// android application.
+const fakeAndroidGrantedPermissions = `
+android.permission.READ_EXTERNAL_STORAGE: granted=true
+android.permission.WRITE_EXTERNAL_STORAGE: granted=true
+`;
+
+// Enable chai-as-promised plugin.
+chai.use(chaiAsPromised);
```
kumar303: Is this making a global change to the test suite? If so, I think it would be better in [`tests/unit/runner.js`](https://github.com/mozilla/web-ext/blob/master/tests/unit/runner.js) -- would that work?

PR #873, opened 2017-03-21
New firefox-profile version

### LINKED ISSUE #870: web-ext run -p <> does not preserve user.js settings
Bug

user.js settings are not taken from --firefox-profile at all, instead they are directly overwritten with just the ones in [preferences.js](https://github.com/mozilla/web-ext/blob/master/src/firefox/preferences.js#L22)
Repro:
```
mkdir test_profile
echo 'user_pref("marionette.defaultPrefs.port", 34209);' > test_profile/user.js
yarn run web-ext run -p test_profile # --keep-profile-changes # is helpful for testing
# exit firefox
grep "marionette.defaultPrefs.port" test_profile/user.js
# is not found, or look in about:config for the setting which isnt there
```
Note that the preferences are ignored even without keep-changes, they will just be copied to a tmpdir though, so harder to track down.

This is an issue for me because I am trying to get a selenium environment up and running with my extension installed via web-ext. [WIP branch here](https://github.com/lutostag/web-ext/tree/selenium_arg_passthrough) I have everything tied together with an [almost working example](https://github.com/lutostag/webextension-formsave). This is basically my attempt at #5 for true e2e testing without stubs.

That preferences of a profile are correctly merged with the given preferences.js not entirely ignored and lost.

* **Firefox version**: Mozilla Firefox 52.0
* **Your OS and version**: Ubuntu 16.10

````
$ node --version && npm --version && yarn run web-ext --version
v6.10.0
3.10.10
0.21.3
````

saadtazi:
When a profile has user preferences, the new firefox-profile version makes sure to load those preferences.

This should fix #870.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "es6-error": "4.0.2",
     "es6-promisify": "5.0.0",
     "event-to-promise": "0.8.0",
-    "firefox-profile": "0.4.9",
+    "firefox-profile": "0.5.0",
     "fx-runner": "1.0.6",
     "git-rev-sync": "1.8.0",
     "minimatch": "3.0.3",


kumar303:
Thanks. I just merged this in our auto-updater https://github.com/mozilla/web-ext/pull/872

PR #885, opened 2017-04-03
chore(package): update sinon to version 2.1.0

kumar303:
Closes #822

https://greenkeeper.io/


kumar303:
I don't see anything that needs changing so this should be good to go. http://sinonjs.org/releases/v2.0.0/migrating-to-2.0/

PR #923, opened 2017-04-25
fix: initiate remote debugger with firefox runner's port

### LINKED ISSUE #821: `web-ext run` does not install extension if another Firefox is open

bug

This is how I reproduced it. 

* I had `web-ext run -f firefoxdeveloperedition` going with one extension
* I fired off `web-ext run -f nightly` inside the [window-manipulator](https://github.com/mdn/webextensions-examples/tree/master/window-manipulator) example.
* Nightly started up but the extension was not installed and there were no obvious errors 

The second extension should have been installed or there should have been an error. When I stopped the first instance of `web-ext run` and retried the second one, the extension was installed as expected.

* **Firefox version**: 53.0a2 (2017-02-24) (64-bit) and 54.0a1 (2017-02-26) (64-bit)
* **Your OS and version**: Mac OS X Sierra

````
````

v6.9.4
3.10.10
master-c08360dafb3e1f3df8454e359992bef6f2171e10 (1.8.1)

Verbose log: 
[web-ext-run-window-manipulator.txt](https://github.com/mozilla/web-ext/files/802733/web-ext-run-window-manipulator.txt)

It looks like maybe there is an issue with retrying the connection to the remote debugger.

### PR DESCRIPTION
**Pull request by ccarruitero**
Fixes #821

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -25,7 +25,7 @@ import {
 import type {FirefoxPreferences} from '../firefox/preferences';
 import type {OnSourceChangeFn} from '../watcher';
 import type {
-  FirefoxProcess, // eslint-disable-line import/named
+  FirefoxProcess, FirefoxInfo, // eslint-disable-line import/named
 } from '../firefox/index';
 import type {
   FirefoxConnectorFn, RemoteFirefox,
@@ -172,15 +172,16 @@ export function defaultReloadStrategy(
 
 export type CreateFirefoxClientParams = {|
   connectToFirefox?: FirefoxConnectorFn,
-  maxRetries: number,
-  retryInterval: number,
+  maxRetries?: number,
+  retryInterval?: number,
+  port: number,
 |};
 
 export function defaultFirefoxClient(
   {
     connectToFirefox = defaultFirefoxConnector,
     // A max of 250 will try connecting for 30 seconds.
-    maxRetries = 250, retryInterval = 120,
+    maxRetries = 250, retryInterval = 120, port,
   }: CreateFirefoxClientParams = {}
 ): Promise<RemoteFirefox> {
   async function establishConnection() {
@@ -188,7 +189,7 @@ export function defaultFirefoxClient(
 
     for (let retries = 0; retries <= maxRetries; retries++) {
       try {
-        return await connectToFirefox();
+        return await connectToFirefox(port);
       } catch (error) {
         if (isErrorWithCode('ECONNREFUSED', error)) {
           // Wait for `retryInterval` ms.
@@ -290,13 +291,13 @@ export default async function run(
     installed = true;
   }
 
-  const runningFirefox = await runner.run(profile);
+  const { firefox: runningFirefox, debuggerPort } = await runner.run(profile);
 
   if (installed) {
     log.debug('Not installing as temporary add-on because the ' +
               'add-on was already installed');
   } else if (requiresRemote) {
-    client = await firefoxClient();
+    client = await firefoxClient({ port: debuggerPort });
 
     try {
       addonId = await runner.installAsTemporaryAddon(client).then(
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/cmd/run.js [hunk 1]
```diff
@@ -177,6 +177,7 @@ export type CreateFirefoxClientParams = {|
 |};
 
 export function defaultFirefoxClient(
+  port: number,
```
kumar303: This argument should be an object key, not a positional argument

src/cmd/run.js
```diff
@@ -296,7 +297,8 @@ export default async function run(
     log.debug('Not installing as temporary add-on because the ' +
               'add-on was already installed');
   } else if (requiresRemote) {
-    client = await firefoxClient();
+    const port = runningFirefox.spawnargs[2];
```
kumar303: I think you can just tell Firefox to start the debugger server on a specific port. You would pass it like [`binaryArgs.push('--listen', '6005')`](https://github.com/ccarruitero/web-ext/blob/4f4b9efd5834f2603c8bb8c4d9d399652567b206/src/cmd/run.js#L425) (see [fx-runner](https://github.com/mozilla-jetpack/node-fx-runner)).

ccarruitero: We already started firefox at this point, `runningFirefox` is the process returned from `fx-runner`. That we need here is retrieve the port in which the debugger server was started.

I was thinking in maybe extend the process. But I don't think that will be easy. So maybe we just can move when [find the port](https://github.com/mozilla/web-ext/blob/master/src/firefox/index.js#L136) and run that before [initiate the runner](https://github.com/ccarruitero/web-ext/blob/4f4b9efd5834f2603c8bb8c4d9d399652567b206/src/cmd/run.js#L271) and then pass it the to the runner and the client.

src/firefox/index.js
```diff
@@ -127,13 +128,13 @@ export async function run(
...
     findRemotePort = defaultRemotePortFinder,
-    firefoxBinary, binaryArgs,
+    firefoxBinary, binaryArgs, port,
   }: FirefoxRunOptions = {}
 ): Promise<FirefoxProcess> {
 
   log.debug(`Running Firefox with profile at ${profile.path()}`);
 
-  const remotePort = await findRemotePort();
+  const remotePort = port || await findRemotePort();
```
kumar303: Huh. I just realized that the call to `await findRemotePort()` means that Firefox was always starting up with a unique remote debugger port. Maybe all we need to do is return the port? Currently it returns the `firefox` instance so maybe it needs to return `{firefox, port}` instead.

src/cmd/run.js
```diff
@@ -290,13 +291,14 @@ export default async function run(
     installed = true;
   }
 
-  const runningFirefox = await runner.run(profile);
+  const remotePort = await runner.findPort();
+  const runningFirefox = await runner.run(profile, remotePort);
```
kumar303: after the change I mentioned below, this could be:

````js
const { firefox: runningFirefox, port: remotePort } = await runner.run(profile, remotePort);
````

File: src/firefox/index.js [hunk 5]
```diff
@@ -109,6 +109,11 @@ export type FirefoxRunnerFn =
   (params: FirefoxRunnerParams) => Promise<FirefoxRunnerResults>;
 
 
+export type FirefoxRunResponse = {|
```
kumar303: We should call this `FirefoxInfo` to make it more specific to the data rather than the method that returns the data.

src/firefox/index.js
```diff
@@ -109,6 +109,11 @@ export type FirefoxRunnerFn =
...
+  firefox: FirefoxProcess,
+  port: number,
```
kumar303: this should be called `debuggerPort` to make it unambiguous.

tests/unit/test-cmd/test.run.js
```diff
@@ -104,6 +105,21 @@ describe('run', () => {
     });
   });
 
+  it('run extension in correct port', () => {
```
kumar303: grammar nitpick: `it('runs the extension...`

ccarruitero:
@kumar303 I just amend the changes. Let me know if there are something else.

caitmuenster:
Thanks so much, @ccarruitero! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#April_2017) and your mozillians.org profile has been vouched for. 

Welcome onboard! I look forward to seeing you around!

PR #937, opened 2017-05-04
fix: set runAsBinary according shouldExitProgram for lint command

### LINKED ISSUE #823: web-ext lint should not exit the process when shouldExitProgram is set to false

bug

When web-ext is used as a library to be integrated into another tool (e.g. gulp, grunt, webpack etc.)
it should not exit the nodejs application that integrates it, which should be achievable using the [`shouldExitProgram`](https://github.com/mozilla/web-ext/blob/4be7a6a4f7ca35c8c31227c5e3778d61a7e5291b/src/program.js#L121-L122) property of the `Program` class.

Unfortunately, when the `web-ext lint` command is executed, web-ext currently passes [`runAsBinary: true`](https://github.com/mozilla/web-ext/blob/4be7a6a4f7ca35c8c31227c5e3778d61a7e5291b/src/cmd/lint.js#L89) to the created `Linter` instance, which is going to make the command to exit the nodejs process once the linting scan and the linting report has been completed.

We should set `runAsBinary` to `false` when `shouldExitProgram` is set to `false`, so that the `web-ext lint` command will have the same behavior of the other commands when integrated into another nodejs application.

ccarruitero:
Fixes #823

### DIFF
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -56,6 +56,7 @@ export type LintCmdParams = {|
 export type LintCmdOptions = {|
   createLinter?: LinterCreatorFn,
   createFileFilter?: FileFilterCreatorFn,
+  shouldExitProgram?: boolean,
 |};
 
 export default function lint(
@@ -66,6 +67,7 @@ export default function lint(
   {
     createLinter = defaultLinterCreator,
     createFileFilter = defaultFileFilterCreator,
+    shouldExitProgram = true,
   }: LintCmdOptions = {}
 ): Promise<void> {
   const fileFilter = createFileFilter({sourceDir, ignoreFiles, artifactsDir});
@@ -86,7 +88,7 @@ export default function lint(
       // which should be the directory to the extension.
       _: [sourceDir],
     },
-    runAsBinary: true,
+    runAsBinary: shouldExitProgram,
   });
   return linter.run();
 }
--- a/src/program.js
+++ b/src/program.js
@@ -147,7 +147,7 @@ export class Program {
         });
       }
 
-      await runCommand(argv);
+      await runCommand(argv, {shouldExitProgram});
 
     } catch (error) {
       const prefix = cmd ? `${cmd}: ` : '';
--- a/tests/unit/test-cmd/test.lint.js
+++ b/tests/unit/test-cmd/test.lint.js
@@ -25,13 +25,14 @@ describe('lint', () => {
       lintResult,
       createLinter,
       runLinter,
-      lint: (params = {}) => {
+      lint: (params = {}, options = {}) => {
         return defaultLintCommand({
           sourceDir: '/fake/source/dir',
           ...params,
         }, {
           createLinter,
           createFileFilter,
+          ...options,
         });
       },
     };
@@ -63,6 +64,14 @@ describe('lint', () => {
     return lint().then(() => {
       const args = createLinter.firstCall.args[0];
       assert.equal(args.runAsBinary, true);
+    });
+  });
+
+  it('sets runAsBinary according shouldExitProgram option', () => {
+    const {lint, createLinter} = setUp();
+    return lint({}, {shouldExitProgram: false}).then(() => {
+      const args = createLinter.firstCall.args[0];
+      assert.strictEqual(args.runAsBinary, false);
     });
   });
 
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -417,6 +417,16 @@ describe('program.main', () => {
         assert.equal(customPrefs.prop2, 'value2');
       });
   });
+
+  it('passes shouldExitProgram option to commands', () => {
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    return execProgram(['lint'], {commands: fakeCommands}).then(() => {
+      const options = fakeCommands.lint.firstCall.args[1];
+      assert.strictEqual(options.shouldExitProgram, false);
+    });
+  });
 });
 
 describe('program.defaultVersionGetter', () => {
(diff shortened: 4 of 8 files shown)


src/cmd/lint.js
```diff
@@ -51,6 +51,7 @@ export type LintCmdParams = {|
   warningsAsErrors?: boolean,
   ignoreFiles?: Array<string>,
   artifactsDir?: string,
+  runAsBinary?: boolean,
```
kumar303: This should move to a function option (which are [passed here](https://github.com/ccarruitero/web-ext/blob/a0a2bfcd3667c868ae729f9a39b83791d263d5f5/src/cmd/lint.js#L57)) rather than a command line parameter.

src/program.js
```diff
@@ -425,6 +428,11 @@ Example: $0 --help run.
         type: 'boolean',
         default: false,
       },
+      'runAsBinary': {
```
kumar303: I don't think it makes sense to define this as a command line option, do you? I can't think of any use for this.

src/program.js
```diff
@@ -146,6 +146,9 @@ export class Program {
           version: getVersion(absolutePackageDir),
         });
       }
+      if (cmd === 'lint') {
+        argv.runAsBinary = this.shouldExitProgram;
```
kumar303: By explicitly setting `runAsBinary` you are making `program.js` care and understand about how the `lint` function operates. It should neither care nor understand :)

Instead, I think it would be better to pass `shouldExitProgram` as an option to `lint` and, actually, to all commands. That would look like:

````js
await runCommand(argv, {shouldExitProgram});
````

With that new optional parameter the `lint` function could decide what to do.

As part of this change, all command handler functions (every `default` export within [`src/cmd`](https://github.com/mozilla/web-ext/tree/master/src/cmd)) will need to accept a boolean `shouldExitProgram` parameter. They can all just ignore it but the Flow object definition should at least define it as a possible parameter.

File: src/program.js (hunk 2 above)
Comment (ccarruitero) [inline src/program.js]: Yeah, it was part of some initial attempt that I did, before ending with the changes in the execute function. I just forgot to clean up this.

**Comment by ccarruitero** [comment]
@kumar303 I made the changes, but I have some doubts about how to test that the commands allow `shouldExitProgram` option. I'm been adding a [test](https://github.com/mozilla/web-ext/pull/937/commits/02e0ed53fadad7493f0b03392bbcfc49e1d02383#diff-7940d43edd1a9801c0f3fc8d5969eabeR323) calling the command with the option for each command, so the flow check will fire an error, but I think is not the best way to test this.
Aside some commands (`sign` and `docs`) not was firing he flow error before I added the option.
Do you have any thought?

**Comment by ccarruitero** [comment]
@kumar303 I just rebase the commits, let me know what you think.

File: tests/unit/test.program.js [hunk 4]
```diff
@@ -417,6 +417,17 @@ describe('program.main', () => {
         assert.equal(customPrefs.prop2, 'value2');
       });
   });
+
+  it('pass shouldExitProgram option to commands', () => {
```
kumar303: grammar nit: 'pass' should be 'passes'

tests/unit/test.program.js
```diff
@@ -417,6 +417,17 @@ describe('program.main', () => {
...
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    return execProgram(['lint'], {commands: fakeCommands}
+    ).then(() => {
+      const options = fakeCommands.lint.firstCall.args[1];
+      assert.equal(options.shouldExitProgram, false);
```
kumar303: Same thing here: this needs to be `assert.strictEqual` otherwise many things will match `false`.

Thanks for adding this test!

tests/unit/test-cmd/test.lint.js
```diff
@@ -66,6 +67,14 @@ describe('lint', () => {
     });
   });
 
+  it('allow deactivate runAsBinary param into linter', () => {
+    const {lint, createLinter} = setUp();
+    return lint({}, {shouldExitProgram: false}).then(() => {
+      const args = createLinter.firstCall.args[0];
+      assert.equal(args.runAsBinary, false);
```
kumar303: This needs to be `assert.strictEqual` otherwise many things will match `false`.

tests/unit/test-cmd/test.lint.js
```diff
@@ -66,6 +67,14 @@ describe('lint', () => {
     });
   });
 
+  it('allow deactivate runAsBinary param into linter', () => {
```
kumar303: I suggest re-wording this to:

it... passes shouldExitProgram option to the linter

tests/unit/test.program.js
```diff
@@ -417,6 +417,17 @@ describe('program.main', () => {
         assert.equal(customPrefs.prop2, 'value2');
       });
   });
+
+  it('pass shouldExitProgram option to commands', () => {
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    return execProgram(['lint'], {commands: fakeCommands}
+    ).then(() => {
```
kumar303: style nit: I think this would be easier to read if it moved up a line like this:

````js
return execProgram(['lint'], {commands: fakeCommands}).then(() => {
````
If that exceeds 80 chars, try to break it more like this:
````js
return execProgram(
  ['lint'], {commands: fakeCommands}).then(() => {
````

tests/unit/test-cmd/test.lint.js
ccarruitero: I'm not sure about this. I think maybe could be `it ... set runAsBinary according shouldExitProgram option` ?

kumar303: sure, that works fine

Issue #887, opened 2017-04-04
Drop support for Node 4


kumar303:
We can drop support for Node 4 soon since it's no longer the active [Long Term Support version of Node](https://github.com/nodejs/LTS). Currently Node 4 causes lots of intermittent appveyor failures (our Windows CI) and I've seen sporadic failures when trying to update various libs (like flow-bin).

PR #942, opened 2017-05-08
feat: Drop testing on node4

wagnerand:
See https://github.com/mozilla/addons-linter/pull/1272

Fixes https://github.com/mozilla/web-ext/issues/887

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 
 # Keep this in sync with appveyor.yml
 node_js:
-- '4'
 - '6'
 
 before_script:
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ cross-platform way. Initially, it will provide a streamlined experience for deve
 ## Installation from source
 
 You'll need:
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -3,7 +3,6 @@ sudo: false
 
 # Keep this in sync with appveyor.yml
 node_js:
-- '4'
 - '6'
```
kumar303: Can you add Node 8 to the list? That will help us prepare for the next [LTS release](https://github.com/nodejs/LTS). If adding that causes problems, we can do it in a separate issue but I don't foresee any issues.

kumar303:
Once you push a new commit (don't rebase or squash) then the build error will be resolved.

.travis.yml
wagnerand: But...there is no Node 8 yet?

kumar303: Hah, sorry, I thought there was. Nevermind then!

kumar303:
Don't sweat the changelog linter, it's being annoying right now. When I squash and merge I'll make sure the final commit is formatted correctly.

PR #945, opened 2017-05-09
fix: upgrade to webpack 2

### LINKED ISSUE #787: Upgrade to webpack 2
We use webpack 1 but webpack 2 is slowly becoming more stable. I don't think it's urgent that we upgrade to it but we will have to eventually.

ccarruitero:
Fixes #787


webpack.config.js
```diff
@@ -65,10 +66,10 @@ module.exports = {
...
   ],
   resolve: {
-    extensions: ['', '.js', '.json'],
-    modulesDirectories: [
-      'src',
-      'node_modules',
+    extensions: ['.js', '.json'],
+    modules: [
+      path.resolve(__dirname, 'node_modules'),
+      path.join(__dirname, 'src'),
```
kumar303: `src` should come before `node_modules` in case one of our local modules shadows a third party module. Hopefully it won't happen but if we give precedence to our local modules it would be less confusing.

Issue #921, opened 2017-04-24
Cannot suspend web-ext


andymckay:
### Is this a bug or feature request?

Bug

### What is the current behavior?

web-ext run

Then press ctrl-z and expect the process to be suspended so you can resume later. It does not.

### What is the expected or desired behavior?

ctrl-z suspends the information.

### Version information (for bug reports)

* **Firefox version**: Nightly
* **Your OS and version**: OS X 10.10.5
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````

v7.8.0
4.2.0
1.9.0

This used to work, might have been a result of the handling for "Press R to reload".

kumar303:
I tried a version of web-ext before and after https://github.com/mozilla/web-ext/issues/705 and I can confirm that after https://github.com/mozilla/web-ext/issues/705, suspending is not working (with Node 6.10.2).

If I'm not mistaken, bash sends [`SIGSTOP`](https://nodejs.org/api/process.html#process_signal_events) when you type ctrl-z. Perhaps the co-routine [running the while loop](https://github.com/mozilla/web-ext/blob/542f9ce8f556d14a43d237c13ac83c1961fff732/src/cmd/run.js#L151) needs to look for this somehow? @saintsebastian @rpl any ideas on how to fix this?

saintsebastian:
Tbh I didn't know about this behaviour and completely ignored it while writing that feature. Sorry about that. Don't see reasons why loop can't expect ctrl-z as and send SIGSTOP. Have to test it first though.

kumar303:
It's possible Node is getting the `SIGTSTP` instead of `SIGSTOP`, I'm not sure.

saintsebastian:
@kumar303 i think it was `SIGTSTP`. I can definitely send it to process, the more interesting question is how one can restart the loop on `fg`. I will think it other and try to come up with some solution and start a PR.

rpl:
@andymckay @kumar303 yeah, it looks like a side-effect of enabling the readline keypress events on the stdin stream, we handled it explicitly for the Ctrl-C keyboard shortcut (https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L156-L157), but we didn't for Ctrl-Z.

@saintsebastian don't worry about how to restore the process with `fg`: it is something that the shell application will take care of.

[...]

PR #948, opened 2017-05-15
fix: unable to suspend web-ext when running in interactive mode

saintsebastian:
Here is my version of #921 ~~which doesn't work with node v7.10.0~~ (it works now)

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -149,7 +149,7 @@ export function defaultReloadStrategy(
     const keypressUsageInfo = 'Press R to reload (and Ctrl-C to quit)';
 
     Promise.resolve().then(async function() {
-      log.info('Press R to reload (and Ctrl-C to quit)');
+      log.info(keypressUsageInfo);
 
       let userExit = false;
 
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -113,6 +113,7 @@ export type ReloadStrategyOptions = {|
   createWatcher?: WatcherCreatorFn,
   createFileFilter?: FileFilterCreatorFn,
   stdin: stream$Readable,
+  kill: typeof process.kill;
 |};
 
 export function defaultReloadStrategy(
@@ -124,6 +125,7 @@ export function defaultReloadStrategy(
     addonReload = defaultAddonReload,
     createWatcher = defaultWatcherCreator,
     stdin = process.stdin,
+    kill = process.kill,
   }: ReloadStrategyOptions = {}
 ): void {
   const watcher: Watchpack = createWatcher({
@@ -143,6 +145,9 @@ export function defaultReloadStrategy(
     // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
     // before the callback attached to the Promise returned by this function (and it allows the `run` function
     // to not be stuck in the while loop).
+
+    const keypressUsageInfo = 'Press R to reload (and Ctrl-C to quit)';
+
     Promise.resolve().then(async function() {
       log.info('Press R to reload (and Ctrl-C to quit)');
 
@@ -156,7 +161,24 @@ export function defaultReloadStrategy(
         if (keyPressed.ctrl && keyPressed.name === 'c') {
           userExit = true;
         } else if (keyPressed.ctrl && keyPressed.name === 'z') {
-          process.kill(process.pid, 'SIGTSTP');
+          // Prepare to suspend.
+
+          // NOTE: Switch the raw mode off before suspending (needed to make the keypress event to work correctly
+          // when the nodejs process is resumed).
+          if (stdin instanceof tty.ReadStream) {
+            stdin.setRawMode(false);
+          }
+
+          log.info('\nweb-ext has been suspended on user request');
+          kill(process.pid, 'SIGTSTP');
+
+          // Prepare to resume.
+
+          log.info(`\nweb-ext has been resumed. ${keypressUsageInfo}`);
+          // Switch the raw mode on on resume.
+          if (stdin instanceof tty.ReadStream) {
+            stdin.setRawMode(true);
+          }
         } else if (keyPressed.name === 'r' && addonId) {
           log.debug('Reloading extension on user request');
           await addonReload({addonId, client});
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -155,6 +155,8 @@ export function defaultReloadStrategy(
 
         if (keyPressed.ctrl && keyPressed.name === 'c') {
           userExit = true;
+        } else if (keyPressed.ctrl && keyPressed.name === 'z') {
+          process.kill(process.pid, 'SIGTSTP');
         } else if (keyPressed.name === 'r' && addonId) {
           log.debug('Reloading extension on user request');
           await addonReload({addonId, client});
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -495,6 +495,9 @@ describe('run', () => {
           assert.equal(setRawMode.firstCall.args[0], true);
           assert.equal(setRawMode.secondCall.args[0], false);
           assert.equal(setRawMode.lastCall.args[0], true);
+        })
+        .then(() => {
+          fakeStdin.emit('keypress', 'c', {name: 'c', ctrl: true});
         });
     });
 
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -469,6 +469,7 @@ describe('run', () => {
     it('shuts down firefox on user request (CTRL+C in shell console)', () => {
       const {firefoxProcess, reloadStrategy} = prepare();
       const fakeStdin = new tty.ReadStream();
+      sinon.spy(fakeStdin, 'setRawMode');
 
       return reloadStrategy({}, {stdin: fakeStdin})
         .then(() => {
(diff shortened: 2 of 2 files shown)


saintsebastian:
@rpl Hey! I am sorry it took me so long, but after `fg` key commands no longer work for me, they just appear on the terminal as if i am typing. I tried it with node v7.10.0 and v6.6.0. I have no idea what is happening, because i basically used master and added this change to it

rpl:
@saintsebastian you are right, while `SIGTSTP` and `SIGCONT` seems to work as expected (the process is sent to the background and then restored with `fg`), after web-ext is restored using `fg`, it seems to not be able to receive the keypress anymore (actually, it is interesting to notice that the keypress events are actually received if we press "enter" after the key, e.g. 'r' + 'enter')

This bug is probably hiding deeper in the nodejs implementation of the keypress event, we definitely need to investigate it deeper.

rpl:
@saintsebastian I took a deeper look in the nodejs sources and it seems that I've been finally able to figure it out:

By looking into the readline sources in the nodejs repo, we can notice how readline restore itself after being suspended here: https://github.com/nodejs/node/blob/master/lib/readline.js#L833-L850

In particular the most important line there are the following:
```js
// disable the raw mode before suspending
this._setRawMode(false);
process.kill(process.pid, 'SIGTSTP');
````

and

```js
// explicitly re-enable "raw mode" ....
self._setRawMode(true);
```

I applied the following changes (which enables and disables the raw mode and prints
some informative messages to the user while suspending and resuming) on top of this PR
and it worked correctly in my local environment (nodejs v.6.10.0 running on Linux):

```diff
diff --git a/src/cmd/run.js b/src/cmd/run.js
index 92955dc..b6d96cd 100644
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -143,8 +143,11 @@ export function defaultReloadStrategy(
     // NOTE: this `Promise.resolve().then(...)` is basically used to spawn a "co-routine" that is executed
     // before the callback attached to the Promise returned by this function (and it allows the `run` function
     // to not be stuck in the while loop).
+
+    const keypressUsageInfo = 'Press R to reload (and Ctrl-C to quit)';
+
     Promise.resolve().then(async function() {
-      log.info('Press R to reload (and Ctrl-C to quit)');
+      log.info(keypressUsageInfo);
 
       let userExit = false;
 
@@ -156,7 +159,19 @@ export function defaultReloadStrategy(
         if (keyPressed.ctrl && keyPressed.name === 'c') {
           userExit = true;
         } else if (keyPressed.ctrl && keyPressed.name === 'z') {
+          // Prepare to suspend.
+
+          // NOTE: Switch the raw mode off before suspending (needed to make the keypress event to work correctly
+          // when the nodejs process is resumed).
+          stdin.setRawMode(false);
+          log.info('\nweb-ext has been suspended on user request');
           process.kill(process.pid, 'SIGTSTP');
+
+          // Prepare to resume.
+
+          log.info(`\nweb-ext has been resumed. ${keypressUsageInfo}`);
+          // Switch the raw mode on on resume.
+          stdin.setRawMode(true);
         } else if (keyPressed.name === 'r' && addonId) {
           log.debug('Reloading extension on user request');
           await addonReload({addonId, client});
```

Can you give it a try on node7 and MacOS to be sure that it works as expected there as well?

rpl:
Once we have verified that the approach works correctly on both Linux and OSX (and across the recent nodejs versions), we should turn the `process.kill(process.pid, 'SIGTSTP')` call into an injected dependencies, so that we can add unit tests for these behaviors.

saintsebastian:
@rpl i've tested on both 6.6.0 and 7.10.0 and it does work apart from some flow errors (we probably just need to reuse readable stream type and add raw mode to it). The same logic we used for `process.stdin` test can be reused for `process.kil`. It s probably a priority since it interferes with a running version of web-ext

rpl:
> @rpl i've tested on both 6.6.0 and 7.10.0 and it does work apart from some flow errors (we probably just need to reuse readable stream type and add raw mode to it). 

That's great! thanks for testing it.

and, yeah, I didn't run the flow check yet on the changes from the above diff, it definitely need the tweak that you mentioned "to make flow check happy again" ;-)

Something like:
```js
          if (stdin instanceof tty.ReadStream) {
            stdin.setRawMode(...);
          }
```

> The same logic we used for process.stdin test can be reused for process.kill...

yep, exactly!

saintsebastian:
@rpl tried to address what you've mentioned. I think it is a good idea to test the action on 'fg' as well, but not quite sure how best to approach it.

tests/unit/test-cmd/test.run.js
```diff
@@ -477,6 +478,18 @@ describe('run', () => {
...
 
+    it('pauses the web-ext process (CTRL+Z in shell console)', () => {
+      const {reloadStrategy, kill} = prepare();
+      const fakeStdin = new tty.ReadStream();
+
+      return reloadStrategy({}, {stdin: fakeStdin})
+        .then(() => {
+          fakeStdin.emit('keypress', 'z', {name: 'z', ctrl: true});
+        }).then(() => {
+          assert.ok(kill.called);
```
rpl: @saintsebastian I've been thinking about how to test this behavior in the unit tests and I think that we can:

- assert that kill is called with the expected pid and the expected signal
- assert that `stdin.setRawMode(false)` has been called on `fakeStdin` **before** the call to the `kill` sinon spy
- assert that `stdin.setRawMode(true)` has been called on `fakeStdin` **after** the call to the `kill` sinon spy

The 3 assertions above should be enough to ensure that the real (non mocked) behavior would be the expected one.

How that sounds to you?

(side note: we should also make sure that we are going to exit the `while` loop related to the keypress processing, https://github.com/saintsebastian/web-ext/blob/c6d78220df21e4f37779643468a8d9bb3876561d/src/cmd/run.js#L156-L186, when we exit these tests)

saintsebastian:
@rpl regarding the side-note: I am not quite sure how to do that. Should we assert that stuff outside of loop is happening (`firefoxProcess.kill()` for example), or should we somehow assert the test time itself?

rpl:
> @rpl regarding the side-note: I am not quite sure how to do that. Should we assert that stuff outside of loop is happening (firefoxProcess.kill() for example), or should we somehow assert the test time itself?

I was thinking about sending the fake 'ctrl-C' keypress event before the test is completed, it should be enough to make the while loop to exit.

saintsebastian:
@rpl i see what you mean!

tests/unit/test-cmd/test.run.js
```diff
@@ -477,6 +479,28 @@ describe('run', () => {
...
+          assert.deepEqual(kill.firstCall.args, [process.pid, 'SIGTSTP']);
+          sinon.assert.callOrder(setRawMode, setRawMode, kill, setRawMode);
+          sinon.assert.calledThrice(setRawMode);
+          assert.equal(setRawMode.firstCall.args[0], true);
+          assert.equal(setRawMode.secondCall.args[0], false);
+          assert.equal(setRawMode.lastCall.args[0], true);
+        })
+        .then(() => {
+          fakeStdin.emit('keypress', 'c', {name: 'c', ctrl: true});
```
rpl: @saintsebastian Thanks!

I've been checking if there are other tests that can leak this file loop and it looks like there is one more, the fakeStdin related to the "can reload when user presses R in shell console" test (at line 454 of this file).

Another scenario that we should cover is when a test fails to reach the `then` resolve callback because of an unexpected rejection in the promise chains (we should probably pack the `fakeStdin.emit(...)` into a test helper and use the helper to handle both resolve and reject in the end of the Promise chain, something like:
```js
function getDoEmitCtrlC(stdin) {
  return () => stdin.emit('keypress', 'c', {name: 'c', ctrl: true});
}

.then(getDoEmitCtrlC(fakeStdin), getDoEmitCtrlC(fakeStdin));
```

maybe with a name that looks nicer than `getDoEmitCtrlC` :-)

saintsebastian: @rpl good call. i added those to all cases that test have the loop, in other cases for this function since `stdin` is not `TTY` they shouldn't enter the loop in the first place.

tests/unit/test-cmd/test.run.js
```diff
@@ -462,19 +467,43 @@ describe('run', () => {
...
 
       return reloadStrategy({}, {stdin: fakeStdin})
         .then(() => {
           fakeStdin.emit('keypress', 'c', {name: 'c', ctrl: true});
         }).then(() => {
           assert.ok(firefoxProcess.kill.called);
-        });
+        })
+        .then(loopExiter(fakeStdin), loopExiter(fakeStdin));
+
```
rpl: Nit: unnecessary empty line to remove

saintsebastian:
@rpl thanks for sticking with me on this one! I am sorry it took so long (i have been without wi-fi for some time).

rpl:
Thanks @saintsebastian !

Hey @kumar303, this should be ready, can you take a final look?

src/cmd/run.js
```diff
@@ -155,6 +160,25 @@ export function defaultReloadStrategy(
 
         if (keyPressed.ctrl && keyPressed.name === 'c') {
           userExit = true;
+        } else if (keyPressed.ctrl && keyPressed.name === 'z') {
+          // Prepare to suspend.
```
kumar303: don-draper-bliss

tests/unit/test-cmd/test.run.js
```diff
@@ -462,19 +467,42 @@ describe('run', () => {
           assert.ok(addonReload.called);
           assert.equal(addonReload.firstCall.args[0].addonId,
             tempInstallResult.addon.id);
-        });
+        })
+        .then(loopExiter(fakeStdin), loopExiter(fakeStdin));
```
kumar303: The error handler is going to ignore any thrown error which means these tests will always pass which would be bad. You need to make sure it throws the error with something like this:

```diff
diff --git a/tests/unit/test-cmd/test.run.js b/tests/unit/test-cmd/test.run.js
index ef51f95..b91bb2d 100644
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -389,6 +389,13 @@ describe('run', () => {
       return () => stdin.emit('keypress', 'c', {name: 'c', ctrl: true});
     }
 
+    function exitLoopOnError(stdin) {
+      return (error) => {
+        loopExiter(stdin);
+        throw error;
+      };
+    }
+
     function prepare() {
       const client = new RemoteFirefox(fakeFirefoxClient());
       const watcher = {
@@ -468,7 +475,7 @@ describe('run', () => {
           assert.equal(addonReload.firstCall.args[0].addonId,
             tempInstallResult.addon.id);
         })
-        .then(loopExiter(fakeStdin), loopExiter(fakeStdin));
+        .then(loopExiter(fakeStdin), exitLoopOnError(fakeStdin));
     });
 
     it('shuts down firefox on user request (CTRL+C in shell console)', () => {
```

kumar303:
Hi @saintsebastian - any chance you'll have some time to finish this out soon? We are getting ready to release a new web-ext and it would be nice to include this fix.

saintsebastian:
@kumar303 I am so sorry, this patch took an embarrassing amount of time! let me know if you spot anything else

PR #954, opened 2017-05-23
feat: Added check for existing package and a new --overwrite-dest CLI option in build command

### LINKED ISSUE #68: Build command should not overwrite existing extension
The build command currently clobbers an existing extension at the destination path. We should probably not do that and maybe provide an option to allow it.

harikishen:
Fixes https://github.com/mozilla/web-ext/issues/68

Done some changes to fix the issue

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -140,6 +140,7 @@ export async function defaultPackageCreator(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
 
+  //Added 'wx' flags to avoid overwriting of existing package
   let stream = createWriteStream(extensionPath, {flags: 'wx'});
 
   stream.write(buffer, () => stream.end());
@@ -184,7 +185,6 @@ export type BuildCmdOptions = {|
   fileFilter?: FileFilter,
   onSourceChange?: OnSourceChangeFn,
   packageCreator?: PackageCreatorFn,
-  overwriteDest?: boolean,
   showReadyMessage?: boolean,
   createFileFilter?: FileFilterCreatorFn,
   shouldExitProgram?: boolean,
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -113,7 +113,7 @@ export async function defaultPackageCreator(
   }: PackageCreatorParams,
   {
     eventToPromise = defaultEventToPromise,
-  }:PackageCreatorOptions = {}
+  }: PackageCreatorOptions = {}
 ): Promise<ExtensionBuildResult> {
   let id;
   if (manifestData) {
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -40,6 +40,7 @@ export type PackageCreatorParams = {|
   sourceDir: string,
   fileFilter: FileFilter,
   artifactsDir: string,
+  overwriteDest: boolean,
   showReadyMessage: boolean
 |};
 
@@ -98,7 +99,12 @@ export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
 async function defaultPackageCreator({
-  manifestData, sourceDir, fileFilter, artifactsDir, showReadyMessage,
+  manifestData,
+  sourceDir,
+  fileFilter,
+  artifactsDir,
+  overwriteDest,
+  showReadyMessage,
 }: PackageCreatorParams): Promise<ExtensionBuildResult> {
   let id;
   if (manifestData) {
@@ -124,11 +130,42 @@ async function defaultPackageCreator({
   const packageName = safeFileName(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
-  const stream = createWriteStream(extensionPath);
+  let destStat;
+  try {
+    destStat = await fs.stat(extensionPath);
+  } catch (error) {
+    destStat = false;
+    throw error;
+  }
+  const destinationFileExists = destStat && destStat.isFile();
+
+  function throwExtensionExists() {
+    throw new UsageError(
+      `Extension exists at the destination path: ${extensionPath}\n` +
+      'Use --overwrite-dest to enable overwriting.');
+  }
+
+  if (destinationFileExists) {
+    if (overwriteDest) {
+      log.info(`Destination exists, overwriting: ${extensionPath}`);
+    } else {
+      throwExtensionExists();
+    }
+  }
+
+  const stream = overwriteDest ?
+            createWriteStream(extensionPath)
+           : createWriteStream(extensionPath, {flags: 'wx'}); // double-check if destination file exists
 
   stream.write(buffer, () => stream.end());
 
-  await eventToPromise(stream, 'close');
+  try {
+    await eventToPromise(stream, 'close');
+  } catch (error) {
+    if (error.code === 'EEXIST') {
+      throwExtensionExists();
+    }
+  }
 
   if (showReadyMessage) {
     log.info(`Your web extension is ready: ${extensionPath}`);
@@ -151,6 +188,7 @@ export type BuildCmdOptions = {|
   fileFilter?: FileFilter,
   onSourceChange?: OnSourceChangeFn,
   packageCreator?: PackageCreatorFn,
+  overwriteDest?: boolean,
   showReadyMessage?: boolean,
   createFileFilter?: FileFilterCreatorFn,
   shouldExitProgram?: boolean,
@@ -168,6 +206,7 @@ export default async function build(
     }),
     onSourceChange = defaultSourceWatcher,
     packageCreator = defaultPackageCreator,
+    overwriteDest = false,
     showReadyMessage = true,
   }: BuildCmdOptions = {}
 ): Promise<ExtensionBuildResult> {
@@ -176,7 +215,12 @@ export default async function build(
   log.info(`Building web extension from ${sourceDir}`);
 
   const createPackage = () => packageCreator({
-    manifestData, sourceDir, fileFilter, artifactsDir, showReadyMessage,
+    manifestData,
+    sourceDir,
+    fileFilter,
+    artifactsDir,
+    overwriteDest,
+    showReadyMessage,
   });
 
   await prepareArtifactsDir(artifactsDir);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -40,7 +40,6 @@ export type PackageCreatorParams = {|
   sourceDir: string,
   fileFilter: FileFilter,
   artifactsDir: string,
-  eventToPromise: typeof defaultEventToPromise,
   overwriteDest: boolean,
   showReadyMessage: boolean
 |};
@@ -49,6 +48,10 @@ export type LocalizedNameParams = {|
   messageFile: string,
   manifestData: ExtensionManifest,
 |}
+
+export type PackageCreatorOptions = {|
+  eventToPromise: typeof defaultEventToPromise,
+|};
 
 // This defines the _locales/messages.json type. See:
 // https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales
@@ -99,15 +102,19 @@ export async function getDefaultLocalizedName(
 export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
-export async function defaultPackageCreator({
-  manifestData,
-  sourceDir,
-  fileFilter,
-  artifactsDir,
-  eventToPromise = defaultEventToPromise,
-  overwriteDest,
-  showReadyMessage,
-}: PackageCreatorParams): Promise<ExtensionBuildResult> {
+export async function defaultPackageCreator(
+  {
+    manifestData,
+    sourceDir,
+    fileFilter,
+    artifactsDir,
+    overwriteDest,
+    showReadyMessage,
+  }: PackageCreatorParams,
+  {
+    eventToPromise = defaultEventToPromise,
+  }:PackageCreatorOptions = {}
+): Promise<ExtensionBuildResult> {
   let id;
   if (manifestData) {
     id = getManifestId(manifestData);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -140,7 +140,7 @@ export async function defaultPackageCreator(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
 
-  //Added 'wx' flags to avoid overwriting of existing package
+  // Added 'wx' flags to avoid overwriting of existing package.
   let stream = createWriteStream(extensionPath, {flags: 'wx'});
 
   stream.write(buffer, () => stream.end());
@@ -148,19 +148,18 @@ export async function defaultPackageCreator(
   try {
     await eventToPromise(stream, 'close');
   } catch (error) {
-    if (isErrorWithCode('EEXIST', error)) {
-      if (!overwriteDest) {
-        throw new UsageError(
-          `Extension exists at the destination path: ${extensionPath}\n` +
-          'Use --overwrite-dest to enable overwriting.');
-      }
-      log.info(`Destination exists, overwriting: ${extensionPath}`);
-      stream = createWriteStream(extensionPath);
-      stream.write(buffer, () => stream.end());
-      await eventToPromise(stream, 'close');
-    } else {
+    if (!isErrorWithCode('EEXIST', error)) {
       throw error;
     }
+    if (!overwriteDest) {
+      throw new UsageError(
+        `Extension exists at the destination path: ${extensionPath}\n` +
+        'Use --overwrite-dest to enable overwriting.');
+    }
+    log.info(`Destination exists, overwriting: ${extensionPath}`);
+    stream = createWriteStream(extensionPath);
+    stream.write(buffer, () => stream.end());
+    await eventToPromise(stream, 'close');
   }
 
   if (showReadyMessage) {
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -130,16 +130,6 @@ async function defaultPackageCreator({
   const packageName = safeFileName(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
-  let destStat;
-  try {
-    destStat = await fs.stat(extensionPath);
-  } catch (error) {
-    destStat = false;
-    if (!isErrorWithCode('ENOENT', error)) {
-      throw error;
-    }
-  }
-  const destinationFileExists = destStat && destStat.isFile();
 
   function throwExtensionExists() {
     throw new UsageError(
@@ -147,11 +137,8 @@ async function defaultPackageCreator({
       'Use --overwrite-dest to enable overwriting.');
   }
 
-  if (!overwriteDest && destinationFileExists) {
-    throwExtensionExists();
-  }
 
-  if (destinationFileExists) {
+  if (overwriteDest) {
     log.info(`Destination exists, overwriting: ${extensionPath}`);
   }
 
@@ -164,7 +151,7 @@ async function defaultPackageCreator({
   try {
     await eventToPromise(stream, 'close');
   } catch (error) {
-    if (error.code === 'EEXIST') {
+    if (isErrorWithCode('EEXIST', error)) {
       throwExtensionExists();
     }
   }
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -11,7 +11,7 @@ import {zipDir} from '../util/zip-dir';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
-import {UsageError} from '../errors';
+import {UsageError, isErrorWithCode} from '../errors';
 import {
   createFileFilter as defaultFileFilterCreator,
   FileFilter,
@@ -135,7 +135,9 @@ async function defaultPackageCreator({
     destStat = await fs.stat(extensionPath);
   } catch (error) {
     destStat = false;
-    throw error;
+    if (!isErrorWithCode('ENOENT', error)) {
+      throw error;
+    }
   }
   const destinationFileExists = destStat && destStat.isFile();
 
@@ -145,12 +147,12 @@ async function defaultPackageCreator({
       'Use --overwrite-dest to enable overwriting.');
   }
 
+  if (!overwriteDest && destinationFileExists) {
+    throwExtensionExists();
+  }
+
   if (destinationFileExists) {
-    if (overwriteDest) {
-      log.info(`Destination exists, overwriting: ${extensionPath}`);
-    } else {
-      throwExtensionExists();
-    }
+    log.info(`Destination exists, overwriting: ${extensionPath}`);
   }
 
   const stream = overwriteDest ?
@@ -180,6 +182,7 @@ export type BuildCmdParams = {|
   sourceDir: string,
   artifactsDir: string,
   asNeeded?: boolean,
+  overwriteDest?: boolean,
   ignoreFiles?: Array<string>,
 |};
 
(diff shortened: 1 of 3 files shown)


src/cmd/build.js
```diff
@@ -124,11 +130,42 @@ async function defaultPackageCreator({
   const packageName = safeFileName(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
-  const stream = createWriteStream(extensionPath);
+  let destStat;
+  try {
+    destStat = await fs.stat(extensionPath);
+  } catch (error) {
+    destStat = false;
+    throw error;
```
rpl: @harikishen it looks like the existent tests are currently failing because fs.stat raises an error when the "extensionPath" does not exist (it raises an error object with code 'ENOENT', e.g. like documented in the following nodejs doc pages: https://nodejs.org/api/errors.html#errors_common_system_errors and https://nodejs.org/api/fs.html#fs_fs_stat_path_callback) and so we should throw the caught error only if the error code is not 'ENOENT' (which is an expected error when we are building a file to a path that doesn't exist yet).

Applying the following change should fix the existent tests:
```diff
diff --git a/src/cmd/build.js b/src/cmd/build.js
index 125eed7..6d123d2 100644
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -11,7 +11,7 @@ import {zipDir} from '../util/zip-dir';
 import getValidatedManifest, {getManifestId} from '../util/manifest';
 import {prepareArtifactsDir} from '../util/artifacts';
 import {createLogger} from '../util/logger';
-import {UsageError} from '../errors';
+import {UsageError, isErrorWithCode} from '../errors';
 import {
   createFileFilter as defaultFileFilterCreator,
   FileFilter,
@@ -135,7 +135,9 @@ async function defaultPackageCreator({
     destStat = await fs.stat(extensionPath);
   } catch (error) {
     destStat = false;
-    throw error;
+    if (!isErrorWithCode('ENOENT', error)) {
+      throw error;
+    }
   }
   const destinationFileExists = destStat && destStat.isFile();
```

src/cmd/build.js
```diff
@@ -124,11 +130,42 @@ async function defaultPackageCreator({
...
+  const destinationFileExists = destStat && destStat.isFile();
+
+  function throwExtensionExists() {
+    throw new UsageError(
+      `Extension exists at the destination path: ${extensionPath}\n` +
+      'Use --overwrite-dest to enable overwriting.');
+  }
+
+  if (destinationFileExists) {
+    if (overwriteDest) {
```
rpl: Instead of nesting two `if` statement into each other we can probably do something like:

```js
if (!overwriteDest && destinationFileExists) {
  throw new UsageError(...);
}

if (destinationFileExists) {
  log.info(`Destination exists, ...`);
}

...
```

harikishen: but won't both conditions be satisfied if this is true `if (!overwriteDest && destinationFileExists)`

harikishen:
@rpl @kumar303 I've added the `--overwrite-dest` option and the check for existing packages in [build.js](https://github.com/harikishen/web-ext/blob/ticket_68/src/cmd/build.js) and [program.js](https://github.com/harikishen/web-ext/blob/ticket_68/src/program.js). Also i have added the corresponding [unit tests](https://github.com/harikishen/web-ext/blob/ticket_68/tests/unit/test-cmd/test.build.js)

rpl:
@harikishen hehe no worries, the build on travis is failing because of the linting checks on the pull request title (which becomes the merged commit message if the PR is composed by more then one commit, and given that we generate the changelog using conventional-changelog, we run [conventinal-changelog-lint](https://github.com/marionebl/conventional-changelog-lint) on every pull request update to prevent us to merge it with a commit message that would make the changelog generation to fail once we release a new web-ext version).

I just fixed the PR title and restarted the travis job, which is now green.

Coveralls is still red, though, which means that there is still something that needs to be tested explicitly.

rpl:
I just looked into the Coveralls report and it looks that the following are the scenarios that are missing a test:

- `await fs.stat(extensionPath)` raises an error that is not an error with code `"ENOENT"` (https://coveralls.io/builds/11654589/source?filename=src%2Fcmd%2Fbuild.js#L138)
- `await eventToPromise(stream, 'close')` raises an error with code `"EEXIST"` (https://coveralls.io/builds/11654589/source?filename=src%2Fcmd%2Fbuild.js#L167)

While the first one can probably be reproduced by forcing build to use a directory where we don't have write permissions (e.g. we used this strategy in this test file: https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.artifacts.js#L60-L123), but in that case we would need to skip the test on windows, on the contrary the build will fail on appveyor (the CI service that we use to run the tests on windows)

The second one is harder, because it would require the test to recreate the race condition of something that create the file between the fs.stat and the actual file creation.

Actually, I'm not sure that we should check that the file exist after we did the `fs.stat` call, because if I get the feature correctly, it should prevent the developer to overwrite a file that is already there, and not protecting the user from the issue that could arise by running the build concurrently on more than one shell, am I right?

@kumar303 what to you think about it?

harikishen:
@rpl i think this line maybe the cause for the issue https://github.com/harikishen/web-ext/blob/ticket_68/src/cmd/build.js#L167 because i didn't follow the steps given by @kumar303 in the reference pull request

rpl:
@harikishen another reasonable way to approach it could be the following:

- do not check the file existence with `fs.stat`, but just use `createWriteStream(extensionPath, {flags: 'wx'})` the first time, so that it raises an exception if the file already exist (the error with code 'EEXIST')
- if the file already exist and the `overwriteDest` option has been specified, rerun `createWriteStream` without the 'wx' flags, or raise the exception if the `overwriteDest` option has not been enabled.

Something like the following:
```
  const stream = createWriteStream(extensionPath, {flags: 'wx'}); // Check if destination file exists first

  stream.write(buffer, () => stream.end());

  try {
    await eventToPromise(stream, 'close');
  } catch (error) {
    if (error.code === 'EEXIST') {
      if (overwriteDest) {
        log.info(`Destination exists, overwriting: ${extensionPath}`);

        // Retry, this time by overwriting the existent file.
        const overwriteStream = createWriteStream(extensionPath);
        stream.write(bugger, () => stream.end());
        await eventToPromise(stream, 'close');
      } else {
        throw new UsageError(
          `Extension exists at the destination path: ${extensionPath}\n` +
            'Use --overwrite-dest to enable overwriting.');
      }
    } else {
       // Re-throw any unexpected exception as it is.
       throw error;
    }
  }
```

This way we achieve the same safety of the current implementation, but with a simplified implementation.

How that sounds to you?

harikishen:
actually i thought of that the first time i saw the code. but @kumar303 mentioned on performing the `fs.stat` check somewhere in the previous PR

rpl:
@harikishen true @kumar303 mentioned `fs.stat` in the issue comments, but while I've been looking at the code in this PR again this morning and thinking about the scenarios, I realized that the first `fs.stat` call doesn't make the "file existence check" safer, it just make it unnecessarily more complex (and create more code branches to cover with the unit tests).

Given that using the 'wx' flags makes the "file existence check" pretty safe, we can use it as the only "file existence check", which will also simplify its testing.

kumar303:
> This way we achieve the same safety of the current implementation, but with a simplified implementation.

I agree with the approach illustrated above. Making less code paths is always better.

harikishen:
@kumar303  @rpl  I've made the said changes.

src/cmd/build.js
```diff
@@ -124,11 +130,31 @@ async function defaultPackageCreator({
...
 
   stream.write(buffer, () => stream.end());
 
-  await eventToPromise(stream, 'close');
+  try {
+    await eventToPromise(stream, 'close');
+  } catch (error) {
+    if (isErrorWithCode('EEXIST', error)) {
+      throwExtensionExists();
+    }
```
rpl: we should re-raise any unexpected error here (basically anything that is not an 'EEXIST' error.

src/cmd/build.js
```diff
@@ -4,14 +4,14 @@ import {createWriteStream} from 'fs';
 
 import {fs} from 'mz';
 import parseJSON from 'parse-json';
-import eventToPromise from 'event-to-promise';
+import {eventToPromise as defaultEventToPromise} from 'event-to-promise';
```
rpl:
```js
import defaultEventToPromise from 'event-to-promise';
```

src/cmd/build.js
```diff
@@ -40,6 +40,8 @@ export type PackageCreatorParams = {|
   sourceDir: string,
   fileFilter: FileFilter,
   artifactsDir: string,
+  eventToPromise: typeof defaultEventToPromise,
```
rpl: This is an injected dependency and by convention we are always separating the regular parameter (with the options that the function is supposed to get usually) from the injected parameters (which are usually redefined only by the unit tests).

And so, `eventToPromise` should be removed from the `PackageCreatorParams` and a new `PackageCreatorOptions` type should be created:

```js
export type PackageCreatorOptions = {|
  eventToPromise: typeof defaultEventToPromise,
|};
```

src/cmd/build.js
```diff
@@ -97,8 +99,14 @@ export async function getDefaultLocalizedName(
 export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
-async function defaultPackageCreator({
-  manifestData, sourceDir, fileFilter, artifactsDir, showReadyMessage,
+export async function defaultPackageCreator({
```
rpl: Like mentioned in the previous comment, eventToPromise should be in a second parameter, not in the first one.

```js
export async function defaultPackageCreator(
  {
    manifestData,
    sourceDir,
    fileFilter,
    artifactsDir,
    overwriteDest,
    showReadyMessage,
  }: PackageCreatorParams,
  {
    eventToPromise = defaultEventToPromise,
  }: PackageCreatorOptions = {}
): Promise<ExtensionBuildResult> {
  ... /* here is where the function body is */
}
```

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +368,27 @@ describe('build', () => {
...
+
+    it('should reject on Unexpected errors', () => {
+      const fakeEventToPromise = sinon.spy(
+        () => Promise.reject(new Error('Unexpected error')));
+      const params = {
+        manifestData: basicManifest,
+        sourceDir: fixturePath('minimal-web-ext'),
+        fileFilter: {wantFile: () => true},
+        artifactsDir: 'artifacts',
+        eventToPromise: fakeEventToPromise,
```
rpl: Once you have applied the changes to the "src/cmd/build.js" module described above, we can fix this test accordingly (by moving the fake `eventToPromise` in the second parameter, and by using a real FileFilter instance to prevent flow to complain at line 385):

```diff
diff --git a/tests/unit/test-cmd/test.build.js b/tests/unit/test-cmd/test.build.js
index 14ce59d..ce54181 100644
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -373,16 +373,23 @@ describe('build', () => {
     it('should reject on Unexpected errors', () => {
       const fakeEventToPromise = sinon.spy(
         () => Promise.reject(new Error('Unexpected error')));
+      const sourceDir = fixturePath('minimal-web-ext');
+      const artifactsDir = 'artifacts';
+      const fileFilter = new FileFilter({sourceDir, artifactsDir});
       const params = {
         manifestData: basicManifest,
-        sourceDir: fixturePath('minimal-web-ext'),
-        fileFilter: {wantFile: () => true},
-        artifactsDir: 'artifacts',
-        eventToPromise: fakeEventToPromise,
+        sourceDir,
+        fileFilter,
+        artifactsDir,
         overwriteDest: false,
         showReadyMessage: false,
-      };// $FLOW_FIXME: flow complains about the fileFiler type
-      return defaultPackageCreator(params)
+      };
+
+      const options = {
+        eventToPromise: fakeEventToPromise,
+      };
+
+      return defaultPackageCreator(params, options)
         .then(() => assert.fail('The expected error has not been raised'))
         .catch ((error) => {
           assert.match(error.message, /Unexpected error/);
```

src/cmd/build.js
rpl: @harikishen you forgot to apply this change in the last update on this patch.

`eventToPromise` is not a named export of the 'event-to-promise' npm package, it is its default export and it should be imported with `defaultEventToPromise` as its name using the syntax:
```js
import defaultEventToPromise from 'event-to-promise';
```

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +321,44 @@ describe('build', () => {
...
+
+      });
+  });
+
+  it('overwrites zip file if it exists', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const testFileName =
+          path.join(tmpDir.path(), 'minimal_extension-1.0.zip');
+        return fs.writeFile(testFileName, 'test')
```
rpl: @harikishen We need to remove this file to fix this issue when this test run on Windows:
https://ci.appveyor.com/project/kumar303/web-ext/build/1.0.518#L1693

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +321,44 @@ describe('build', () => {
...
+          .then(() =>
+        build({
```
rpl: @harikishen can you indent this `build({...})` a bit to make this lines nicer to read?

tests/unit/test-cmd/test.build.js
```diff
@@ -376,8 +373,12 @@ describe('build', () => {
 
     it('should reject on Unexpected errors', () => withTempDir(
       (tmpDir) => {
-        const fakeEventToPromise = sinon.spy(
-          () => Promise.reject(new Error('Unexpected error')));
+        const fakeEventToPromise = sinon.spy(async (stream) => {
+          await defaultEventToPromise(stream, 'close');
+          await fs.unlink(
+            path.join(tmpDir.path(), 'minimal_extension-1.0.zip'));
```
rpl: This test fails because the created zip file is not named 'minimal_extension-1.0.zip'.

Let's delete the file by discovering its name inside this async function, so that if the file name changes
the test is still able to remove the tmpDir path correctly:

```diff
diff --git a/tests/unit/test-cmd/test.build.js b/tests/unit/test-cmd/test.build.js
index ca6e0e8..b483057 100644
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -375,8 +375,10 @@ describe('build', () => {
       (tmpDir) => {
         const fakeEventToPromise = sinon.spy(async (stream) => {
           await defaultEventToPromise(stream, 'close');
-          await fs.unlink(
-            path.join(tmpDir.path(), 'minimal_extension-1.0.zip'));
+          const files = await fs.readdir(tmpDir.path());
+          for (const file of files) {
+            await fs.unlink(path.join(tmpDir.path(), file));
+          }
           return Promise.reject(new Error('Unexpected error'));
         });
         const sourceDir = fixturePath('minimal-web-ext');
```

src/cmd/build.js
```diff
@@ -97,9 +102,19 @@ export async function getDefaultLocalizedName(
...
+    manifestData,
+    sourceDir,
+    fileFilter,
+    artifactsDir,
+    overwriteDest,
+    showReadyMessage,
+  }: PackageCreatorParams,
+  {
+    eventToPromise = defaultEventToPromise,
+  }:PackageCreatorOptions = {}
```
rpl: This line has a lint warning (a space is missing between `:` and `PackageCreatorOptions`):

```
$ npm run lint
> grunt lint

Running "lint" task

Running "eslint:js" (eslint) task

/.../web-ext/src/cmd/build.js
  114:3  warning  There must be a space after "{
    eventToPromise = defaultEventToPromise,
  }:PackageCreatorOptions" parameter type annotation colon  flowtype/space-after-type-colon

✖ 1 problem (0 errors, 1 warning)
```

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,44 @@ describe('build', () => {
...
   ));
 
+  it('raises a UsageError if zip file exists', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const testFileName =
+          path.join(tmpDir.path(), 'minimal_extension-1.0.zip');
+        return fs.writeFile(testFileName, 'test')
+          .then(() =>
+            build({
```
rpl: can you move `build({` to the previous line? it makes much more clear that the arrow function is going to return the result of the `build` call.

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,44 @@ describe('build', () => {
...
+              sourceDir: fixturePath('minimal-web-ext'),
+              artifactsDir: tmpDir.path(),
+            }))
+          .catch ((error) => {
+            assert.instanceOf(error, UsageError);
+          });
+
```
rpl: Nit: remove this empty line

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,44 @@ describe('build', () => {
...
+      });
+  });
+
+  it('overwrites zip file if it exists', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const testFileName =
```
rpl: same coding style issue mentioned in the previous comment.

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,44 @@ describe('build', () => {
...
+          path.join(tmpDir.path(), 'minimal_extension-1.0.zip');
+        return fs.writeFile(testFileName, 'test')
+          .then(() =>
+            build({
```
rpl: same coding style issue mentioned in one of the previous comments.

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,44 @@ describe('build', () => {
...
+              sourceDir: fixturePath('minimal-web-ext'),
+              artifactsDir: tmpDir.path(),
+              overwriteDest: true,
+            }))
+          .then((buildResult) => {
+            assert.match(buildResult.extensionPath,
+                         /minimal_extension-1\.0\.zip$/);
+            return buildResult.extensionPath;
```
rpl: This return is probably not really needed.

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +369,41 @@ describe('build', () => {
 
   });
 
+  describe('defaultPackageCreator', () => {
+
+    it('should reject on Unexpected errors', () => withTempDir(
```
rpl: we should keep the coding style of this test consistent with the other tests in the same file:
```js
it('should reject on Unexpected errors', () => {
  return withTempDir(
     (tmpDir) => {
       ...
     });
});
```

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +369,41 @@ describe('build', () => {
...
+      (tmpDir) => {
+        const fakeEventToPromise = sinon.spy(async (stream) => {
+          await defaultEventToPromise(stream, 'close');
+          const files = await fs.readdir(tmpDir.path());
```
rpl: we should add an inline comment here to explain why we are removing the content of the tmpDir here.

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +369,41 @@ describe('build', () => {
...
+          eventToPromise: fakeEventToPromise,
+        };
+
+        return defaultPackageCreator(params, options)
+          .then(() => assert.fail('The expected error has not been raised'))
+          .catch ((error) => {
+            assert.match(error.message, /Unexpected error/);
+          });
+      }
+  ));
```
rpl: This looks like it should be indented with 2 more spaces.

harikishen: the linter tells the indentation is fine, tried adding two but caused error

harikishen:
@rpl  i've made the said changes

tests/unit/test-cmd/test.build.js
rpl: @harikishen this one is not yet fixed.

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +369,42 @@ describe('build', () => {
 
   });
 
+  describe('defaultPackageCreator', () => {
+    it('should reject on Unexpected errors', () => {
+      return withTempDir(
+      (tmpDir) => {
```
rpl: This should be indented as the other tests in this file.

harikishen: hope it's fine now

src/cmd/build.js
```diff
@@ -124,11 +139,29 @@ async function defaultPackageCreator({
   const packageName = safeFileName(
     `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
-  const stream = createWriteStream(extensionPath);
+
+  //Added 'wx' flags to avoid overwriting of existing package
```
rpl: Nit, The comments should have a space between `//` and the comment, and ends with a '.'

src/cmd/build.js
```diff
@@ -124,11 +139,29 @@ async function defaultPackageCreator({
...
+  let stream = createWriteStream(extensionPath, {flags: 'wx'});
 
   stream.write(buffer, () => stream.end());
 
-  await eventToPromise(stream, 'close');
+  try {
+    await eventToPromise(stream, 'close');
+  } catch (error) {
+    if (isErrorWithCode('EEXIST', error)) {
```
rpl: @harikishen By re-looking at this block I think that we can reduce the nesting of the `if` blocks by reversing the check and make it nicer:

```js
} catch (error) {
  if (!isErrorWithCode('EEXIST', error)) {
    throw error;
  }

  if (!overwriteDest) {
    ...
  }

  log.info(`Destination exists, overwriting: ${extensionPath}`);
  ...
} 
```

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,43 @@ describe('build', () => {
     }
   ));
 
+  it('raises an UsageError if zip file exists', () => {
+    return withTempDir(
+      (tmpDir) => {
+        const testFileName =
```
rpl: This needs to be reformatted as the similar one that is at line 346

tests/unit/test-cmd/test.build.js
harikishen: @rpl i've changed it

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +368,42 @@ describe('build', () => {
 
   });
 
+  describe('defaultPackageCreator', () => {
+    it('should reject on Unexpected errors', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const fakeEventToPromise = sinon.spy(async (stream) => {
+            await defaultEventToPromise(stream, 'close');
+            //Remove contents of tmpDir before removal of directory
```
rpl: Same coding style issue of one of the previous comments (Add a space between `//` and the comment and ends it with a `.`).

harikishen: @rpl done

harikishen:
@rpl done i guess

src/program.js
```diff
@@ -287,6 +287,10 @@ Example: $0 --help run.
           describe: 'Watch for file changes and re-build as needed',
           type: 'boolean',
         },
+        'overwrite-dest': {
```
kumar303: Can you add `alias: 'o'` to this? This will be a commonly used option so I think it needs an alias.

tests/unit/test-cmd/test.build.js
```diff
@@ -329,4 +367,42 @@ describe('build', () => {
...
+            artifactsDir,
+            overwriteDest: false,
+            showReadyMessage: false,
+          };
+          const options = {
+            eventToPromise: fakeEventToPromise,
+          };
+
+          return defaultPackageCreator(params, options)
+            .then(() => assert.fail('The expected error has not been raised'))
```
kumar303: You can import [this helper](https://github.com/mozilla/web-ext/blob/master/tests/unit/helpers.js#L116) and use it like `.then(makeSureItFails()).catch(...)`

tests/unit/test-cmd/test.build.js
```diff
@@ -320,6 +322,42 @@ describe('build', () => {
...
+        const testFileName = path.join(tmpDir.path(),
+                                       'minimal_extension-1.0.zip');
+        return fs.writeFile(testFileName, 'test')
+          .then(() => build(
+            {
+              sourceDir: fixturePath('minimal-web-ext'),
+              artifactsDir: tmpDir.path(),
+            }))
+          .catch ((error) => {
+            assert.instanceOf(error, UsageError);
```
kumar303: In addition to asserting the error type, this should also match something specific about the exception message because there are a lot of possible `UsageError`s that can be raised. It doesn't have to match the full message. For example:
````js
assert.match(error.message, /Extension exists at the destination path/);
````

harikishen: @kumar303 I've made the changes you requested

Issue #809, opened 2017-02-15
Use web-ext to run an extension in Google Chrome


kumar303:
### Is this a bug or feature request?

feature

### What is the current behavior?

You can use [web-ext run](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run) to launch an extension for development in Firefox. It creates a temporary profile and will automatically reload the extension when you change the source.

### What is the expected or desired behavior?

It would be great to use `web-ext` to launch the extension like this in Chrome. I'm not sure if we can support auto-reloading or temporary profiles but at the very least we could launch [chromium --load-extension=/path/to/source-dir](http://stackoverflow.com/questions/20782557/load-unpacked-chrome-extension-programatically) which does a similar temporary installation for development.

kumar303:
If Chrome doesn't have remote debugger support for auto-reloading, here is one idea for an epic, horrendous hack:
* create a helper Chrome Extension that can listen to commands from `web-ext`
* install the helper extension along with the extension under test
* whenever the source changes, `web-ext` would tell the Chrome Extension to automatically click the 'reload' button on the `chrome://extensions/` page (Extensions may not have access to run content scripts on this page though)

AgrimPrasad:
Hacky idea for auto-reloading which was previously implemented for Chrome: https://60devs.com/hot-reloading-for-chrome-extensions.html

kumar303:
Wow, that's cool. Thanks for the link. I guess the downside is that one would have to modify their extension source to use this approach.

josephfrazier:
For [OctoLinker](https://github.com/OctoLinker/browser-extension), we use [chrome-launch](https://github.com/hughsk/chrome-launch) to start an isolated instance of Chrome with the extension loaded. Here's the code, if you're interested in adapting it: https://github.com/OctoLinker/browser-extension/blob/08bd7aebe3fc43c96690c2ca0dab14398810c424/scripts/chrome-launch.js

lutostag:
Chrome (well a symlink to chromium on my rig) does reload the extension when I run with the following...
`chrome --temp-profile --password-store=basic --load-extension=<dir> about:blank`
Have used it [for active debugging here](https://github.com/lutostag/webextension-formsave/blob/master/formsave/manifest.json#L50).

version: `Chromium 58.0.3029.110 Built on Ubuntu , running on Ubuntu 17.04`

DamienCassou:
[...]

PR #962, opened 2017-05-31
chore: Refactoring out extension runners from the 'cmd/run' module

rpl:
The goal of this PR is refactoring the 'cmd/run' module so that we can more easily include:

- support for multiple extension runner targets, as Firefox for Android (Pull Request #868) and Chrome (Feature Request #809)

- support for multiple extensions' source dirs, e.g. so that we can make it easier to develop WebExtensions Experiment API addons (Pull Request #830), but instead of an additional `--experiments` CLI option, it would be achieved by using more than one `-s extensionSourceDir` CLI options

To make it possible to more easily plug additional "extension runner" definitions (as well as allow to run multiple extension runner at once), this PR introduces the following changes:

- Move the 'connect with retries' helper into the `"src/firefox/remote"` module, it is basically a wrap of the `connect` helper provide by that module with the retry logic, it should belong to that module.

- Move the ExtensionRunner into its own set of modules, and prepare it to:
  - provide the shared flow types and IExtensionRunner interface (which is the interface that
    all the ExtensionRunner implementation have to provide) 
  - support multiple extensions sourceDirs (by taking an array of extensions as one of its parameters)
  - support multiple targets (by abstracting into a MultiTargetsExtensionRunner the logic to create,
    start and interact multiple IExtensionRunner instances)
  - make `"src/cmd/run"` module more "target agnostic", e.g. by removing from it any direct reference to the addonId, because it is a "target" specific concept (e.g. it would not work the same for the Firefox and Chrome targets), the IExtensionRunner provides two methods to reload an extension: all the extensions (e.g. when the user press R in the console), one extension selected by its sourceDir (and the specific IExtensionRunner implementation will keep track internally of the addonId assigned to a defined extension sourceDir if and when needed, e.g. the FirefoxDesktopExtensionRunner keep track of the addonId assigned to a sourceDir to be able to reload the extension)

The "watch & reload" and "listen for user input & reload" logic has not been included in the IExtensionRunner, because we can run them once for all the targets in `"src/cmd/run"` and then use the created MultiTargetsExtensionRunner instance to reload the extensions (or exit the IExtensionRunner) everytime the source changes or the user press a key in the console.

Side Notes: Given that this PR was already changing a lot of lines on a couple of the existent test file, I took the opportunity to rewrite those tests into async functions (and I've also fixed the flow type name convention in a couple of files).


rpl:
Hi @kumar303, I'm sorry this patch is not a small one, follows some additional note that could help a bit during the review:

- there are at least 4 files that Github doesn't load the diff for by default (because they are large diffs):
  - "src/cmd/run.js" (most of the changes are removals)
  - "tests/unit/test-cmd/test.run.js" (most of the changes are removals)
  - "tests/unit/test-firefox/test.remote.js" (most of the changes are related to the rewrite into async functions)
  - "tests/unit/test-util/test.extension-runners.firefox-desktop.js" (that is the longest of the two new unit test files)

- the module "src/util/extension-runners/base.js" only contains flow types (it is a separate js files only to not create a 'fake' cyclic dependencies by importing the common type from the "index.js" module of the "src/util/extension-runners" dir

- I created two new unit tests and so the unit test coverage is consistent with the previous coverage level, and the new implementation should present mostly the same behavior of the current one, but given the size/impact of the change I'm still looking for any further behavior regression that has not been yet caught by the existent tests (especially given that most of the tests are unit tests which inject fake versions of the dependencies)

src/util/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,248 @@
...
+      for (const url of urls) {
+        binaryArgs.push('--url', url);
+      }
+    }
+
+    this.runningInfo = await firefoxApp.run(this.profile, {
+      firefoxBinary, binaryArgs,
+    });
+
+    this.runningInfo.firefox.on('close', () => {
```
kumar303: Since this is a little different than before, it would be nice to have some debug logging saying that Firefox closed.

rpl:
Hi @kumar303 
Thank you so much for the previous round of review comments, they have been very helpful to "refactor the refactoring", I think that the result is definitely nicer 🎉 

I have been able to remove also a bunch of code from this patch in the process (which is good, there are less test cases to review in the updated version ;-))

Let me know how the rest looks to you!

In the meantime I'll keep looking for any further "change of behavior" that I could have introduced with this heavy refactoring of the "src/cmd/run" module.

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+export class FirefoxDesktopExtensionRunner {
+  params: FirefoxDesktopExtensionRunnerParams;
+  reloadableExtensions: Map<string, string>;
+  cleanupCallbacks: Set<Function>;
+  profile: FirefoxProfile;
+  runningInfo: FirefoxInfo;
+  remoteFirefox: RemoteFirefox;
+
+  constructor(
+    params: FirefoxDesktopExtensionRunnerParams,
```
kumar303: this can fit on one line now:
```js
  constructor(params: FirefoxDesktopExtensionRunnerParams) {
```

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+const log = createLogger(__filename);
+
+/**
+ * This module provide an ExtensionRunner subclass that manage an extension executed
+ * in a Firefox for Desktop instance.
+ */
+
+export class FirefoxDesktopExtensionRunner {
+  params: FirefoxDesktopExtensionRunnerParams;
+  reloadableExtensions: Map<string, string>;
```
kumar303: Hmm, is there a way to make it more obvious that this maps a source directory to an addon ID? A comment maybe?

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+
+    if (reloadErrors.size > 0) {
+      return Promise.reject(new MultipleExtensionsReloadError(reloadErrors));
+    }
+  }
+
+  async reloadExtensionBySourceDir(extensionSourceDir: string): Promise<void> {
+    const addonId = this.reloadableExtensions.get(extensionSourceDir);
+    if (!addonId) {
+      throw new WebExtError('Extension not reloadable');
```
kumar303: I think it would be better to make this message more explicit, like:

```js
`No addonId has been mapped to extension at source dir "${extensionSourceDir}"`
```

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+        binaryArgs.push('--url', url);
+      }
+    }
+
+    this.runningInfo = await firefoxApp.run(this.profile, {
+      firefoxBinary, binaryArgs,
+    });
+
+    this.runningInfo.firefox.on('close', () => {
+      for (const fn of this.cleanupCallbacks) {
```
kumar303: Minor suggestion for readability: `for (const cleanUp of this.cleanupCallbacks)`

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+        try {
+          fn();
+        } catch (error) {
+          log.debug(`Exception on executing cleanup callback: ${error}`);
```
kumar303: Should this be `log.error()` instead? Would it be typical for a cleanup function to throw an error that one can ignore?

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+            throw new WebExtError(
+              'Unexpected missing addonId in the installAsTemporaryAddon result'
+            );
+          }
+
+          this.reloadableExtensions.set(extension.sourceDir, addonId);
+        } catch (error) {
+          if (error instanceof RemoteTempInstallNotSupported) {
+            log.debug(`Caught: ${error}`);
+            throw new WebExtError(
```
kumar303: Hmm, I wonder why this wasn't a `UsageError` before. Probably nothing to worry about in this patch.

src/extension-runners/index.js
```diff
@@ -0,0 +1,67 @@
...
+} from './base';
+
+export type MultipleTargetsExtensionRunnerParams = {
+  runners: Array<IExtensionRunner>,
+};
+
+// Export everything exported by the firefox-desktop runner.
+export * from './firefox-desktop';
+
+export class MultipleTargetsExtensionRunner {
```
kumar303: You could call this `MultiExtensionRunner` to make it less wordy

src/extension-runners/index.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    await Promise.all(promises);
+  }
+
+  registerCleanup(fn: Function): void {
+    const promises = [];
+    for (const runner of this.extensionRunners) {
+      promises.push(new Promise((resolve) => {
+        runner.registerCleanup(resolve);
```
kumar303: This is really confusing to me. Can't you just do `runner.registerCleanup(fn)` ? I'm also confused by `Promise.all(promises).then(fn, fn)` so maybe I'm missing it? Can you add a comment explaining what this is doing? :P

tests/unit/test-cmd/test.run.js
```diff
@@ -1,35 +1,43 @@
...
 // Fake result for client.installTemporaryAddon().then(installResult => ...)
 const tempInstallResult = {
   addon: {id: 'some-addon@test-suite'},
 };
-// Fake missing addon id result for client.installTemporaryAddon
-const tempInstallResultMissingAddonId = {
-  addon: {id: null},
-};
 
+function createFakeExtensionRunner(params, deps, overriddenMethods = {}) {
```
kumar303: It looks like `params` and `deps` are merged together in the real implementation now.

tests/unit/helpers.js
```diff
@@ -267,3 +275,39 @@ export class ErrorWithCode extends Error {
     this.code = code || 'SOME_CODE';
   }
 }
+
+/*
+ * A class that implements an empty IExtensionRunner interface.
+ */
+export class FakeExtensionRunner {
+  params: any;
+  deps: any;
```
kumar303: This no longer has deps, right?

tests/unit/helpers.js
```diff
@@ -267,3 +275,39 @@ export class ErrorWithCode extends Error {
     this.code = code || 'SOME_CODE';
   }
 }
+
+/*
+ * A class that implements an empty IExtensionRunner interface.
+ */
+export class FakeExtensionRunner {
+  params: any;
```
kumar303: Couldn't this also have a real constructor like:

```js
export class FakeExtensionRunner {
  constructor(params) {
    this.params = params;
  }
}
```
That way you could call it more naturally from `createFakeExtensionRunner()`:

```js
const runner = new FakeExtensionRunner(params);
```

tests/unit/test-cmd/test.run.js
```diff
@@ -358,43 +226,31 @@ describe('run', () => {
...
-    }
-
-    function prepare() {
-      const client = new RemoteFirefox(fakeFirefoxClient());
+    function prepare({stubExtensionRunner} = {}) {
       const watcher = {
         close: sinon.spy(() => {}),
       };
+      const extensionRunner = createFakeExtensionRunner({}, {},
+                                                        stubExtensionRunner);
```
kumar303: I find code like this hard to read. I know it tends to be more verbose but I'd rather see named params passed like:

```js
createFakeExtensionRunner({
  params: {}, overiddenMethods: stubRunnerExtensionRunner,
});
```
In most cases here I don't think it would actually add much verbosity.

tests/unit/test-extension-runners/test.extension-runners.firefox-desktop.js
```diff
@@ -0,0 +1,407 @@
...
+  });
+  const firefoxProcess = new StubChildProcess();
+
+  return {
+    remoteFirefox,
+    firefoxProcess,
+    params: {
+      extensions: [{
+        sourceDir: '/fake/sourceDir',
+        manifestData: {
```
kumar303: I think it would be helpful to extend this from [`basicManifest`](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.manifest.js#L15), at least just for documentation purposes.

tests/unit/test-extension-runners/test.extension-runners.firefox-desktop.js
```diff
@@ -0,0 +1,407 @@
...
+      preInstall: false,
+      noReload: false,
+      firefoxApp: getFakeFirefox({
+        run: sinon.spy(() => {
+          return Promise.resolve({
+            debuggerPort,
+            firefox: firefoxProcess,
+          });
+        }),
+        ...fakeFirefoxApp,
```
kumar303: is this in the right order? I'm wondering if the `run` definition needs to come after it

src/extension-runners/base.js
```diff
@@ -0,0 +1,23 @@
...
+  keepProfileChanges: boolean,
+  startUrl: ?string | ?Array<string>,
+};
+
+export interface IExtensionRunner {
+  run(): Promise<void>,
+  reloadAllExtensions(): Promise<void>,
+  reloadExtensionBySourceDir(extensionSourceDir: string): Promise<void>,
+  registerCleanup(fn: Function): void,
+  exit(): Promise<void>
```
rpl: yeah, definitely, and I always add the trailing comma, but I forgot to add it here, it is a pity that eslint doesn't check this on the flow types definitions.

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+  RemoteFirefox,
+} from '../firefox/remote';
+import type {
+  ExtensionRunnerParams,
+} from './base';
+import type {
+  FirefoxPreferences,
+} from '../firefox/preferences';
+import type {
+  FirefoxInfo, // eslint-disable-line import/named
```
rpl: we already have the same issue elsewhere (in one of the test files if I recall correctly).

I also don't get "the why" yet, while working on the refactoring I decided to defer further investigation, but I'm going to take a look again (as soon as I've applied the rest of the changes needed based on this round of review comments)

src/extension-runners/firefox-desktop.js
```diff
@@ -0,0 +1,229 @@
...
+} from '../firefox/index';
+
+export type FirefoxDesktopExtensionRunnerParams = ExtensionRunnerParams & {
```
rpl: No, unfortunately the exact object types (the pipes) and the intersection types (the '&') do not play well together :-(

src/extension-runners/index.js
rpl: This is basically creating a promise for every one of the created runners, each promise is going to be resolved when the related runner calls its registered cleanup functions.

The goal is be able to wait that all the runners have called their registered cleanup callbacks, so that we can finally call the cleanup callbacks registered on the "multi targets extension runner" runner.

I'll add a comment in the next update.

tests/unit/helpers.js
rpl: yes, these test helpers have a number of leftovers from the previous version.

src/extension-runners/firefox-desktop.js
rpl: actually, it looks that we can do something that works similar to this and which also supports the exact types:

- https://github.com/facebook/flow/issues/2626#issuecomment-290107574

I'm going to try this approach as part of the next round of updates on this PR.

src/extension-runners/firefox-desktop.js
kumar303: Huh, I think we have it working on `addons-frontend`. Well, no biggie, I was just confused.

src/extension-runners/index.js
kumar303: Oh! I see

src/extension-runners/firefox-desktop.js
rpl: @kumar303 I just tried and it works (the tests will require a bunch of additional $FLOW_IGNORE on some of the sinon spy, but, on the bright side, flow is able to catch a lot more potential issue in the code).

We have to update flow-bin to use the feature from the above link, which currently raises a warning during npm install because of the peer dependencies currently defined in grunt-flowbin:

https://github.com/kumar303/grunt-flowbin/blob/master/package.json#L39

we should update the peer dependency to `"flow-bin": ">=0.47.0"`, while we are still working on this review, so that when this PR is ready we can update the grunt-flowbin version in the web-ext devDependencies and remove the warning.

kumar303: oh, that's annoying about grunt-flowbin. I proposed this: https://github.com/kumar303/grunt-flowbin/pull/77

tests/unit/test-extension-runners/test.extension-runners.firefox-desktop.js
rpl: The reason why I added it after the `run` definition is to be able to override the run method in the tests (if we need to).

rpl:
@kumar303 This should be now ready to be reviewed again.

tests/unit/helpers.js
```diff
@@ -267,3 +275,62 @@ export class ErrorWithCode extends Error {
     this.code = code || 'SOME_CODE';
   }
 }
+
+/*
+ * A basic manifest fixtus using in unit tests.
```
kumar303: typo: fixture

tests/unit/test-extension-runners/test.extension-runners.firefox-desktop.js
```diff
@@ -0,0 +1,428 @@
...
+      params: {
+        profilePath: '/path/to/profile',
+        keepProfileChanges: true,
+      },
+    });
+
+    const runnerInstance = new FirefoxDesktopExtensionRunner(params);
+    await runnerInstance.run();
+
+    // $FLOW_IGNORE: ignored 'property not found' on sinon spy.
```
kumar303: oof. I think moving to [sinon.assert](https://github.com/mozilla/web-ext/issues/972) methods will make Flow happy.

src/extension-runners/index.js
```diff
@@ -0,0 +1,299 @@
...
+const log = createLogger(__filename);
+
+
+export type MultiExtensionRunnerParams = {|
+  runners: Array<IExtensionRunner>,
+  desktopNotifications: typeof defaultDesktopNotifications,
+|};
+
+// Export everything exported by the firefox-desktop runner.
+export * from './firefox-desktop';
```
rpl: It is not really necessary, but I like the pattern:

if the "index.js" module of a directory is the one that exports everything to the "the external importers" (e.g. mostly like "src/program" imports all the commands from "src/cmd" instead of loading every single submodule in the "src/cmd" dir), at least when (as in this case) the various modules in that directory are very related to each other (which is true for "src/cmd" and "src/extension-runners", but it is not for "src/utils").

This way, if we decide to re-organize the modules defined into that directory, the "external importers" could remain unchanged most of the times (at least if we don't rename the exported names).

kumar303: oh, sorry, I totally read this backwards. I thought you were exporting all of `src/extension-runners` from `src/extension-runners/firefox-desktop` which didn't seem useful to me.

Ship it! I agree that exporting everything from `src/extension-runners` to conceal the sub-module implementation is a useful pattern.

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -0,0 +1,584 @@
...
+      try {
+        await reloadStrategy({}, {stdin: fakeStdin});
+        fakeStdin.emit('keypress', 'r', {name: 'r', ctrl: false});
+
+        // Wait for one tick.
+        await Promise.resolve();
+
+        sinon.assert.called(fakeStdin.setRawMode);
+        sinon.assert.called(extensionRunner.reloadAllExtensions);
+      } finally {
```
rpl: @kumar303 now that these tests are implemented as async functions, I opted for `try...finally` as the strategy to ensure that we always exit the "keypress processing loop" (instead of the previous strategy based on the `loopExiter` and `exitLoopOnError` test helpers).

rpl:
@kumar303 thanks again! your review comments have been super useful as usually.

PR #965, opened 2017-06-06
add programatical use documentation

### LINKED ISSUE #381: Document the non-CLI version (JS API for web-ext)
Is there a way to do a `require("web-ext")` and build + sign web extensions using Node.js instead of using this via Terminal?

iaptsiauri:
Fixes #381
@kumar303 please take a look

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,33 @@ need to relink it.
     git pull
     npm run build
 
+## Programmatic use
+You can use the web-ext programmatically as well 
+````javascript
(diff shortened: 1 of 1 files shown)


kumar303:
Hi. I'm really sorry but somehow I missed this pull request. Since you've submitted it, we have an [easier way](https://github.com/mozilla/web-ext/pull/1028) to make programmatic calls so now your docs aren't quite right.

PR #968, opened 2017-06-08
Fix #932 - Modified useProfile() to check for profile paths

### LINKED ISSUE #932: Can't run with specified profile name and --keep-profile-changes
Bug.

```
D:\Dev\in-my-pocket> web-ext run -s build -p test --keep-profile-changes --verbose

[src\program.js][debug] Getting version from the git revision
[src\program.js][info] Version: master-36cc19fe3a71d912296fdc458033091633fcf0e7
[src\cmd\run.js][info] Running web extension from D:\Dev\in-my-pocket\build
[src\util\manifest.js][debug] Validating manifest at D:\Dev\in-my-pocket\build\manifest.json
[src\cmd\run.js][debug] Using Firefox profile from test
[src\program.js][error] 
run: Error: ENOENT: no such file or directory, open 'D:\Dev\in-my-pocket\test\user.js'
    at Object.fs.openSync (fs.js:584:18)
    at Object.fs.writeFileSync (fs.js:1316:33)
    at FirefoxProfile._writeUserPrefs (D:\Dev\web-ext\node_modules\firefox-profile\lib\firefox_profile.js:517:6)
    at FirefoxProfile.updatePreferences (D:\Dev\web-ext\node_modules\firefox-profile\lib\firefox_profile.js:347:8)
    at configureProfile (D:\Dev\web-ext\dist\webpack:\src\firefox\index.js:227:11)
    at Object._callee3$ (D:\Dev\web-ext\dist\webpack:\src\firefox\index.js:250:16)
    at tryCatch (D:\Dev\web-ext\node_modules\regenerator-runtime\runtime.js:65:40)
    at Generator.invoke [as _invoke] (D:\Dev\web-ext\node_modules\regenerator-runtime\runtime.js:303:22)
    at Generator.prototype.(anonymous function) [as next] (D:\Dev\web-ext\node_modules\regenerator-runtime\runtime.js:117:21)
    at step (D:\Dev\web-ext\dist\webpack:\~\babel-runtime\helpers\asyncToGenerator.js:17:1)

[src\program.js][error] run: Error code: ENOENT
```
However, both
```
web-ext run -s build -p test
web-ext run -s build -p path/to/test/profile --keep-profile-changes
```
work fine.

I think it should convert the profile name to file path.

* **Firefox version**: 54.a2
* **Your OS and version**: Windows 7 x64

````
````
```
D:\Dev\in-my-pocket>node --version && npm --version && web-ext --version
v7.9.0
4.6.1
master-36cc19fe3a71d912296fdc458033091633fcf0e7
```

### PR DESCRIPTION
**Pull request by harikishen**
@rpl @kumar303 Is this enough ?

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -251,7 +251,54 @@ export async function useProfile(
     customPrefs = {},
   }: UseProfileParams = {},
 ): Promise<FirefoxProfile> {
-  const profile = new FirefoxProfile({destinationDirectory: profilePath});
+  let profile;
+  try {
+    const dirExists = await isDirectory(profilePath);
+    var finder = new FirefoxProfile.Finder();
+
+    if (dirExists) {
+      log.debug(`Copying profile directory from "${profilePath}"`);
+      finder.getPath("default", function(err, profileDirectory) {
+        if (err) {
+          throw err;
+        }
+        if (profileDirectory === profilePath) {
+          throw new UsageError(
+            `Cannot use named profile "${profilePath}"`
+          );
+        } else {
+          finder.getPath("default-dev-edition", function(err, profileDirectory) {
+            if (err) {
+              throw err;
+            }
+            if (profileDirectory === profilePath) {
+              throw new UsageError(
+                `Cannot use named profile "${profilePath}"`
+              );
+            } else {
+              profile = new FirefoxProfile({destinationDirectory: profilePath});
+            }
+          });
+        }
+      });
+    } else {
+      log.debug(`Assuming ${profilePath} is a named profile`);
+      if (profilePath === 'default' || profilePath === 'default-dev-edition') {
+        throw new UsageError(
+          `Cannot use named profile "${profilePath}"`
+        );
+      }
+      finder.getPath(profilePath, function(err, profileDirectory) {
+        if (err) {
+          throw err;
+        }
+        profile = new FirefoxProfile({destinationDirectory: profileDirectory});
+      });
+    }
+  } catch (error) {
+    throw new WebExtError(
+      `Could not copy Firefox profile from ${profilePath}: ${error}`);
+  }
   return await configureThisProfile(profile, {app, customPrefs});
 }
 
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -258,7 +258,8 @@ export async function useProfile(
 
     if (dirExists) {
       log.debug(`Copying profile directory from "${profilePath}"`);
-      finder.getPath("default", function(err, profileDirectory) {
+      finder.getPath("default")
+      .then((err, profileDirectory) => {
         if (err) {
           throw err;
         }
@@ -266,8 +267,10 @@ export async function useProfile(
           throw new UsageError(
             `Cannot use named profile "${profilePath}"`
           );
-        } else {
-          finder.getPath("default-dev-edition", function(err, profileDirectory) {
+        }
+      })
+      .then(finder.getPath("default-dev-edition"))
+      .then((err, profileDirectory) => {
             if (err) {
               throw err;
             }
@@ -278,8 +281,9 @@ export async function useProfile(
             } else {
               profile = new FirefoxProfile({destinationDirectory: profilePath});
             }
-          });
-        }
+          })
+      .catch((error) => {
+
       });
     } else {
       log.debug(`Assuming ${profilePath} is a named profile`);
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -258,33 +258,15 @@ export async function useProfile(
 
     if (dirExists) {
       log.debug(`Copying profile directory from "${profilePath}"`);
-      finder.getPath("default")
-      .then((err, profileDirectory) => {
-        if (err) {
-          throw err;
-        }
-        if (profileDirectory === profilePath) {
-          throw new UsageError(
-            `Cannot use named profile "${profilePath}"`
-          );
-        }
-      })
-      .then(finder.getPath("default-dev-edition"))
-      .then((err, profileDirectory) => {
-            if (err) {
-              throw err;
-            }
-            if (profileDirectory === profilePath) {
-              throw new UsageError(
-                `Cannot use named profile "${profilePath}"`
-              );
-            } else {
-              profile = new FirefoxProfile({destinationDirectory: profilePath});
-            }
-          })
-      .catch((error) => {
-
-      });
+      const defaultProfilePath = finder.getPath('default');
+      const defaultDevProfilePath = finder.getPath('default-dev-edition');
+      if (profilePath === defaultProfilePath ||
+        profilePath === defaultDevProfilePath) {
+        throw new UsageError(
+          `Cannot use named profile "${profilePath}"`
+        );
+      }
+      profile = new FirefoxProfile({destinationDirectory: profilePath});
     } else {
       log.debug(`Assuming ${profilePath} is a named profile`);
       if (profilePath === 'default' || profilePath === 'default-dev-edition') {

### CODE REVIEW

**Comment by harikishen** [comment]
the failure seems to be because of the title of the PR.

**Comment by rpl** [comment]
Hi @harikishen 
First of all, my apologies for the time I needed to be able to look into this PR.

Currently the tests are only failing on the PR title linting, nevertheless, most of the code from this PR is currently not covered by the tests (in other words: "it is currectly not tested"), as you can see from the coveralls failures and related reports, e.g.:

- https://coveralls.io/builds/11889712/source?filename=src%2Ffirefox%2Findex.js#L259

By looking in a bit more details into the diff, the first thing that I notice is that you are currently trying to use finder.getPath('default') as a function that synchronously return a result, while it is actually a function that call a callback asynchronously and pass the result as a parameter to it.

To make the code more readable I have probably already suggested you, in our last chat on IRC about this change, to use es6-promisify (which is already in the web-ext dependencies and we are currently using it to wrap the zip-dir library methods with promise-based wrapper functions), but it looks that we are not yet wrapping the finder object using es6-promisify (and we should also use `await` on these calls to make the `async` function able to wait the returned promises to be resolved).

To learn more about how to use es6-promisify, you can look at the project README here:
- https://github.com/digitaldesignlabs/es6-promisify

To summarize the changes that are still needed on this PR to put it on the right direction (at least from my current point of view):

- using es6-promisfy to wrap the `finder` object methods  
- using `await` to wait a promise to be resolved or rejected
- defining additional tests to cover the new scenarios (and the new lines of code) introduced by this PR

File: src/firefox/index.js [hunk 1]
```diff
@@ -251,7 +251,37 @@ export async function useProfile(
     customPrefs = {},
   }: UseProfileParams = {},
 ): Promise<FirefoxProfile> {
-  const profile = new FirefoxProfile({destinationDirectory: profilePath});
+  let profile;
+  const finder = new FirefoxProfile.Finder();
```
rpl: we will need to be able to override the path where the profiles will be searched by the profile Finder helper class, to be able to more easily test these behaviors in the unit tests, and so this function should take an additional `searchProfilePath` (or another similar identifier name) parameter which will default to undefined (so that the default behavior of the Finder class will be preserved when this method is executed without passing the search profile path explicitly).

src/firefox/index.js
```diff
@@ -251,7 +251,37 @@ export async function useProfile(
...
+  const finderGetPath = promisify(finder.getPath.bind(finder));
```
rpl: `finderGetPath` will reject with an Error when the Finder instance is not able to find the requested profile name, which is something that we expect to happen (because the user could have no profile with that name yet).

To be able to only call `findGetPath` when we know that it will resolve to the path instead of being rejected with an Error, we can use the finder.readProfiles method and check if the named profile that we need actually exists before trying to resolve them into paths, e.g. something like the following snippet should do the trick:

```js
  ...
  const finder = new FirefoxProfile.Finder(searchProfilesPath);
  const finderGetPath = promisify(finder.getPath, finder);
  const finderReadProfiles = promisify(finder.readProfiles, finder);

  // Read the profiles list from the profiles.ini file from `searchProfilesPath`.
  await finderReadProfiles();

  // Helper function which returns true if the profile name exists in the profiles.ini file.
  const hasProfileName = (profileName) => {
    return finder.profiles.filter(
      (profileDef) => profileDef.Name === profileName
    ).length !== 0;
  };

  ...
  const defaultProfilePath = (
    hasProfileName('default') && await finderGetPath('default')
  );
  const defaultDevProfilePath = (
    hasProfileName('default') && await finderGetPath('dev-edition-default')
  );
```

This way `defaultProfilePath` and `defaultDevProfilePath` will be the profile path (which is a string), or `undefined` if the profile doesn't exist in the `profiles.ini` list.

src/firefox/index.js
```diff
@@ -251,7 +251,37 @@ export async function useProfile(
...
+  try {
+    const dirExists = await isDirectory(profilePath);
+
+    if (dirExists) {
+      log.debug(`Copying profile directory from "${profilePath}"`);
+      const defaultProfilePath = await finderGetPath('default');
+      const defaultDevProfilePath = await finderGetPath('default-dev-edition');
```
rpl: The default name of the Firefox DevEdition profile is `'dev-edition-default'` (instead of `'default-dev-edition'`)

harikishen:
@rpl could you look into it now ?

harikishen:
i've written 3 tests

harikishen:
@rpl coveralls fails because of hasProfileName(). How do I write a test for it ?

src/firefox/index.js
```diff
@@ -249,9 +250,51 @@ export async function useProfile(
     app,
     configureThisProfile = configureProfile,
     customPrefs = {},
+    searchProfilesPath = '',
   }: UseProfileParams = {},
 ): Promise<FirefoxProfile> {
-  const profile = new FirefoxProfile({destinationDirectory: profilePath});
+  let profile;
+  const finder = new FirefoxProfile.Finder(searchProfilesPath);
```
rpl: Given how we are using and testing the `Finder` class, I'm wondering if it would be better to create an `createProfileFinder` helper which creates the profile finder and wraps the methods that we need with promisify, something like

```js
function defaultCreateProfileFinder() {
  const finder = new FirefoxProfile.Finder();
  const readProfiles = promisify(finder.readProfiles, finder);
  return {
    getPath: promisify(finder.getPath, finder),
    hasProfileName: async () => {
       await finder.readProfiles();
       return finder.profiles.filter(
          (profileDef) => profileDef.Name === profileName
       ).length !== 0;
    }
  }
}
```

So that we can simplify `useProfile`:

```js
export async function useProfile(
   profilePath: string,
   {
      ...
      createProfileFinder: defaultCreateProfileFinder,
   }: UseProfileParams = P{,
): Promise<FirefoxProfile> {
   const finder = createProfileFinder();

   ...
}
```

and more easily test its behavior, e.g. because we can override `createProfileFinder` with a function that returns a "fake" version of our profile finder wrapper (and so we dont' even need to create the real "profiles.ini" file or the fake profiles directory in tmpDir).

src/firefox/index.js
```diff
@@ -249,9 +250,51 @@ export async function useProfile(
...
+    const dirExists = await isDirectory(profilePath);
+    const defaultProfilePath = (
+      hasProfileName('default') && await finderGetPath('default')
+    );
+    const defaultDevProfilePath = (
+      hasProfileName('dev-edition-default') &&
+        await finderGetPath('dev-edition-default')
+    );
+    if (dirExists) {
+      log.debug(`Copying profile directory from "${profilePath}"`);
```
rpl: This message does not seem right here, `useProfile` is never going to copy a profile.

src/firefox/index.js
```diff
@@ -249,9 +250,51 @@ export async function useProfile(
...
+      if (profilePath === defaultProfilePath ||
+        profilePath === defaultDevProfilePath) {
+        throw new UsageError(
+          `Cannot use profile at "${profilePath}"`
+        );
+      }
+      profile = new FirefoxProfile({destinationDirectory: profilePath});
+    } else {
+      log.debug(`Assuming ${profilePath} is a named profile`);
```
rpl: I'm wondering if we should make this to always fail if the requested profile name is `default` or `dev-edition-default`.

src/firefox/index.js
```diff
@@ -249,9 +250,51 @@ export async function useProfile(
...
+            profileDirectory === defaultDevProfilePath) {
+        throw new UsageError(
+          `Cannot use named profile "${profilePath}"`
+        );
+      }
+      profile = new FirefoxProfile({destinationDirectory: profileDirectory});
+    }
+  } catch (error) {
+    throw new WebExtError(
+      `Could not copy Firefox profile from ${profilePath}: ${error}`);
```
rpl: As the message logged at line 275, this error message is also not the right one.

src/firefox/index.js
harikishen: @rpl what do you propose ?

rpl: something like:
```
if (profilePath === 'default' || profilePath === 'dev-edition-default') {
  throw new WebExtError(`Cannot use the blacklisted named profile "${profilePath}" `);
}
```

tests/unit/test-firefox/test.firefox.js
```diff
@@ -324,6 +324,149 @@ describe('firefox', () => {
...
+        const app = 'fennec';
+        const profilePath = tmpDir.path();
+        const profileFinder = () => {
+          return {
+            getPath: sinon.spy((pathToProfile) =>
+                                    Promise.resolve(pathToProfile)),
+            hasProfileName: () => Promise.resolve(true),
+          };
+        };
+        const spy = sinon.spy(profileFinder, "getPath");
```
rpl: @harikishen this is probably a leftover, but it is making the CI to fail on the eslint checks and it never reaches the unit tests:

- https://travis-ci.org/mozilla/web-ext/builds/244601162#L1351-L1353

tests/unit/test-firefox/test.firefox.js
```diff
@@ -324,6 +324,149 @@ describe('firefox', () => {
       }
     ));
 
+    it('configures a named profile', () => withTempDir(
+      (tmpDir) => {
+        const configureProfile =
+          sinon.spy((profile) => Promise.resolve(profile));
+        const app = 'fennec';
+        const profilePath = tmpDir.path();
+        const profileFinder = () => {
```
rpl: @harikishen `profileFinder` is a function that creates the fake `Finder` objects, and not a `Finder` object itself, that is the reason why the assertion commented on line 351 is not working for you.

You should do something like:
```js
const profileFinder = {
  getPath: sinon.spy(...),
  ...
};

const createProfileFinder = () => profileFinder;

const profile = await firefox.useProfile(profilePath, {
   ...
   configureThisProfile: configureProfile,
   createProfileFinder: createProfileFinder,
});

assert.equal(configureProfile.called, true);
...
assert.equal(profileFinder.getPath.callCount, 2)
```

harikishen:
@rpl what do you think of the changes in index.js ?

src/firefox/index.js
```diff
@@ -239,8 +244,34 @@ export type UseProfileParams = {
...
 
+export function defaultCreateProfileFinder() {
+  const finder = new FirefoxProfile.Finder();
+  const readProfiles = promisify(finder.readProfiles, finder);
+  return {
+    getPath: promisify(finder.getPath, finder),
+    hasProfileName: async (profileName: string) => {
+      try {
+        await fs.stat(path.join(FirefoxProfile.Finder.locateUserDirectory(),
+                                                              'profile.ini'));
```
rpl: "profiles.ini"

src/firefox/index.js
```diff
@@ -239,8 +244,34 @@ export type UseProfileParams = {
...
+      } catch (error) {
+        if (isErrorWithCode('ENOENT', error)) {
+          log.info('No firefox profiles exist');
```
rpl: Here, we have already checked that "profiles.ini" doesn't exist (which should probably be a warning instead of an info message), but we are going to execute `await readProfiles()` at line 266, which will raises an exception because of that.

And so I think that it would be reasonable to `return false` after logging a warning (no profile name can be found if no `"profiles.ini"` has been found).

src/firefox/index.js
harikishen: Actually I was trying the case if the file did not exist... forgot to fix it

src/firefox/index.js
harikishen: What is the call for warning...???

rpl: unsurprisingly `log.warn` :-) (https://github.com/mozilla/web-ext/blob/97404a9d3b803376dcc732a28e26aae63d0b3aa9/tests/unit/helpers.js#L174)

harikishen:
@rpl i've done the requested changes. What else remains to be done ?

tests/unit/test-firefox/test.firefox.js
```diff
@@ -311,48 +322,84 @@ describe('firefox', () => {
 
     it('configures a profile', () => withBaseProfile(
       (baseProfile) => {
-        const configureThisProfile =
+        const configureProfile =
           sinon.spy((profile) => Promise.resolve(profile));
+        const profileFinder = () => {
```
rpl: it would be nicer if we can keep the same naming convention in all the tests, e.g. this profileFinder identifier is the function that creates the fake Finder instance in some of the tests and the fake Finder instance in other tests.

Also, if you name these identifier like the params property names, you can use the ES6 short syntax, eg.

```js
const createProfileFinder = () => ...;
const configureThisProfile = () => ...;

...

return firefox.useProfile(profilePath, {app, configureThisProfile, createProfileFinder});
```

tests/unit/test-firefox/test.firefox.js
```diff
@@ -311,48 +322,84 @@ describe('firefox', () => {
...
+          return {
+            getPath: (profilePath) => Promise.resolve(profilePath),
+            hasProfileName: () => Promise.resolve(true),
+          };
+        };
         const app = 'fennec';
         const profilePath = baseProfile.path();
-        return firefox.useProfile(profilePath, {app, configureThisProfile})
+        return firefox.useProfile(profilePath,
```
rpl: most of these tests would be easier to read if we use the `await` syntax here
```js
const profile = await firefox.useProfile(...);

... // test assertions
```

src/firefox/index.js
```diff
@@ -239,8 +244,32 @@ export type UseProfileParams = {
   app?: PreferencesAppName,
   configureThisProfile?: ConfigureProfileFn,
   customPrefs?: FirefoxPreferences,
+  createProfileFinder?: typeof defaultCreateProfileFinder,
 };
 
+export function defaultCreateProfileFinder() {
```
rpl: All the new tests are using a fake `createProfileFinder`, and so (like the code coverage reports confirm) it is time to create some tests specifically for defaultCreateProfileFinder).

To make this function more easily testable we should add an optional `userDirectoryPath` (a string) parameter to it, so that we can customize the directory where the `Finder` class will search the `profiles.ini` file and the profile directories.

src/firefox/index.js
```diff
@@ -249,9 +278,45 @@ export async function useProfile(
...
+      defaultProfilePath = await finder.getPath('default');
+    }
+    if (await finder.hasProfileName('dev-edition-default')) {
+      defaultDevProfilePath = await finder.getPath('dev-edition-default');
+    }
+    if (dirExists) {
+      log.debug(`Using profile directory from "${profilePath}"`);
+      if (profilePath === defaultProfilePath ||
+        profilePath === defaultDevProfilePath) {
+        throw new WebExtError(
```
rpl: By looking at the coverage reports, this line doesn't seem to be currently executed by any of the current tests.

src/firefox/index.js
harikishen: what tests do you propose ?

harikishen:
@rpl The issue seems to be with some sort of race condition with `fs.stat()` call and the `await readProfiles()` call. Correct me if I'm wrong.

src/firefox/index.js
```diff
@@ -239,8 +244,32 @@ export type UseProfileParams = {
...
 };
 
+export function defaultCreateProfileFinder(userDirectoryPath: string = '') {
+  const finder = new FirefoxProfile.Finder(userDirectoryPath);
+  const readProfiles = promisify(finder.readProfiles, finder);
+  return {
+    getPath: promisify(finder.getPath, finder),
+    hasProfileName: async (profileName: string) => {
+      try {
+        await fs.stat(path.join(FirefoxProfile.Finder.locateUserDirectory(),
```
rpl: @harikishen the tests are failing because on an issue in the `hasProfileName` implementation:

- while we are configuring the Finder to use the specified parameter as the profiles search path (as expected)
  at line 251, here we are always doing an fs.stat on the default search path the `FirefoxProfile.Finder.locateUserDirectory()` provides (but it should use the same path that the `Finder` is going to use

And so it should be something like:
```js
        const profilesIniPath = path.join(
          userDirectoryPath || FirefoxProfile.Finder.locateUserDirectory(),
          'profiles.ini'
        );
        const res = await fs.stat(profilesIniPath);
```

src/firefox/index.js
```diff
@@ -239,8 +244,32 @@ export type UseProfileParams = {
   app?: PreferencesAppName,
   configureThisProfile?: ConfigureProfileFn,
   customPrefs?: FirefoxPreferences,
+  createProfileFinder?: typeof defaultCreateProfileFinder,
 };
 
+export function defaultCreateProfileFinder(userDirectoryPath: string = '') {
```
rpl: this should be `string?` (a string that can be null) and it should not default to an empty string
(it should be undefined when not specified, and so we can just remove the explicit assigned default)

src/firefox/index.js
```diff
@@ -99,6 +99,11 @@ export interface FirefoxProcess extends events$EventEmitter {
   kill: Function;
 }
 
+export interface IProfileFinder {
```
rpl: This flow type is currently unused and it should be used for the `defaultCreateProfileFinder` return value type annotation (we should also move it near that function, that is where it is going to be used).

Also, the type signatures are wrong: both `hasProfileName` and `getPath` take a string as parameter
(and so they should be `hasProfileName(string): Promise<boolean>`, and `getPath` needs the same kind of fix).

src/firefox/index.js
rpl: This `defaultCreateProfileFinder` function should have a type annotation of the return value:
```js
export function defaultCreateProfileFinder(userDirectoryPath: ?string): IProfileFinder {
  ...
}
```

Once the return value is "type annotated" correctly, flow will notice some new errors that needs to be fixed, in particular:

- `hasProfileName` needs the type annotation of the return value
```js
  hasProfileName: async (profileName: string): Promise<boolean> => {
    ...
  },
```

- we should only catch the exception of `fs.stat` in the `try...catch` at line 256.
- we should re-throw any exception that as been caught by the above `try...catch` which do not verify the `isErrorWithCode('ENOENT', error)` check
- in the test file `"test.firefox.js"` the fake `getPath` defined at lines 417 and 444 should resolve to a string value:
```diff
diff --git a/tests/unit/test-firefox/test.firefox.js b/tests/unit/test-firefox/test.firefox.js
index da6a612..caa54b8 100644
--- a/tests/unit/test-firefox/test.firefox.js
+++ b/tests/unit/test-firefox/test.firefox.js
@@ -414,7 +414,7 @@ describe('firefox', () => {
           sinon.spy((profile) => Promise.resolve(profile));
         const createProfileFinder = () => {
           return {
-            getPath: () => Promise.resolve(),
+            getPath: () => Promise.resolve('/fake/profilePath'),
             hasProfileName: () => Promise.resolve(true),
           };
         };
@@ -441,7 +441,7 @@ describe('firefox', () => {
           sinon.spy((profile) => Promise.resolve(profile));
         const createProfileFinder = () => {
           return {
-            getPath: () => Promise.resolve(),
+            getPath: () => Promise.resolve('/fake/profilePath'),
             hasProfileName: () => Promise.resolve(true),
           };
         };
```

harikishen:
@kumar303 @rpl Could you guys check this out when your free ?

src/firefox/index.js
```diff
@@ -239,8 +244,35 @@ export type UseProfileParams = {
...
+        const profilesIniPath = path.join(
+          userDirectoryPath || FirefoxProfile.Finder.locateUserDirectory(),
+          'profiles.ini');
+
+        await fs.stat(profilesIniPath);
+        await readProfiles();
+        return finder.profiles.filter(
+          (profileDef) => profileDef.Name === profileName).length !== 0;
+      } catch (error) {
+        if (isErrorWithCode('ENOENT', error)) {
```
rpl: This try catch should only surround the `await fs.stat` call and it should re-throw any error that does not verify `isErrorWithCode('ENOENT', error)` (or, on the contrary, this function is going to return `undefined` on all the other kind of exception that can be caught from this `try...catch`).

src/firefox/index.js
```diff
@@ -249,9 +281,45 @@ export async function useProfile(
...
 ): Promise<FirefoxProfile> {
-  const profile = new FirefoxProfile({destinationDirectory: profilePath});
+  let profile;
+  let defaultProfilePath = '';
+  let defaultDevProfilePath = '';
+  const finder = createProfileFinder();
+  try {
+    const dirExists = await isDirectory(profilePath);
+    if (await finder.hasProfileName('default')) {
+      defaultProfilePath = await finder.getPath('default');
```
rpl: As an alternative approach to this two `if (await finder.hasProfileName(...)) { ... = await finder.getPath(...); }` we can opt to return from `createProfileFinder` a single function that returns a string (the profile path) if there is such a profile name and `undefined` (or `null`) otherwise 
(in other words: we can merge the `hasProfileName` and `getPath` function defined at line 253 into a single helper function with the behavior described above).

This change can make this method implementation cleaner and simpler, turning it into something like:

```js
const getProfilePath = createProfileFinder();
const dirExists = await isDirectory(profilePath);

let destinationDirectory;

if (dirExists) {
  ...
  if (profilePath === getProfilePath('default') ||
      profilePath === getProfilePath('dev-edition-default')) {
    throw new WebExtError(...);
  }
  
  destinationDirectory = profilePath;
} else {
  ...
  if (profilePath === 'default' ||
      profilePath === 'dev-edition-default') {
    throw new WebExtError(...);
  }

  destinationDirectory = getProfilePath(profilePath);

  if (!destinationDirectory) {
    throw new UserError(
      `The request "${profilePath}" profile name cannot be resolved to a profile path`
    );
  }
}

....

const profile = new FirefoxProfile({destinationDirectory});
return await configureThisProfile(profile, {app, customPrefs});
```

src/firefox/index.js
```diff
@@ -249,9 +281,45 @@ export async function useProfile(
...
+      if (profilePath === 'default' ||
+            profilePath === 'dev-edition-default') {
+        throw new WebExtError(
+          `Cannot use the blacklisted named profile "${profilePath}"`
+        );
+      }
+      const profileDirectory = await finder.getPath(profilePath);
+      profile = new FirefoxProfile({destinationDirectory: profileDirectory});
+    }
+  } catch (error) {
```
rpl: I'm not sure that this giant `try...catch` is actually useful, at least by looking currently we can just leave any other exception to be propagated to the callers.

harikishen:
@rpl  I've done some of the changes mentioned. Please look into it and tell me whether i'm on the right track.

harikishen:
@rpl I have a lot of doubts regarding rewriting the tests. So I've commented them for now.

rpl:
> @rpl I have a lot of doubts regarding rewriting the tests. So I've commented them for now.

Hi @harikishen, sorry for the waiting time (I've been traveling for a conference),
Can you please remove the comments added on these tests?

Commenting the tests that fail is worst than let them fail on travis, by leaving them to fail on travis I would be able to look at the failures and help you to figure out what to do about the failing tests.

In general, the main point of writing tests is not to make them pass, on the contrary, it is to see them fail when something has been changed and needs more changes to be handled correctly, so the main point of writing tests is to read the failures logs ;-)

PR #974, opened 2017-06-12
fix: Fixed Convert test suite to use sinon assertions

ahsanbarkati:


ahsanbarkati:
Fixed it

tests/unit/test-cmd/test.run.js
```diff
@@ -459,9 +459,9 @@ describe('run', () => {
         })
         .then(() => {
           assert.ok(fakeStdin.setRawMode.called);
-          assert.ok(addonReload.called);
-          assert.equal(addonReload.firstCall.args[0].addonId,
-            tempInstallResult.addon.id);
+          sinon.assert.calledWith(
+          addonReload, sinon.match({addonId: tempInstallResult.addon.id})
```
kumar303: this line should be indented two spaces to show that it's a continuation of the call.

kumar303:
Thanks, it's looking good. Let me know when it's ready for another review.

rpl:
Closing as incomplete, the issue is still valid (some tests still use manual assertion where sinon assertion could be used) but unfortunately this PR has been branched out quite some time ago and it is unlikely to be resumed from this old branch.

PR #984, opened 2017-06-22
fix: Add no-default-ignore-files cmd option (#799)

### LINKED ISSUE #799: Add --no-default-ignore-files option

feature

After https://github.com/mozilla/web-ext/issues/131 you can use the [--ignore-files](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference?document_saved=true#--ignore-files_-i) option. However, there is no way to reset the default ignore rules. 

You should be able to clear all the defaults and specify your own rules. You may want to do this, for example, if your extension source should contain `.zip` files.

It would look something like this:
````
web-ext --no-default-ignore-files --ignore-files "**/node_modules" build
````

### PR DESCRIPTION
**Pull request by ccarruitero**
Fixes #799

### DIFF
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -27,6 +27,7 @@ export type FileFilterOptions = {|
   ignoreFiles?: Array<string>,
   sourceDir: string,
   artifactsDir?: string,
+  noDefaultIgnoreFiles?: boolean
 |};
 
 /*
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/util/file-filter.js [hunk 1]
```diff
@@ -48,13 +49,16 @@ export class FileFilter {
...
     artifactsDir,
+    noDefaultIgnoreFiles = false,
   }: FileFilterOptions = {}) {
     sourceDir = path.resolve(sourceDir);
 
     this.filesToIgnore = [];
     this.sourceDir = sourceDir;
 
-    this.addToIgnoreList(baseIgnoredPatterns);
+    if (!noDefaultIgnoreFiles) {
```
ccarruitero: Not sure about this name for the option in the file filter, maybe can use something like `overrideDefaultIgnoreFiles`?

Aside, the filter has the option `baseIgnorePatterns`, that require an array. But I couldn't find an easy way to use in this case.

Rob--W:
FIxed by #846.

PR #993, opened 2017-07-05
chore: added some sinon assertions

Jp-Rivera:
Some more sinon assertions for review.

### DIFF
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -66,11 +66,9 @@ describe('build', () => {
     };
     return build(params, {packageCreator, createFileFilter}).then(() => {
       // ensure sourceDir, artifactsDir, ignoreFiles is used
-      assert.ok(createFileFilter.called);
-      assert.deepEqual(createFileFilter.firstCall.args[0], params);
+      sinon.assert.calledWith(createFileFilter, sinon.match(params));
       // ensure packageCreator received correct fileFilter
-      assert.ok(packageCreator.called);
-      assert.equal(packageCreator.firstCall.args[0].fileFilter, fileFilter);
+      sinon.assert.calledWith(packageCreator, sinon.match({fileFilter}));
     });
   });
 
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -66,9 +66,9 @@ describe('build', () => {
     };
     return build(params, {packageCreator, createFileFilter}).then(() => {
       // ensure sourceDir, artifactsDir, ignoreFiles is used
-      sinon.assert.calledWith(createFileFilter, sinon.match(params));
+      sinon.assert.calledWithMatch(createFileFilter, params);
       // ensure packageCreator received correct fileFilter
-      sinon.assert.calledWith(packageCreator, sinon.match({fileFilter}));
+      sinon.assert.calledWithMatch(packageCreator, {fileFilter});
     });
   });
 
--- a/tests/unit/test-cmd/test.docs.js
+++ b/tests/unit/test-cmd/test.docs.js
@@ -10,8 +10,7 @@ describe('docs', () => {
   it('passes the correct url to docs', () => {
     const openUrl = sinon.spy((urlToOpen, callback) => callback(null));
     return defaultDocsCommand({}, {openUrl}).then(() => {
-      sinon.assert.calledOnce(openUrl);
-      sinon.assert.calledWith(openUrl, sinon.match(url));
+      sinon.assert.calledWith(openUrl, url);
     });
   });
 
--- a/tests/unit/test-cmd/test.docs.js
+++ b/tests/unit/test-cmd/test.docs.js
@@ -10,8 +10,8 @@ describe('docs', () => {
   it('passes the correct url to docs', () => {
     const openUrl = sinon.spy((urlToOpen, callback) => callback(null));
     return defaultDocsCommand({}, {openUrl}).then(() => {
-      assert.ok(openUrl.called);
-      assert.equal(openUrl.firstCall.args[0], url);
+      sinon.assert.calledOnce(openUrl);
+      sinon.assert.calledWith(openUrl, sinon.match(url));
     });
   });
 
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -64,8 +64,8 @@ describe('run', () => {
     const cmd = prepareRun();
     const FirefoxDesktopExtensionRunner = sinon.spy(FakeExtensionRunner);
     await cmd.run({firefox}, {FirefoxDesktopExtensionRunner});
-    sinon.assert.calledWith(
-     FirefoxDesktopExtensionRunner, sinon.match({firefoxBinary: firefox})
+    sinon.assert.calledWithMatch(
+      FirefoxDesktopExtensionRunner, {firefoxBinary: firefox}
     );
   });
 
@@ -76,8 +76,8 @@ describe('run', () => {
 
     await cmd.run({startUrl: expectedStartUrls},
                   {FirefoxDesktopExtensionRunner});
-    sinon.assert.calledWith(
-     FirefoxDesktopExtensionRunner, sinon.match({startUrl: expectedStartUrls})
+    sinon.assert.calledWithMatch(
+      FirefoxDesktopExtensionRunner, {startUrl: expectedStartUrls}
     );
   });
 
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -64,9 +64,9 @@ describe('run', () => {
     const cmd = prepareRun();
     const FirefoxDesktopExtensionRunner = sinon.spy(FakeExtensionRunner);
     await cmd.run({firefox}, {FirefoxDesktopExtensionRunner});
-    assert.equal(FirefoxDesktopExtensionRunner.called, true);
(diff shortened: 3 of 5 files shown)


tests/unit/test-cmd/test.build.js
```diff
@@ -66,11 +66,9 @@ describe('build', () => {
     };
     return build(params, {packageCreator, createFileFilter}).then(() => {
       // ensure sourceDir, artifactsDir, ignoreFiles is used
-      assert.ok(createFileFilter.called);
-      assert.deepEqual(createFileFilter.firstCall.args[0], params);
+      sinon.assert.calledWith(createFileFilter, sinon.match(params));
```
kumar303: This should be `sinon.assert.calledWithMatch(createFileFilter, params)`

tests/unit/test-cmd/test.build.js
```diff
@@ -66,11 +66,9 @@ describe('build', () => {
...
       // ensure packageCreator received correct fileFilter
-      assert.ok(packageCreator.called);
-      assert.equal(packageCreator.firstCall.args[0].fileFilter, fileFilter);
+      sinon.assert.calledWith(packageCreator, sinon.match({fileFilter}));
```
kumar303: This can use `sinon.assert.calledWithMatch()` too

tests/unit/test-cmd/test.docs.js
```diff
@@ -10,8 +10,8 @@ describe('docs', () => {
   it('passes the correct url to docs', () => {
     const openUrl = sinon.spy((urlToOpen, callback) => callback(null));
     return defaultDocsCommand({}, {openUrl}).then(() => {
-      assert.ok(openUrl.called);
-      assert.equal(openUrl.firstCall.args[0], url);
+      sinon.assert.calledOnce(openUrl);
+      sinon.assert.calledWith(openUrl, sinon.match(url));
```
kumar303: Does this need to use `sinon.match()`?

tests/unit/test-cmd/test.docs.js
```diff
@@ -10,8 +10,8 @@ describe('docs', () => {
   it('passes the correct url to docs', () => {
     const openUrl = sinon.spy((urlToOpen, callback) => callback(null));
     return defaultDocsCommand({}, {openUrl}).then(() => {
-      assert.ok(openUrl.called);
-      assert.equal(openUrl.firstCall.args[0], url);
+      sinon.assert.calledOnce(openUrl);
```
kumar303: Is there a reason to check the call count here? I don't think you need to.

tests/unit/test-cmd/test.run.js
```diff
@@ -64,9 +64,9 @@ describe('run', () => {
     const cmd = prepareRun();
     const FirefoxDesktopExtensionRunner = sinon.spy(FakeExtensionRunner);
     await cmd.run({firefox}, {FirefoxDesktopExtensionRunner});
-    assert.equal(FirefoxDesktopExtensionRunner.called, true);
-    assert.equal(FirefoxDesktopExtensionRunner.firstCall.args[0].firefoxBinary,
-                 firefox);
+    sinon.assert.calledWith(
+     FirefoxDesktopExtensionRunner, sinon.match({firefoxBinary: firefox})
```
kumar303: This should be indented two spaces, not one.

tests/unit/test-cmd/test.run.js
```diff
@@ -64,9 +64,9 @@ describe('run', () => {
     const cmd = prepareRun();
     const FirefoxDesktopExtensionRunner = sinon.spy(FakeExtensionRunner);
     await cmd.run({firefox}, {FirefoxDesktopExtensionRunner});
-    assert.equal(FirefoxDesktopExtensionRunner.called, true);
-    assert.equal(FirefoxDesktopExtensionRunner.firstCall.args[0].firefoxBinary,
-                 firefox);
+    sinon.assert.calledWith(
```
kumar303: This can use `sinon.assert.calledWithMatch()`

tests/unit/test-cmd/test.run.js
```diff
@@ -76,10 +76,9 @@ describe('run', () => {
 
     await cmd.run({startUrl: expectedStartUrls},
                   {FirefoxDesktopExtensionRunner});
-
-    assert.ok(FirefoxDesktopExtensionRunner.called);
-    assert.deepEqual(FirefoxDesktopExtensionRunner.firstCall.args[0].startUrl,
-                     expectedStartUrls);
+    sinon.assert.calledWith(
+     FirefoxDesktopExtensionRunner, sinon.match({startUrl: expectedStartUrls})
```
kumar303: This should be indented two spaces. All code should be indented two spaces but I don't think we have a lint rule for it when it applies to extending function calls.

tests/unit/test-cmd/test.run.js
```diff
@@ -76,10 +76,9 @@ describe('run', () => {
 
     await cmd.run({startUrl: expectedStartUrls},
                   {FirefoxDesktopExtensionRunner});
-
-    assert.ok(FirefoxDesktopExtensionRunner.called);
-    assert.deepEqual(FirefoxDesktopExtensionRunner.firstCall.args[0].startUrl,
-                     expectedStartUrls);
+    sinon.assert.calledWith(
```
kumar303: This can use `sinon.assert.calledWithMatch()`

tests/unit/test-extension-runners/test.firefox-desktop.js
```diff
@@ -103,10 +103,10 @@ describe('util/extension-runners/firefox-desktop', () => {
     await runnerInstance.run();
 
     assert.ok(runnerInstance.getName(), 'Firefox Desktop');
-    assert.ok(remoteFirefox.installTemporaryAddon.calledOnce);
-    assert.equal(remoteFirefox.installTemporaryAddon.firstCall.args[0],
-                 params.extensions[0].sourceDir);
-
+    sinon.assert.calledOnce(remoteFirefox.installTemporaryAddon);
+    sinon.assert.calledWith(remoteFirefox.installTemporaryAddon,
```
kumar303: This can use `sinon.assert.calledWithMatch()`

tests/unit/test.program.js
```diff
@@ -73,8 +73,8 @@ describe('program.Program', () => {
       shouldExitProgram: true,
     })
       .then(() => {
-        assert.equal(fakeProcess.exit.called, true);
-        assert.equal(fakeProcess.exit.firstCall.args[0], 1);
+        sinon.assert.calledOnce(fakeProcess.exit);
+        sinon.assert.calledWith(fakeProcess.exit, sinon.match(1));
```
kumar303: does this need to use `sinon.match()`?

tests/unit/test.program.js
```diff
@@ -73,8 +73,8 @@ describe('program.Program', () => {
       shouldExitProgram: true,
     })
       .then(() => {
-        assert.equal(fakeProcess.exit.called, true);
-        assert.equal(fakeProcess.exit.firstCall.args[0], 1);
+        sinon.assert.calledOnce(fakeProcess.exit);
+        sinon.assert.calledWithMatch(fakeProcess.exit, 1);
```
kumar303: Why does this need to use `calledWithMatch()`? Can it just use `calledWith()`?

Jp-Rivera: The sinon.js documentation shows sinon.match(number) requires the value to be "==" to the given number. In this case, it seems like we are checking for a specific number value.

calledWith() passes and throws a similar error message as calledWithMatch(), but I am not sure that it is performing the "==" check. 

To put it another way…here are the tests and errors when the values do not match.

1.Test -  sinon.assert.calledWith(fakeProcess.exit, 0);
1.Error - "AssertError: expected exit to be called with arguments 1 0"

2.Test - sinon.assert.calledWith(fakeProcess.exit, sinon.match(0));
2.Error - "AssertError: expected exit to be called with arguments 1 match(0)"

3.Test -  sinon.assert.calledWithMatch(fakeProcess.exit, 0);
3.Error - "AssertError: expected exit to be called with match 1 0”

I know that #2 and #3 are identical, other than that #3’s error message reads better. After digging into this a bit more. I am unsure of what sets #1 apart from the other two. I chose to use calledWithMatch() to make sure I was checking for a specific value, not just a number.

kumar303: `assert.calledWithMatch()` is really only good for cases where you don't need to check all function arguments or if you only want to allow arbitrary object parameters in the value that you're checking. For all other cases you should use `assert.calledWith()`. 

As a general rule of thumb, first start with `assert.calledWith()` and if the test fails because the check is too strict (e.g. for comparing objects), consider switching to `assert.calledWithMatch()`.

In this case, the test should definitely use `assert.calledWith()` because it needs to check that `exit()` is called exactly with the number `1`. 

Here is a patch you can try:

```diff
diff --git a/src/program.js b/src/program.js
index 07559c5..fed0a93 100644
--- a/src/program.js
+++ b/src/program.js
@@ -161,7 +161,8 @@ export class Program {
       }
 
       if (this.shouldExitProgram) {
-        systemProcess.exit(1);
+        // $FLOW_IGNORE
+        systemProcess.exit(true);
       } else {
         throw error;
       }
```

This should cause the test to fail but when using `assert.calledWithMatch()` it passes since it's using a loose equality check (`==`).

As you can see, Flow helps us mitigate problems like this -- I was only using this as an example to show how the tests will need to avoid loose equality. The `assert.calledWith()` method will use strict equality (`===`).

Jp-Rivera: I see, my confusion was thinking that (==) was strict equality. Thank you for clearing this up.

PR #1002, opened 2017-07-12
fix: Don't show the command in error messages

### LINKED ISSUE #994: Don't show the command in error messages

feature

If you encounter an error, web-ext will always prefix the command to the error message, like:

```
$ web-ext run --target firefox-android -v --android-device FA6AX0301736

run: UsageError: Select one of the packages using --firefox-apk

```

I don't think we need to show the command. It can just be:

```
$ web-ext run --target firefox-android -v --android-device FA6AX0301736

UsageError: Select one of the packages using --firefox-apk

```

The command name is added as a [prefix here](https://github.com/mozilla/web-ext/blob/4a4641c561ab43d6293b1e00e1ad617e9134f4c3/src/program.js#L153). When removed, it should still be added for verbose mode, like:

```js
log.debug(`Command executed: ${cmd}`);
```

Originally, the command prefix was added to help in bug reports but really we always need a verbose log for debug reports anyway. Also, you can look at the command line to seed what command was run.

tsl143:
Fixes #994

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -150,15 +150,16 @@ export class Program {
       await runCommand(argv, {shouldExitProgram});
 
     } catch (error) {
-      const prefix = cmd ? `${cmd}: ` : '';
       if (!(error instanceof UsageError) || argv.verbose) {
-        log.error(`\n${prefix}${error.stack}\n`);
+        log.error(`\n${error.stack}\n`);
       } else {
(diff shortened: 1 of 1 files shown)


tsl143:
Hahaha, even I forgot this :P
Thanks for merge :)

PR #1003, opened 2017-07-13
docs: Add instructions for a devDependencies install

clarkbw:
I've seen this issue come up a number of times where people miss the installation step of `web-ext` https://github.com/facebook/react-devtools/issues/847

Also when installed as one of the `devDependencies` then services like greenkeeper will keep the packaged updated where as global installations tend to never be updated by developers and so new versions can be missed and teams get out of sync.

In general I avoid the global install approach but I added it as an option here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,30 @@ cross-platform way. Initially, it will provide a streamlined experience for deve
 
 ## Installation from npm
 
+### Global
+
+You can install this command onto your machine globally.
+
     npm install --global web-ext
+
+### Dependency
+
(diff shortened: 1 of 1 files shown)


kumar303:
Thanks Bryan! That's a good idea. I hadn't thought of it since I always add any `npm` task dependency to my `package.json` but I could totally see how others may not know about this pattern.

Don't worry about the test failure, it was just a formatting issue for auto-generating our changelog which will be moot once I merge.

clarkbw:
Thanks @kumar303!

PR #1007, opened 2017-07-17
fix: Prevent --keep-profile-changes on default Firefox profiles

### LINKED ISSUE #1005: Using a default profile with --keep-profile-changes will break Firefox

bug

Consider a command like this (DO NOT run this):

```
web-ext run --firefox-profile=default --keep-profile-changes
```

Here's what web-ext will do to your default profile:
- It will disable automatic updates to Firefox
- It will disable updates to all add-ons
- It will disable the prompt for accepting a remote debugging connection -- this means anyone can silently connect a remote debugger to your Firefox
- It will disable signature checks for add-ons (but only in pre-release builds of Firefox)
- It will allow side loading add-ons which is how a lot of crapware installers inject their add-ons without user consent

This really shouldn't be possible at all. It should raise an error. We could possibly allow it with an option like `--allow-web-ext-to-completely-break-my-profile` because it's possible that someone may not care about trashing their default profile? I don't know, I'm leaning to not allowing this at all.

* **Firefox version**: N/A
* **Your OS and version**: N/A

````
v6.10.3
3.10.10
master-7b390e45f3a998a57238ddb9f163da4a61dfb08a
````

### PR DESCRIPTION
**Pull request by rpl**
This PR extracts (and expands) the "checks for the Firefox default profile" pieces that we are working on in #968 to fix #1005 

- [x] additional checks needed to prevent the usage on the --keep-profile-changes option on a default Firefox profile (when specified using the name, a full path or a relative path)
- [x] additional unit tests

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -199,6 +199,7 @@ const DEFAULT_PROFILES_NAMES = [
 export type IsDefaultProfileFn = (
   profilePathOrName: string,
   ProfileFinder?: typeof FirefoxProfile.Finder,
+  fsStat?: typeof fs.stat,
 ) => Promise<boolean>;
 
 /*
@@ -209,7 +210,8 @@ export type IsDefaultProfileFn = (
  */
 export async function isDefaultProfile(
   profilePathOrName: string,
-  ProfileFinder?: typeof FirefoxProfile.Finder = FirefoxProfile.Finder
+  ProfileFinder?: typeof FirefoxProfile.Finder = FirefoxProfile.Finder,
+  fsStat?: typeof fs.stat = fs.stat,
 ): Promise<boolean> {
   if (DEFAULT_PROFILES_NAMES.includes(profilePathOrName)) {
     return true;
@@ -218,9 +220,11 @@ export async function isDefaultProfile(
   const baseProfileDir = ProfileFinder.locateUserDirectory();
   const profilesIniPath = path.join(baseProfileDir, 'profiles.ini');
   try {
-    await fs.stat(profilesIniPath);
+    await fsStat(profilesIniPath);
   } catch (error) {
     if (isErrorWithCode('ENOENT', error)) {
+      log.debug(`profiles.ini not found: ${error}`);
+
       // No profiles exist yet, default to false (the default profile name contains a
       // random generated component).
       return false;
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -336,9 +336,10 @@ export async function useProfile(
   const isForbiddenProfile = await isFirefoxDefaultProfile(profilePath);
   if (isForbiddenProfile) {
     throw new UsageError(
-      'Using --keep-profile-changes' +
-      ` on the Firefox default profile "${profilePath}" is forbidden.` +
-      '\n(See https://github.com/mozilla/web-ext/issues/1005)'
+      'Cannot use --keep-profile-changes on a default profile' +
+      ` ("${profilePath}")` +
+      ' because web-ext will make it insecure and unsuitable for daily use.' +
+      '\nSee https://github.com/mozilla/web-ext/issues/1005'
     );
   }
   const profile = new FirefoxProfile({destinationDirectory: profilePath});
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -211,7 +211,7 @@ export async function isDefaultProfile(
   profilePathOrName: string,
   ProfileFinder?: typeof FirefoxProfile.Finder = FirefoxProfile.Finder
 ): Promise<boolean> {
-  if (DEFAULT_PROFILES_NAMES.indexOf(profilePathOrName) >= 0) {
+  if (DEFAULT_PROFILES_NAMES.includes(profilePathOrName)) {
     return true;
   }
 
@@ -243,7 +243,7 @@ export async function isDefaultProfile(
   for (const profile of finder.profiles) {
     // Check if the profile dir path or name is one of the default profiles
     // defined in the profiles.ini file.
-    if (DEFAULT_PROFILES_NAMES.indexOf(profile.Name) >= 0 ||
+    if (DEFAULT_PROFILES_NAMES.includes(profile.Name) ||
         profile.Default === '1') {
       let profileFullPath;
 
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -189,6 +189,86 @@ export async function run(
 }
 
 
+// isDefaultProfile types and implementation.
+
+const DEFAULT_PROFILES_NAMES = [
+  'default',
+  'dev-edition-default',
+];
+
+export type IsDefaultProfileFn = (
+  profilePathOrName: string,
+  ProfileFinder?: typeof FirefoxProfile.Finder,
+) => Promise<boolean>;
+
+/*
+ * Tests if a profile is a default Firefox profile (both as a profile name or
+ * profile path).
+ *
+ * Returns a promise that resolves to true if the profile is one of default Firefox profile.
+ */
+export async function isDefaultProfile(
+  profilePathOrName: string,
+  ProfileFinder?: typeof FirefoxProfile.Finder = FirefoxProfile.Finder
+): Promise<boolean> {
+  if (DEFAULT_PROFILES_NAMES.indexOf(profilePathOrName) >= 0) {
+    return true;
+  }
+
+  const baseProfileDir = ProfileFinder.locateUserDirectory();
+  const profilesIniPath = path.join(baseProfileDir, 'profiles.ini');
+  try {
+    await fs.stat(profilesIniPath);
+  } catch (error) {
+    if (isErrorWithCode('ENOENT', error)) {
+      // No profiles exist yet, default to false (the default profile name contains a
+      // random generated component).
+      return false;
+    }
+
+    // Re-throw any unexpected exception.
+    throw error;
+  }
+
+  // Check for profile dir path.
+  const finder = new ProfileFinder(baseProfileDir);
+  const readProfiles = promisify(finder.readProfiles, finder);
+
+  await readProfiles();
+
+  const normalizedProfileDirPath = path.normalize(
+    path.join(path.resolve(profilePathOrName), path.sep)
+  );
+
+  for (const profile of finder.profiles) {
+    // Check if the profile dir path or name is one of the default profiles
+    // defined in the profiles.ini file.
+    if (DEFAULT_PROFILES_NAMES.indexOf(profile.Name) >= 0 ||
+        profile.Default === '1') {
+      let profileFullPath;
+
+      // Check for profile name.
+      if (profile.Name === profilePathOrName) {
+        return true;
+      }
+
+      // Check for profile path.
+      if (profile.IsRelative === '1') {
+        profileFullPath = path.join(baseProfileDir, profile.Path, path.sep);
+      } else {
+        profileFullPath = path.join(profile.Path, path.sep);
+      }
+
+      if (path.normalize(profileFullPath) === normalizedProfileDirPath) {
+        return true;
+      }
+    }
+  }
+
+  // Profile directory not found.
+  return false;
+}
+
 // configureProfile types and implementation.
 
 export type ConfigureProfileOptions = {|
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rpl** [comment]
@kumar303 This should be ready to be reviewed now. (I'm going to also test it manually on Windows as I did on Linux in the meantime).

@saintsebastian I've also added you as a reviewer, because I'm touching files that could probably impact some of your open pull requests, and a pair of additional eyes are always appreciated  ;-)

File: src/firefox/index.js [hunk 1]
```diff
@@ -189,6 +189,86 @@ export async function run(
 }
 
 
+// isDefaultProfile types and implementation.
+
+const DEFAULT_PROFILES_NAMES = [
```
saintsebastian: I'm no expert on firefox profiles, but could that be that those names are misleading? As in I could have created a profile which i called default but don't actually use if as default.

src/firefox/index.js
```diff
@@ -248,9 +329,18 @@ export async function useProfile(
   {
     app,
     configureThisProfile = configureProfile,
+    isFirefoxDefaultProfile = isDefaultProfile,
     customPrefs = {},
   }: UseProfileParams = {},
 ): Promise<FirefoxProfile> {
+  const isForbiddenProfile = await isFirefoxDefaultProfile(profilePath);
```
saintsebastian: Kumar proposed option like `--allow-web-ext-to-completely-break-my-profile`, i think that might be a good idea if we show a warning first, since if users want to do something potentially dangerous that's on them.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -35,6 +35,37 @@ function withBaseProfile(callback) {
...
+  return FakeProfileFinder;
+}
+
+async function createFakeProfilesIni(dirPath, profilesDefs) { // eslint-disable-line
+  let content = '';
+
+  for (const [idx, profile] of profilesDefs.entries()) {
+    content += `[Profile${idx}]\n`;
+    for (const k of Object.keys(profile)) {
+      // $FLOW_FIXME: test123
```
saintsebastian: I want to do that very often. Maybe a more descriptive reason for why fixme is needed would make it easier to fix it later.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +294,113 @@ describe('firefox', () => {
...
+           'dev-edition-default'
+         );
+         assert.equal(isDevEditionDefault, true);
+       });
+
+    it('allows profile name if is not listed as default in profiles.ini',
+      async () => {
+        return withTempDir(async (tmpDir) => {
+          const profilesDirPath = tmpDir.path();
+          const FakeProfileFinder = createFakeProfileFinder(profilesDirPath);
```
kumar303: Instead of using a fake `ProfileFinder`, could you just pass in an explicit `baseProfileDir` value to `firefox.isDefaultProfile()`? That would bypass `ProfileFinder.locateUserDirectory()` but continue to use a real `ProfileFinder` which would provide better test coverage.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +294,113 @@ describe('firefox', () => {
...
+              Path: 'fake-default-profile',
+              IsRelative: 1,
+              Default: 1,
+            },
+          ]);
+
+          const isDefault = await firefox.isDefaultProfile(
+            'manually-set-default', FakeProfileFinder
+          );
+          const isNotDefault = await firefox.isDefaultProfile(
```
kumar303: This seems like it should be in a separate unit test (unless it's sharing the same setup?) to make test failures easier to deal with

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +294,113 @@ describe('firefox', () => {
...
+            'unkown-profile-name', FakeProfileFinder
+          );
+
+          assert.equal(isDefault, true);
+          assert.equal(isNotDefault, false);
+        });
+      });
+
+    it('allows profile path if is not listed as default in profiles.ini',
```
kumar303: nit: allows profile path if it is not ...

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +294,113 @@ describe('firefox', () => {
...
+         const isDefault = await firefox.isDefaultProfile('default');
+         assert.equal(isDefault, true);
+
+         const isDevEditionDefault = await firefox.isDefaultProfile(
+           'dev-edition-default'
+         );
+         assert.equal(isDevEditionDefault, true);
+       });
+
+    it('allows profile name if is not listed as default in profiles.ini',
```
kumar303: nit: allows profile name if it is not ...

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +294,113 @@ describe('firefox', () => {
...
+           );
+           assert.equal(isDevEditionDefaultPath, true);
+
+           const isManuallyDefault = await firefox.isDefaultProfile(
+             '/tmp/fake-manually-default-profile',
+             FakeProfileFinder
+           );
+           assert.equal(isManuallyDefault, true);
+
+           const isNotDefault = await firefox.isDefaultProfile(
```
kumar303: This case looks like it's covered up above. Is there something different about testing it again here?

src/firefox/index.js
```diff
@@ -189,6 +189,86 @@ export async function run(
...
+  await readProfiles();
+
+  const normalizedProfileDirPath = path.normalize(
+    path.join(path.resolve(profilePathOrName), path.sep)
+  );
+
+  for (const profile of finder.profiles) {
+    // Check if the profile dir path or name is one of the default profiles
+    // defined in the profiles.ini file.
+    if (DEFAULT_PROFILES_NAMES.indexOf(profile.Name) >= 0 ||
```
mstriemer: It could be less noisy to use `.includes()`, then it seems simple enough to keep together if it makes sense.

tests/unit/test-firefox/test.firefox.js
rpl: @kumar303 @saintsebastian it is an acronym :-) it means: "Damn... we will have a chat you and me... later... for now just shut up" :-P

Thank you both for spotting it, I'm going to have a chat with flow about this and get back to you with more info about it :-D

tests/unit/test-firefox/test.firefox.js
```diff
@@ -35,6 +35,37 @@ function withBaseProfile(callback) {
...
+  });
+
+  FakeProfileFinder.locateUserDirectory = sinon.spy(() => {
+    return profilesDirPath;
+  });
+
+  return FakeProfileFinder;
+}
+
+async function createFakeProfilesIni(dirPath, profilesDefs) { // eslint-disable-line
```
rpl: definitely not needed, it is a leftover while I was rewriting some of the new tests.

src/firefox/index.js
```diff
@@ -248,9 +329,18 @@ export async function useProfile(
...
     app,
     configureThisProfile = configureProfile,
+    isFirefoxDefaultProfile = isDefaultProfile,
     customPrefs = {},
   }: UseProfileParams = {},
 ): Promise<FirefoxProfile> {
+  const isForbiddenProfile = await isFirefoxDefaultProfile(profilePath);
+  if (isForbiddenProfile) {
+    throw new UsageError(
+      'Using --keep-profile-changes' +
```
rpl: :+1:

src/firefox/index.js
rpl: @mstriemer :+1: yeah, I did it mostly for compatibility, but to be fair we are already transpiling this files with babel and it is completely possible that the polyfill is already there and supported across the nodejs versions, I'll change it to use `.includes()`.

rpl: @kumar303 if I'm not wrong, we would miss the default profiles when specified by path and not marked as Default in the profiles.ini file

src/firefox/index.js
```diff
@@ -189,6 +189,86 @@ export async function run(
...
+    await fs.stat(profilesIniPath);
+  } catch (error) {
+    if (isErrorWithCode('ENOENT', error)) {
+      // No profiles exist yet, default to false (the default profile name contains a
+      // random generated component).
+      return false;
+    }
+
+    // Re-throw any unexpected exception.
+    throw error;
```
rpl: you are right... "never trust the coverage report" :-) it is not marked as missed (nor executed actually) in the html report generated locally.

src/firefox/index.js
```diff
@@ -189,6 +189,86 @@ export async function run(
...
+  }
+
+  const baseProfileDir = ProfileFinder.locateUserDirectory();
+  const profilesIniPath = path.join(baseProfileDir, 'profiles.ini');
+  try {
+    await fs.stat(profilesIniPath);
+  } catch (error) {
+    if (isErrorWithCode('ENOENT', error)) {
+      // No profiles exist yet, default to false (the default profile name contains a
+      // random generated component).
```
rpl: yeah I agree

tests/unit/test-firefox/test.firefox.js
rpl: we are actually overriding only `locateUserDirectory` static method (which is only used directly in our helper, because we manually pass it to the created `ProfileFinder` and so it will never be discovered again), all the other methods are the real ones.

I also remembered right now that I added the sinon spies to add some additional assertion... that I should add :-)

src/firefox/index.js
rpl: I would prefer to keep this as strict as "you can use a different profile name instead of these ones"

mostly because the incorrect usage can be implemented in a script (like the react-devtools automation script mentioned in #1005) and so the user could be not in charge of that "potentially harmful" decision.

src/firefox/index.js
rpl: @saintsebastian @kumar303 I'm open to implement such option if we decide to go for it, but I vote against it :-)

tests/unit/test-firefox/test.firefox.js
rpl: @kumar303 @saintsebastian flow asked me to tell you that everything is ok now and it is happy again ;-) we agreed on something :-P (faac3d5)

tests/unit/test-firefox/test.firefox.js
rpl: you are right, it is a leftover of the rewrite

tests/unit/test-firefox/test.firefox.js
rpl: @kumar303 How about adding some assertion messages where useful instead? (added in 3c1c0a3)

src/firefox/index.js
kumar303: I thought about it more (especially when reviewing this patch) and I don't know that introducing the `--allow-web-ext-to-completely-break-my-profile` option is the right thing to do. 

It's nice because it gives a clear indication of not only keeping profile changes but also letting web-ext save its own settings. However, it will just be annoying to those who know what they are doing. Plus, there is really no valid use case (other than laziness) that requires someone to use their default profile with web-ext.

src/firefox/index.js
kumar303: > we would miss the default profiles when specified by path and not marked as Default in the profiles.ini file

Oh, right! In that case, should the [early return statement](https://github.com/mozilla/web-ext/pull/1007/files#diff-e20d3de9931a6adfc2906b995bc90df6R215) be removed?

rpl: @kumar303 the early return should ensure that the usage error with the "Cannot use default profile with --keep-profile-changes..."  message is logged also when the profiles.ini file doesn't exist yet and the user is using "default" or "dev-edition-default" as the firefox profile parameter.

rpl:
@kumar303 this should be ready now, it can also help to have a second pair of eyes that check the new behavior locally

tests/unit/test-firefox/test.firefox.js
```diff
@@ -263,6 +295,160 @@ describe('firefox', () => {
...
+           const FakeProfileFinder = createFakeProfileFinder(profilesDirPath);
+           const fakeFsStat = sinon.spy(() => {
+             return Promise.reject(new Error('Fake fs stat error'));
+           });
+
+           let exception;
+           try {
+             await firefox.isDefaultProfile(
+               '/tmp/my-custom-profile-dir',
+               FakeProfileFinder,
```
kumar303: I prefer to see optional function parameters passed as an object, like:
```js
firefox.isDefaultProfile('/tmp/my-custom-profile-dir', {
  profileFinder: FakeProfileFinder,
  fsStat: fakeFsStat,
}); 
```
However, this preference was born in a world without Flow so I can see how it's not such a big deal anymore. Anyway, I just wanted to point it out since it would be the first time we have a function in web-ext that does optional params like this.

PR #1009, opened 2017-07-20
test: async tests for build cmd

### LINKED ISSUE #755: Move all test code to async/await

internal feature of the architecture

Web-ext's program code uses async/await but all of the tests still use promises. They are hard to read and some things like using try/catch/finally is really difficult.

All tests should use async/await

saintsebastian:
Fixes #755 
My first attempt to rewrite tests for this issue. I feel it wasn't the most graceful and I've missed something. 

Side note: `test.lint.js` is being rewritten as a part of the [PR 817](https://github.com/mozilla/web-ext/pull/817)

### DIFF
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -71,7 +71,6 @@ describe('build', () => {
       });
       assert.match(buildResult.extensionPath,
                    /name_of_the_extension-1\.0\.zip$/);
-      buildResult.extensionPath;
     }
   ));
 
@@ -169,6 +168,7 @@ describe('build', () => {
 
   it('can build an extension without an ID', () => withTempDir(
     async (tmpDir) => {
+      // Make sure a manifest without an ID doesn't throw an error.
       await build({
         sourceDir: fixturePath('minimal-web-ext'),
         artifactsDir: tmpDir.path(),
@@ -199,7 +199,6 @@ describe('build', () => {
       assert.match(
         buildResult.extensionPath,
         /the_extension-0\.0\.1\.zip$/);
-      buildResult.extensionPath;
     }
   ));
 
@@ -235,7 +234,7 @@ describe('build', () => {
       });
       // Make sure we still have a build result.
       assert.match(buildResult.extensionPath, /\.zip$/);
-      assert.equal(onSourceChange.called, true);
+      sinon.assert.called(onSourceChange);
       const args = onSourceChange.firstCall.args[0];
       assert.equal(args.sourceDir, sourceDir);
       assert.equal(args.artifactsDir, artifactsDir);
@@ -270,7 +269,7 @@ describe('build', () => {
       }, {
         manifestData: basicManifest, onSourceChange, packageCreator,
       });
-      assert.equal(onSourceChange.called, true);
+      sinon.assert.called(onSourceChange);
       assert.equal(packageCreator.callCount, 1);
 
       const {onChange} = onSourceChange.firstCall.args[0];
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -78,12 +78,14 @@ describe('build', () => {
   it('handles repeating localization keys', () => withTempDir(
     async (tmpDir) => {
       const messageFileName = path.join(tmpDir.path(), 'messages.json');
-      await fs.writeFile(messageFileName,
+      await fs.writeFile(
+        messageFileName,
         `{"extensionName": {
             "message": "example extension",
             "description": "example description"
           }
-        }`);
+        }`
+      );
 
       const manifestWithRepeatingPattern = {
         name: '__MSG_extensionName__ __MSG_extensionName__',
@@ -110,7 +112,7 @@ describe('build', () => {
         makeSureItFails();
       } catch (error) {
         assert.instanceOf(error, UsageError);
-        assert.match(error.message, /Unexpected token '"' at 1:15/);
+        assert.match(error.message, /Unexpected string in JSON at position 14/);
         assert.match(error.message, /^Error parsing messages.json/);
         assert.include(error.message, messageFileName);
       }
@@ -121,7 +123,8 @@ describe('build', () => {
     async (tmpDir) => {
       const messageFileName = path.join(tmpDir.path(), 'messages.json');
       //This is missing the 'message' key
-      await fs.writeFile(messageFileName,
+      await fs.writeFile(
+        messageFileName,
         `{"extensionName": {
             "description": "example extension"
             }
@@ -139,7 +142,8 @@ describe('build', () => {
         makeSureItFails();
       } catch (error) {
         assert.instanceOf(error, UsageError);
-        assert.match(error.message,
+        assert.match(
+          error.message,
           /The locale file .*messages\.json is missing key: extensionName/);
       }
     }
@@ -155,7 +159,8 @@ describe('build', () => {
     } catch (error) {
       log.info(error);
       assert.instanceOf(error, UsageError);
-      assert.match(error.message,
+      assert.match(
+        error.message,
         /Error: ENOENT: no such file or directory, open .*messages.json/);
       assert.match(error.message, /^Error reading messages.json/);
       assert.include(error.message, '/path/to/non-existent-dir/messages.json');
@@ -191,7 +196,8 @@ describe('build', () => {
       }, {
         manifestData: basicManifest,
       });
-      assert.match(buildResult.extensionPath,
+      assert.match(
+        buildResult.extensionPath,
         /the_extension-0\.0\.1\.zip$/);
       buildResult.extensionPath;
     }
@@ -275,7 +281,8 @@ describe('build', () => {
         onChange();
         makeSureItFails();
       } catch (error) {
-        assert.include(error && error.message,
+        assert.include(
+          error && error.message,
           'Simulate an error on the second call to packageCreator()');
       }
     }
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -17,7 +17,6 @@ import {withTempDir} from '../../../src/util/temp-dir';
 import {
   basicManifest,
   fixturePath,
-  makeSureItFails,
   ZipFile,
 } from '../helpers';
 import {manifestWithoutApps} from '../test-util/test.manifest';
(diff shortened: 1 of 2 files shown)


tests/unit/test-cmd/test.build.js
```diff
@@ -28,32 +27,25 @@ const log = createLogger(__filename);
...
+  it('zips a package', () => withTempDir(
+    async (tmpDir) => {
+      const zipFile = new ZipFile();
+      const buildResult = await build({
+        sourceDir: fixturePath('minimal-web-ext'),
+        artifactsDir: tmpDir.path(),
+      });
+      assert.match(buildResult.extensionPath,
+                   /minimal_extension-1\.0\.zip$/);
+      const extensionPath = await buildResult.extensionPath;
```
rpl: @saintsebastian no need to `await` here, `buildResult.extensionPath` should already be a string and you can pass it directly to zipFile.open at line 40.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
-        };
+  it('gives the correct name to a localized extension', () => withTempDir(
+    async (tmpDir) => {
+      const buildResult = await build({
+        sourceDir: fixturePath('minimal-localizable-web-ext'),
+        artifactsDir: tmpDir.path(),
+      });
+      assert.match(buildResult.extensionPath,
+                   /name_of_the_extension-1\.0\.zip$/);
+      await buildResult.extensionPath;
```
rpl: @saintsebastian This last await is not needed (the `return` that it is replacing was also not needed)

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
-          .then((result) => {
-            assert.match(result, /example extension example extension/);
-          });
-      }
-    );
-  });
+  it('handles repeating localization keys', () => withTempDir(
+    async (tmpDir) => {
+      const messageFileName = path.join(tmpDir.path(), 'messages.json');
+      fs.writeFileSync(messageFileName,
```
rpl: nit, we could also rewrite this using `await fs.writeFile(...)`, instead of the sync version of the node fs method.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
-    return withTempDir(
-      (tmpDir) => {
-        const messageFileName = path.join(tmpDir.path(), 'messages.json');
-        fs.writeFileSync(messageFileName,
-          '{"simulated:" "json syntax error"');
-        return getDefaultLocalizedName({
+  it('checks locale file for malformed json', () => withTempDir(
+    async (tmpDir) => {
+      const messageFileName = path.join(tmpDir.path(), 'messages.json');
+      fs.writeFileSync(messageFileName, '{"simulated:" "json syntax error"');
```
rpl: nit, same as above.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
+      let expectedError;
+      try {
+        await getDefaultLocalizedName({
           messageFile: messageFileName,
           manifestData: manifestWithoutApps,
-        })
-          .then(makeSureItFails())
```
rpl: we can also map the current test strategy by using `makeSureItFails` inside the try block to make sure that the catch block is always executed and then put the assertions directly inside the catch block:

```
try {
  await getDefaultLocalizedName(...);
  makeSureItFails();
} catch (error) {
  assert.instanceOf(error, UsageError);
  ...
}
```

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
-        const basicLocalizedManifest = {
-          name: '__MSG_extensionName__',
-          version: '0.0.1',
-        };
-        return getDefaultLocalizedName({
+  it('checks locale file for incorrect format', () => withTempDir(
+    async (tmpDir) => {
+      const messageFileName = path.join(tmpDir.path(), 'messages.json');
+      //This is missing the 'message' key
+      fs.writeFileSync(messageFileName,
```
rpl: nit, same as for the other `fs.writeFileSync`.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
+        name: '__MSG_extensionName__',
+        version: '0.0.1',
+      };
+      let expectedError;
+      try {
+        await getDefaultLocalizedName({
           messageFile: messageFileName,
           manifestData: basicLocalizedManifest,
-        })
-          .then(makeSureItFails())
```
rpl: Same as the above `try { ... } catch { ... }` and the `makeSureItFails` test helper.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
-      })
-        .then((buildResult) => {
-          assert.match(buildResult.extensionPath,
-                       /the_extension-0\.0\.1\.zip$/);
-          return buildResult.extensionPath;
-        })
+      });
+      assert.match(buildResult.extensionPath,
+        /the_extension-0\.0\.1\.zip$/);
+      await buildResult.extensionPath;
```
rpl: `await` not needed here

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
+      assert.typeOf(args.shouldWatchFile, 'function');
+      args.shouldWatchFile('/some/path');
+      assert.equal(fileFilter.wantFile.called, true);
+
+      // Remove the built extension.
+      await fs.unlink(buildResult.extensionPath);
+      // Execute the onChange handler to make sure it gets built
+      // again. This simulates what happens when the file watcher
+      // executes the callback.
+      await args.onChange();
```
rpl: I'm not sure that args.onChange returns a promise (we can probably just call it without the `await`, at least if we don't also need to "wait for a tick").

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
+      assert.equal(onSourceChange.called, true);
+      assert.equal(packageCreator.callCount, 1);
+
+      const {onChange} = onSourceChange.firstCall.args[0];
+      packageResult = Promise.reject(new Error(
+        'Simulate an error on the second call to packageCreator()'));
+      // Invoke the stub packageCreator() again which should throw an error
+      let expectedError;
+      try {
+        await onChange();
```
rpl: as the above `await onChange()`, let's check if we really need the await here.

tests/unit/test-cmd/test.build.js
```diff
@@ -64,301 +56,271 @@ describe('build', () => {
...
+    }
+  ));
 
-          const {onChange} = onSourceChange.firstCall.args[0];
-          packageResult = Promise.reject(new Error(
-            'Simulate an error on the second call to packageCreator()'));
-          // Invoke the stub packageCreator() again which should throw an error
-          return onChange();
-        })
-        .then(makeSureItFails())
```
rpl: as for the other `try...catch`, we can opt to use `makeSureItfails` instead of the external `expectedError` variable.

tests/unit/test-cmd/test.build.js
saintsebastian: i like this idea quite a bit

tests/unit/test-cmd/test.build.js
saintsebastian: This breaks the test `Error: ENOENT: no such file or directory, stat '/var/folders/y_/qr18jygs6z56qqtmwnsn7nmh0000gn/T/tmp-web-ext-5923H6qh7ZsBc4oF/the_extension-0.0.1.zip'`.
I think we need to `await` or rewrite this test

saintsebastian: This breaks the test Error: ENOENT: no such file or directory, stat '/var/folders/y_/qr18jygs6z56qqtmwnsn7nmh0000gn/T/tmp-web-ext-5923H6qh7ZsBc4oF/the_extension-0.0.1.zip'.
I think we need to await or rewrite this test

saintsebastian:
@rpl @kumar303 just a ping to remind about this PR, I rebased it and it's ready to be reviewed

tests/unit/test-cmd/test.build.js
```diff
@@ -64,313 +56,269 @@ describe('build', () => {
...
-          assert.match(buildResult.extensionPath, /\.zip$/);
-          return fs.stat(buildResult.extensionPath);
-        })
-        .then((stat) => {
-          assert.equal(stat.isFile(), true);
-        });
+      });
+      // Make sure we still have a build result.
+      assert.match(buildResult.extensionPath, /\.zip$/);
+      assert.equal(onSourceChange.called, true);
```
rpl: we rewrote some of these kind of assertion to use the sinon assertions some times ago (http://sinonjs.org/releases/v4.1.4/assertions/, e.g. an assertion in the form `assert.equal(aSinonSpy.called, true)` has been rewritten into `sinon.assert.called(aSinonSpy);`).

Let's restore the usage of the sinon assertions here (also for the other assertions that were using the sinon assertions in the previous version)

tests/unit/test-cmd/test.build.js
```diff
@@ -64,313 +56,269 @@ describe('build', () => {
...
+      assert.typeOf(args.shouldWatchFile, 'function');
+      args.shouldWatchFile('/some/path');
+      assert.equal(fileFilter.wantFile.called, true);
+
+      // Remove the built extension.
+      await fs.unlink(buildResult.extensionPath);
+      // Execute the onChange handler to make sure it gets built
+      // again. This simulates what happens when the file watcher
+      // executes the callback.
+      await args.onChange();
```
rpl: @saintsebastian based on our last exchange about this ([from this comment](https://github.com/mozilla/web-ext/pull/1009/files#r130421956)) I guess that the following version should also work:

```js
args.onChange(); // <--- execute the callback
await Promise.resolve(); // <--- "wait fot 1 tick" trick 
```

tests/unit/test-cmd/test.build.js
```diff
@@ -64,313 +56,269 @@ describe('build', () => {
...
-          return onChange();
-        })
-        .then(makeSureItFails())
-        .catch((error) => {
-          assert.include(
-            error.message,
-            'Simulate an error on the second call to packageCreator()');
-        });
+      });
+      assert.equal(onSourceChange.called, true);
```
rpl: Let's use the sinon assetions in this test (as in the previous version of the same test).

tests/unit/test-cmd/test.build.js
saintsebastian: Unfortunately this version throws `ENOENT`, so i guess not tick is happening as we'd expect

tests/unit/test-cmd/test.build.js
```diff
@@ -64,313 +56,269 @@ describe('build', () => {
...
-  });
+  it('gives the correct name to a localized extension', () => withTempDir(
+    async (tmpDir) => {
+      const buildResult = await build({
+        sourceDir: fixturePath('minimal-localizable-web-ext'),
+        artifactsDir: tmpDir.path(),
+      });
+      assert.match(buildResult.extensionPath,
+                   /name_of_the_extension-1\.0\.zip$/);
+      buildResult.extensionPath;
```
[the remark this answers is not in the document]
saintsebastian: This indeed helps a lot!

saintsebastian:
@rpl Thanks for the review! This old PR needed some love <3

rpl:
Closing as incomplete/do not merge, the issue is still technically valid (a few old test may not have been rewritten as async functions) but unfortunately this PR was branched quite some time ago and so unlikely to be resumed from this branch.

Issue #683, opened 2016-12-17
Feature request: JS API for web-ext


futpib:
JS API would make it possible to use `web-ext` in build scripts (like `gulpfile.js`) without calls to system shell.

kumar303:
Did you try it? It should work. We will be documenting this in https://github.com/mozilla/web-ext/issues/381

Try this example and let me know if it's working for you or not: https://github.com/mozilla/web-ext/issues/381#issuecomment-234089616

futpib:
Oh, ok, it's there. It's just that I required `web-ext` and saw only `main` function, and instantly jumped to conclusion that it's all API we've got.

Is there a way to require it the ES5 way though? I see how it could work If my gulpfile was compiled with webpack/browserify + babel, but not right out of the box.

```
$ node -e "require('web-ext/src/cmd/build')"
/home/futpib/code/policeman/node_modules/web-ext/src/cmd/build.js:2
import path from 'path';
^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:543:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at [eval]:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:26:33)
```

kumar303:
Oh crap, you're right. In ES5 there is currently no way to use web-ext programmatically. We should fix that.

The reason is because in ES5 you'll have to use the compiled source which indeed only exposes `main()`:
````
$  node -e "console.log(require('web-ext/dist/web-ext'))"
{ main: [Function: main] }
````
The ES6 sources are accessible but that's not helpful:
````
$  node -e "console.log(require('web-ext/src/cmd/build'))"
/Users/kumar/tmp/web-ext-api-test/node_modules/web-ext/src/cmd/build.js:2
import path from 'path';
^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:513:28)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.require (module.js:468:17)
[...]

PR #1018, opened 2017-08-01
Exposing run command from node package

chrmod:
Relates to https://github.com/mozilla/web-ext/issues/683

web-ext can be useful not only as a cli but also being integrated to other build tools.
This simple PR exposes `run` command from node module entry point, so other project can use it:
```js
const path = require('path');
const run = require('web-ext').run;

run({
  noReload: true,
  sourceDir: path.resolve('.'),
  artifactsDir: path.resolve('.', 'web-ext-artifacts'),
});
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,4 +1,5 @@
 /* @flow */
 import {main} from './program';
+import cmd from './cmd';
 
-export {main};
+export {main, cmd};


chrmod:
Hi. It makes total sense to me. Have submitted simplest patch to start the discussion.

kumar303:
This needed a new test suite and I wanted to experiment with the API a little bit. How does this work for you? https://github.com/mozilla/web-ext/pull/1028

chrmod:
closing in favor of https://github.com/mozilla/web-ext/pull/1028/files as it work as fine as this one.

PR #1019, opened 2017-08-01
feat: run command handles bootstap addons

chrmod:
on top of: https://github.com/mozilla/web-ext/pull/1018
and this PR, web-ext can be use to `run` boostrapped/system addons.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -67,7 +67,17 @@ export default async function run(
     noReload = true;
   }
 
-  const manifestData = await getValidatedManifest(sourceDir);
+  let manifestData;
+
+  try {
+    manifestData = await getValidatedManifest(sourceDir);
+  } catch (e) {
+    // in case it is not a webExtension
+    manifestData = {
+      name: '',
+      version: '',
+    };
+  }
 
   const firefoxDesktopRunnerParams = {
     // Common options.


kumar303:
Hi. Thanks for taking the time to submit a patch. We do not intend to support bootstrap add-ons with web-ext because there are too many special cases that we'd need to accommodate.

I'm going to close this issue but you can still work with us to help you solve the issue at hand. I suggest trying some experimental patches (like this one) in your local build of web-ext and see how far you get. Are you able to run the extension? What needed to change? Perhaps this can guide us in extracting some useful parts of web-ext into a new bootstrap extension runner or exposing some new things programmatically so you can write a script. 

Also, you may wish to try out [`jpm run`](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/jpm#jpm_run) which I believe supports bootstrap extensions.

chrmod:
got the web-ext `run` command working fine for bootstrap addon with those 3 commits:
* https://github.com/cliqz-oss/web-ext/commit/0c4b3e50b8ef9a51272862c6a0d023ab89b3a187
* https://github.com/cliqz-oss/web-ext/commit/adc45aa604d6d1a92e3cb852e3d9a1d55c1cabee
* https://github.com/cliqz-oss/web-ext/commit/7daa417992c7788982e3ec67e098519166e4a8b0

First two you know already, the last one makes it possible to reload the extension on code changes in case web-ext gets integrated with other build tool.

My team should start to work with it by Tomorrow, so we will get some feedback.

As changes are so minimal I would rather prefer to keep using web-ext, but if you really don't want to support bootstrap/system addons, that we can discuss, which part of code we can extract to separate package.

kumar303:
How many tests are failing with those patches? I'd have to look closer at the implications of [changing this return value](https://github.com/cliqz-oss/web-ext/commit/7daa417992c7788982e3ec67e098519166e4a8b0). 

As for [allowing invalid manifests](https://github.com/cliqz-oss/web-ext/commit/adc45aa604d6d1a92e3cb852e3d9a1d55c1cabee), this is just going to cause pain and sorrow for people who have a syntax error in their manifest. Supporting invalid WebExtension manifests (i.e. bootstrap add-ons) is going to be tricky and it will be hard to keep up support for that as the app changes. I would possibly accept a patch that allows you to use dependency injection like this:

```js
// Set up a stub in your script:
const getFakeManifest: () => ({
  name: '',
  version: '',
});
// Patch `run()` to accept the stub function like this:
run(params, {getValidatedManifest: getFakeManifest})
  .then(...);
```

chrmod:
there was no failing test after changing run return type. This PR makes the change and adds some tests https://github.com/mozilla/web-ext/pull/1027/

Can we also have this one merged https://github.com/mozilla/web-ext/pull/1018 ?

PR #1027, opened 2017-08-03
feat: cmd run returns extension runner

chrmod:

PR #1028, opened 2017-08-03
feat: You can now use web-ext programatically, e.g. `webExt.cmd.run()`

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/683

This only supports minimal programmatic use. You can just execute raw functions without any argument validation, like:

```js
import webExt from 'web-ext';

webExt.cmd.run({shouldExitProgram: false, firefox: '...', sourceDir: '...'})
  .then((extensionRunner) => {
    console.log(extensionRunner);
  });
```

One would have to set up verbose logging manually:

```js
webExt.util.logger.consoleStream.makeVerbose();
webExt.cmd.run({shouldExitProgram: false, firefox: '...', sourceDir: '...'});
```

You can also disable the use of standard input:

```js
webExt.cmd.run({shouldExitProgram: false, noInput: true});
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,5 +1,10 @@
 /* @flow */
 import {main} from './program';
 import cmd from './cmd';
+import * as logger from './util/logger';
 
-export default {main, cmd};
+// This only exposes util/logger so far.
+// Do we need anything else?
+const util = {logger};
+
+export default {main, cmd, util};
--- a/src/main.js
+++ b/src/main.js
@@ -1,4 +1,5 @@
 /* @flow */
 import {main} from './program';
+import cmd from './cmd';
 
-export {main};
+export default {main, cmd};
--- a/bin/web-ext
+++ b/bin/web-ext
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
 
+var webExt = require('../dist/web-ext').default;
 var path = require('path');
 var absolutePackageDir = path.join(path.resolve(__dirname), '..');
-require('../dist/web-ext').main(absolutePackageDir);
+
+webExt.main(absolutePackageDir);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -175,6 +175,7 @@ export type BuildCmdParams = {|
   sourceDir: string,
   artifactsDir: string,
   asNeeded?: boolean,
+  noInput?: boolean,
   overwriteDest?: boolean,
   ignoreFiles?: Array<string>,
 |};
@@ -196,6 +197,7 @@ export default async function build(
     asNeeded = false,
     overwriteDest = false,
     ignoreFiles = [],
+    noInput = false,
   }: BuildCmdParams,
   {
     manifestData,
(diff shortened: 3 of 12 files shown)


kumar303:
> not sure why the logger is needed. 

I was thinking that the caller may want to have some control over the log output. The could also replace `webExt.util.logger.consoleStream.write` with a no-op if desired.

> should there be any doc for it?

Yeah, definitely. The way we do it is once the issue lands and web-ext gets deployed with the change, we document it [on MDN](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext).

kumar303:
... although, hmm, maybe the README should have some docs since this usage is more developer focused.

chrmod:
if there would be a way to silence/capture the logs through `logger` object, it is definitely very useful for embedding purposes.   

Dev oriented docs could be published at readthedocs, what do you think?

kumar303:
> Dev oriented docs could be published at readthedocs, what do you think?

kumar303:
> Dev oriented docs could be published at readthedocs, what do you think?

I don't want to maintain and manage another service. I think we can try it in our current documentation and see how hard it is.

src/main.js
```diff
@@ -1,4 +1,10 @@
 /* @flow */
 import {main} from './program';
+import cmd from './cmd';
+import * as logger from './util/logger';
 
-export {main};
+// This only exposes util/logger so far.
+// Do we need anything else?
```
rpl: @kumar303 :+1: Nice! This is pretty neat!

The change seems ok to me, but I'm wondering if we should make some tweak to the defaultReloadStrategy, mostly to make the caller of the API able to disable or customize the "interactive" behavior (e.g. reading from stdin when used as a library doesn't seem a great idea, it is probably something that the caller may want to easily disable and maybe manage with an API object which allow it to know when a change has been detected, trigger the addon reload and exit the runner).

What do you think?

kumar303:
Thanks @rpl , good idea about disabling input. Could you take another look? I added `noInput` as a global param to all commands but I didn't hook it up in yargs yet. We could do that in a separate patch so that we have a global `--no-input` option.

rpl:
@kumar303 :+1: looks great

Let's file the following as followups:
- expose noInput through yargs (maybe also aliased to `'--no-interactive'`?)
- return a useful API object from cmd.run (e.g. an object which provides a `reload()` method which reloads the extension which rejects if the reload fails, with the same error logged in the console, and an `exit()` method which exits the managed Firefox instance and the watcher)

kumar303:
> return a useful API object from cmd.run

It's currently return `extensionRunner` which already does what you're proposing, right? I see methods like `reloadAllExtensions()` and `exit()`.

kumar303:
Feature request for `--no-input`: https://github.com/mozilla/web-ext/issues/1030

rpl:
> It's currently returning extensionRunner which already does what you're proposing, right? I see methods like reloadAllExtensions() and exit().

currently we are returning `firefoxApp`, which was what we used to return before the refactoring,
but I agree that returning the `extensionRunner` would do it.

It seems to be what #1027 is aiming to, though, and so there is no need to file another issue for it.

PR #1033, opened 2017-08-08
chore: Upgrade linter and adjust eslint rules

kumar303:
Closes #992 

This upgrades the linter but also adjusts some eslint rules. For some mysterious reason, the linter triggered eslint errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,19 @@
 {
   "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "arrowFunctions": true,
+      "blockBindings": true,
+      "classes": true,
+      "destructuring": true,
+      "defaultParams": true,
+      "modules": true,
+      "restParams": true,
+      "spread": true
+    },
+  },
   "plugins": [
     "async-await",
     "flowtype",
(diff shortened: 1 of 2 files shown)


kumar303:
@mstriemer I made some progress but I'm a bit stumped on the new lint errors. It looks like our `flowtype` plugin is not getting applied somehow. We have specific rules like [`define-flow-type`](https://github.com/gajus/eslint-plugin-flowtype#define-flow-type) which don't seem to be taking affect.

mstriemer:
Hmm, a lot of these seem like eslint isn't understanding the es6 code properly. Not necessarily related to flow.

kumar303:
The errors I looked at were all for Flow types (not valid javascript). The [define-flow-type](https://github.com/gajus/eslint-plugin-flowtype#define-flow-type) rule is supposed to suppress those. It says: "Used to suppress no-undef reporting of type identifiers."

mstriemer:
Looks like the errors are still somehow caused by addons-linter. Leaving it at 0.22.3 passes the build.

mstriemer:
Looks like the errors start with https://github.com/mozilla/addons-linter/compare/0.20.3...0.20.4 which has dispensary updating eslint to version 4, which is in its dependencies likely by accident. I filed https://github.com/mozilla/dispensary/pull/398 to move that to devDependencies.

Should figure out how to test if that fixes it locally, but either way that change should likely go out.

kumar303:
Thanks. The eslint 4 dep sounds suspiciously related but still weird.

mstriemer:
Updating to addons-linter@0.22.8 fixes this for me locally. I think we should still get the Firefox 56 schemas in an addons-linter release before we cut a new web-ext, though.

kumar303:
@mstriemer thanks for your help on this! Give me a ping when the schemas are released and we'll cut a release.

PR #1035, opened 2017-08-10
Logger: listeners

chrmod:
I'm  using web-ext to start Firefox in continuous integration and run tests inside.

One very straightforward way to report test results is to print them to stdout with `dump`. But as Firefox tends to output quite a lot,  there should be a way to filter the logs before they hit stdout.

That is how it works, with a simple wrapper program:
```js
const webExt = require('web-ext');

const logStream = webExt.util.logger.consoleStream

logStream.makeVerbose();
logStream.startCapturing();

logStream.addListener(msg => {
  const prefix = '[firefox/index.js][debug] Firefox stdout: TAP:  ';
  if (msg.indexOf(prefix) === 0) {
    const log = msg.substr(prefix.length).replace(/\nTAP:  /g, '\n');
    log.split('\n').forEach(m => console.log(m));
  }
  logStream.clear();
});

webExt.run();
```

I don't really like the `clear` method, but there has to be a way to avoid duplicated messages. In this use case I need log streaming.

If this makes sense, please help me refine it and I will cover it with tests.

### DIFF
--- a/src/util/logger.js
+++ b/src/util/logger.js
@@ -43,11 +43,13 @@ export class ConsoleStream {
   verbose: boolean;
   isCapturing: boolean;
   capturedMessages: Array<string>;
+  listeners: Set<(string) => void>;
 
   constructor({verbose = false}: ConsoleStreamParams = {}) {
     this.verbose = verbose;
     this.isCapturing = false;
     this.capturedMessages = [];
+    this.listeners = new Set();
   }
 
   format({name, msg, level}: BunyanLogEntry): string {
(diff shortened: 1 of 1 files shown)


kumar303:
Can't you already just hijack the write method? Maybe with something like this:

```js
const webExt = require('web-ext');

const logStream = webExt.util.logger.consoleStream
logStream.write = (packet) => {
  process.stdout.write(`hijacked: ${logStream.format(packet)}`);
};

webExt.run();
```

chrmod:
that makes total sense. thanks

PR #1036, opened 2017-08-10
chore(package): update webpack to version 3.5.3

kumar303:
Closes #967


kumar303:
I tested this locally. Looks like it's working. They say there aren't any breaking changes unless you develop plugins, which we don't.

PR #1037, opened 2017-08-10
chore(package): update sinon to version 3.2.0

kumar303:
Closes #1025


kumar303:
I checked each breaking change and I don't think any apply to us

PR #1038, opened 2017-08-10
fix(package): update addons-linter to version 0.22.8

kumar303:
Closes #992 
Closes #1033


kumar303:
oops, I forgot we need a change in web-ext's eslintrc. See https://github.com/mozilla/web-ext/pull/1033

PR #1040, opened 2017-08-12
Fixes #1030

### LINKED ISSUE #1030: Add --no-input option to disable standard input

feature

`web-ext run` will allow you to press keys (such as `R`) to trigger a reload.

It should be possible to disable this if desired with something like `web-ext run --no-input`.

Adding this as a [global option](https://github.com/mozilla/web-ext/blob/e48ac75745b4d25aade4290b97401550d2fd4ce3/src/program.js#L248) should be really easy because the `noInput` parameter was already added in https://github.com/mozilla/web-ext/pull/1028

serendicoder:


src/program.js
```diff
@@ -276,6 +276,11 @@ Example: $0 --help run.
       requiresArg: true,
       type: 'array',
     },
+    'no-input': {
+      alias: 'no-interactive',
```
kumar303: I don't think we need to add an alias. You can remove this. Even though `--no-input` isn't perfect, it's concise. The less aliases we have the better.

src/program.js
```diff
@@ -276,6 +276,11 @@ Example: $0 --help run.
...
+      describe: 'Disable the interactive behaviour',
```
kumar303: Let's make this more specific to what it will do. How about this:

Disable all features that require standard input

serendicoder: I have incorporated the changes. Please let me know if this works for you.

serendicoder:
Thanks @kumar303 . I had to create a new PR [here](https://github.com/mozilla/web-ext/pull/1044) because I had amended the old commit and then rebased with master which created some issues on local repo.

PR #1059, opened 2017-08-24
chore: sinon assertions for test.program.js and test.lint.js

Jp-Rivera:
@kumar303 Here are some more sinon assertions.


kumar303:
Thanks JP. The diff is showing unrelated commits. I think it's because you haven't pushed your master branch to your forked repository? Could you try doing that? I think `git push origin master` should do it as long as `origin` is `Jp-Rivera`.

Jp-Rivera:
Thanks! The push removed the unrelated commits. Is there a way to fix the commit message, or will I need to make another commit?

PR #1060, opened 2017-08-24
fix: BOM encoding fix

### LINKED ISSUE #1013: web-ext build/run command does not work with manifest.json in utf-8 with BOM encoding
bug

When running the command "web-ext build" or "web-ext run" in an extension folder where the manifest.json file has the encoding of utf-8 with BOM, it returns an error saying
InvalidManifest: Error parsing manifest.json at manifest.json: JSONError: Unexpected token '?' at 1:1 in manifest.json
?{
^
( I have omitted the absolute paths in the error message)

The commands work as normal when the manifest.json has ANSI encoding.

I think "web-ext build" and "web-ext run" should work as normal when the manifest.json is in utf-8 with BOM encoding.

* **Firefox version**: 54.0.1 (64-bit)
* **Your OS and version**: Windows 7 Home Premium SP1

````
````
v8.2.1
5.3.0
master-19bdd384191dbeb51b7376feb3ce47e01347977e

### PR DESCRIPTION
**Pull request by tsl143**
Fixes #1013

### DIFF
--- a/src/util/json.js
+++ b/src/util/json.js
@@ -1,0 +1,23 @@
+/* @flow */
+import parseJSONLib from 'parse-json';
+
+import {InvalidJSON} from '../errors';
+
+export function parseJSON(jsonContents: Object, jsonFile: string) {
+  try {
+    jsonContents = stripBOM(jsonContents);
+  } catch (error) {
+    throw new InvalidJSON(
+      `Could not read json file at ${jsonFile}: ${error}`);
+  }
+
+  return parseJSONLib(jsonContents, jsonFile);
+}
+
+function stripBOM(content) {
+  content = content.toString();
+  if (content.charCodeAt(0) === 0xFEFF) {
+    content = content.slice(1);
+  }
+  return content;
+}
--- a/src/util/manifest.js
+++ b/src/util/manifest.js
@@ -46,6 +46,7 @@ export default async function getValidatedManifest(
   let manifestData;
 
   try {
+    manifestContents = stripBOM(manifestContents);
     manifestData = parseJSON(manifestContents, manifestFile);
   } catch (error) {
     throw new InvalidManifest(
@@ -82,4 +83,12 @@ export default async function getValidatedManifest(
 export function getManifestId(manifestData: ExtensionManifest): string | void {
   return manifestData.applications ?
     manifestData.applications.gecko.id : undefined;
+}
+
+function stripBOM(content) {
+  content = content.toString();
+  if (content.charCodeAt(0) === 0xFEFF) {
+    content = content.slice(1);
+  }
+  return content;
 }
--- a/src/util/manifest.js
+++ b/src/util/manifest.js
@@ -2,8 +2,8 @@
 import path from 'path';
 
 import {fs} from 'mz';
-import parseJSON from 'parse-json';
 
+import {parseJSON} from './json';
 import {InvalidManifest} from '../errors';
 import {createLogger} from './logger';
 
@@ -46,7 +46,6 @@ export default async function getValidatedManifest(
   let manifestData;
 
   try {
-    manifestContents = stripBOM(manifestContents);
     manifestData = parseJSON(manifestContents, manifestFile);
   } catch (error) {
     throw new InvalidManifest(
@@ -83,12 +82,4 @@ export default async function getValidatedManifest(
 export function getManifestId(manifestData: ExtensionManifest): string | void {
   return manifestData.applications ?
     manifestData.applications.gecko.id : undefined;
-}
-
-function stripBOM(content) {
-  content = content.toString();
-  if (content.charCodeAt(0) === 0xFEFF) {
-    content = content.slice(1);
-  }
-  return content;
 }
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/util/manifest.js [hunk 1]
```diff
@@ -83,3 +84,11 @@ export function getManifestId(manifestData: ExtensionManifest): string | void {
   return manifestData.applications ?
     manifestData.applications.gecko.id : undefined;
 }
+
+function stripBOM(content) {
```
kumar303: This needs tests. There's a manifest with a BOM [attached here](https://github.com/mozilla/web-ext/issues/1013#issuecomment-319105758) that you can use for the tests.

src/util/manifest.js
```diff
@@ -46,6 +46,7 @@ export default async function getValidatedManifest(
   let manifestData;
 
   try {
+    manifestContents = stripBOM(manifestContents);
```
kumar303: I think it would be helpful to wrap this in its own try/catch so that the error is more specific:

```js
  try {
    manifestContents = stripBOM(manifestContents);
  } catch (error) {
    throw new InvalidManifest(
      `Error stripping BOM from manifest.json at ${manifestFile}: ${error}`);
  }
```

src/util/manifest.js
kumar303: I think we need a more general approach to stripping the BOM since it applies to more than just manifest files. I suggest searching for this:

```js
import parseJSON from 'parse-json'
```

Next, create a `web-ext` wrapper for it called `parseJSON` (the same name) and put it somewhere like `src/util/json.js`. Make that function strip the BOM from any JSON file. Replace all old calls to `parseJSON` with the new one.

tests/unit/test-util/test.json.js
```diff
@@ -0,0 +1,27 @@
...
+describe('util.parseJSON', () => {
+
+  it('returns a parsed JSON', () => {
+    const parsedJSON = parseJSON(JSON.stringify(basicManifest));
+    assert.deepEqual(parsedJSON, basicManifest);
+  });
+
+  it('returns a parsed JSON from file with BOM Chanracter', () => {
+    const dummyJson = { manifest_version: 2 };
+    fs.readFile('./test.manifest.json', function read(err, data) {
```
kumar303: I'd like to see `test.manifest.json` moved to `tests/unit/test-util/data/manifest-with-bom.js`

kumar303: because you're using `fs` from `mz`, it returns a promise. You can check out other examples of `fs.readFile()` in the test suite to see how to use it as a promise.

tests/unit/test-util/test.json.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('util.parseJSON', () => {
+
+  it('returns a parsed JSON', () => {
+    const parsedJSON = parseJSON(JSON.stringify(basicManifest));
+    assert.deepEqual(parsedJSON, basicManifest);
+  });
+
+  it('returns a parsed JSON from file with BOM Chanracter', () => {
+    const dummyJson = { manifest_version: 2 };
```
kumar303: to make this a bit more clear, how about naming this `expectedJson` ?

src/util/json.js
```diff
@@ -0,0 +1,23 @@
...
+import parseJSONLib from 'parse-json';
+
+import {InvalidJSON} from '../errors';
+
+export function parseJSON(jsonContents: Object, jsonFile: string) {
+  try {
+    jsonContents = stripBOM(jsonContents);
+  } catch (error) {
+    throw new InvalidJSON(
+      `Could not read json file at ${jsonFile}: ${error}`);
```
kumar303: This needs test coverage. Actually, you can just move this test into the new `test-util/test.json.js` suite: https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.manifest.js#L53-L69

tests/unit/test-util/test.json.js
```diff
@@ -0,0 +1,27 @@
+import {it, describe} from 'mocha';
```
kumar303: Add this comment at the top to opt into Flow parsing: `/* @flow */`

tests/unit/test-util/test.json.js
```diff
@@ -0,0 +1,27 @@
...
+import {fs} from 'mz';
+
+import {parseJSON} from '../../../src/util/json';
+import {basicManifest} from '../helpers';
+
+
+describe('util.parseJSON', () => {
+
+  it('returns a parsed JSON', () => {
+    const parsedJSON = parseJSON(JSON.stringify(basicManifest));
```
kumar303: You will get a Flow error here because you aren't passing enough arguments to `parseJSON`. I think it would be better to test this with a temp file. You can see how it's done here: https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.manifest.js#L53-L69

tests/unit/test-util/test.json.js
kumar303: Typically, we write temp files in the tests but for this case I think it's ok to use a real file.

kumar303:
Hi @tsl143 . Let me know if you have any questions about how to proceed. Thanks for tackling this bug!

rpl:
Superseded by #1712

PR #1062, opened 2017-08-25
docs: Add summary to docs and fix command descriptions

kumar303:
The command descriptions now follow the new extension terminology: https://wiki.mozilla.org/Add-ons/Terminology

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,16 +17,16 @@ cross-platform way. Initially, it will provide a streamlined experience for deve
 * [Getting started with web-ext](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext)
 * [Command reference](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference)
 
-Here are the commands you can run. Click on each one for detailed documentation or use `--help` on the command line, such as `web-ext --help build`.
+Here are the commands you can run. Click on each one for detailed documentation or use `--help` on the command line, such as `web-ext build --help`.
 
-* [`build`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build)
-  * Create an extension package from source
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -17,6 +17,19 @@ cross-platform way. Initially, it will provide a streamlined experience for deve
 * [Getting started with web-ext](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext)
 * [Command reference](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference)
 
+Here are the commands you can run. Click on each one for detailed documentation or use `--help` on the command line, such as `web-ext --help build`.
```
rpl: nit, not a big deal (and I'm pretty sure that with yargs the order doesn't matter), but as a command line user I'm more used to see the opposite `web-ext build --help`

README.md
```diff
@@ -17,6 +17,19 @@ cross-platform way. Initially, it will provide a streamlined experience for deve
...
+
+* [`build`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build)
```
rpl: I'm wondering if it would be better to order them based on how much they are actually used (and useful) or just ordering them alphabetically.

If we opt for ordering them by "number of times you will actually use the command", I would say:
- run, build, lint, sign, docs
or
- run, lint, build, sign, docs

README.md
kumar303: Good call. That's more intuitive too since you may start typing `web-ext run ...` and then realize you don't know the name of some option.

README.md
kumar303: Sure. The first ordering is what I would think. I'll change it to that.

kumar303: well, actually, `build` really isn't so useful. This comes to mind after more thought: run, lint, sign, build, docs

Issue #730, opened 2017-01-03
Discover config files in multiple places


kumar303:
This is a stepping stone toward implementing all of https://github.com/mozilla/web-ext/issues/176

After #744 and #1080, let's look for config files in multiple places. The load order will look like this:

1. Look for `~/.web-ext-config.js` and load that file
2. Look for `./web-ext-config.js` and load that, overwriting any previous config values
3. Look for `--config=any-file.js` and load that (implemented in https://github.com/mozilla/web-ext/issues/728), once again overwriting previous values

All discovery can be disabled with `--no-config-discovery`

PR #1080, opened 2017-09-16
feat: basic implementation of command line config option

### LINKED ISSUE #775: Configure the Program class to use config parsing helpers
This is a stepping stone toward implementing all of #176

After landing the config helper objects in https://github.com/mozilla/web-ext/pull/744, let's hook them up to the Program class so that you can get config handling working on the command line.

saintsebastian:
Fixes #775

Basic version. Some question:
1. would it be a good idea to use `./web-ext-config.js` as default value for `--config` option since it needs one?
2. is `program.js` a good place to parse other config files from expected places like `~/web-ext-config.js`and `./web-ext-config.js`?


rpl:
I took a look at what is the behavior currently implemented by the src/config.js module landed by #744 
and here is some initial feedback related to the question:

1. if I remember correctly, there was the idea of loading the config values from "./.web-ext-config.js" and "~/.web-ext-config.js" if one of these files actually exist, but (at least in my opinion) the "config loading behavior" should be sightly different in the two scenarios: 
    - when the config file option has been explicitly passed over the command line, we should fail if the file doesn't exist 
    - but if we are loading some config file implicitly (e.g. from the current directory or from the home directory), then we should not fail if the file doesn't exist (nevertheless, we should emit a warning or a debug message)

2. yes, I think that program.js is where the config loading behaviors are going to be implemented (and based on the description in #775 I think that it is where @kumar303 is suggesting to implement it) 

Another detail to take into account if we are going to load a config file implicitly is related to the default values displayed in the "web-ext --help" output, which can be different from the actual default values set once we load the config file implicitly, and it is likely to be confusing for the user
(and so we should at least make sure that we print the new loaded defaults and/or update the default values in the "web-ext --help" output).

@kumar303 Let me know how that sounds to you, e.g. if I'm not missing anything related to the plans for the "config file loading" behaviors.

kumar303:
> @kumar303 Let me know how that sounds to you

Sounds good!

I would suggesting keeping the features in this patch small. For example, a great start would be to simply allow an option like `--config-file` that takes a path to a file, that's it. There are a lot of things to consider with auto-detecting config files -- we should look in the home directory, in the current directory, and so on. These are all features that could come later.

rpl:
> Sounds good!

Yeah, I agree completely, it is better to keep this PR simpler and defer that behaviors as follow up issues.

@saintsebastian  do you mind to file this part (the "implicit config file loading") as a separate issue? (and mention #775 and this PR into its description so that they will be linked together by github).

saintsebastian:
@rpl Hi! I assumed this issue [730](https://github.com/mozilla/web-ext/issues/730) covers the file loading

rpl:
@saintsebastian yeah, it is definitely the issue that I wanted, it was linked to a pull request that has been closed in favor of the one that has been landed, I've also updated its description to mention this PR.

src/program.js
```diff
@@ -132,6 +133,20 @@ export class Program {
...
+    if (argv.config) {
+      const configObject = loadJSConfigFile(argv.config);
+      try {
+        applyConfigToArgv({
+          argv,
+          configObject,
+          options: this.options,
+          configFileName: argv.config,
+        });
+      } catch (error) {
```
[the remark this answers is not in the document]
saintsebastian: @rpl tbh I wrapped it this way because instead of throwing i would get `UnhandledPromiseRejectionWarning` despite `applyConfigToArgv` being a sync function.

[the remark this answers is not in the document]
rpl: @saintsebastian that's because the `execute` method is an async function, it should be enough to just move the code that uses `loadJSConfigFile` and `applyConfigToArgv` inside the existent `try...catch` at line 150 of this patch (it was at line 135 before the changes applied by this patch), because it already implements the behavior that we want.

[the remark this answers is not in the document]
saintsebastian: @rpl ah! now i see it. thanks! I will add tests and that should probably be it for this patch

tests/unit/test.program.js
```diff
@@ -430,6 +430,38 @@ describe('program.main', () => {
...
+        commands: fakeCommands,
+        runOptions: {
+          applyConfigToArgv: fakeApplyConfigToArgv,
+          loadJSConfigFile: fakeLoadJSConfigFile,
+        },
+      })
+      .then(() => {
+        const options = fakeCommands.lint.firstCall.args[0];
+        assert.strictEqual(options.config, fakePath);
+        assert.ok(fakeLoadJSConfigFile.called);
```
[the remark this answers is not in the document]
rpl: we should probably also add some additional test cases to cover the following scenarios:

- the mocked loadJSConfigFile raises a UsageError or an unexpected exception
- the mocked applyConfigToArgv raises a UsageError or a WebExtError or an unexpected exception

And check that the behavior is the one that we expect (e.g. we can assert that these errors should make the call to reject with the expected error, and we could also check that they have been printed on the stderr, e.g. by injecting a logStream defined as `const logStream = fake(new ConsoleStream());` into `execProgram`)

saintsebastian:
@rpl this is now ready for a review. I think new tests are very repetitive, so maybe you have some ideas how to make them better. Also there is Travis failure on this and my other pr, it seems to be npm related.

tests/unit/test.program.js
```diff
@@ -430,6 +433,203 @@ describe('program.main', () => {
...
+        sinon.assert.calledOnce(fakeApplyConfigToArgv);
+        sinon.assert.calledWith(fakeLoadJSConfigFile,
+          sinon.match(resolvedFakePath));
+        sinon.assert.calledWith(fakeApplyConfigToArgv, sinon.match({
+          configFileName: resolvedFakePath,
+        }));
+      });
+  });
+
+  it('logs and throws when config file was not loaded', () => {
```
rpl: how about changing the description into something like 'throws an UsageError when the config file can't be loaded'?

tests/unit/test.program.js
```diff
@@ -430,6 +433,203 @@ describe('program.main', () => {
...
+      .then(makeSureItFails())
+      .catch(onlyInstancesOf(UsageError, (error) => {
+        assert.match(error.message, /bad path/);
+        sinon.assert.notCalled(fakeApplyConfigToArgv);
+        sinon.assert.notCalled(fakeProcess.exit);
+        sinon.assert.calledOnce(logger.error);
+      }));
+  });
+
+  it('throws when config file loading fails', () => {
```
rpl: to make it a bit more clear which is the difference between this test and the previous one,
we could rephrase this description to "throws when config file loading raises an unexpected error" 
(given that we only expect it to raise a `UsageError` in the known error cases)

tests/unit/test.program.js
```diff
@@ -430,6 +433,203 @@ describe('program.main', () => {
...
+        if (!(error instanceof UsageError)) {
+          assert.match(error.message, /some error/);
+          sinon.assert.notCalled(fakeApplyConfigToArgv);
+          sinon.assert.notCalled(fakeProcess.exit);
+          sinon.assert.calledOnce(logger.error);
+        }
+      });
+  });
+
+  it('logs and throws when fakeApplyConfigToArgv throws UsageError', () => {
```
rpl: how about "throws a UsageError when the loaded config can't be applied"?

tests/unit/test.program.js
```diff
@@ -430,6 +433,203 @@ describe('program.main', () => {
...
+      })
+      .then(makeSureItFails())
+      .catch(onlyInstancesOf(UsageError, (error) => {
+        assert.match(error.message, /bad config option/);
+        sinon.assert.notCalled(fakeProcess.exit);
+        sinon.assert.calledOnce(logger.error);
+      }));
+  });
+
+  it('logs and throws when fakeApplyConfigToArgv throws WebExtError', () => {
```
rpl: @saintsebastian I'm wondering if we are really supposed to throw a WebExtError when a config file option is missing the type:

https://github.com/mozilla/web-ext/blob/177709b59137b74ca933014e843894b169be6487/src/config.js#L52-L55

while we are raising a UsageError when the specified type is incorrect:

https://github.com/mozilla/web-ext/blob/177709b59137b74ca933014e843894b169be6487/src/config.js#L61-L65

it seems to me that we could/should throw a UsageError for both, @kumar303 what do you think about it?
(e.g. I'm wondering if this has been already evaluated when the config helper module has been created)

tests/unit/test.program.js
```diff
@@ -430,6 +433,203 @@ describe('program.main', () => {
...
+      })
+      .then(makeSureItFails())
+      .catch(onlyInstancesOf(WebExtError, (error) => {
+        assert.match(error.message, /bad config option/);
+        sinon.assert.notCalled(fakeProcess.exit);
+        sinon.assert.calledOnce(logger.error);
+      }));
+  });
+
+  it('logs and throws when fakeApplyConfigToArgv throws WebExtError', () => {
```
rpl: This seems to be testing mostly the same scenario of the test above it, but it should probably be the test case which explicitly test what is going to happen if "fakeApplyConfigToArgv throws an unexpected error", am I right?

saintsebastian:
Hey @rpl sorry it took me so long to get to this PR
1. I tried using `consoleStreeam.startCapturing()`, this however still only works with injected `logger` dependency.
2. about your concern on use of `WebExtError` I added the last commit with the changes you suggested, which we can always throw away in case @kumar303 thinks there is a good reason to use this type of error.

Thanks for cleaning up my wording and my tests!

tests/unit/test.program.js
```diff
@@ -430,6 +432,190 @@ describe('program.main', () => {
...
+    const fakeProcess = createFakeProcess();
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    const fakePath = 'path/to/web-ext-config.js';
+    const fakeLoadJSConfigFile = sinon.spy(() => {
+      throw new UsageError('bad path');
+    });
+    const fakeApplyConfigToArgv = sinon.spy(() => {});
+    const capturingConsoleStream = new ConsoleStream();
```
rpl: @saintsebastian The `ConsoleStream` instance used by the loggers created using the `createLogger` helper function is also exported by the `"util/logger"` module (as `consoleStream`: https://github.com/mozilla/web-ext/blob/master/src/util/logger.js#L94).

Here is a diff applied to this test to make it able to use the default logger instance to make the assertions on the logged messages:

```diff
diff --git a/tests/unit/test.program.js b/tests/unit/test.program.js
index ac73535..ee1a32e 100644
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -19,7 +19,10 @@ import {
   makeSureItFails,
   ErrorWithCode,
 } from './helpers';
-import {createLogger, ConsoleStream} from '../../src/util/logger';
+import {
+  createLogger,
+  consoleStream, ConsoleStream,
+} from '../../src/util/logger';

 describe('program.Program', () => {

@@ -474,11 +477,11 @@ describe('program.main', () => {
       throw new UsageError('bad path');
     });
     const fakeApplyConfigToArgv = sinon.spy(() => {});
-    const capturingConsoleStream = new ConsoleStream();
-    capturingConsoleStream.startCapturing();
-    const logger = createLogger(__filename, {
-      logStream: capturingConsoleStream,
-    });
+
+    // Empty any existent capturedMessages.
+    consoleStream.stopCapturing();
+    // Start collecting capturedMessages.
+    consoleStream.startCapturing();

     return execProgram(
       ['lint', '--config', fakePath],
@@ -489,13 +492,16 @@ describe('program.main', () => {
           systemProcess: fakeProcess,
           applyConfigToArgv: fakeApplyConfigToArgv,
           loadJSConfigFile: fakeLoadJSConfigFile,
-          logger,
         },
       })
       .then(makeSureItFails())
       .catch(onlyInstancesOf(UsageError, (error) => {
+        const {capturedMessages} = consoleStream;
+        // empty any existent capturedMessages and stop capturing further logs.
+        consoleStream.stopCapturing();
+
         assert.match(error.message, /bad path/);
-        assert.match(capturingConsoleStream.capturedMessages[0], /bad path/);
+        assert.match(capturedMessages[0], /bad path/);
         sinon.assert.notCalled(fakeApplyConfigToArgv);
         sinon.assert.notCalled(fakeProcess.exit);
```

saintsebastian: ah! i missed this somehow

src/program.js
```diff
@@ -26,7 +30,9 @@ type ExecuteOptions = {
   checkForUpdates?: Function,
   systemProcess?: typeof process,
   logStream?: typeof defaultLogStream,
-  getVersion?: Function,
+  getVersion?: VersionGetterFn,
```
rpl: @saintsebastian :heart: thank you so much! this type annotation is way better!

src/program.js
```diff
@@ -176,6 +195,8 @@ type VersionGetterOptions = {
   globalEnv?: string,
 };
 
+export type VersionGetterFn = (absolutePackageDir: string) => string;
```
rpl: @saintsebastian nit: do you mind to move this flow type definition before the first usage at line 33?

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    const fakePath = 'path/to/web-ext-config.js';
+    const fakeLoadJSConfigFile = sinon.spy(() => {});
+    const fakeApplyConfigToArgv = sinon.spy(() => {
+      throw new Error('some error');
+    });
+
+    consoleStream.stopCapturing();
```
rpl: @saintsebastian nit: we should probably add a comment near here to explain that:

```js
consoleStream.stopCapturing();
```

are used here to ensure that we empty any captured log.

src/util/logger.js
```diff
@@ -110,12 +110,16 @@ export type CreateBunyanLogParams = {|
 export type CreateBunyanLogFn = (params: CreateBunyanLogParams) => Logger;
 
 export type CreateLoggerOptions = {|
-  createBunyanLog: CreateBunyanLogFn,
+  createBunyanLog?: CreateBunyanLogFn,
+  logStream?: ConsoleStream,
```
rpl: @saintsebastian it looks that the changes applied on this file are not necessary anymore to make our assertions on the produced logs, am I right? in that case we can remove these changes from this PR.

saintsebastian:
Thanks for your patience and thoughtful reviews :) hope it's a step towards full config implementation

kumar303:
This is exciting! When it's ready for me to take a second look just request a review so it shows up on my github dashboard.

tests/unit/test.config.js
```diff
@@ -657,8 +657,8 @@ describe('config', () => {
       };
       assert.throws(() => {
         applyConf({...params, configObject});
-      }, WebExtError,
-        'WebExtError: Option: apiUrl was defined without a type.');
+      }, UsageError,
+        'UsageError: Option: apiUrl was defined without a type.');
     });
 
     it('throws an error when the type of one of them is in config' +
```
rpl: @saintsebastian This test seems basically the same of the above test, I see that you have only changed the expected error here, but given that you looked into this test recently I'm wondering if you could check if I'm right and this test case is redundant.

src/util/logger.js
```diff
@@ -110,7 +110,7 @@ export type CreateBunyanLogParams = {|
 export type CreateBunyanLogFn = (params: CreateBunyanLogParams) => Logger;
 
 export type CreateLoggerOptions = {|
-  createBunyanLog: CreateBunyanLogFn,
+  createBunyanLog?: CreateBunyanLogFn,
```
rpl: @saintsebastian is this change to the type annotation needed?

tests/unit/test.program.js
```diff
@@ -9,15 +9,20 @@ import {assert} from 'chai';
...
 import {
   createFakeProcess,
   fake,
   makeSureItFails,
   ErrorWithCode,
 } from './helpers';
-import {ConsoleStream} from '../../src/util/logger';
-
+import {
+  consoleStream,
```
rpl: @saintsebastian nit, an inline comment here could be useful here (just to confirm to the next developer that reads this test file that we are importing both consoleStream and ConsoleStream on purpose, one to be able to inspect the messages logged by the logger instance created internally in the tested module and the other to create new fake console streams)

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
+        runOptions: {
+          shouldExitProgram: false,
+          systemProcess: fakeProcess,
+          applyConfigToArgv: fakeApplyConfigToArgv,
+          loadJSConfigFile: fakeLoadJSConfigFile,
+        },
+      })
+      .then(makeSureItFails())
+      .catch((error) => {
+        if (!(error instanceof UsageError)) {
```
rpl: @saintsebastian This should be:

```js
const {capturedMessages} = consoleStream;
consoleStream.stopCapturing();

if (!(error instanceof UsageError)) {
  throw error;
}

assert.match(error.message, /some error/);
...
```

to be sure that the test still fails if an UsageError is received here.

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
```
rpl: same as the previous comment (at line 539)

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
+    consoleStream.stopCapturing();
+    consoleStream.flushCapturedLogs();
+    consoleStream.startCapturing();
+
+    return execProgram(
+      ['lint', '--config', fakePath],
+      {
+        commands: fakeCommands,
+        runOptions: {
+          shouldExitProgram: false,
```
rpl: This seems that it should be `shouldExitProgram: true` and then we should assert `sinon.assert.calledOnce(fakeProcess.exit);` to be sure that web-ext return an exit code 1 in this scenario.

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
```
rpl: This seems that it should be `shouldExitProgram: true` and then we should assert `sinon.assert.calledOnce(fakeProcess.exit);` to be sure that web-ext return an exit code 1 in this scenario.

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
```
rpl: This seems that it should be `shouldExitProgram: true` and then we should assert `sinon.assert.calledOnce(fakeProcess.exit);` to be sure that web-ext return an exit code 1 in this scenario.

tests/unit/test.program.js
```diff
@@ -430,6 +435,203 @@ describe('program.main', () => {
...
```
rpl: This seems that it should be `shouldExitProgram: true` and then we should assert `sinon.assert.calledOnce(fakeProcess.exit);` to be sure that web-ext return an exit code 1 in this scenario.

saintsebastian:
@rpl just to clarify, you would want `shouldExitProgram: true` and not to make any error assertions because of [that](https://github.com/saintsebastian/web-ext/blob/af2e9a9654ede0fd5d6a818509f9f4f2e8405904/src/program.js#L183-L187)?

rpl:
@saintsebastian no, actually you are right, we already have a test for that behavior at line 70 (`it exits 1 on a thrown error`), and so the only tweak needed here are the ones related to these two review comments https://github.com/mozilla/web-ext/pull/1080#discussion_r147722064 and https://github.com/mozilla/web-ext/pull/1080#discussion_r147723553

saintsebastian:
@rpl thanks for your reviews! hope this is ready for last reviews

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+    });
+    const fakePath = 'path/to/web-ext-config.js';
+    const fakeLoadJSConfigFile = sinon.spy(() => {
+      throw new UsageError('bad path');
+    });
+    const fakeApplyConfigToArgv = sinon.spy(() => {});
+
+    consoleStream.stopCapturing();
+    consoleStream.flushCapturedLogs();
+    consoleStream.startCapturing();
```
kumar303: This is interesting. I see what you're doing here but I'd like to see it moved only to a single test instead of multiple tests. The test should be an assertion of how `Program()` will log *any* `UsageError` to the console. Actually, I was surprised that we did not have test coverage of this already. Good catch.

After you add a single test for this logic, your assertion of the `UsageError` messages (like `assert.match(error.message, /bad path/)`) will be enough. This is because we will know a thrown `UsageError` will always be logged -- you don't have to test the console stream every time.

Also, you should add a comment around the stream capturing/re-capturing logic because it's not obvious that it is working around the test suite itself.

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+        },
+      })
+      .then(() => {
+        const options = fakeCommands.lint.firstCall.args[0];
+        assert.strictEqual(options.config, fakePath);
+        sinon.assert.calledOnce(fakeLoadJSConfigFile);
+        sinon.assert.calledWith(fakeLoadJSConfigFile,
+                                sinon.match(resolvedFakePath));
+        sinon.assert.calledOnce(fakeApplyConfigToArgv);
+        sinon.assert.calledWith(fakeApplyConfigToArgv, sinon.match({
```
kumar303: I think it would be better to check all parameters for the call to `fakeApplyConfigToArgv` because Flow isn't going to catch all configuration problem, especially the `options` parameter.

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+          configFileName: resolvedFakePath,
+        }));
+      });
+  });
+
+  it('throws a UsageError when the config file can\'t be loaded', () => {
```
kumar303: I don't think you need this test because the behavior of handling thrown errors is already well defined. I can see that perhaps you wanted to be sure that the config checking code is executed from within the catch block but I don't think it's necessary to cover that logic. If one day we encounter a bug where someone accidentally moved it outside of the catch block then we could add a test for it. Until then, I think code review is a better way to make sure the config logic is executed with the catch block.

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+        assert.ok(
+          capturedMessages.some(
+            (message) => message.match(/bad path/)
+        ));
+        sinon.assert.notCalled(fakeApplyConfigToArgv);
+        sinon.assert.notCalled(fakeProcess.exit);
+      }));
+  });
+
+  it('throws when config file loading raises an unexpected error', () => {
```
kumar303: Similar to my last comment, I also don't think you need this test.

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+        assert.match(error.message, /some error/);
+        assert.ok(capturedMessages.some(
+          (message) => message.match(/some error/)
+        ));
+        sinon.assert.notCalled(fakeApplyConfigToArgv);
+        sinon.assert.notCalled(fakeProcess.exit);
+      });
+  });
+
+  it('throws a UsageError when the loaded config can\'t be applied', () => {
```
kumar303: Same here :) I say delete the test.

tests/unit/test.program.js
```diff
@@ -430,6 +435,205 @@ describe('program.main', () => {
...
+        assert.match(error.message, /bad config option/);
+        assert.ok(
+          capturedMessages.some(
+            (message) => message.match(/bad config option/)
+        ));
+        sinon.assert.notCalled(fakeProcess.exit);
+      }));
+  });
+
+  it('throws when fakeApplyConfigToArgv throws an unexpected error', () => {
```
kumar303: This can be deleted. I don't think it's helpful to check that an error is thrown.

src/config.js
```diff
@@ -50,7 +50,7 @@ export function applyConfigToArgv({
         `an unknown option: "${option}"`);
     }
     if (options[decamelizedOptName].type === undefined) {
-      throw new WebExtError(
+      throw new UsageError(
```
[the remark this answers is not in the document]
rpl: @kumar303 ah, you are definitely right, and this change is my fault, it is related to this comment https://github.com/mozilla/web-ext/pull/1080#discussion_r141647124 (Sorry @saintsebastian).

Nevertheless, to prevent this check from be read wrong in the future, how about adding an inline comment near this `throw` to briefly explain what this error is about?

Another interesting option (I can file a bug so that we can handle it in a follow up) would be to use a more specific type in the flow annotations here:

https://github.com/mozilla/web-ext/blob/a55ee4c9ef59c590f0016be7984ac7ca63b27ea0/src/program.js#L96

so that we can make flow to raise an error during its static analysis if any of the options defined in program.js is missing the type property.

tests/unit/test.program.js
saintsebastian: I assume `options` will always be the same (set in `program.js`), but i added a check for `argv`.

tests/unit/test.program.js
saintsebastian: That makes a lot of sense. I was not sure what source of error should be, so I made command itself throw, let me know if it should be a dependency.

saintsebastian:
@rpl @kumar303 Thanks for the review! Let me know if something else is missing

src/config.js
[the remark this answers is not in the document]
kumar303: @rpl good idea about defining the shape of `options`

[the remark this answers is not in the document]
saintsebastian: Is this PR a good place to add that type? @kumar303 @rpl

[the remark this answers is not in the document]
rpl: @kumar303 @saintsebastian we could definitely define and add the additional flow types for the options in this PR, the amount of changes should not be that big, but it could be reasonable to handle it in as a separate issue and pull requests, so that we are free to discuss the details of it without blocking this PR.

In the meantime, I added an issue related to it as #1144 (which also contains some additional info about the change, e.g. where these new flow types would be used and what their flow type definitions could probably look like).

saintsebastian:
@kumar303 I fixed one broken test, but didn't notice any other errors so far, please let me know if you see any inconsistencies.

src/program.js
```diff
@@ -145,7 +155,19 @@ export class Program {
         });
       }
 
-      await runCommand(argv, {shouldExitProgram});
+      let argvFromConfig = { ...argv };
+      if (argv.config) {
+        const configFileName = path.resolve(argv.config);
+        const configObject = loadJSConfigFile(configFileName);
+        argvFromConfig = applyConfigToArgv({
```
kumar303: Thanks for correcting this. It also needs test coverage. When I comment out your fix, the tests all pass -- at least one should fail.

kumar303:
How are you testing it out? I made a config file like this:

```js
module.exports = {
  sourceDir: 'extension',
  run: {
    firefox: 'firefoxdeveloperedition',
  },
};
```

When I execute it inside [sound-control](https://github.com/kumar303/sound-control), I hit a couple problems with the `sourceDir` option. My command:
```
web-ext run --verbose -c web-ext-config.js
```

For sanity, you can run the same thing without a config and watch it succeed:
```
web-ext run --verbose -s extension -f firefoxdeveloperedition
```

The first thing I am noticing is that option recursion is not implemented right. In this case, when `{run: { firefox: {...}}}` is parsed, the `newArgv.sourceDir` value gets erased. This patch fixes it but try to write a failing test before adding this patch:

```diff
diff --git a/src/config.js b/src/config.js
index edcf65b..4aecd70 100644
--- a/src/config.js
+++ b/src/config.js
@@ -36,7 +36,7 @@ export function applyConfigToArgv({
       typeof configObject[option] === 'object') {
       // Descend into the nested configuration for a sub-command.
       newArgv = applyConfigToArgv({
-        argv,
+        argv: newArgv,
         configObject: configObject[option],
         options: options[option],
         configFileName});
```

The next thing I noticed is that option coercion needs to be implemented. Take a look at the [`coerce`](https://github.com/mozilla/web-ext/blob/master/src/program.js#L240) callback for `sourceDir`. The config file needs to invoke these somehow, like:

```diff
diff --git a/src/config.js b/src/config.js
index edcf65b..701b27a 100644
--- a/src/config.js
+++ b/src/config.js
@@ -36,7 +36,7 @@ export function applyConfigToArgv({
       typeof configObject[option] === 'object') {
       // Descend into the nested configuration for a sub-command.
       newArgv = applyConfigToArgv({
-        argv,
+        argv: newArgv,
         configObject: configObject[option],
         options: options[option],
         configFileName});
@@ -87,6 +87,11 @@ export function applyConfigToArgv({
     }

     newArgv[option] = configObject[option];
+    const coerce = options[decamelizedOptName].coerce;
+    if (coerce) {
+      log.debug(`Calling coerce() on ${option}`);
+      newArgv[option] = coerce(newArgv[option]);
+    }
   }
   return newArgv;
 }
```

This patch would need test coverage too.

I hope that helps!

It was tricky to track this down. I had to add a breakpoint in `config.js` and run:

```
node --inspect --debug-brk `which web-ext` run --verbose -c web-ext-config.js
```

kumar303:
Hi @saintsebastian , good news! @rpl and I are at the Mozilla All-Hands right now (we are missing you). We decided that we have time to work on the config features so I'm going to land the patch and help with the fixes. Thanks so much for all of your work on this. Since you added the much needed *glue* to hook things up, it revealed a couple hidden bugs :)

saintsebastian:
@kumar303 @rpl wish you a great time in Austin! Hope to join you next time:) Glad the config features are moving forward, and feel free to ping me.

PR #1082, opened 2017-09-19
fix(package): update addons-linter to version 0.26.1

kumar303:
Closes #1076

Here's an explanation of the problem. Eslint needs to load the `eslint-recommended` plugin. Before the latest `addons-linter` upgrade, it worked like this:

```
$ eslint --debug 'src/**/*.js'
...
  eslint:config-file Loading /Users/kumar/dev/web-ext/node_modules/eslint/conf/eslint-recommended.js +7ms
```

This works fine because there's only one `eslint` in the tree:

```
$ yarn list eslint
yarn list v1.1.0
└─ eslint@3.19.0
```

After upgrading `addons-linter`, we introduce a new `eslint` and it loads a different version of `eslint-recommended`:

```
$ eslint --debug 'src/**/*.js'
...
  eslint:config-file Loading /Users/kumar/dev/web-ext/node_modules/addons-linter/node_modules/eslint/conf/eslint-recommended.js +4ms
```

Here's where you can see the conflicting packages:

```
$ yarn list eslint
yarn list v1.1.0
├─ addons-linter@0.26.1
│  └─ eslint@4.7.1
└─ eslint@3.19.0
```

Why is `eslint` loading the wrong `eslint-recommended`? These bugs could be related but in our case it fails with `npm`, not just `yarn`:

* https://github.com/eslint/eslint/issues/8547
* https://github.com/eslint/eslint/issues/7338

### DIFF
--- a/tasks/configs/copy.js
+++ b/tasks/configs/copy.js
@@ -1,0 +1,11 @@
+module.exports = {
+  'dist-files-to-artifacts-dir': {
+    files: [
+      {
+        expand: true,
+        src: ['package.json', 'dist/**', 'bin/**'],
+        dest: 'artifacts/production',
+      },
+    ],
+  },
+};
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,16 @@ before_script:
 
 # Keep this in sync with appveyor.yml
 script:
+## run eslint, flow and the unit test suite.
 - COVERAGE=y NODE_ENV=production npm test
+
+## run functional test suite in a npm production environment
+## (See #1082 for rationale).
+- npm run copy-dist-files-to-artifacts-dir
+- cd artifacts/production && npm install --production && cd -
+- TEST_WEB_EXT_BIN=./artifacts/production/bin/web-ext npm run test:functional
+
+## lint the github PR title.
 - npm run travis-pr-title-lint
 
 after_script: npm run publish-coverage
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -30,13 +30,12 @@ module.exports = function(grunt) {
     'webpack:functional_tests',
   ]);
 
-  grunt.registerTask('test', 'run linting and test suites', function() {
+  grunt.registerTask('test', 'run linting and the unit test suite', function() {
     var tasks = [
       'lint',
       'flowbin:check',
       'build-tests',
       'mochaTest:unit',
-      'mochaTest:functional',
     ];
 
     // TODO: enable the flowbin:check task on AppVeyor (mozilla/web-ext#773)
@@ -46,6 +45,27 @@ module.exports = function(grunt) {
     }
 
     grunt.task.run(tasks);
+  });
+
+  grunt.registerTask('test:functional', 'run functional test suites', [
+    'build-tests',
+    'mochaTest:functional',
+  ]);
+
+  grunt.registerTask('copy-dist-files-to-artifacts-dir', function() {
+    const distFile = grunt.file.expand('./dist/**').filter((filename) => {
+      return filename.endsWith('web-ext.js');
+    });
+
+    if (distFile.length === 0) {
+      // By default grunt will exit here, unless the `--force` grunt option has been
+      // specified on the command line.
+      grunt.fail.warn(
+        'Required dist/ file missing. Run the build command first.'
+      );
+    }
+
+    grunt.task.run('copy:dist-files-to-artifacts-dir');
   });
 
   grunt.registerTask('develop', [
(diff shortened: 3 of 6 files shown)


EnTeQuAk:
It seems that his is hellarously broken. I wonder if it's easier for web-ext to upgrade to eslint 4.x? This isn't a great situation but sounds like a reasonable workaround to unblock any new release.

rpl:
@kumar303 @EnTeQuAk the change in 88b4ead adds an explicit devDependencies between web-ext and eslint 3.19.0, this is enough to fix the unexpected eslint linting errors

Unfortunately it also unveils another issue:

- addons-linter has a strong dependency on eslint 4, because it is needed for more than just linting the addons-linter sources, it is part of the dependencies that addons-linter is going to use at runtime

- web-ext is currently (loosely) depending from eslint 3, by adding it explicitly to the web-ext devDependencies, we ensure that (in a developer installation) the web-ext sources will be linted using the expected eslint version

- when web-ext runs the functional tests on travis, the "node_modules/" dir has been created by including the web-ext devDependencies (which is not true when web-ext is installed by a user or its dependencies installed using "npm install --production"), and so the eslint version that the addons-linter is going to use is eslint 3.19.0 instead of the eslint 4.6.x expected by updated addons-linter sources and the change in https://github.com/mozilla/addons-linter/commit/9d4f45bb7cc518f2fdf9745f0b9e7f57f4e72a89#diff-2bf56d6974b3fd35628de87b7c500747R75 doesn't work on eslint 3

Locally I've manually verified that in a user installation "web-ext lint" is going to use eslint 4 as expected and so the linting command works correctly, nevertheless on travis the related test is going to fail because of eslint 3 used at runtime (which to be fair is a limitation/issue of the functional tests which should really use a node_modules/ dir generated as it would be with "npm install --production", to prevent this issue but also to be able to catch other issue related to differences in the "node_modules/" dir content between development and production installations).

EnTeQuAk:
Hmm, so https://github.com/gajus/eslint-plugin-flowtype/issues/261#issuecomment-327387222 seems to workaround that for now. Maybe worth looking into that?

rpl:
@EnTeQuAk unfortunately manually declaring all our flow types as globals doesn't seem a be great options for us (in that case the reporter seems to have issues only on a restricted number of flow types which are defined internally by flow, in our case eslint is complaining about the flow types that we are declaring in the web-ext sources itself)

and we have also the issue with the `no-shadow` rule :-(

rpl:
@kumar303 @EnTeQuAk the patch in d1d9fb7 is a (super gross) way to show what I've observed locally and it is still not "the workaround that we can use in the meantime" (installing an exact version of eslint manually in the travis and appveyor files is not something that we can really use as a workaround).

The best workaround/solution (at least in my current vision) would be to be able to run the functional tests on a web-ext build that is using the "node_modules/" dir created using "npm install --production",
e.g. because besides fixing the pressing problem of the two different eslint version that we would need to use right now, it will also allow to catch other similar issues related to the different "node_modules/" layout that is created by "npm install" (developer) vs. "npm install --production" (user).

rpl:
@kumar303 @EnTeQuAk In ba556cc there is a prototype of the approach described in my previous comment (and it works on both travis and appveyor, as well as locally):

- using a grunt task we create a clone of the web-ext dir with just the "package.json", the bundled files from "dist/" and the binary file form "bin/"
- in .travis.yml and appveyor.yml the npm packages in the cloned "artifacts/production" dir are installed using "npm install --production"
- then we point the functional tests to the web-ext executable in "artifacts/production/bin/" (which is going to use exactly the same dependencies that a user is going to use)

How that sounds to you?

tasks/configs/copy.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  'create-artifacts-production': {
```
kumar303: Just a small nit but the full command would read better to me as `copy:dist-files-to-artifacts-dir`.

tasks/configs/copy.js
```diff
@@ -0,0 +1,11 @@
...
+    files: [
+      {
+        expand: true,
+        src: ['package.json', 'dist/**', 'bin/**'],
```
kumar303: Is there an easy way to make this throw an error if `dist` is empty? A simple message like 'run the build command first' might prevent some confusion.

kumar303: I looked at [`grunt-contrib-copy`](https://github.com/gruntjs/grunt-contrib-copy) but I don't see anything obvious. It's not worth spending too much time on. Maybe just add a comment to this file that you need to build first?

PR #1105, opened 2017-10-13
fix: Fix deploy script so that it always publishes a production build.

kumar303:
I think this will work but we'll have to cut a bug fix release to fully test it. If this works I'll file an issue to follow-up with some better safety checks so it doesn't happen again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,12 @@ script:
 
 ## lint the github PR title.
 - npm run travis-pr-title-lint
-## In case Travis is building a release tag, make sure we have a production build
-## before publishing to npm
-- NODE_ENV=production npm run build
 
 after_script: npm run publish-coverage
+
+before_deploy:
+## Make sure we have a production build.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,6 +22,9 @@ script:
 
 ## lint the github PR title.
 - npm run travis-pr-title-lint
+## In case Travis is building a release tag, make sure we have a production build
+## before publishing to npm
+- NODE_ENV=production npm run build
```
rpl: :+1:

I guess that we can use something like:
```
before_deploy:
- NODE_ENV=production npm run build
```
to run it only when we are release a new version on a tag build (https://docs.travis-ci.com/user/deployment/npm/#Running-commands-before-and-after-deploy)

kumar303: oh nice! I didn't know about that hook.

kumar303: That will keep the build running fast when we aren't doing a deploy

PR #1121, opened 2017-10-25
Allow legacy add-ons for Firefox 57+ (DevEdition)

mai-cliqz:


kumar303:
Could you help me understand what problem are you trying to solve? I'm confused because `web-ext` is designed for developing [Web Extensions](https://developer.mozilla.org/en-US/Add-ons/WebExtensions) which are not legacy extensions.

mai-cliqz:
@kumar303 I was developing a legacy add-on for FF57 and I found out that we need this pref to run my add-on in FF57+. I forgot that with web-ext add-ons we don't need this pref. Thanks for reminding me. Hence, I close this PR.

Issue #1100, opened 2017-10-09
Allow connecting to already running Firefox when using web-ext run


kewisch:
It would be great if we could use web-ext to install an add-on in an already running Firefox. This is especially useful if you are developing an add-on for a page that requires a log-in. I know I can quit Firefox and web-ext run using the existing profile, or I can just log in on the newly created profile, but it would be more convenient if I can just have it connect to my current Firefox instance.

kumar303:
This is similar to the approach we take in the Android patch (https://github.com/mozilla/web-ext/pull/868). However, I don't think what you're proposing will be possible because web-ext needs to make modifications to the profile in order to function correctly. Thus, web-ext pretty much always needs to alter the profile before Firefox starts.

Thanks for taking the time to file an issue!

kewisch:
I actually have a working patch for this, it just needs polish and some design decisions. All the developer really needs to do afaik is open gcli and enter e.g. `listen 6060`, which can easily be documented. I am not sure the other preferences are really needed to function, can you point out which those would be?

kumar303:
Yeah, I think all web-ext needs is [access to the debugger](https://github.com/mozilla/web-ext/blob/master/src/firefox/preferences.js#L30). It seems like this will confuse users who are less experienced and who don't read docs but I suppose if you have a patch then we can consider it. I will admit that other people have asked for it but mostly it was because they didn't know about other ways to do it.

kumar303:
> it would be more convenient if I can just have it connect to my current Firefox instance

Why? The case you described sounds like it would be really convenient as `web-ext run --firefox-profile default` (if *default* was the profile name). You wouldn't have to quit your running Firefox. This command would copy your profile into a temporary profile and would allow you not to pollute that profile.

kewisch:
That is indeed handy, but my profile is 1GB large, and I don't want to copy that around each time I test an add-on.

kewisch:
If you don't believe this feature is useful I don't have any bad feelings if you want to close this. I am fine just keeping the WiP patch applied locally and will use it for myself.

kumar303:
Wow, 1GB? Is the `--firefox-profile` option noticeably slow?

kewisch:
Not major, likely due to the SSD. I can make way with this as well, I don't debug add-ons often enough that this would really be a problem for me, but copying the whole profile just for a debugging session where I can instead use the browser I have running sounds like a workaround to me.

decadent:
[...]

PR #1122, opened 2017-10-27
feat: WiP - Allow connecting to an existing Firefox instance

kewisch:
Hey @kumar303, this is my WiP for #1100. It is mostly unpolished because I do have a few design questions open:

* `FirefoxInfo` has a `FirefoxProcess` member that is checked against `null` in some of the code. When connecting to a remote port without a process, there is nothing sensible to set `FirefoxProcess` to. Right now what I am doing is creating a mock `runningInfo` that just has the debuggerPort and the process' `kill` function which is used for cleanup. I can imagine better ways to fix this, but I'd appreciate your input on what you prefer. There could be some refactoring involved in this.
* Code right now assumes the only reason a process will interrupt is if the process is closed. I think it would be good to change/adapt this to react when the connection to the devtools server is interrupted. I'm not yet sure the event for this is exposed, but I would investigate. This way if you close the Firefox you've connected to it will properly disconnect. Does this sound reasonable?
* A feature I should add is that if e.g. Ctrl+C is pressed, the last action before the connection to the devtools server is interrupted is to uninstall the temporary add-on. The code for this would have to live in the mock `kill` function and would be asynchronous instead. I'd likely have to adapt the `FirefoxDesktopExtensionRunner::exit` function to take care of the asynchronous shutdown. Does this sound right?
* There are a few options not compatible with the `--firefox-port` option, e.g. `--firefox-profile`, `--firefox-binary`, `--keep-profile-changes`, and `--pre-install`. I could reasonably make `--pref`, `--browser-console`, and `--start-url` work, but I think that would be bonus. Is there an existing way to flag incompatible options, and if not anything I should consider while implementing? 
* I've added some text explaining this is an advanced feature to the option description. Do you consider this enough?

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -341,6 +341,15 @@ Example: $0 --help run.
         demand: false,
         type: 'string',
       },
+      'firefox-port': {
+        alias: 'P',
+        describe: 'Attach to a running Firefox on this developer tools ' +
+                  'port. To do so, open the developer toolbar and enter ' +
+                  'e.g. `listen 6060`. This flag usually does not need to ' +
+                  'be set and is only for advanced use.',
+        demand: false,
+        type: 'number'
+      },
       'keep-profile-changes': {
         describe: 'Run Firefox directly in custom profile. Any changes to ' +
                   'the profile will be saved.',
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -27,6 +27,7 @@ export type CmdRunParams = {|
   pref?: FirefoxPreferences,
   firefox: string,
   firefoxProfile?: string,
+  firefoxPort?: number,
   ignoreFiles?: Array<string>,
   keepProfileChanges: boolean,
   noInput?: boolean,
@@ -54,6 +55,7 @@ export default async function run(
     pref,
     firefox,
     firefoxProfile,
+    firefoxPort,
     keepProfileChanges = false,
     ignoreFiles,
     noInput = false,
@@ -93,6 +95,7 @@ export default async function run(
 
     // Firefox specific CLI options.
     firefoxBinary: firefox,
+    firefoxPort: firefoxPort,
     profilePath: firefoxProfile,
     customPrefs,
     browserConsole,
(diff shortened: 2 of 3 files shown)


kumar303:
> I think it would be good to change/adapt this to react when the connection to the devtools server is interrupted. I'm not yet sure the event for this is exposed, but I would investigate. This way if you close the Firefox you've connected to it will properly disconnect. Does this sound reasonable?

Sounds reasonable but I don't know if it's possible. If it's complicated, don't worry about it.

kumar303:
> A feature I should add is that if e.g. Ctrl+C is pressed, the last action before the connection to the devtools server is interrupted is to uninstall the temporary add-on. The code for this would have to live in the mock kill function and would be asynchronous instead. I'd likely have to adapt the FirefoxDesktopExtensionRunner::exit function to take care of the asynchronous shutdown. Does this sound right?

I think you already added that, right?

kumar303:
> Is there an existing way to flag incompatible options, and if not anything I should consider while implementing?

Just throw a `WebExtError`

src/program.js
```diff
@@ -341,6 +341,15 @@ Example: $0 --help run.
         demand: false,
         type: 'string',
       },
+      'firefox-port': {
```
kumar303: This name should make it more obvious that you are connecting to a remote Firefox instance. Maybe `remote-firefox-port`.

src/program.js
```diff
@@ -341,6 +341,15 @@ Example: $0 --help run.
...
+        alias: 'P',
+        describe: 'Attach to a running Firefox on this developer tools ' +
+                  'port. To do so, open the developer toolbar and enter ' +
```
kumar303: We should also show these instructions on the console in case a connection cannot be made

src/program.js
```diff
@@ -341,6 +341,15 @@ Example: $0 --help run.
         demand: false,
         type: 'string',
       },
+      'firefox-port': {
+        alias: 'P',
+        describe: 'Attach to a running Firefox on this developer tools ' +
```
kumar303: This should explain things more from the perspective of what web-ext will do. Maybe something like "Connect to this debugger port on a running Firefox instance to install the browser extension."

rpl:
Closing as "status: do not merge", the related issue is still technically not covered by web-ext but this PR has been branched out quite some time ago and so it may be easier to recreate it from scratch by cherry picking the proposed changes from this PR than rebasing this branch.

PR #1128, opened 2017-11-03
fix: upgrade eslint to version 4.10.0 and fix new linting errors

rpl:
This PR contains changes needed to the package.json, the .eslintrc to upgrade eslint to the new v.4 release
(and also upgrade some of the related npm dependencies, e.g. eslint-plugin-flowtype and grunt-eslint).

The new eslint v4 is much more strict on some of the linting rules (e.g. the new indent rule is able to catch a good number of indentations issues in the existent code) and so this PR contains also changes that are needed to fix the new linting errors.


src/cmd/build.js
```diff
@@ -175,7 +179,6 @@ export type BuildCmdParams = {|
   sourceDir: string,
   artifactsDir: string,
   asNeeded?: boolean,
-  noInput?: boolean,
```
kumar303: I think it would be unfortunate to remove this from the type. It's true to say that it will be a possible param since that's how yargs will send it. I wonder if we could solve it instead by defining all global options in one place and referencing it? Example:

```js
import type {GlobalOptions} from '../somewhere';

export type BuildCmdParams = {|
  asNeeded?: boolean,
  overwriteDest?: boolean,
  ignoreFiles?: Array<string>,
  ...GlobalOptions,
|};
```

kumar303:
I'll fix up the global option type thing. Let's get this in!

PR #1140, opened 2017-11-30
fix(package): update addons-linter to version 0.30.0 and ensure that CI tests fail on production npm binary components in dependencies

rpl:
Closes #1112

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,24 @@ script:
 ## run eslint, flow and the unit test suite.
 - COVERAGE=y NODE_ENV=production npm test
 
+## NOTE: by setting the configured python to /bin/false we are forcing the production mode tests
+## to fail if any of the dependencies is a binary dependency that is build using node-gyp.
+- npm config set python /bin/false
+
 ## run functional test suite in a npm production environment
 ## (See #1082 for rationale).
 - npm run copy-dist-files-to-artifacts-dir
-- cd artifacts/production && npm install --production && cd -
-- TEST_WEB_EXT_BIN=./artifacts/production/bin/web-ext npm run test:functional
+- cd artifacts/production
+- npm install --production; export NPM_INSTALL_EXIT_CODE=$?
+- cd -
+- test $NPM_INSTALL_EXIT_CODE -eq 0 &&
+  TEST_WEB_EXT_BIN=./artifacts/production/bin/web-ext npm run test:functional
+
+## NOTE: remove the custom python path from the npm config, and also
+## remove production package.json (which seems to convince npm run travlis-pr-title-lint
+## to look for grunt in the production node_modules/ dir).
+- npm config delete python
+- rm artifacts/production/package.json
 
 ## lint the github PR title.
 - npm run travis-pr-title-lint
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,8 @@
+## Force 64bit platform and nodejs, the addons-linter has dependencies which
+## do not provide a pre-compiled 32bit version of their nodejs binary module.
+platform: x64
 install:
-  - ps: Install-Product node $env:nodejs_version
+  - ps: Install-Product node $env:nodejs_version x64
   - set CI=true
   - set PATH=%APPDATA%\npm;%PATH%
   - npm install
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,13 @@ test_script:
   ## (See #1082 for rationale).
   - npm run copy-dist-files-to-artifacts-dir
   - cd artifacts\production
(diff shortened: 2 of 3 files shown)


rpl:
@kumar303 I've recreated the pull request with addons-linter updated to 0.29.1 instead of 0.29.0 and that fixes the "nsp check" issues and so all the tests pass on Travis.

Unfortunately they still fail on windows, and I took a brief look and it seems to be because of a new dependency of the addons-linter, introduced in mozilla/addons-linter#1581

Based on the error logged at this line: https://ci.appveyor.com/project/kumar303/web-ext/build/1.0.1113#L26 

It seems that the `sharp` npm dependency used to lint the addon and raise validation errors on corrupted images, and it seems that it is pre-compiled only for 64bit and on 32bit needs a manual installation.

(I've not yet tested this version on windows 32, but my guess is that the same npm dependency issue that is making the tests to fail on appveyor would also be an issue for a user that is installing or updating web-ext windows32 :-( )

asamuzaK:
See also [Can't install addons-linter anymore if python is not installed · Issue #1668 · mozilla/addons-linter](https://github.com/mozilla/addons-linter/issues/1668 "Can't install addons-linter anymore if python is not installed · Issue #1668 · mozilla/addons-linter")
Even if you merge this, it can't be installed on Windows, unless end user install Python.

rpl:
@kumar303 in 7e6e89c there are some additional changes to the .travis.yml and appveyor.yml files, these changes ensure that the CI jobs are going to fail if any of the npm dependencies is trying to build a nodejs native module in production mode (so that we can be sure that we are going to catch installation issues for the user similar to the one related to the addon-linter 0.29.1 that we had investigated in this PR).

Follows links to the failures to expect in this case:

- appveyor: https://ci.appveyor.com/project/kumar303/web-ext/build/1.0.1134#L2521
- travis: https://travis-ci.org/mozilla/web-ext/builds/312919780#L2887-L2888

rpl:
@kumar303 this is ready for you now (the only changes that are actually interesting from this PR are the one in the .travis.yml and appveyor.yml files).

.travis.yml
```diff
@@ -14,11 +14,24 @@ script:
...
+
 ## run functional test suite in a npm production environment
 ## (See #1082 for rationale).
 - npm run copy-dist-files-to-artifacts-dir
-- cd artifacts/production && npm install --production && cd -
-- TEST_WEB_EXT_BIN=./artifacts/production/bin/web-ext npm run test:functional
+- cd artifacts/production
+- npm install --production; export NPM_INSTALL_PROD=$?
+- cd -
+- test $NPM_INSTALL_PROD -eq 0 &&
```
kumar303: This looks like it will do nothing if `npm install --production` fails, correct? Will the test still fail in that case?

.travis.yml
```diff
@@ -14,11 +14,24 @@ script:
...
+## to fail if any of the dependencies is a binary dependency that is build using node-gyp.
+- npm config set python /bin/false
+
 ## run functional test suite in a npm production environment
 ## (See #1082 for rationale).
 - npm run copy-dist-files-to-artifacts-dir
-- cd artifacts/production && npm install --production && cd -
-- TEST_WEB_EXT_BIN=./artifacts/production/bin/web-ext npm run test:functional
+- cd artifacts/production
+- npm install --production; export NPM_INSTALL_PROD=$?
```
kumar303: How about something more specific like `NPM_INSTALL_EXIT_CODE` ?

appveyor.yml
```diff
@@ -27,7 +30,13 @@ test_script:
   ## (See #1082 for rationale).
   - npm run copy-dist-files-to-artifacts-dir
   - cd artifacts\production
+  ## NOTE: by setting the configured python to a batch file that exit with an error,
```
kumar303: nit: ...a batch file that exits...

appveyor.yml
```diff
@@ -27,7 +30,13 @@ test_script:
...
+  ## we are forcing the production mode tests to fail if any of the dependencies is
+  ## a binary dependency that is build using node-gyp.
```
kumar303: nit: ...that is built...

appveyor.yml
```diff
@@ -27,7 +30,13 @@ test_script:
...
+  - echo "exit /b 1" > C:\Python-forbidden-in-production-install.cmd
```
kumar303: nice!

.travis.yml
rpl: yes, when `npm install --production` fails, the travis job will still fail, at line 27 when we test the return value.

Here is a test on travis which show the failure:

- https://travis-ci.org/mozilla/web-ext/builds/313421003#L2903

.travis.yml
rpl: :+1:

PR #1149, opened 2017-12-10
feat: support of named profiles

### LINKED ISSUE #932: Can't run with specified profile name and --keep-profile-changes

saintsebastian:
Fixes #932 

I shamelessly reused bits of code from [previous attempt to solve this issue](https://github.com/mozilla/web-ext/pull/968).


kumar303:
Thanks! Let me know when you've added tests and I can take a look.

saintsebastian:
@kumar303 Hi, sorry it took me so long to get to this. 
I am having troubles figuring out how to test the [returned function](https://github.com/saintsebastian/web-ext/blob/4bef5869f7a3b05664c254f43ab924a04fc48a75/src/firefox/index.js#L336), maybe you have ideas?

saintsebastian:
@kumar303 @rpl this is pretty much ready for initial review, and maybe your suggestions for further testing. Let me know if you have any thoughts

rpl:
@saintsebastian To test `defaultCreateProfileFinder` it seems that we can inject a fake `FxProfile` (and `userDirectoryPath` for the test case that tests the behavior when a custom user directory path has been specified) when we call the function, then we can call the function returned by `defaultCreateProfileFinder` and verify the internal behavior of that function by checking that the sinon spies the we include in the fake `FxProfile` have been called (or not called) with the expected parameters.

Inside the returned function there is also a call to `fs.stat`, and so we have to ensure the in the test we point it out to a temporary test directory, or we can opt to add `fs` as an additional "injected dependencies" to the `defaultCreateProfileFinder` method and in the test case redefine it to suit its expected behavior.

e.g. the test case could be something like the following:

```js
// A fake Finder instance.
const fakeFinder = {
  readProfiles: sinon.spy(() => { ... }),
  getPath: sinon.spy(() => { ... }),
}

// A fake Finder constructor which returns the fake Finder instance.
const FakeFinder = sinon.spy(() => fakeFinder);

// Add a fake localUserDirectory static method to the fake Finder constructor.
FakeFinder.locateUserDirectory = sinon.spy(() => "/tmp/fakeUserDir";

// Define the fake FxProfile to inject as the fake dependency into defaultCreateProfileFinder.
const fakeFxProfile = {Finder: sinon.spy(() => fakeFinder)};

// Create a fakeFS object which provides a fake `stat` method (wrapped by a sinon spy).
const fakeFs = ...

const profileFinder = defaultCreateProfileFinder({FxProfile: fakeFxProfile, fs: fakeFs});

const res = await profileFinder("aFakeProfileName");

sinon.assert.calledOnce(FakeFinder.locateUserDirectory)
...
```

tests/unit/test-firefox/test.firefox.js
```diff
@@ -511,20 +511,136 @@ describe('firefox', () => {
...
+          const createProfileFinder = () => {
+            return (profilePath) => Promise.resolve(profilePath);
+          };
+          const profile = await firefox.useProfile(baseProfile.path(), {
+            app,
+            configureThisProfile,
+            createProfileFinder,
           });
+          assert.instanceOf(profile, FirefoxProfile);
+        } catch (error) {
```
rpl: This `try...catch` is just re-throwing the caught error, it doesn't seem to be needed (same for the other ones in the other new tests below).

src/firefox/index.js
```diff
@@ -317,13 +317,48 @@ export function configureProfile(
...
+  return async (profileName: string): Promise<string | void> => {
+    try {
+      await readProfiles();
+      const hasProfileName = finder.profiles.filter(
+        (profileDef) => profileDef.Name === profileName).length !== 0;
+      if (hasProfileName) {
+        return await getPath(profileName);
+      }
+    } catch (error) {
+      if (isErrorWithCode('ENOENT', error)) {
```
rpl: Do you mind to reverse this if and remove the else block?
```js
if (!isErrorWithCode('ENOENT', error)) {
  throw error;
}
log.warn('...');
```

Also, I'm thinking that we should tweak the warning message a bit, e.g.  
```js
log.warn('Unable to find Firefox profiles.ini');
```

saintsebastian:
@rpl thanks for the review, i think it really helped! what do you think of those tests?

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+         }];
+
+         const userDirectoryPath = '/non/existent/path';
+
+         const getter = firefox.defaultCreateProfileFinder({
+           userDirectoryPath,
+           FxProfile,
+         });
+
+         try {
```
rpl: Let's rewrite this using `assert.isRejected` and the approach described in one of the comments above.

(as a side note, when we use the `try...catch` strategy to assert a raised exception, we always have to be sure that the test fails if the expected exception has not be raised, e.g. by just get a reference to the actual error from the catch block and then call the assertion after the `try...catch`).

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+
+      sinon.assert.called(FxProfile.Finder);
+      sinon.assert.calledWith(
+        FxProfile.Finder,
+        sinon.match(userDirectoryPath),
+      );
+    });
+
+    it('returns a finder that looks for a default profile', async () => {
+      const FxProfile = {
```
rpl: If I'm reading this test correctly, we could make this a bit more readable by doing the opposite:

```js
         const fakeReadProfiles = sinon.spy(() => {
           return Promise.reject(
             new Error('unspecified error'));
         });
         const fakeGetPath = sinon.spy(() => Promise.resolve());
         const fakeProfiles = [{Name: 'someName'}];

         const FxProfile = {
           Finder: function () {
             return {
               readProfiles: fakeReadProfiles,
               getPath: fakeGetPath,
               profiles: fakeProfiles,
             };
          },
         };

         ....
```

This is a pattern used in a couple of tests, and so the same comment is valid for the tests below, it may be even worth to turn it into a test helper function defined in this test file and reuse it in all the tests.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
 
   });
 
+  describe('defaultCreateProfileFinder', () => {
+
+    it('creates a finder', async () => {
+      const FxProfile = {
+        Finder: sinon.spy(() => () => Promise.resolve()),
```
rpl: `Finder` is a constructor function and so it seems that its spy should be something like:

```js
const FxProfile = {
  Finder: sinon.spy(function () { 
    return {}; 
  }),
};
```

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+      };
+      firefox.defaultCreateProfileFinder({FxProfile});
+      sinon.assert.called(FxProfile.Finder);
+      sinon.assert.calledWith(
+        FxProfile.Finder,
+        sinon.notOk,
```
rpl: This `sinon.assert.calledwith(FxProfile.Finder, sinon.notOk);` assertion doesn't seem right (or it is not very readable), shouldn't it be something like the following?

```js
firefox.defaultCreateProfileFinder({FxProfile});
sinon.assert.calledWith(FxProfile.Finder, sinon.match(undefined));
```

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+      );
+    });
+
+    it('creates finder based on userDirectoryPath if present', async () => {
+      const FxProfile = {
+        Finder: sinon.spy(() => () => Promise.resolve()),
+      };
+      const userDirectoryPath = '/non/existent/path';
+
```
rpl: Nit, remove this empty line.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+      firefox.defaultCreateProfileFinder({userDirectoryPath, FxProfile});
+
+      sinon.assert.called(FxProfile.Finder);
+      sinon.assert.calledWith(
+        FxProfile.Finder,
+        sinon.match(userDirectoryPath),
+      );
+    });
+
+    it('returns a finder that looks for a default profile', async () => {
```
rpl: uhm... I think that the description of this test should probably be `'returns a finder that resolves a profile name'` or something like that (instead of 'looks for a default profile').

tests/unit/test-firefox/test.firefox.js
```diff
@@ -545,6 +646,136 @@ describe('firefox', () => {
...
+         assert.equal(
+           res,
+           undefined,
+           'Got an undefined result when the profiles.ini file does not exist');
+
+         sinon.assert.called(fakeReadProfiles);
+         sinon.assert.notCalled(fakeGetPath);
+       });
+
+    it('returns a finder that throws other errors',
```
rpl: Nit, s/other errors/unexpected errors/

saintsebastian:
@rpl thanks for the thorough as usual review! I think this is pretty close to done

tests/unit/test-firefox/test.firefox.js
```diff
@@ -511,40 +512,240 @@ describe('firefox', () => {
...
+        FxProfile,
+        userDirectoryPath,
+      } = prepareReaderTest(Promise.resolve());
+
+      const getter = firefox.defaultCreateProfileFinder({
+        userDirectoryPath,
+        FxProfile,
+      });
+
+      await getter('someName');
```
rpl: Just a minor tweak to this test:
we could test that the result value of getter('someName') is the same one that `fakeGetPath` returns.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -511,40 +512,240 @@ describe('firefox', () => {
...
-            assert.equal(configureThisProfile.firstCall.args[1].app, app);
           });
       }
     ));
 
   });
 
+  describe('defaultCreateProfileFinder', () => {
+
+    function prepareReaderTest(readProfileReturns) {
```
rpl: Nit, how about renaming this to `prepareProfileFinderTest`?

also it think that it may be easier to guess what the parameters of this test helper do if we pass them as object properties, and it would also allow us to expand the helper for the test that needs it (e.g. we may add a `getPathResult` to inject an expected result into the out test spy and complete the test "returns a finder that resolves a profile name" as suggested below), I mean something like:

```js
function prepareReaderTest({readProfileResult, getPathResult}) {
  ...
}

it('returns a finder that resolves a profile name', async () => {
      const {
        ...
      } = prepareReaderTest({
       readProfileResult: ...,
       getPathResult: ...,
      });
      
      ...
```

rpl:
@saintsebastian I forgot to mention: don't worry about the failure on appveyor, it is an unrelated (and pretty annoying to be honest) intermittent that is happening (too often :-( ) on the windows CI service.

tests/unit/test-firefox/test.firefox.js
rpl: These changes have been applied in the follow up commit 7d8b703.

tests/unit/test-firefox/test.firefox.js
rpl: These changes have been applied in the follow up commit 7d8b703.

rpl:
This PR has been merged as c85aa6e (rebased and squashed version of the patches in this PR) 
and 7d8b703 (which contains the small tweaks suggested in the last round of review comments).

Issue #1151, opened 2017-12-13
Fix integration issues with --config option


kumar303:
This fixes a few integration issues related to https://github.com/mozilla/web-ext/issues/775

- Nested config options are erasing configured global options
- Configured option values are not getting coerced the same way they do on the CLI

PR #1152, opened 2017-12-13
fix: Fix config parsing issues

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/1151

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,7 +36,7 @@ export function applyConfigToArgv({
       typeof configObject[option] === 'object') {
       // Descend into the nested configuration for a sub-command.
       newArgv = applyConfigToArgv({
-        argv,
+        argv: newArgv,
         configObject: configObject[option],
         options: options[option],
         configFileName});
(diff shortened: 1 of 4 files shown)


tests/unit/test.program.js
```diff
@@ -472,46 +472,30 @@ describe('program.main', () => {
...
-        sinon.assert.calledOnce(fakeApplyConfigToArgv);
-        sinon.assert.calledWith(fakeApplyConfigToArgv, sinon.match({
-          configFileName: resolvedFakePath,
-          configObject,
-          argv: {
-            _: ['lint'],
-            config: fakePath,
-          },
-        }));
+        // This makes sure that the config object was applied
```
kumar303: This part of the code didn't have test coverage before.

src/config.js
```diff
@@ -87,6 +87,13 @@ export function applyConfigToArgv({
     }
 
     newArgv[option] = configObject[option];
+
+    const coerce = options[decamelizedOptName].coerce;
```
rpl: It would be nice to cleanup this method further in a followup (e.g. `options[decamelizedOptName]` could be assigned to a variable near line 48 and then use it where we are currently using `options[decamelizedOptName]`, only to make the rest of this method a bit more readable).

PR #1158, opened 2017-12-14
chore(package): update sinon to version 4.1.3

kumar303:
Closes #1086


kumar303:
We're not affected by breaking changes: http://sinonjs.org/guides/migrating-to-4.0

PR #1159, opened 2017-12-14
chore(package): update webpack to version 3.10.0

kumar303:
Closes #1072


kumar303:
I tested this locally and it seems to be working fine. It feels faster too.

PR #1161, opened 2017-12-15
feat: Discover config files in home dir, working dir

kumar303:
Fixes https://github.com/mozilla/web-ext/issues/730

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -16,7 +16,7 @@ type ApplyConfigToArgvParams = {|
   // This is the argv object which will get updated by each
   // config applied.
   argv: Object,
-  // This is the argv that only has CLI values applies to it.
+  // This is the argv that only has CLI values applied to it.
   argvFromCLI: Object,
   configObject: Object,
   options: Object,
@@ -157,14 +157,14 @@ export async function discoverConfigFiles(
         log.debug(
           `Discovered config "${resolvedFileName}" does not ` +
           'exist or is not readable');
-        return false;
+        return undefined;
       }
     }
   ));
 
   const existingConfigs = [];
   configs.forEach((f) => {
-    if (f) {
+    if (typeof f === 'string') {
       existingConfigs.push(f);
     }
   });
--- a/src/config.js
+++ b/src/config.js
@@ -1,17 +1,23 @@
 /* @flow */
+import os from 'os';
 import path from 'path';
 
 import requireUncached from 'require-uncached';
 import camelCase from 'camelcase';
 import decamelize from 'decamelize';
 
+import fileExists from './util/file-exists';
 import {createLogger} from './util/logger';
 import {UsageError, WebExtError} from './errors';
 
 const log = createLogger(__filename);
 
 type ApplyConfigToArgvParams = {|
+  // This is the argv object which will get updated by each
+  // config applied.
   argv: Object,
+  // This is the argv that only has CLI values applies to it.
+  argvFromCLI: Object,
   configObject: Object,
   options: Object,
   configFileName: string,
@@ -19,6 +25,7 @@ type ApplyConfigToArgvParams = {|
 
 export function applyConfigToArgv({
   argv,
+  argvFromCLI,
   configObject,
   options,
   configFileName,
@@ -26,9 +33,9 @@ export function applyConfigToArgv({
   let newArgv = {...argv};
 
   for (const option in configObject) {
-
     if (camelCase(option) !== option) {
-      throw new UsageError(`The config option "${option}" must be ` +
+      throw new UsageError(
+        `The config option "${option}" must be ` +
         `specified in camel case: "${camelCase(option)}"`);
     }
 
@@ -37,6 +44,7 @@ export function applyConfigToArgv({
       // Descend into the nested configuration for a sub-command.
       newArgv = applyConfigToArgv({
         argv: newArgv,
+        argvFromCLI,
         configObject: configObject[option],
         options: options[option],
         configFileName});
@@ -74,15 +82,19 @@ export function applyConfigToArgv({
       }
     }
 
-    // we assume the value was set on the CLI if the default value is
-    // not the same as that on the argv object as there is a very rare chance
-    // of this happening
+    // This is our best effort (without patching yargs) to detect
+    // if a value was set on the CLI instead of in the config.
+    // It looks for a default value and if the argv value is
+    // different, it assumes that the value was configured on the CLI.
 
-    const wasValueSetOnCLI = typeof(argv[option]) !== 'undefined' &&
-      (argv[option] !== defaultValue);
+    const wasValueSetOnCLI =
+      typeof argvFromCLI[option] !== 'undefined' &&
+      argvFromCLI[option] !== defaultValue;
     if (wasValueSetOnCLI) {
-      log.debug(`Favoring CLI: ${option}=${argv[option]} over ` +
+      log.debug(
+        `Favoring CLI: ${option}=${argvFromCLI[option]} over ` +
         `configuration: ${option}=${configObject[option]}`);
+      newArgv[option] = argvFromCLI[option];
       continue;
     }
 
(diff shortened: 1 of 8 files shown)


tests/unit/test.config.js
```diff
@@ -474,6 +505,40 @@ describe('config', () => {
       assert.strictEqual(newArgv.apiKey, cmdApiKey);
     });
 
+    it('can load multiple configs', () => {
```
rpl: This test looks pretty similar to the one at line 252, but on the sign command instead of a fake command, aand with commantOpt instead of globalOpt. 

if we need both these tests, it would be nice to extract the common part into a test helper function defined in this file.

tests/unit/test.program.js
```diff
@@ -480,21 +507,154 @@ describe('program.main', () => {
...
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+    const configObject = {
+      lint: {
+        selfHosted: true,
+      },
+    };
+    // Instead of loading/parsing a real file, just return an object.
+    const fakeLoadJSConfigFile = sinon.spy(() => {
```
rpl: This is defined as a `sinon.spy`, we could use it to assert that it is called with the expected filename, am I right?

tests/unit/test.program.js
```diff
@@ -480,21 +507,154 @@ describe('program.main', () => {
...
+      return configObject;
+    });
+
+    await execProgram(
+      ['lint'],
+      {
+        commands: fakeCommands,
+        runOptions: {
+          discoverConfigFiles: async () => ['fake/config.js'],
```
rpl: we could also wrap this `discoverConfigFiles` into a sinon.spy and check that it has been called as expected.

tests/unit/test.program.js
```diff
@@ -480,21 +507,154 @@ describe('program.main', () => {
...
+    assert.equal(
+      options.selfHosted, configObject.lint.selfHosted);
+  });
+
+  it('lets you disable config discovery', async () => {
+    const fakeCommands = fake(commands, {
+      lint: () => Promise.resolve(),
+    });
+
+    const discoverConfigFiles = sinon.spy(() => Promise.resolve([]));
```
rpl: Not mandatory, but we could do the same for loadJSConfigFile (defining it as a sinon.spy and assert that it has never been called).

tests/unit/test.program.js
```diff
@@ -480,21 +507,154 @@ describe('program.main', () => {
...
+        runOptions: {
+          discoverConfigFiles: async () => [
+            globalConfig, projectConfig,
+          ],
+          loadJSConfigFile,
+        },
+      }
+    );
+
+    const options = fakeCommands.lint.firstCall.args[0];
```
rpl: This tests should test that the config file are loaded and applied in the expected order, how about wrapping `loadJSConfigFile` in a sinon.spy and assert that it has been called with the filenames in the expected order?

src/config.js
```diff
@@ -118,3 +130,43 @@ export function loadJSConfigFile(filePath: string): Object {
...
+        log.debug(
+          `Discovered config "${resolvedFileName}" does not ` +
+          'exist or is not readable');
+        return false;
+      }
+    }
+  ));
+
+  const existingConfigs = [];
+  configs.forEach((f) => {
```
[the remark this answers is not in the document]
kumar303: I'll take another look but I think `filter()` was failing because Flow doesn't seem to understand that the returned array will not contain undefined values (or something like that).

kumar303:
> I was also wondering if it could make sense to apply the tweaks to the flow type annotations related to the global and command options as part of this pull request

I think we should defer this because I tried it and fell down a rabbit hole. It's a tough change because the test suite needs a ton of fixes to pass the right default CLI params to each command callback. This is something we should have done all along, whoops.

src/config.js
```diff
@@ -118,3 +130,43 @@ export function loadJSConfigFile(filePath: string): Object {
...
+  {getHomeDir = os.homedir}: DiscoverConfigFilesParams = {}
+): Promise<Array<string>> {
+  const magicConfigName = 'web-ext-config.js';
+
+  // Config files will be loaded in this order.
+  const possibleConfigs = [
+    // Look for a magic hidden config (preceded by dot) in home dir.
+    path.join(getHomeDir(), `.${magicConfigName}`),
+    // Look for a magic config in the current working directory.
+    path.join(process.cwd(), magicConfigName),
```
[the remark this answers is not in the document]
kumar303: Good idea. Filed: https://github.com/mozilla/web-ext/issues/1166

src/config.js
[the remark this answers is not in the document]
kumar303: This is the best workaround I found:

```js
  const files = (
    (configs.filter((f) => typeof f === 'string'): Array<any>): Array<string>
  );
  return files;
```

That's too confusing though. I think it's probably best to leave it as is.

The Flow bug: https://github.com/facebook/flow/issues/1414

tests/unit/test-util/test.file-exists.js
```diff
@@ -0,0 +1,49 @@
...
+import {assert} from 'chai';
+import {describe, it} from 'mocha';
+import {fs} from 'mz';
+
+import fileExists from '../../../src/util/file-exists';
+import {withTempDir} from '../../../src/util/temp-dir';
+import {ErrorWithCode} from '../helpers';
+
+
+describe('util/file-exists', () => {
```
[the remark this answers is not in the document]
kumar303: whoops, good catch!

tests/unit/test.config.js
kumar303: Yeah, they are both necessary. One will apply a global option twice and the other will apply a sub-command option twice. The sub-command test covers some recursion.

I added some comments and improved the test specs to make it more clear. I couldn't really see any code that could be abstracted in a helpful way.

tests/unit/test.program.js
kumar303: I prefer to check things like that implicitly when possible. For this test case, if `discoverConfigFiles` is not called, you get this error:

```
  1) program.main discovers config files:
     AssertionError: expected false to equal true
      at _callee5$ (dist/webpack:/tests/unit/test.program.js:556:12)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at step (dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:1)
      at dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:1
```

When you look up the line of the assertion failure you will see this:

```js
    // This makes sure that the config object was applied
    // to the lint command options.
    assert.equal(
      options.selfHosted, configObject.lint.selfHosted);
```

I think that will be helpful to diagnose the problem.

tests/unit/test.program.js
kumar303: The test you commented on is just about disabling config discovery. I think it's enough coverage to just assert that the discovery function is not called. Without that call, there is no way to load a discovered file.

tests/unit/test.program.js
kumar303: Good catch, fixed

tests/unit/test.program.js
kumar303: Oh, yeah, thanks. This test wasn't checking the order at all. Fixed.

src/config.js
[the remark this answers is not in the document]
rpl: yeah, I agree,  better to leave it as it is (which is also probably much more readable to a contributor that is not so used to the flow syntaxes).

tests/unit/test.program.js
rpl: yeah, I was thinking more about a possible "future regression" where some new code could try to call loadJSConfigFile without having called discoverConfigFile first, but that's more likely to be just "me being too paranoid" :-P

tests/unit/test.config.js
rpl: :+1: the comments and the changes to the test names are useful and more then enough to quickly notice how the two tests are different in purpose (so that we don't end up to remove them by mistake by missing their different purpose), looks good to me.

PR #1171, opened 2018-01-05
chore(package): update mocha to version 4.1.0

kumar303:
Closes #1098


kumar303:
It looks like these breaking changes don't affect us: https://github.com/mochajs/mocha/releases/tag/v4.0.0

PR #1178, opened 2018-01-05
fix(package): update firefox-profile to version 1.1.0

kumar303:
Closes #896


kumar303:
This breaking change doesn't affect us because we don't use `profile.encode(...)` https://github.com/saadtazi/firefox-profile-js/blob/master/CHANGELOG.md#100

PR #1181, opened 2018-01-05
fix(package): update parse-json to version 4.0.0

kumar303:
Closes #1052


kumar303:
It doesn't look like there are any breaking changes affecting us:

* https://github.com/sindresorhus/parse-json/compare/v2.2.0...v3.0.0
* https://github.com/sindresorhus/parse-json/compare/v3.0.0...v4.0.0

PR #1183, opened 2018-01-05
chore(package): update conventional-changelog-lint to version 2.1.1

kumar303:
Closes #996


kumar303:
The breaking change here had to do with wildcards which we don't use http://marionebl.github.io/commitlint/#/guides-upgrade?id=version-1-to-2

PR #1192, opened 2018-01-08
chore(package): update flow-bin to version 0.63.1

kumar303:
Closes #1187

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "eslint-plugin-flowtype": "2.41.0",
     "eslint-plugin-import": "2.8.0",
     "firefox-client": "0.3.0",
-    "flow-bin": "0.62.0",
+    "flow-bin": "0.63.1",
     "grunt": "1.0.1",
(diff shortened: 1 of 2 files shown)


rpl:
@kumar303 it looks that flow 0.63.1 is smarter and it is now complaining that we were "lying" in the `fake` test helper :-)

- https://travis-ci.org/mozilla/web-ext/builds/326406542#L1906-L1909

I propose to workaround the issue by using an additional $FLOW_IGNORE, similarly to the one we are already using to stop flow from complaining about the `fake` function type signature: 

- https://github.com/mozilla/web-ext/blob/d16f9c04e8b3f599022d450fa540f2a1e585e006/tests/unit/helpers.js#L154

kumar303:
Maybe this would work? I haven't tried it yet.

```js
export function fake<T>(...): $Shape<T>
```

I have a feeling it won't work though because things like `Object.keys()` tend to confuse Flow easily.
 
I am fine with a `$FLOW_IGNORE` comment too.

PR #1199, opened 2018-01-09
fix: Config file array files should be recognized as options of array type

### LINKED ISSUE #1198: Array values are not recognized as is in config file
a bug

config file does not recognize, as expected, parameters with array as value. For example, the following config file:
```javascript
module.exports = {
    ignoreFiles: [
        'web-ext-config.js'
    ]
};
```
Following command
```
web-ext build
```
raises the following error:
```
Applying config file: ./web-ext-config.js

UsageError: The config file at .../web-ext-config.js specified the type of "ignoreFiles" incorrectly as "object" (expected type: "array")
```

parameter ignoreFiles recognized its value as an array.

* **Firefox version**: irrelevant
* **Your OS and version**: MacOS 10.13.2

````
````
node: v8.2.1
npm: 5.4.2
web-ext: 2.3.0

### PR DESCRIPTION
**Pull request by rpl**
This PR contains the changes (and additional test case) to fix #1198

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -66,7 +66,11 @@ export function applyConfigToArgv({
     const expectedType = options[decamelizedOptName].type ===
       'count' ? 'number' : options[decamelizedOptName].type;
 
-    const optionType = typeof configObject[option];
+    const optionType = (
+      Array.isArray(configObject[option]) ?
+        'array' : typeof configObject[option]
+    );
+
     if (optionType !== expectedType) {
       throw new UsageError(`The config file at ${configFileName} specified ` +
         `the type of "${option}" incorrectly as "${optionType}"` +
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/config.js [hunk 1]
```diff
@@ -66,7 +66,11 @@ export function applyConfigToArgv({
     const expectedType = options[decamelizedOptName].type ===
       'count' ? 'number' : options[decamelizedOptName].type;
 
-    const optionType = typeof configObject[option];
+    const optionType = (
+      Array.isArray(configObject[option]) ?
```
kumar303: Oh man, duh. 🧀

kumar303: While you're in here, could you also add support for number types [like this one](https://github.com/mozilla/web-ext/blob/cdb5b003afceea9426490fc21897cbd6e8fc9a3d/src/program.js#L393)?

kumar303: Nevermind, sorry, I thought that it would need special handling but it does not.

PR #1217, opened 2018-01-18
fix: Granted android permissions are only needed for android >= 23

### LINKED ISSUE #1211: web-ext run is unable to run on Android 5 because of external storage permissions discovery

bug

When the "web-ext run" target is "firefox-android", web-ext detects an android version and if it is >= android-21 it also checks that the external storage permissions are granted for the selected Firefox for Android APK.

Unfortunately on Android 5 (android-21 and android-22) it seems that even if the permission exists,
it is not optional and so it cannot be granted and "web-ext run" is currently unable to run successfully. 

"web-ext run -t firefox-android" should be able to run an extension on Android 5 and Android 5.1

* **Firefox version**: any
* **Your OS and version**: Android 5 and 5.1

rpl:
This pull request contains the changes needed to fix #1211 

The android "external storage permissions" are optional only starting from Marshmallow (android-23), and so they cannot be "granted" on the previous android versions.

This patch also fixes a bug in the test suite (and in the code), the previous version of the test was wrongly returning undefined from the `getAndroidVersionNumber` spy and so the test wasn't really testing the behavior on the expected android version number (and the implementation wasn't raising an error for an unexpected "not number" android version).

### DIFF
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -400,7 +400,13 @@ export class FirefoxAndroidExtensionRunner {
       selectedAdbDevice
     );
 
-    if (androidVersion < 21) {
+    if (typeof androidVersion !== 'number' || isNaN(androidVersion)) {
+      throw new WebExtError(`Invalid Android version: ${androidVersion}`);
+    }
+
+    log.debug(`Detected Android version ${androidVersion}`);
+
+    if (androidVersion < 23) {
       return;
     }
 
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -400,7 +400,7 @@ export class FirefoxAndroidExtensionRunner {
       selectedAdbDevice
     );
 
-    if (typeof androidVersion !== 'number' || isNaN(androidVersion)) {
+    if (typeof androidVersion !== 'number' || Number.isNaN(androidVersion)) {
       throw new WebExtError(`Invalid Android version: ${androidVersion}`);
     }
 
@@ -414,7 +414,8 @@ export class FirefoxAndroidExtensionRunner {
               `on ${selectedFirefoxApk}...`);
 
     // Runtime permission needed to be able to run Firefox on a temporarily created profile
-    // on android versions >= 21.
+    // on android versions >= 23 (Android Marshmallow, which is the first version where
+    // these permissions are optional and have to be granted explicitly).
     await adbUtils.ensureRequiredAPKRuntimePermissions(
       selectedAdbDevice, selectedFirefoxApk, [
         'android.permission.READ_EXTERNAL_STORAGE',
--- a/tests/unit/test-extension-runners/test.firefox-android.js
+++ b/tests/unit/test-extension-runners/test.firefox-android.js
@@ -704,37 +704,99 @@ describe('util/extension-runners/firefox-android', () => {
          sinon.assert.calledOnce(anotherCallback);
        });
 
-    it('checks the granted android permissions on Android >= 21',
+    it('raised an error when unable to find an android version number',
        async () => {
-         const {
-           params, fakeADBUtils,
-         } = prepareSelectedDeviceAndAPKParams();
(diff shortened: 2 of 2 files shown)


tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -704,37 +704,99 @@ describe('util/extension-runners/firefox-android', () => {
...
-         fakeADBUtils.getAndroidVersionNumber = sinon.spy(() => {
-           Promise.resolve(21);
-         });
+           fakeADBUtils.getAndroidVersionNumber = sinon.spy(() => {
+             return version;
+           });
 
-         const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+           const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+           const promise = runnerInstance.run();
```
kumar303: Shouldn't this be `const promise = await runnerInstance.run()` so that `assert.isRejected()` gets the finished promise? I can't quite tell from the chai-as-promised docs how it works. Alternatively, maybe `expectInvalidVersionError()` needs to return `promise` so that the test always waits for completion?

src/extension-runners/firefox-android.js
```diff
@@ -400,7 +400,13 @@ export class FirefoxAndroidExtensionRunner {
...
     );
 
-    if (androidVersion < 21) {
+    if (typeof androidVersion !== 'number' || isNaN(androidVersion)) {
+      throw new WebExtError(`Invalid Android version: ${androidVersion}`);
+    }
+
+    log.debug(`Detected Android version ${androidVersion}`);
+
+    if (androidVersion < 23) {
```
kumar303: Could you adjust [this comment](https://github.com/rpl/web-ext/blob/a68cb531f07eed5bd2dcb223a16ad6a8f061f2f8/src/extension-runners/firefox-android.js#L416)? It still says android-21

src/extension-runners/firefox-android.js
```diff
@@ -400,7 +400,13 @@ export class FirefoxAndroidExtensionRunner {
       selectedAdbDevice
     );
 
-    if (androidVersion < 21) {
+    if (typeof androidVersion !== 'number' || isNaN(androidVersion)) {
```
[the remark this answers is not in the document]
rpl: ah! I see, yeah, let's use `Number.isNaN` (at that point we are actually sure that `androidVersion` is a number and so it seems that it should not make a difference, but it is not going to harm neither if we do use `Number.isNaN` :-))

src/extension-runners/firefox-android.js
rpl: yeah, definitely, I should have run a grep :-)

tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -704,37 +704,99 @@ describe('util/extension-runners/firefox-android', () => {
          sinon.assert.calledOnce(anotherCallback);
        });
 
-    it('checks the granted android permissions on Android >= 21',
+    it('raised an error when unable to find an android version number',
```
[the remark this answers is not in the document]
rpl: thanks, that is exactly what I meant to write :-)

tests/unit/test-extension-runners/test.firefox-android.js
rpl: ah, great catch!

You are right, I forgot to re-add the await on the `assert.isRejected(..)` calls when I refactored the test to make it easier to test all the values that I wanted to test, the right way to use it is like we are using it in the existent test.adb.js tests:

https://github.com/mozilla/web-ext/blob/1b22d60aefa31e2869e15d18ec52b39afa0de405/tests/unit/test-util/test.adb.js#L106-L107

(unfortunately the current version of the chai-as-promised docs do not have any example that uses the async functions, on the contrary Jest provides a similar `rejects` assertion and it also provide an example of its usage in an async function: https://facebook.github.io/jest/docs/en/expect.html#rejects)

We can't use `await runnerInstance.run(...)` because we expect a rejection and the assertion would never be reached.

PR #1218, opened 2018-01-19
feat: Added package.json config to set option defaults

### LINKED ISSUE #1166: Apply package.json config to set option defaults

feature

You can define a custom config file named `web-ext-config.js` (as of https://github.com/mozilla/web-ext/issues/730) in your project directory to set option defaults.

It would be nice to specify a custom key in `package.json` like `webExt` that has configuration. This would cut down on the amount of config files littered in one's project directory.

For example, we want to support a case where a developer puts something like this in their `package.json`:

```json
{
  "name": "some-package",
  "version": "1.0.0",
  "webExt": {
    "sourceDir": "./src",
    "verbose": true,
    "run": {
      "firefox": "nightly"
    }
  }
}
```

This `package.json` file contains a custom key named `webExt` which has configuration data. We want to load that data as if it were in a config file. Read this to learn how to do it: https://stackoverflow.com/questions/10065564/add-custom-metadata-or-config-to-package-json-is-it-valid

Config tracker: https://github.com/mozilla/web-ext/issues/176

ChujunGeng:
Fixes #1166

I stumbled upon this repository while trying to get myself involved in open-source development. Wish I could be of any help.

Basically what I did was adding `./package.json` as an option for config files and creating a new function `loadJSONConfigFile()`, which is just a modified version of `loadJSConfigFile()`, to load config options from `./package.json`

As a result, I was able to load configs from `./package.json` under the `webExt` key. However, `discoverConfigFiles` tests are now failing because `./package.json` was not expected as an option.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -135,6 +135,24 @@ export function loadJSConfigFile(filePath: string): Object {
   return configObject;
 }
 
+export function loadJSONConfigFile(filePath: string): Object {
+  const resolvedFilePath = path.resolve(filePath);
+  log.debug(
+    `Loading JSON config file: "${filePath}" ` +
+    `(resolved to "${resolvedFilePath}")`);
+  let configObject;
+  try {
+    const topLevelConfig = requireUncached(resolvedFilePath);
+    configObject = topLevelConfig.webExt || {};
+  } catch (error) {
+    log.debug('Handling error:', error);
+    throw new UsageError(
+      `Cannot read config file: ${resolvedFilePath}\n` +
+      `Error: ${error.message}`);
+  }
+  return configObject;
+}
+
 type DiscoverConfigFilesParams = {|
   getHomeDir: () => string,
 |};
@@ -150,6 +168,8 @@ export async function discoverConfigFiles(
     path.join(getHomeDir(), `.${magicConfigName}`),
     // Look for a magic config in the current working directory.
     path.join(process.cwd(), magicConfigName),
+    // Look for webExt key in package.json file
+    path.join(process.cwd(), 'package.json'),
   ];
 
   const configs = await Promise.all(possibleConfigs.map(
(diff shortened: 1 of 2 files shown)


ChujunGeng:
Thanks for your feedback! I think that was what I implemented.

I added package.json as an option in the possibleConfigs, but created another function to look for the webExt key inside package.json, instead of using the original loadJSONConfigFile() function.

I tested it with this package.json file and was able to successfully load the configs in it:
{
  "name": "foo",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "path": "^0.12.7",
    "require-uncached": "^1.0.3"
  },
  **"webExt": {
    "verbose": true,
    "build": {
      "overwriteDest": true
    },
    "run": {
      "firefox": "nightly"
    }
  }**
}


The problem here is that, tests are failing because they don't expect package.json to be included in the possibleConfigs array. So I'm not sure if I should modify the test cases or to modify my implementation to make it compatible with existing tests.

src/config.js
```diff
@@ -135,6 +135,24 @@ export function loadJSConfigFile(filePath: string): Object {
   return configObject;
 }
 
+export function loadJSONConfigFile(filePath: string): Object {
```
kumar303: Adding a new function for this is not necessary. You can re-use the same config loading function. You may need to make some adjustments to `loadJSConfigFile` to handle the JSON.

src/program.js
```diff
@@ -191,7 +194,8 @@ export class Program {
       }
 
       configFiles.forEach((configFileName) => {
-        const configObject = loadJSConfigFile(configFileName);
+        const configObject = configFileName.endsWith('package.json') ?
```
kumar303: Since you don't need a custom loader you could use this opportunity to look for the `webExt` key:

```js
let configObject = loadJSConfigFile(configFileName);
if (configFileName.endsWith('package.json')) {
  log.debug('Looking for webExt key in package.json file');
  configObject = configObject.webExt || {};
}
```

ChujunGeng:
Thanks again for the great feedback! I'll get onto it right away.

Issue #1214, opened 2018-01-18
Verbose option ignored when declared in config file


dessant:
### Is this a feature request or a bug?
bug

### What is the current behavior?
web-ext-config.js
```js
module.exports = {
  verbose: true
};
```

```
$ web-ext run -s dist
Applying config file: ./web-ext-config.js
Running web extension from /media/s1/dev/pr/clear-browsing-data/dist
Use --verbose or open Tools > Web Developer > Browser Console to see logging
Installed /media/s1/dev/pr/clear-browsing-data/dist as a temporary add-on
The extension will reload if any source file changes
Press R to reload (and Ctrl-C to quit)
```

### What is the expected or desired behavior?
The process should switch to verbose output after the config file is parsed, or the limitation should be documented.

### Version information (for bug reports)

* **Firefox version**: 57.0.4
* **Your OS and version**: Arch Linux

````
$ node --version && npm --version && web-ext --version
v8.9.1
5.5.1
2.3.1

````

kumar303:
Oof, I don't know how we missed this. Thanks for the bug report. Fixed in https://github.com/mozilla/web-ext/pull/1221

PR #1221, opened 2018-01-19
fix: Do not treat an array of config values as a sub-command config section

### LINKED ISSUE #1213: Crash when pref option is an array in config file
bug

web-ext-config.js
```js
module.exports = {
  run: {
    pref: [
      'devtools.webconsole.persistlog=true',
      'extensions.webextensions.remote=true'
    ]
  }
};
```

```
$ web-ext run -s dist --verbose
[program.js][debug] Getting the version from package.json
[program.js][info] Version: 2.3.1
[program.js][debug] Getting the version from package.json
[program.js][debug] Discovering config files. Set --no-config-discovery to disable
[config.js][debug] Discovered config "/home/prc/.web-ext-config.js" does not exist or is not readable
[program.js][info] Applying config file: ./web-ext-config.js
[config.js][debug] Loading JS config file: "/media/s1/dev/pr/clear-browsing-data/web-ext-config.js" (resolved to "/media/s1/dev/pr/clear-browsing-data/web-ext-config.js")
[program.js][error] 
UsageError: The config file at /media/s1/dev/pr/clear-browsing-data/web-ext-config.js specified an unknown option: "0"
    at applyConfigToArgv (/media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/src/config.js:57:13)
    at applyConfigToArgv (/media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/src/config.js:45:17)
    at applyConfigToArgv (/media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/src/config.js:45:17)
    at /media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/src/program.js:195:24
    at Array.forEach (<anonymous>)
    at Program._callee$ (/media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/src/program.js:193:19)
    at tryCatch (/media/s1/dev/pr/clear-browsing-data/node_modules/regenerator-runtime/runtime.js:62:40)
    at Generator.invoke [as _invoke] (/media/s1/dev/pr/clear-browsing-data/node_modules/regenerator-runtime/runtime.js:296:22)
    at Generator.prototype.(anonymous function) [as next] (/media/s1/dev/pr/clear-browsing-data/node_modules/regenerator-runtime/runtime.js:114:21)
    at step (/media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:1)
    at /media/s1/dev/pr/clear-browsing-data/node_modules/web-ext/dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:1
    at <anonymous>

[program.js][debug] Command executed: run

```

The browser preferences should be applied without errors.

* **Firefox version**: 57.0.4
* **Your OS and version**: Arch Linux

````
v8.9.1
5.5.1
2.3.1

````

### PR DESCRIPTION
**Pull request by rpl**
This PR contains a set of proposed changes to fix #1213, #1214 and #1215 

#1213 is fixed by prevent `applyConfigToArgv` to recurse on an array of config values (by consider it as a sub-command config section by mistake)

#1214 is fixed by checking if `adjustedArgv.verbose` is `true` after we loaded and applied the config file and then calling `logStream.makeVerbose` accordingly.

#1215 proposed fix is (as describe in [this comment](https://github.com/mozilla/web-ext/issues/1215#issuecomment-359062606)) to mark explicitly as array the options that may makes sense as an array of values in the config file

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -401,7 +401,7 @@ Example: $0 --help run.
                   'run against multiple targets.',
         default: 'firefox-desktop',
         demand: false,
-        type: 'string',
+        type: 'array',
       },
       'firefox': {
         alias: ['f', 'firefox-binary'],
@@ -447,7 +447,7 @@ Example: $0 --help run.
                   'preference.',
         demand: false,
         requiresArg: true,
-        type: 'string',
+        type: 'array',
         coerce: coerceCLICustomPreference,
       },
       'start-url': {
@@ -455,7 +455,7 @@ Example: $0 --help run.
         describe: 'Launch firefox at specified page',
         demand: false,
         requiresArg: true,
-        type: 'string',
+        type: 'array',
       },
       'browser-console': {
         alias: ['bc'],
--- a/src/program.js
+++ b/src/program.js
@@ -50,6 +50,7 @@ export class Program {
   yargs: any;
   commands: { [key: string]: Function };
   shouldExitProgram: boolean;
+  verboseEnabled: boolean;
   options: Object;
 
   constructor(
@@ -69,6 +70,7 @@ export class Program {
     // config (See web-ext#469 for rationale).
     const yargsInstance = yargs(argv, absolutePackageDir);
 
+    this.verboseEnabled = false;
     this.shouldExitProgram = true;
     this.yargs = yargsInstance;
     this.yargs.strict();
@@ -121,6 +123,19 @@ export class Program {
     return this;
   }
 
+  enableVerboseMode(
+    logStream: typeof defaultLogStream,
+    version: string
+  ): void {
+    if (this.verboseEnabled) {
+      return;
+    }
+
+    logStream.makeVerbose();
+    log.info('Version:', version);
+    this.verboseEnabled = true;
+  }
+
   async execute(
     absolutePackageDir: string,
     {
@@ -135,21 +150,17 @@ export class Program {
       globalEnv = WEBEXT_BUILD_ENV,
     }: ExecuteOptions = {}
   ): Promise<void> {
-
     this.shouldExitProgram = shouldExitProgram;
     this.yargs.exitProcess(this.shouldExitProgram);
 
     const argv = this.yargs.argv;
     const cmd = argv._[0];
 
+    const version = getVersion(absolutePackageDir);
     const runCommand = this.commands[cmd];
 
-    let versionLogged = false;
-
     if (argv.verbose) {
-      logStream.makeVerbose();
-      log.info('Version:', getVersion(absolutePackageDir));
-      versionLogged = true;
+      this.enableVerboseMode(logStream, version);
     }
 
     let adjustedArgv = {...argv};
@@ -206,11 +217,8 @@ export class Program {
       });
 
       if (adjustedArgv.verbose) {
-        logStream.makeVerbose();
-
-        if (!versionLogged) {
-          log.info('Version:', getVersion(absolutePackageDir));
-        }
+        // Ensure that the verbose is enabled when specified in a config file.
+        this.enableVerboseMode(logStream, version);
       }
 
       await runCommand(adjustedArgv, {shouldExitProgram});
--- a/src/program.js
+++ b/src/program.js
@@ -144,10 +144,15 @@ export class Program {
 
     const runCommand = this.commands[cmd];
 
+    let versionLogged = false;
+
     if (argv.verbose) {
       logStream.makeVerbose();
       log.info('Version:', getVersion(absolutePackageDir));
+      versionLogged = true;
     }
+
+    let adjustedArgv = {...argv};
 
     try {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: tests/unit/test.config.js [hunk 1]
```diff
@@ -296,6 +296,30 @@ describe('config', () => {
...
+               type: 'string',
+             },
+           },
+         });
+
+         const configObject = {
+           pref: ['pref1=true', 'pref2=false'],
+         };
+
+         // TODO: expect a raised exception array is not a string
```
kumar303: What does this todo comment mean? Was there another test you wanted to add?

tests/unit/test.config.js
```diff
@@ -296,6 +296,30 @@ describe('config', () => {
...
     });
 
+    it('does not mistake an array config values for a sub-command',
+       () => {
+         const params = makeArgv({
+           userCmd: ['fakecommand'],
+           globalOpt: {
+             pref: {
+               demand: false,
+               type: 'string',
```
kumar303: Shouldn't this be `type: 'array'` so that it tests the bug you were fixing more directly? I'm a bit confused because the exception message in your assertion doesn't seem to match the spec of this test.

src/program.js
```diff
@@ -401,7 +401,7 @@ Example: $0 --help run.
                   'run against multiple targets.',
         default: 'firefox-desktop',
         demand: false,
-        type: 'string',
+        type: 'array',
```
kumar303: Can you also update the Flow type:

https://github.com/rpl/web-ext/blob/3848c777b290b4c5ed6ad1af0436afe5f952b005/src/cmd/run.js#L44

It should now just be `target?: Array<string>`.

src/program.js
```diff
@@ -447,15 +447,15 @@ Example: $0 --help run.
                   'preference.',
         demand: false,
         requiresArg: true,
-        type: 'string',
+        type: 'array',
```
kumar303: Could you also update this Flow type:

https://github.com/rpl/web-ext/blob/3848c777b290b4c5ed6ad1af0436afe5f952b005/src/firefox/preferences.js#L120

It should now just be `cliPrefs: Array<string>`. I think you may need to update some tests after that.

src/program.js
```diff
@@ -447,15 +447,15 @@ Example: $0 --help run.
...
         coerce: coerceCLICustomPreference,
       },
       'start-url': {
         alias: ['u', 'url'],
         describe: 'Launch firefox at specified page',
         demand: false,
         requiresArg: true,
-        type: 'string',
+        type: 'array',
```
kumar303: This Flow type needs an update too 

https://github.com/rpl/web-ext/blob/3848c777b290b4c5ed6ad1af0436afe5f952b005/src/cmd/run.js#L43

src/program.js
rpl: I applied these changes to the flow types in 
025db7c

rpl:
@kumar303 This should be ready for another review, besides applying the changes suggested in the previous review, I've also introduced a fix (and a new test case) for #1214 in 
6ded415

src/program.js
```diff
@@ -201,10 +205,18 @@ export class Program {
         });
       });
 
+      if (adjustedArgv.verbose) {
+        logStream.makeVerbose();
+
+        if (!versionLogged) {
+          log.info('Version:', getVersion(absolutePackageDir));
```
kumar303: Oof. This duplication is unfortunate. I can't think of another way to do it though. I suppose you could pull it into a shared function just for a bit more maintainability.

tests/unit/test.program.js
```diff
@@ -674,6 +674,39 @@ describe('program.main', () => {
...
+      lint: () => Promise.resolve(),
+    });
+
+    const customConfig = path.resolve('custom/web-ext-config.js');
+
+    const finalSourceDir = path.resolve('final/source-dir');
+    const loadJSConfigFile = makeConfigLoader({
+      configObjects: {
+        [customConfig]: {
+          sourceDir: finalSourceDir,
```
kumar303: This doesn't seem like it's necessary for the test. You could remove it along with the variable.

src/program.js
rpl: yeah, I agree, I've refactored the part that enable the verbose mode and print the version into a class method and reused it into the two places where we have to enable the verbose mode (when it is specified on the command line and when it is only specified in a config file).

Let me know how it looks to you.

tests/unit/test.program.js
rpl: 👍 removed

kumar303:
Let's ship it

Issue #877, opened 2017-03-23
web-ext sign fails if last version uploaded via website was listed


dstillman:
### Is this a bug or feature request?

Bug

### What is the current behavior?

I have an add-on with both an unlisted and a listed version, the latter awaiting review. If the last version uploaded via the website was listed, `web-ext sign` fails with a generic "The WebExtension could not be signed". Using verbose mode and loading the validation URL, I can see that the underlying error is `"applications.gecko.update_url" is not allowed for Mozilla-hosted add-ons.`. 

### What is the expected or desired behavior?

Using `web-ext sign` should always create an unlisted version, regardless of the last-used mode on AMO. (See also: https://github.com/mozilla/web-ext/issues/804 for adding a separate `web-ext submit`.)

### Version information (for bug reports)

* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````
v7.7.2
4.4.4
1.8.1

kumar303:
Huh, that's odd. Thanks for the bug report. If it's convenient for you to reproduce this, could you attach a verbose log?

dstillman:
Sure: https://gist.githubusercontent.com/dstillman/d9cf22f7ff6655b55bbe7a6faa92f8d6/raw/3d8582453b084ee5ccd03c0657ad8f317feab9f6/sign_failure.txt

kumar303:
Thanks. The response property `automated_signing: false` means that the API thinks it's dealing with a listed add-on which doesn't seem right.

kumar303:
@mstriemer do you think https://github.com/mozilla/addons-server/issues/958 would fix this?

mstriemer:
I think this is due to supporting listed and unlisted for the same add-on. It looks like a `channel` parameter has been added to the API which can be either `"listed"` or `"unlisted"`, if it isn't specified then last the last versions channel is used [1].

So I'm guessing web-ext needs to be updated to pass the channel flag when requested (or possibly just set it to unlisted unless it is called with `--listed` or something) .

[1] https://github.com/mozilla/addons-server/blob/38b2901df38372462c86574b06751e385e0f7688/src/olympia/signing/views.py#L189-L196

kumar303:
Oh, hey, there is even a [channel param for the API endpoint now](http://addons-server.readthedocs.io/en/latest/topics/api/signing.html#put--api-v3-addons-[string-addon-id]-versions-[string-version]-).

[...]

PR #1250, opened 2018-02-12
feat: Added release channel as optional sign parameter

mi-g:
Fixes https://github.com/mozilla/web-ext/issues/877

Added release channel as optional sign parameter.

To be effective, this requires a version of the [sign-addon](https://github.com/mozilla/sign-addon) module patched with PR [#249](https://github.com/mozilla/sign-addon/pull/249) or [#250](https://github.com/mozilla/sign-addon/pull/250).

### DIFF
--- a/tests/unit/test-cmd/test.sign.js
+++ b/tests/unit/test-cmd/test.sign.js
@@ -288,6 +288,34 @@ describe('sign', () => {
     }
   ));
 
+  it('passes the channel parameter to the signer', () => withTempDir(
+    (tmpDir) => {
+      const stubs = getStubs();
+      const artifactsDir = path.join(tmpDir.path(), 'some-artifacts-dir');
+      const applications: ExtensionManifestApplications =
+        stubs.preValidatedManifest.applications || {gecko: {}};
+      return sign(tmpDir, stubs, {extraArgs: {
+        artifactsDir,
+        channel: 'unlisted',
+      }})
+        .then(() => {
+          sinon.assert.called(stubs.signAddon);
+          sinon.assert.calledWithMatch(stubs.signAddon, {
+            apiKey: stubs.signingConfig.apiKey,
+            apiProxy: stubs.signingConfig.apiProxy,
+            apiSecret: stubs.signingConfig.apiSecret,
+            apiUrlPrefix: stubs.signingConfig.apiUrlPrefix,
+            downloadDir: artifactsDir,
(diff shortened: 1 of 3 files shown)


mi-g:
I added the test for channel parameter passing to the signer. I hope this is ok.

stoically:
It might be worth to consider merging this in the light of the soon to be deactivated beta channel on AMO, since it helps with getting signed xpis for the selfhosted channel on the commandline without having to use the AMO UI.

sneakypete81:
Is there anything preventing this from being merged @kumar303? 
Merging this would make hosting beta versions much more pleasant.

All I can think of is to reduce the number test assertions to only those that we care about - something more like [this test](https://github.com/mozilla/web-ext/blob/2d83d8c53306e82a4212415fb8ee995b091d10e0/tests/unit/test-cmd/test.sign.js#L302).

tests/unit/test-cmd/test.sign.js
```diff
@@ -288,6 +288,34 @@ describe('sign', () => {
...
+            apiKey: stubs.signingConfig.apiKey,
+            apiProxy: stubs.signingConfig.apiProxy,
+            apiSecret: stubs.signingConfig.apiSecret,
+            apiUrlPrefix: stubs.signingConfig.apiUrlPrefix,
+            downloadDir: artifactsDir,
+            id: applications.gecko.id,
+            timeout: stubs.signingConfig.timeout,
+            version: stubs.preValidatedManifest.version,
+            xpiPath: stubs.buildResult.extensionPath,
+            channel: 'unlisted',
```
kumar303: This is the only important part so you can remove all other parameter assertions.

sneakypete81:
> Other than that, we would need an issue on file so we can give it the needs: docs label.

How about #877?

kumar303:
Yep, that looks like the same issue. I updated the description so it will close that one. Thanks.

sneakypete81:
@mi-g - I hope you don't mind that I applied @kumar303's review comment in PR #1400. I'm keen to get this feature merged and released.

mi-g:
> @mi-g - I hope you don't mind that I applied @kumar303's review comment in PR #1400. I'm keen to get this feature merged and released.

No problem.

kumar303:
This patch was amended and merged via https://github.com/mozilla/web-ext/pull/1400

PR #1258, opened 2018-02-16
chore: Ignore Hawk security advisory because we don't use Hawk

kumar303:


.nsprc
```diff
@@ -1,4 +1,3 @@
 {
-  // https://github.com/nodesecurity/nsp#exceptions
```
kumar303: We can't use comments anymore. Boo. https://github.com/nodesecurity/nsp/issues/207

PR #1266, opened 2018-02-21
feat: added package.json as an option for configs

### LINKED ISSUE #1166: Apply package.json config to set option defaults

ChujunGeng:
Fixes #1166

This time I modified the `loadJSConfigFile()` function to parse package.json, and I also added test coverage for it.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -131,6 +131,10 @@ export function loadJSConfigFile(filePath: string): Object {
       `Cannot read config file: ${resolvedFilePath}\n` +
       `Error: ${error.message}`);
   }
+  if (filePath.endsWith('package.json')) {
+    log.debug('Looking for webExt key inside package.json file');
+    configObject = configObject.webExt || {};
+  }
   if (Object.keys(configObject).length === 0) {
     log.debug(`Config file ${resolvedFilePath} did not define any options. ` +
       'Did you set module.exports = {...}?');
@@ -151,6 +155,8 @@ export async function discoverConfigFiles(
   const possibleConfigs = [
     // Look for a magic hidden config (preceded by dot) in home dir.
     path.join(getHomeDir(), `.${magicConfigName}`),
+    // Look for webExt key inside package.json file
+    path.join(process.cwd(), 'package.json'),
     // Look for a magic config in the current working directory.
     path.join(process.cwd(), magicConfigName),
   ];
--- a/tests/unit/test.config.js
+++ b/tests/unit/test.config.js
@@ -905,12 +905,47 @@ describe('config', () => {
         });
     });
 
+    it('parses package.json file correctly', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'package.json');
+          fs.writeFileSync(
+            configFilePath,
+            `{
+                "name": "dummy-package-json",
+                "version": "1.0.0",
+                "webExt": {
+                  "sourceDir": "path/to/fake/source/dir"
+                }
+            }`
+          );
+          const configObj = loadJSConfigFile(configFilePath);
+          assert.equal(configObj.sourceDir, 'path/to/fake/source/dir');
+        });
+    });
+
     it('does not throw an error for an empty config', () => {
       return withTempDir(
         (tmpDir) => {
           const configFilePath = path.join(tmpDir.path(), 'config.js');
           fs.writeFileSync(configFilePath, 'module.exports = {};');
           loadJSConfigFile(configFilePath);
+        });
+    });
+
+    it('returns an empty object when webExt key is not in package.json', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'package.json');
+          fs.writeFileSync(
+            configFilePath,
+            `{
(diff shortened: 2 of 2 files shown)


src/config.js
```diff
@@ -153,6 +157,8 @@ export async function discoverConfigFiles(
     path.join(getHomeDir(), `.${magicConfigName}`),
     // Look for a magic config in the current working directory.
     path.join(process.cwd(), magicConfigName),
+    // Look for webExt key inside package.json file
+    path.join(process.cwd(), 'package.json'),
```
kumar303: I think we should move this up one position in the array so that it happens before `path.join(process.cwd(), magicConfigName)`. This will make `package.json` load before `web-ext-config.js`. My rationale for this is that `package.json` is a project-wide type of configuration file and `web-ext-config.js` is a config file that is specific only to `web-ext`.

tests/unit/test.config.js
```diff
@@ -905,6 +905,24 @@ describe('config', () => {
...
     });
 
+    it('parses package.json file correctly', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'package.json');
+          fs.writeFileSync(
+            configFilePath,
+            `{
+                "name": "dummyPackage.json",
```
kumar303: technically, a `version` is required to make this valid so I think you should add `"version": "1.0.0"`

tests/unit/test.config.js
```diff
@@ -913,6 +931,21 @@ describe('config', () => {
...
     });
+
+    it('returns an empty object when webExt key is not in package.json', () => {
+      return withTempDir(
+        (tmpDir) => {
+          const configFilePath = path.join(tmpDir.path(), 'package.json');
+          fs.writeFileSync(
+            configFilePath,
+            `{
+              "name": "dummyPackage.json"
```
kumar303: `version` should be added here too

tests/unit/test.config.js
```diff
@@ -927,6 +960,7 @@ describe('config', () => {
     it('finds a config in your home directory', () => {
       return withTempDir(
         async (tmpDir) => {
+          const packageJSON = path.join(process.cwd(), 'package.json');
```
kumar303: Instead of editing this test, you should make a new one. The test you edited is specific to checking that a config in your home directory gets loaded. You need a new test specific to checking that a `package.json` file gets loaded. The test will probably look more like [this one](https://github.com/ChujunGeng/web-ext/blob/204fed982bd8333a25dd33c6b19de17c6374d77f/tests/unit/test.config.js#L979).

ChujunGeng: I edited this test because web-ext itself's `package.json` will be loaded (apparently because it's inside current working directory), and that's not expected by the original test. Unless we cd to a temporary directory, and make another temporary directory home directory to test out `.web-ext-config.js`, this test will always fail for the aforementioned reason.

src/config.js
ChujunGeng: That makes sense. I'll change that asap.

tests/unit/test.config.js
kumar303: Oh, I see. It would be helpful if you can add a comment to this test explaining the situation.

tests/unit/test.config.js
```diff
@@ -927,6 +962,10 @@ describe('config', () => {
     it('finds a config in your home directory', () => {
       return withTempDir(
         async (tmpDir) => {
+          // This is actually web-ext itself's package.json file, which
```
kumar303: Thanks for this comment.

caitmuenster:
Thanks so much, @ChujunGeng! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#March_2018). 

Would you be interested in creating an account on [mozillians.org](https://mozillians.org/)? I'd be happy to vouch for your contribution! 

Thanks again, and welcome onboard!

ChujunGeng:
@caitmuenster Thanks! I'm honored to join the mozillians, and I just created an account on it ([https://mozillians.org/en-US/u/chujungeng/](https://mozillians.org/en-US/u/chujungeng/)).

caitmuenster:
\o/ Yay! Your profile is vouched. 

We're happy to have you with us! I look forward to seeing you around the project. :)

freaktechnik:
[dev-doc-needed]

Issue #1282, opened 2018-03-07
web-ext run broken with Firefox Nightly 60.0a1 (2018-03-07)


eeejay:
### Is this a feature request or a bug?
Bug

### What is the current behavior?
I get the following stack:
```
WebExtError: Remote Firefox: listTabs() error: unknownError: error occurred while processing 'listTabs: TypeError: this.mPanelContainer is null
Stack: TabBrowser@chrome://browser/content/tabbrowser.js:25:5
@chrome://browser/content/browser.js:148:46
get@resource://gre/modules/XPCOMUtils.jsm:194:21
BrowserTabList.prototype._getSelectedBrowser@resource://devtools/shared/base-loader.js -> resource://devtools/server/actors/webbrowser.js:223:3
BrowserTabList.prototype.getList@resource://devtools/shared/base-loader.js -> resource://devtools/server/actors/webbrowser.js:264:23
onListTabs@resource://devtools/shared/base-loader.js -> resource://devtools/server/actors/root.js:309:27
Async*onPacket@resource://devtools/shared/base-loader.js -> resource://devtools/server/main.js:1762:15
_onJSONObjectReady/<@resource://devtools/shared/base-loader.js -> resource://devtools/shared/transport/transport.js:482:11
exports.makeInfallible/<@resource://devtools/shared/base-loader.js -> resource://devtools/shared/ThreadSafeDevToolsUtils.js:109:14
exports.makeInfallible/<@resource://devtools/shared/base-loader.js -> resource://devtools/shared/ThreadSafeDevToolsUtils.js:109:14
Line: 25, column: 5
    at /opt/node/lib/node_modules/web-ext/dist/webpack:/src/firefox/remote.js:97:25
    at /opt/node/lib/node_modules/web-ext/node_modules/firefox-client/lib/client-methods.js:73:9
    at Client.handleMessage (/opt/node/lib/node_modules/web-ext/node_modules/firefox-client/lib/client.js:161:7)
    at Client.readMessage (/opt/node/lib/node_modules/web-ext/node_modules/firefox-client/lib/client.js:220:10)
    at Client.onData (/opt/node/lib/node_modules/web-ext/node_modules/firefox-client/lib/client.js:186:16)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at addChunk (_stream_readable.js:250:12)
    at readableAddChunk (_stream_readable.js:237:11)
    at Socket.Readable.push (_stream_readable.js:195:10)
[...]

PR #1284, opened 2018-03-13
fix: Fixed "web-ext run" regression on Firefox 60

chrmod:
closes: https://github.com/mozilla/web-ext/issues/1282

by implementing:
https://github.com/mozilla/web-ext/issues/1282#issuecomment-371538235

cc: @rpl

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,8 @@
     "opera"
   ],
   "dependencies": {
+    "@cliqz-oss/firefox-client": "0.3.1",
+    "@cliqz-oss/node-firefox-connect": "1.2.1",
     "adbkit": "2.11.0",
     "addons-linter": "0.36.1",
     "babel-polyfill": "6.26.0",
@@ -65,7 +67,6 @@
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
     "mz": "2.7.0",
-    "node-firefox-connect": "1.2.0",
     "node-notifier": "5.2.1",
     "open": "0.0.5",
     "parse-json": "4.0.0",
@@ -104,7 +105,6 @@
     "eslint-plugin-async-await": "0.0.0",
     "eslint-plugin-flowtype": "2.46.1",
     "eslint-plugin-import": "2.8.0",
-    "firefox-client": "0.3.0",
     "flow-bin": "0.65.0",
     "grunt": "1.0.2",
     "grunt-contrib-clean": "1.1.0",
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -1,7 +1,7 @@
 /* @flow */
-import defaultFirefoxConnector from 'node-firefox-connect';
+import defaultFirefoxConnector from '@cliqz-oss/node-firefox-connect';
 // RemoteFirefox types and implementation
-import type FirefoxClient from 'firefox-client'; //eslint-disable-line import/no-extraneous-dependencies
+import type FirefoxClient from '@cliqz-oss/firefox-client';
 
 import {createLogger} from '../util/logger';
 import {


rpl:
@chrmod Thanks a lot for working on a fix for this issue, this is basically the approach that we were already going to follow as a first step (fixing the regression in a minor release by publishing a namespaced version of the package with the fix applied).

Do you mind to also push the changes applied to both these npm packages into the related github forked repos?
(https://github.com/cliqz-oss/node-firefox-connect and https://github.com/cliqz-oss/firefox-client/tree/master)

I'm pretty sure that they are composed by a bump of the package version and the dependency firefox-client on node-firefox-connect, and the fix and bump of the package version on firefox-client but, while we prepare to migrate away from these deprecated dependencies, it would still be better if we could also keep track of the fixes we needed and applied on their forked version.

rpl:
@kumar303 I just looked into the diff between the old "node-firefox-connect / firefox-client" dependencies and the new "@cliqz-oss/node-firefox-connect / @cliqz-oss/firefox-client" and the following are the actual differences:

- node-firefox-connect => @cliqz-oss/node-firefox-connect
  - package.json: changes related to the package name, the package version and the list of the package mantainers and the @cliqz-oss/firefox-client dependency
  - index.js: changes related to require @cliqz-oss/firefox-client dependency (instead of the previous non-namespaced one)

- firefox-client => @cliqz-oss/firefox-client
  - package.json: changes related to the package name, the package version and the list of the package mantainers
  - lib/client.js: applied the fix suggested in https://github.com/mozilla/web-ext/issues/1282#issuecomment-371538235
  - lib/dom.js, lib/domnode.js and lib/stylesheets.js: added for some additional RDP request types (not needed to fix web-ext run on Nightly, but they should not do any harm neither)
  
The Travis Job is currently failing because of the commit message which doesn't pass the changelog-lint rules, but besides that all the tests are passing as expected.

:+1: from me on merging this PR and release a minor bugfix release.

chrmod:
@rpl I've pushed code to forked repos (forgot to that in the earlier, have just pushed the tags). The commit message is also updated so linter check should pass this time. 
Please let me know if there is anything else I can do to help with this one.

package.json
```diff
@@ -104,7 +104,7 @@
     "eslint-plugin-async-await": "0.0.0",
     "eslint-plugin-flowtype": "2.46.1",
     "eslint-plugin-import": "2.8.0",
-    "firefox-client": "0.3.0",
+    "@cliqz-oss/firefox-client": "0.3.1",
```
kumar303: Can you move this to `dependencies` and re-enable the lint check by removing `/* eslint-disable-line import/no-extraneous-dependencies */` ?

I know it's not your fault but I just noticed it -- I think that lint rule was skipped by mistake. It only works by chance because of how npm/yarn flattens modules.

chrmod: Can do an update, but I don't think it is a mistake. `firefox-client` is a devDependency as it is only imported for flow to figure out the types. As flow syntax gets removed during build so is the import and the `firefox-client` is not required by `web-ext` directly. Does this explanation make sense?

rpl: @chrmod yeah, you are right, the historical reasons are definitely that:
- we started to use the flow "import types" feature to have an higher coverage of the flow type checks
- later (in a different pull request) we added the additional eslint rules that checks the imports, eslint started to complain about that import and the 'eslint-disable-line' has been used as a workaround

Anyway, I agree with Kumar, I would probably prefer an additional direct dependency (in any case the dependency is already there, even if it is not currently used directly) and avoid the eslint-disable-line workaround if possible.

Do you mind to also move the new namespaced dependencies on the top of the dependencies?
```
   "dependencies": {
    "@cliqz-oss/firefox-client": "0.3.1",
    "@cliqz-oss/node-firefox-connect": "1.2.1",
     "adbkit": "2.11.0",
    ...
```

(so that the next 'npm install --save/--save-dev is not going to have to reorder them in an unrelated change)

Thanks again for your help on quickly fixing this regression.

chrmod:
@kumar303 @rpl have pushed another update.

PR #1290, opened 2018-03-16
fix: Set langpack and selfHosted values

### LINKED ISSUE #1289: web-ext.cmd.lint seems to be broken
When I run the below code, I got 'Not enough non-option arguments: got 0, need at least 1' error.   Though it might be a yargs bug, passing 'langpack' option to linter.createInstance() solved the error. 

```
#!/usr/bin/env node

const webExt = require('web-ext').default;

webExt.cmd.lint({
  sourceDir: '/home/ikezoe/animation-icon',
  boring: false,
  output: 'text',
  pretty: false,
  selfHosted: false,
  warningsAsErrors: false,
}, {
  shouldExitProgram: false,
});
```

hiikezoe:
Fixes #1289

### DIFF
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -69,7 +69,7 @@ export default function lint(
     output,
     pretty,
     sourceDir,
-    selfHosted,
+    selfHosted = false,
     verbose,
     warningsAsErrors,
   }: LintCmdParams,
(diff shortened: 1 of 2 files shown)


hiikezoe:
Does the failure mean the PR title should have prefix?

hiikezoe:
Or setting the default value to the ctor of ManifestJSONParser instead of calling getConfig()?   I thought the reason why getConfig() is used there is to avoid scattering the default values, but if it's OK, it would be a cleaner way.   I am fine with either way.

hiikezoe:
@kumar303 any suggestions on this?

hiikezoe:
Let's re-run.

hiikezoe:
I am going to fix addons-linter. https://github.com/mozilla/addons-linter/pull/1926

PR #1292, opened 2018-03-19
chore: Drop unused getFakeFirefox and getFakeRemoteFirefox

hiikezoe:
Actually MultiExtensionRunnerParams does not have such parameters at all.


hiikezoe:
@kumar303 r?

PR #1294, opened 2018-03-19
fix: Allow comments in manifest.json

### LINKED ISSUE #1285: manifest.json parser too strict
The manifest.json is parsed using the `json-parser` package, which emits nicer errors, but does not support comments. This was discovered with an error when trying to sign an extension with comments in the manifest by @AntoineTurmel.

The parser should be able to handle comments in the manifest.json file.

hiikezoe:
Fixes #1285.

### DIFF
--- a/tests/unit/test-util/test.manifest.js
+++ b/tests/unit/test-util/test.manifest.js
@@ -158,6 +158,38 @@ describe('util/manifest', () => {
       }
     ));
 
+    it('allows comments in manifest JSON', () =>
+      withTempDir(async (tmpDir) => {
+        const manifestWithComments = `{
+          "name": "the extension",
+          "version": "0.0.1" // comments
+        }`;
+        const manifestFile = path.join(tmpDir.path(), 'manifest.json');
+        await fs.writeFile(manifestFile, manifestWithComments);
+        const manifestData = await getValidatedManifest(tmpDir.path());
+
+        assert.deepEqual(manifestData, manifestWithoutApps);
+      })
+    );
+
+    it('reports an error with line number in manifest JSON with comments', () =>
+      withTempDir(async (tmpDir) => {
+        const invalidManifestWithComments = `{
+          // a comment in its own line
+          // another comment on its own line
+          "name": "I'm an invalid JSON Manifest
+        }`;
+        const manifestFile = path.join(tmpDir.path(), 'manifest.json');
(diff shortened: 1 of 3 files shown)


tests/unit/test-util/test.manifest.js
```diff
@@ -158,6 +158,21 @@ describe('util/manifest', () => {
       }
     ));
 
+    it('allows comments in manifest JSON', () => withTempDir(
+      (tmpDir) => {
```
rpl: We should also add an additional test that asserts that an error raised in a manifest with comments is still reported on the expected line number, we can probably use an assertion similar to the one used in the other test above:

```
assert.include(error.message, 'Unexpected token  in JSON at position 49');
```

and use a manifest with a comment that would potentially change the line, e.g. using something like:

```js
const invalidManifestWithComments = `{
  // a comment in its own line
// another other comment on its own line
  "name": "the extension
}`;
```

Besides that, do you mind to rewrite these new test cases using an async function? it would make these new tests much more readable (don't worry about the other test cases in this file, there's no need to rewrite all of them in this pull request, but it would be nice to don't add more of them).

it should be something like:

```js
it('allows comments in manifest JSON', () => 
  withTempDir(async (tmpDir) => {
    ...
    await fs.writefile(manifestfile, manifestWithComments);
    const manifestData = await getValidatedManifest(tmpDir.path());
    ...
  }));
```

hiikezoe: Fixed.  Your review comment is quite descriptive, so it's easy to understand for beginners in this area like me. :)
Thank you!

tests/unit/test-util/test.manifest.js
```diff
@@ -158,6 +158,40 @@ describe('util/manifest', () => {
...
+    it('reports an error with line number in manifest JSON with comments', () =>
+      withTempDir(async (tmpDir) => {
+        const invalidManifestWithComments = `{
+          // a comment in its own line
+          // another comment on its own line
+          "name": "I'm an invalid JSON Manifest
+        }`;
+        const manifestFile = path.join(tmpDir.path(), 'manifest.json');
+        await fs.writeFile(manifestFile, invalidManifestWithComments);
+        await getValidatedManifest(tmpDir.path())
```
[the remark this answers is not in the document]
hiikezoe: oh neat.  I was thinking promise.catch looks nicer than try..catch but still bit awkward.  yeah this version looks quite nice.

PR #1293, opened 2018-03-19
fix: Fix the second argument of TempAddonCallback

hiikezoe:
It shoudl be a string.

This will fix #1129 partially.

PR #1295, opened 2018-03-22
chore(package): update flow-bin to version 0.68.0 and fixed new related errors

rpl:
This pull request supersedes #1259 by updating flow-bin to version 0.68.0 and contextually fixing a new flow check error that the previous flow version wasn't yet able to catch (basically flow 0.68.0 is figuring out the `stdin.isTTY` is going to be available only if we already checked that it is an instance of `tty.ReadStream`, which is actually true and so reversing the order of those checks fixes the flow error).

Thanks to @hiikezoe for fixing in #1292 and #1293 the other two issues that were making the flow-check step to fail with flow-bin 0.68.0 (flow is definitely becoming smarter and it is able to catch more type issues at every new version, and it is definitely worth to update it sooner than later).

(there is also a small change, in a separate commit, applied to src/utils/temp-dir.js which is actually unrelated, but given that it is just some minor coding style fixes related to empty lines we could just "rebase and merge" this pull request to fix both, and still keeping them into two separate commits).

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "eslint-plugin-async-await": "0.0.0",
     "eslint-plugin-flowtype": "2.46.1",
     "eslint-plugin-import": "2.8.0",
-    "flow-bin": "0.65.0",
+    "flow-bin": "0.68.0",
     "grunt": "1.0.2",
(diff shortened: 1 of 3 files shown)


hiikezoe:
\o/

Issue #1129, opened 2017-11-03
Issues running web-ext tests on node >= 8


rpl:
### Is this a feature request or a bug?

Bug

### What is the current behavior?

Running the tests on node >= 8 currently fails due to the following errors:

- 'RangeError [ERR_INVALID_FD]: "fd" must be a positive integer: undefined' errors raised
  while running the following unit tests:
  - tests/unit/test-extension-runners/test.extension-runners.js
  - tests/unit/test-util/test.logger.js

```
 RangeError [ERR_INVALID_FD]: "fd" must be a positive integer: undefined
      at new ReadStream (tty.js:41:11)
      at _callee10$ (dist/webpack:/tests/unit/test-extension-runners/test.extension-runners.js:489:25)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:117:21)
      at step (dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:1)
      at dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:35:1
      at new Promise (<anonymous>)
      at new F (dist/webpack:/node_modules/core-js/library/modules/_export.js:35:1)
      at Context.<anonymous> (dist/webpack:/node_modules/babel-runtime/helpers/asyncToGenerator.js:14:1)
```

- 'TypeError: "cwd" must be a string' errors raised from the functional tests

```
TypeError: "cwd" must be a string
      at normalizeSpawnArguments (child_process.js:403:11)
      at exports.spawn (child_process.js:496:38)
      at execWebExt (dist/webpack:/tests/functional/common.js:97:26)
      at dist/webpack:/tests/functional/test.cli.build.js:12:20
      at dist/webpack:/tests/functional/common.js:52:16
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

The complete failures logs are available in the following travis job:

- https://travis-ci.org/rpl/web-ext/jobs/296748492

### What is the expected or desired behavior?

The tests (and web-ext) should be able to run successfully on node >= 8

### Version information (for bug reports)

OS: any
node: v8.9.0
npm: 5.5.1

hiikezoe:
[...]

PR #1310, opened 2018-04-30
fix(test): Fixed remaining issues for run test successfully on node >= 8

rpl:
This PR contains some additional changes to the test units needed to successfully run all the tests on node >= 8 (and fix the remaining issues from #1129).

In particular it fixes the ` RangeError [ERR_INVALID_FD]: "fd" must be a positive integer: undefined` errors due to the ReadStream/WriteStream instances created in tests (e.g. to mock the stdin stream).

On this fixes we currently need to suppress some flow type errors (due to the nodejs type definitions already defined in flow not yet including this changes between node6 and node8).

Once ReadStream/WriteStream type definitions are updated in flow itself, flow is going to complain that these suppress comments are not suppressing any actual error and we will know that we can remove the suppress comments from there.

### DIFF
--- a/tests/unit/test-extension-runners/test.extension-runners.js
+++ b/tests/unit/test-extension-runners/test.extension-runners.js
@@ -1,7 +1,5 @@
 /* @flow */
-
 import stream from 'stream';
-import tty from 'tty';
 
 import {describe, it} from 'mocha';
 import {assert} from 'chai';
@@ -15,6 +13,7 @@ import {
 } from '../../../src/extension-runners';
 import {
   FakeExtensionRunner,
+  FakeStdin,
 } from '../helpers';
 import type {
   IExtensionRunner, // eslint-disable-line import/named
@@ -488,7 +487,7 @@ describe('util/extension-runners', () => {
     it('can reload when user presses R in shell console', async () => {
       const {extensionRunner, reloadStrategy} = prepare();
 
-      const fakeStdin = new tty.ReadStream();
+      const fakeStdin = new FakeStdin();
       sinon.spy(fakeStdin, 'setRawMode');
       sinon.spy(extensionRunner, 'reloadAllExtensions');
 
@@ -510,7 +509,7 @@ describe('util/extension-runners', () => {
       const {extensionRunner, reloadStrategy} = prepare();
       sinon.spy(extensionRunner, 'registerCleanup');
 
-      const fakeStdin = new tty.ReadStream();
+      const fakeStdin = new FakeStdin();
       sinon.spy(fakeStdin, 'pause');
       sinon.spy(fakeStdin, 'setRawMode');
 
@@ -537,7 +536,7 @@ describe('util/extension-runners', () => {
            },
          });
 
-         const fakeStdin = new tty.ReadStream();
+         const fakeStdin = new FakeStdin();
          sinon.spy(fakeStdin, 'setRawMode');
 
          try {
@@ -567,7 +566,8 @@ describe('util/extension-runners', () => {
              async exit() {},
            },
          });
-         const fakeStdin = new tty.ReadStream();
+
+         const fakeStdin = new FakeStdin();
 
          try {
            await reloadStrategy({}, {stdin: fakeStdin});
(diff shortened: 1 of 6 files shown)


rpl:
@Rob--W do you have a Windows machine nearby to you? I'm currently traveling and I'd like to double-check if this failures on appveyor are also reproducible when running the web-ext test suite locally or if they are somehow related to something specific to the appveyor CI environment:

```
...
Error: EBADF: bad file descriptor, uv_tty_init
      at FakeWritableStream.WriteStream (tty.js:71:13)
...
```
(e.g. https://ci.appveyor.com/project/kumar303/web-ext/build/1.0.1629/job/xyobnujidha90k8l#L746)

Rob--W:
@rpl Reprodued in a Windows 7 VM. Investigating.

Rob--W:
Oh duh. I don't need to debug further. I noticed that you're trying to instantiate a `WriteStream` for file descriptor 0 (stdin). Try 1 (stdout) instead.

```javascript
new (require("tty").WriteStream)(0) // EBADF: bad file descriptor, uv_tty_init
new (require("tty").WriteStream)(1) // OK.
```

yan12125:
Oh, seems my previous attempt at https://github.com/yan12125/web-ext/commit/51cdf95591b455c35dd1737ea9771ff3c6e29964#diff-556363c9e4a07cd7d847e92df7a3d30c is indeed broken :)

tests/unit/test-util/test.logger.js
```diff
@@ -43,7 +43,8 @@ describe('logger', () => {
         }
       }
 
-      const fakeWritableStream = new FakeWritableStream();
+      // $FLOW_FIXME: flow doesn't yet recognize 0 as a valid parameter of WriteStream.
+      const fakeWritableStream = new FakeWritableStream(0);
```
Rob--W: This 0 should be 1 (don't forget the comment above).

rpl:
@Rob--W :+1: Thanks Rob! That definitely makes sense.

I'm still on the plane, but it looks that I'm already able to push an update to github and let appveyor re-run the job with the fix applied.

Rob--W:
@rpl Oddly, Appveyor somehow failed again. When I run the unit tests locally with `1` instead of `0` (after building and manually then running `node_modules\.bin\mocha dist\unit-tests.js -f logger`), it passes (Win7, Node.js 8.11.1).

yan12125:
I did some more tests on Windows. With node.js 8.11.1 x64 on my Windows 7 VM (in VirtualBox), the following command succeeded:
```
npm test
```
while the following failed:
```
npm test > out
```
In ```out```, there are 8 EBADF errors, just like the case on AppVeyor. I guess the later redirects stdout, too.

rpl:
ah! Finally I got back home, "I discussed the issue a bit with Windows" and we reached a "reasonable understanding" ;-)

The only remaining issue was related to the `FakeWritableStream` class used by the "test.logger.js" unit test as a fake stdout stream:

In the last pushed version `FakeWritableStream` extends `stream.Writable` and defines a fake write method on it (`stream.Writable` doesn't implement it, it just provides the interface that the various Writable streams should provide): https://github.com/mozilla/web-ext/pull/1310/files#diff-556363c9e4a07cd7d847e92df7a3d30cR40

This is enough for the purpose of these unit tests, and it also makes Windows "happy" again.

Thanks a lot @Rob--W and @yan12125, your help have been very valuable to speed up the "Windows"-related investigations and then fix this "test-only" issue in a way that works across the supported platforms.

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -488,7 +488,8 @@ describe('util/extension-runners', () => {
     it('can reload when user presses R in shell console', async () => {
       const {extensionRunner, reloadStrategy} = prepare();
 
-      const fakeStdin = new tty.ReadStream();
+      // $FLOW_FIXME: flow doesn't yet recognize 0 as a valid parameter of tty.ReadStream.
+      const fakeStdin = new tty.ReadStream(0);
```
Rob--W: Wouldn't it be easier to maintain if you create a new `FakeReadableStream` or `FakeStdin` class?

rpl: :+1: good point, let's move it into the test helpers module.

PR #1319, opened 2018-05-17
replace open with opn

ccarruitero:
open has a [security vulnerability](https://nodesecurity.io/advisories/663) and seems to be unmaintained.

Since is just used to open docs I think couldn't be a problem to change for an alternative package.

http://www.npmtrends.com/open-vs-opener-vs-opn
https://npmcompare.com/compare/open,opener,opn

Currently [opn](https://www.npmjs.com/package/opn) seems to be better maintained package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "mkdirp": "0.5.1",
     "mz": "2.7.0",
     "node-notifier": "5.2.1",
-    "open": "0.0.5",
+    "opn": "^5.3.0",
     "parse-json": "4.0.0",
     "regenerator-runtime": "0.11.1",
     "require-uncached": "1.0.3",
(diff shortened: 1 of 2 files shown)


Rob--W:
The security vulnerability in `open` is only relevant if the input contains potentially unsafe characters.

In our code base, the input is a static, safe string:

https://github.com/mozilla/web-ext/blob/0865b9f4b5b78019ebcf6c025ba5e94ea6b9f423/src/cmd/docs.js#L17-L24

Nevertheless, replacing `open` with `opn` seems like a good idea, since the latter is maintained and still a small package with few dependencies.

rpl:
@ccarruitero thank you very much! I'm closing this in favor of #1323, which includes your change (with the small changes described in the pull request description) and also updates the other dependencies that are making nsp check to raise errors on every travis job.

PR #1323, opened 2018-05-23
fix: nsp check errors on travis (related to open, sign-addon and fx-runner)

rpl:
This PR updates the npm dependencies that are currently detected by nsp check as potential vulnerabilities
(which makes every travis job to fail before running any actual test).

This PR also contains the patch from #1319 with just two small changes:
- the exact version on the opn dependency (needed so that greenkeeper can create PR once new versions of that package are released on npm)
- small change on the commit message (so that i follows conventional-changelog conventions)

On top of that patch this PR also updates:
- fx-runner to version 1.0.9 (which fixes the nsp check errors related to lodash)
- sign-addon to version 0.3.1 (which fixes nsp check/npm audit errors related to hoek)


rpl:
Besides being finally back to green on travis (yay! \o/), I also manually verified locally that the docs, run and sign commands are working correctly with the updated dependencies.

PR #1330, opened 2018-06-05
Drop support for Node 6

ExE-Boss:
addons-linter 1.0 has dropped support for Node 6, which is causing recent builds to fail (especially #1326), this drops support for it.


rpl:
Closing, as nodejs 6 support has been dropped as part #1490.

PR #1338, opened 2018-06-25
feat: Allowed detection of given firefox APK

khaled-cliqz:
This commit is to allow extension development on custom forks of
firefox for android, not just on firefox and fennec

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -99,7 +99,8 @@ export default class ADBUtils {
   }
 
   async discoverInstalledFirefoxAPKs(
-    deviceId: string
+    deviceId: string,
+    firefoxApk?: string
   ): Promise<Array<string>> {
     log.debug(`Listing installed Firefox APKs on ${deviceId}`);
 
@@ -110,6 +111,11 @@ export default class ADBUtils {
     return pmList.split('\n')
       .map((line) => line.replace('package:', '').trim())
       .filter((line) => {
+        // Look for an exact match if firefoxApk is defined.
+        if (firefoxApk) {
+          return line === firefoxApk;
+        }
+        // Match any package name that starts with the package name of a Firefox for Android browser.
         return (
           line.startsWith('org.mozilla.fennec') ||
             line.startsWith('org.mozilla.firefox')
--- a/tests/unit/test-util/test.adb.js
+++ b/tests/unit/test-util/test.adb.js
@@ -15,6 +15,8 @@ import ADBUtils from '../../../src/util/adb';
 const fakeADBPackageList = `
 package:org.mozilla.fennec
 package:org.mozilla.firefox
+package:com.some.firefox.fork
+package:com.some.firefox.fork2
 package:org.some.other.software
 `;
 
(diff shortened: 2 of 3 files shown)


khaled-cliqz:
@rpl, could you please review this, and let me know if there's any questions or need any changes?

src/util/adb.js
```diff
@@ -111,8 +112,12 @@ export default class ADBUtils {
       .map((line) => line.replace('package:', '').trim())
       .filter((line) => {
         return (
-          line.startsWith('org.mozilla.fennec') ||
-            line.startsWith('org.mozilla.firefox')
+          firefoxApk ?
+            line.startsWith(firefoxApk) :
```
Rob--W: Use `line === firefoxApk` instead of `line.startsWith(firefoxApk)`.
In practice there is no difference because the list is filtered again in the end, but let's be more explicit that this an exact match is intended:
https://github.com/mozilla/web-ext/blob/f9db35b30a7477cadd3f9101e1ee593efb2fe875/src/extension-runners/firefox-android.js#L340-L367

tests/unit/test-util/test.adb.js
```diff
@@ -236,6 +237,52 @@ describe('utils/adb', () => {
...
     });
+
+    it('resolves only the given firefox APK', async () => {
+      const adb = getFakeADBKit({
+        adbClient: {
+          shell: sinon.spy(() => Promise.resolve('')),
+        },
+        adbkitUtil: {
+          readAll: sinon.spy(() => {
+            return Promise.resolve(new Buffer(fakeADBPackageList));
```
Rob--W: The `Buffer` constructor is deprecated. Please use `Buffer.from(fakeADBPackageList)` instead.

tests/unit/test-util/test.adb.js
```diff
@@ -15,6 +15,7 @@ import ADBUtils from '../../../src/util/adb';
 const fakeADBPackageList = `
 package:org.mozilla.fennec
 package:org.mozilla.firefox
+package:com.some.firefox.fork
```
Rob--W: For test coverage of prefix matching vs exact matching, please add another line such as:
`package:com.some.firefox.fork2`

tests/unit/test-util/test.adb.js
```diff
@@ -236,6 +237,52 @@ describe('utils/adb', () => {
...
+    });
+
+    it('resolves the given custom firefox fork APK', async () => {
+      const adb = getFakeADBKit({
+        adbClient: {
+          shell: sinon.spy(() => Promise.resolve('')),
+        },
+        adbkitUtil: {
+          readAll: sinon.spy(() => {
+            return Promise.resolve(new Buffer(fakeADBPackageList));
```
Rob--W: Again, use the `Buffer.from` method.

tests/unit/test-util/test.adb.js
khaled-cliqz: I don't think I need to, I'm going to use 2 test cases one with the full package name and assert that it resolves to something, and another with not the full package name and assert that it resolves to nothing

khaled-cliqz: oh, I get it now, will do 👍

tests/unit/test-util/test.adb.js
```diff
@@ -236,6 +238,52 @@ describe('utils/adb', () => {
       sinon.assert.calledOnce(adb.util.readAll);
       assert.deepEqual(packages, ['org.mozilla.fennec', 'org.mozilla.firefox']);
     });
+
+    it('resolves only the given firefox APK', async () => {
```
rpl: It seems that these two new test cases are both testing same "exact match" scenario, it doesn't seem that we actually need both (at least if I'm not missing something that makes these two tests actually different).

If we feel that we need to test explicitly this scenario on both "org.mozilla.fennec" and "com.some.firefox.fork", we could at least extract one of these test functions into an helper function which takes the package name as a parameter and then we can call it for both these package names in the same test case.

@khaled-cliqz how that sounds to you?

khaled-cliqz:
@rpl, could you help me with figuring out why this PR is failing?
I think that the test failing is unrelated:
`should accept: --source-dir SRCDIR --api-url-prefix URL`

Rob--W:
@khaled-cliqz The failing test (`web-ext sign`) is a known issue, and @rpl is already looking into it.

rpl:
@khaled-cliqz yeah, that is an unrelated intermittent, I'm going to take another look on this PR asap (and don't worry about that failure, I'm already fighting against it separately from this review).

khaled-cliqz:
@Rob--W, @rpl, thanks guys and best of luck 👍

caitmuenster:
Thank you, @khaled-cliqz! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#July_2018). 

Would you be interested in creating an account on [mozillians.org](https://mozillians.org/)? I'd be happy to vouch for you!

khaled-cliqz:
Thanks, @caitmuenster!
That would be great, here's the link: https://mozillians.org/de/u/BezrA/

caitmuenster:
Yay! Your profile is vouched. 

Thanks again for the patch, @khaled-cliqz! Hope to see you around the project. :)

PR #1345, opened 2018-07-16
Replace npm token

### LINKED ISSUE #1344: Replace NPM Token
Since all NPM upload tokens have been revoked a new one needs to be created and added to travis.yml

muffinresearch:
Fixes  #1344


Rob--W:
The test failures is unrelated to your patch and a known issue.

rpl:
as mentioned by @Rob--W one of the failure is unrelated and a known issue with the "web-ext sign" functional test that has recently started to fail intermittently.

The other issue is related to the commit message which doesn't follow the format required by conventional-changelog-lint, and I can fix that issue by tweaking the message in the merged commit.

PR #1348, opened 2018-07-19
feat: Allow negation in --ignore-files

### LINKED ISSUE #1347: Allow including some files from node_modules

I guess this is feature request.

We would like WebExtension authors to install [libdweb](http://github.com/mozilla/libdweb) via npm and then add entries to the manifest to include experimental APIs.

Sadly it becomes impossible to use `web-ext build` to create an add-on such that it would include experimental APIs from libdweb. 

Following should include path negated in the ignored pattern.

```
web-ext build --ignore-files '!node_modules/libdweb/src/**'
```

```
v9.3.0
5.5.1
2.7.0
````

### PR DESCRIPTION
**Pull request by Gozala**
This change swaps minimatch for multimatch to allow opt-in inclusion subset of node_modules

Fix #1347

### CODE REVIEW

**Comment by Rob--W** [comment]
We considered the removal of `**/node_modules/**/*'` (because `**/node_modules'`) is already in the list):
https://github.com/mozilla/web-ext/blob/0fd2dd3cf8015c9c22b37d7657b69c1fff249ad9/src/util/file-filter.js#L46

Would that solve the original issue?

If not, and this change is necessary, can you include a unit test that fails before this PR and passes with your proposed change?

**Comment by rpl** [comment]
@Rob--W looking at the following code:
https://github.com/mozilla/web-ext/blob/bc4eb1cad9f9b17c8823e6874bd87a484694ba3c/src/util/file-filter.js#L107-L113

removing `**/node_modules/**/*` does not seem to be enough, because we are going to filter out the file as soon as we find the first pattern that matches.

I (very briefly) looked to the `multimatch` package and it seems that what it basically does is to "extend minimatch to support matching on a set of patterns", and it depends from [minimatch ^3.0.0](https://github.com/sindresorhus/multimatch/blob/7681aca457f1d6ee0cc6c3b42ea7d3884ebd280b/package.json#L38).

@Gozala I agree with @Rob--W that some additional unit tests are needed for this change.
(I would also like if you could revert some of the changes to the indentation and spacing, to be consistent with how the file was initially, I guess that most of those changes may have been done automatically by `prettier` executed automatically by the editor on save).

**Comment by Gozala** [comment]
> Would that solve the original issue?
> 

Nope it would not as intention is to include some libraries installed through npm by manually whitelisting them. 

>
> If not, and this change is necessary, can you include a unit test that fails before this PR and passes with your proposed change?

I will update this pull with to include that.

> @Gozala I agree with @Rob--W that some additional unit tests are needed for this change.
> (I would also like if you could revert some of the changes to the indentation and spacing, to be consistent with how the file was initially, I guess that most of those changes may have been done automatically by prettier executed automatically by the editor on save).

Thanks for pointing that out. I thought I've disabled prettier to avoid that but I guess it still end up changing things.

**Comment by Gozala** [comment]
@Rob--W @rpl 

I have updated this pull request such that:

1. First commit contains test that illustrates issues #1347 and fails
2. Second commit contains changes that add support for negation in `--ignore-files` and passes test introduced in first commit.
3. Got rid of all formatting changes that were present previously.

File: src/util/file-filter.js [hunk 1]
```diff
@@ -88,7 +88,12 @@ export class FileFilter {
    */
   addToIgnoreList(files: Array<string>) {
     for (const file of files) {
-      this.filesToIgnore.push(this.resolveWithSourceDir(file));
+      if (file.charAt(0) === NEGATION) {
+        const resolvedFile = this.resolveWithSourceDir(file.substr(1));
+        this.filesToIgnore.push(`${NEGATION}${resolvedFile}`);
```
Rob--W: I think that the code is as readable if you use `!` instead of `${NEGATION}`.

tests/unit/test-util/test.file-filter.js
```diff
@@ -131,4 +131,31 @@ describe('util/file-filter', () => {
...
+      assert.equal(filter.wantFile('/src/node_modules/libdweb/src.js'), false);
+      assert.equal(filter.wantFile('/src/node_modules/libdweb/src'), false);
+    });
+
+    it('includes paths captured by negation', () => {
+      assert.equal(
+        filter.wantFile('/src/node_modules/libdweb/src/lib.js'),
+        true);
+      assert.equal(
+        filter.wantFile('/src/node_modules/libdweb/src/sub/lib.js'),
```
Rob--W: Please add a test for `/src/node_modules/some/other/node_modules/index.js` (I guess that the path is included because of the negation, but I would not be surprised if it got ignored anyway because of the deeper `node_modules/`.

Gozala:
@Rob--W updated pull request to address your comments above.

Gozala:
hmm, I think travis is failing because of pull request title

Gozala:
@Rob--W Alright it's all green now, as just a pull request title. Can you please merge it in ?

Thanks

Gozala:
@rpl Thanks for thorough review, you were right that:

1. My intention was to use `--ignore-files !node_modules/pkg1/**` to opt-into including `pkg1` into bundle.
2. And that patch as it is would not accomplish that. Instead I'd need to do `-i '!node_modules' -i '!node_modules/pkg1' -i '!node_modules/pkg1/**'`

Which is unfortunate, but give how [node-zip-dir](https://github.com/jsantell/node-zip-dir#options works) I am not sure there is a better way. Only other option I can imagine is add negation rules for all parent dirs automatically, so say ignoreFiles is `['!node_modules/pkg1/**']` expand that into `['!node_modules', '!node_modules/pkg1, '!node_modules/pkg1/**]` automatically, but that isn't ideal either given that:

1. You might end up with zip containing bunch of empty directories.
2. You might have more evolved patterns like `/foo/**/bar/*.js` which would be difficult to translate.

Given the constraints I think manually negating parent dirs is probably good compromise as it still can achieve goal of including specific directories without introducing too much complexity.

Probably best solution would be to change [@jsantell/node-zip-dir](https://github.com/jsantell/node-zip-dir#options) so that it understands globs itself, but again I think this pull request is probably a good enough compromise.

P.S. I have added test you've conveniently provided in the comment, except I modified it slightly to capture how ignore negation should be used to actually include desired package.

Gozala:
@rpl also it seems that pull did not picked up on requested changes being addressed and I don't seem to see a way to re-trigger another review request, so just posting this comment instead.

rpl:
> Which is unfortunate, but give how [node-zip-dir](https://github.com/jsantell/node-zip-dir#options works) I am not sure there is a better way. Only other option I can imagine is add negation rules for all parent dirs automatically, so say ignoreFiles is ['!node_modules/pkg1/**'] expand that into ['!node_modules', '!node_modules/pkg1, '!node_modules/pkg1/**] automatically, but that isn't ideal either given that:
> 1. You might end up with zip containing bunch of empty directories.
> 2. You might have more evolved patterns like /foo/**/bar/*.js which would be difficult to translate.

@Gozala Yeah, I definitely agree, I definitely prefer to avoid adding implicitly a bunch of empty directories to workaround that (and yeah, it wouldn't work with more complex negated patterns, which is one more reason to avoid the workaround).

> Given the constraints I think manually negating parent dirs is probably good compromise as it still can achieve goal of including specific directories without introducing too much complexity.

:+1: I mainly wanted to be sure that we double-checked the actual behavior and agreed that it would be good enough for now.

We will need to document the behavior expected when negated pattern are used in the ignore file params or configs in any case and we can mention this behavior in those additions to the docs (I added the "needs: docs" label on both this PR and the original issue, to make sure that we document this on the MDN doc pages for the web-ext tool).

> Probably best solution would be to change @jsantell/node-zip-dir so that it understands globs itself, but again I think this pull request is probably a good enough compromise.

Sounds good to me too, we can defer that for further follow ups.

PR #1361, opened 2018-08-27
fix: Permit single-line JSON comments

### LINKED ISSUE #1352: Web-ext still does not allow comments in _locales JSONs
bug

```js
$ ../node_modules/web-ext/bin/web-ext build
Building web extension from […]/offline-qr-code/src

UsageError: Error parsing messages.json JSONError: Unexpected token / in JSON at position 4 while parsing near '{  // manifest.json  "...' in […]/offline-qr-code/src/_locales/en/messa
```

No error.

Browsers also repect comments in these JSON files. Sources:
* https://discourse.mozilla.org/t/manifest-json-should-mention-it-allows-comments/27529
* similar issues: https://github.com/WeblateOrg/weblate/issues/2043

* **Your OS and version**: Fedora 28

````
v8.11.3
5.6.0
2.8.0
````

### PR DESCRIPTION
**Pull request by Rob--W**
Fixes #1352 .

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -4,6 +4,7 @@ import {createWriteStream} from 'fs';
 
 import {fs} from 'mz';
 import parseJSON from 'parse-json';
+import stripJsonComments from 'strip-json-comments';
 import defaultEventToPromise from 'event-to-promise';
 
 import defaultSourceWatcher from '../watcher';
@@ -71,14 +72,14 @@ export async function getDefaultLocalizedName(
   let extensionName: string = manifestData.name;
 
   try {
-    messageContents = await fs.readFile(messageFile);
+    messageContents = await fs.readFile(messageFile, {encoding: 'utf-8'});
   } catch (error) {
     throw new UsageError(
       `Error reading messages.json file at ${messageFile}: ${error}`);
   }
 
   try {
-    messageData = parseJSON(String(messageContents), messageFile);
+    messageData = parseJSON(stripJsonComments(messageContents), messageFile);
   } catch (error) {
     throw new UsageError(
       `Error parsing messages.json ${error}`);
--- a/src/util/manifest.js
+++ b/src/util/manifest.js
@@ -38,7 +38,7 @@ export default async function getValidatedManifest(
   let manifestContents;
 
   try {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -75,7 +75,6 @@
     "sign-addon": "0.3.1",
     "source-map-support": "0.5.3",
     "stream-to-promise": "2.2.0",
-    "strip-json-comments": "2.0.1",
```
rpl: @Rob--W Do you mind to bring back the `"strip-json-comments"` dependency in this PR?

`"strip-json-comments"` removes the comments stripped [by replacing them with white spaces](https://github.com/sindresorhus/strip-json-comments/tree/f9dcc6f3f18dc41f666cb9aa0f16694828a9948b#options), which is nice because it ensures that any JSON parsing error is going to point to the right place of the original JSON file (which is also confirmed by the change you have to apply in the "tests/unit/test-util/test.manifest.js" test case below to make it pass).

It is true that `"strip-json-comments"` supports both line (`// ...`) and multiline comments (`/* ... */`) whereas Firefox only supports line comments, but it is not a big deal and `web-ext lint` has its own validation rules and it is reporting the expected error on a json file that contains multiline comments (whereas web-ext strips the comments only to be able to read the JSON file and retrieve some information from that file).

I would prefer #384 to ensure that an extension is completely valid before running it (which also have an open PR, #817, currently it has some conflicts but I'm going to take a look at it asap)

Rob--W:
@rpl Done.

src/cmd/build.js
```diff
@@ -71,14 +72,14 @@ export async function getDefaultLocalizedName(
...
-    messageContents = await fs.readFile(messageFile);
+    messageContents = await fs.readFile(messageFile, {encoding: 'utf-8'});
   } catch (error) {
     throw new UsageError(
       `Error reading messages.json file at ${messageFile}: ${error}`);
   }
 
   try {
-    messageData = parseJSON(String(messageContents), messageFile);
+    messageData = parseJSON(stripJsonComments(messageContents), messageFile);
```
rpl: @Rob--W Do you mind to add an additional test case for this fix in 'tests/unit/test-cmd/test.build.js'?

Rob--W:
@rpl Done again.

PR #1385, opened 2018-10-30
chore: Run tests on windows using travis

### LINKED ISSUE #1380: Consider using TravisCI for Windows testing

feature

We use Appveyor for Windows CI but it's a little flaky.

TravisCI just announced Windows support: https://blog.travis-ci.com/2018-10-11-windows-early-release

Maybe it will be less flaky?

rpl:
Fix #1380


rpl:
There is currently one unit test that wasn't failing on appveyor (and it doesn't fail when running locally), but it fails (on both nodejs 6 and nodejs 8) with the following error message:

```
1) util/extension-runners
       defaultReloadStrategy
         allows you to disable input:
     Uncaught Error: read EISDIR
      at exports._errnoException (util.js:1020:11)
      at TTY.onread (net.js:580:26)
```

(e.g. See https://travis-ci.org/rpl/web-ext/jobs/448006862#L1981-L1986)

Currently I've added the failing unit test to the ones that are currently skipped when running on windows (but I'm thinking that we could opt to detect the windows travis worker instead and only skip it there), anyway I'm prone to think that it is still reasonable to proceed with skipping the failing test in this PR and file a follow up to investigate the failure and re-enable it on travis windows jobs.

rpl:
@kumar303 it looks that we have to also explicitly disable appveyor in the settings of the github repo, otherwise appveyor is going to still try to run on every pull request (and fail consistently because of the removed appveyor.yml file).

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -504,7 +504,18 @@ describe('util/extension-runners', () => {
       }
     });
 
-    it('allows you to disable input', async () => {
+    it('allows you to disable input', async function() {
+      // Skip this test only on a travis windows worker
+      // (because it actually runs fine when running locally).
+      if (process.env.TRAVIS_OS_NAME === 'windows') {
+        /* TODO(rpl): figure out the weird failure on travis:
```
Rob--W: Please file a bug and add a reference to the bug.

PR #1389, opened 2018-11-02
chore: Skipped nsp-check and travis-pr-title-lint npm scripts on travis windows workers

rpl:
This PR skips some redundant steps from the travis windows workers (basically skipping steps that are not really needed to run on the slower travis window workers, because the exact same step is already executed in the faster travis linux workers and so an issue is likely already been caught by them).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
 
 before_script:
 # If this command fails and you can't fix it, file an issue and add an exception to .nsprc
-- npm run nsp-check
+- if [[ "$TRAVIS_OS_NAME" != 'windows' ]]; then npm run nsp-check; fi
 
 # Keep this in sync with appveyor.yml
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ matrix:
 
 before_script:
 # If this command fails and you can't fix it, file an issue and add an exception to .nsprc
-- npm run nsp-check
+- if [[ "$TRAVIS_OS_NAME" != 'windows' ]]; then npm run nsp-check; fi
```
Rob--W: This can be shortened to:

`[[ "$TRAVIS_OS_NAME" == windows ]] || npm run nsp-check`

(similarly below)

rpl: Unfortunately travis doesn't like the shortcut syntax (my guess is that the `|` is being considered part of the yaml syntax), I'll land it as it is for now (we may double-check if there is a way to make the shortcut syntax to work in a travis yaml file).

Issue #1130, opened 2017-11-03
Are there any concrete plans to start interoperating or incorporating web-ext with Google Chrome, Opera, and other platforms?


tommai78101:
I understand the primary focus for web-ext is to improve and build upon its features for Mozilla Firefox. 

As noted here:

> By creating a new tool that focuses on the [emerging] web extension standard, we have a better chance of interoperating with other platforms, such as Google Chrome or Opera. It would be hard to do that while preserving compatibility in jpm.

I would like to know if there are any concrete plans to start interoperating with other platforms? Or it still needs to take a while before the main focus can be shifted towards this?

Thanks.

rpl:
@tommai78101 yes, we have plan to add support for Chrome and Opera to web-ext run. 

Part of the refactoring have been already landed, with some additional tweaks that are going to be part of the support for Firefox for Android (#868), which uses the same underlying abstraction that we are going to be used to add the additional chrome/opera "extension runners".

tommai78101:
Thank you. Reason I'm asking is because the quoted part isn't clear if there are plans to add Chrome and Opera support, and it's more like "this tool has potential to do XYZ."

tgvdinesh:
Has this tool started supporting chrome?

Pomax:
I guess same question still: is there a `--target=[chrome|edge|firefox]` on the near horizon that will not try to solve unsolvable differences, but automatically does `browser. => chrome.` substitutions and the like?

kumar303:
> Is there a `--target=[chrome|edge|firefox]` on the near horizon

You can track https://github.com/mozilla/web-ext/issues/809 for Chrome support in web-ext

> ...automatically does browser. => chrome. substitutions...

This is probably out of scope for web-ext. I'd suggest using https://github.com/mozilla/webextension-polyfill which is well supported.

> ...automatically does ...manifest changes...

That one is a bit tricky. I'd suggest using something like webpack ([example](https://github.com/mdn/webextensions-examples/blob/master/react-es6-popup/webpack.config.js)) and configuring multiple targets for your output. Maybe there is a better strategy.

matanster:
Seems to work with latest chrome releases ... at least for me.

PR #1392, opened 2018-11-03
feat: Support chromium-based browsers as web-ext run targets

rpl:
This PR introduces a new `'chromium'` target and a new `--chromium-binary` cli option 
(which covers #809 and part of #1130).

Internally it uses the `chrome-launcher` npm package to manage the chromium instance and so by default it tries to discover and launch the Chrome or Chromium version installed in the system.

Nevertheless other Chromium-based browsers that support extensions (like Opera and Vivaldi) can be launched using this target (with the addition of the `--chromium-binary` option):

[Quick demo of the `web-ext run` used to run an extension on Chrome, Opera and Vivaldi](https://www.youtube.com/watch?v=L4sLn1OD7vM)

TODO:

- [x] add tests for the chromium extension-runner module

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "addons-linter": "1.10.0",
     "bunyan": "1.8.12",
     "camelcase": "5.3.1",
+    "chrome-launcher": "0.10.7",
     "debounce": "1.2.0",
     "decamelize": "3.2.0",
     "es6-error": "4.1.1",
@@ -75,6 +76,7 @@
     "tmp": "0.1.0",
     "update-notifier": "3.0.1",
     "watchpack": "1.6.0",
+    "ws": "7.0.1",
     "yargs": "13.2.4",
     "zip-dir": "1.0.2"
   },
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "addons-linter": "1.10.0",
     "bunyan": "1.8.12",
     "camelcase": "5.3.1",
-    "chrome-launcher": "0.10.7",
+    "chrome-launcher": "0.11.2",
     "debounce": "1.2.0",
     "decamelize": "3.2.0",
     "es6-error": "4.1.1",
@@ -76,7 +76,7 @@
     "tmp": "0.1.0",
     "update-notifier": "3.0.1",
     "watchpack": "1.6.0",
-    "ws": "7.0.1",
+    "ws": "7.1.2",
     "yargs": "13.2.4",
     "zip-dir": "1.0.2"
   },
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -13,6 +13,7 @@ import mkdirp from 'mkdirp';
 import {
   LaunchedChrome,
   launch as defaultChromiumLaunch,
+  default as ChromeLauncher,
 } from 'chrome-launcher';
 import WebSocket from 'ws';
 
@@ -103,7 +104,10 @@ export class ChromiumExtensionRunner {
       log.debug(`(chromiumBinary: ${chromiumBinary})`);
     }
 
-    const chromeFlags = [`--load-extension=${extensions}`];
+    const chromeFlags = ChromeLauncher.defaultFlags()
+      .filter((flag) => flag !== '--disable-extensions');
+
+    chromeFlags.push(`--load-extension=${extensions}`);
 
     if (this.params.args) {
       chromeFlags.push(...this.params.args);
@@ -126,6 +130,8 @@ export class ChromiumExtensionRunner {
       chromePath: chromiumBinary,
       chromeFlags,
       startingUrl,
+      // Ignore default flags to keep the extension enabled.
+      ignoreDefaultFlags: true,
     });
 
     this.chromiumInstance.process.once('close', () => {
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -295,7 +295,6 @@ export class ChromiumExtensionRunner {
       this.wss = null;
     }
 
-
     // Call all the registered cleanup callbacks.
     for (const fn of this.cleanupCallbacks) {
       try {
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -175,20 +175,20 @@ export class ChromiumExtensionRunner {
 
     const wssInfo = this.wss.address();
 
-    const bgPage = `(function bgPage(webExtParams) {
+    const bgPage = `(function bgPage() {
       async function getAllDevExtensions() {
-        const [managerExtension, allExtensions] = await Promise.all([
-          new Promise(r => chrome.management.getSelf(r)),
-          new Promise(r => chrome.management.getAll(r)),
-        ]);
+        const allExtensions = await new Promise(
+          r => chrome.management.getAll(r));
 
         return allExtensions.filter((extension) => {
           return extension.installType === "development" &&
-            extension.id !== managerExtension.id;
+            extension.id !== chrome.runtime.id;
         });
       }
 
       const setEnabled = (extensionId, value) =>
+        chrome.runtime.id == extensionId ?
+        new Promise.resolve() :
         new Promise(r => chrome.management.setEnabled(extensionId, value, r));
 
       async function reloadExtension(extensionId) {
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -234,9 +234,8 @@ export class ChromiumExtensionRunner {
   async reloadAllExtensions(): Promise<Array<ExtensionRunnerReloadResult>> {
     const runnerName = this.getName();
 
-    // TODO(rpl): file github issue to improve the chromium extension runner
-    // to actually wait for the wssBroadcast to be processed by the connected
-    // client (See https://github.com/mozilla/web-ext/pull/1392#discussion_r231535200).
+    // TODO(rpl): wait for the wssBroadcast to be processed by the connected
+    // client (https://github.com/mozilla/web-ext/issues/1686).
     this.wssBroadcast({
       type: 'webExtReloadAllExtensions',
     });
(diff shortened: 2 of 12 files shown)


src/program.js
```diff
@@ -433,6 +433,15 @@ Example: $0 --help run.
         demand: false,
         type: 'string',
       },
+      'chromium-binary': {
+        describe: 'Path or alias to a Firefox executable such as firefox-bin ' +
```
Rob--W: Ehm, his description looks a bit unusual given the context ;)

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,266 @@
...
+   * Reloads all the extensions, collect any reload error and resolves to
+   * an array composed by a single ExtensionRunnerReloadResult object.
+   */
+  async reloadAllExtensions(): Promise<Array<ExtensionRunnerReloadResult>> {
+    const runnerName = this.getName();
+
+    this.wssBroadcast({
+      type: 'webExtReloadAllExtensions',
+    });
+
```
Rob--W: Shouldn't we wait until the reload has been processed by the browser before resolving the promise?

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Reloads a single extension, collect any reload error and resolves to
+   * an array composed by a single ExtensionRunnerReloadResult object.
+   */
+  async reloadExtensionBySourceDir(
+    extensionSourceDir: string // eslint-disable-line no-unused-vars
+  ): Promise<Array<ExtensionRunnerReloadResult>> {
+    // TODO(rpl): file github issue to improve the manager extension
+    // to make it able to detect the extension ids assigned to the
+    // target extensions and map it to the extensions source dir.
```
Rob--W: The extension ID of a Chrome extension can be determined without helper extension:
It may be set through `manifest.json` (via the `key` attribute). If unset, one is automatically generated based on the file path (see https://cs.chromium.org/Extension::InitExtensionID%20file:cc ).

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,266 @@
...
+        version: '1.0',
+        permissions: ['management', 'tabs'],
+        background: {
+          scripts: ['bg.js'],
+        },
+      })
+    );
+
+    const bgPage = `function bgPage(webExtParams) {
+      async function getAllDevAddons() {
```
Rob--W: Nit: In Chrome, the terminology is "extension", not "addon" / "add-on",

src/extension-runners/chromium.js
rpl: Ideally yes, but I'm currently leaving it as one of the improvement we may introduce in a followup.

src/extension-runners/chromium.js
rpl: Noted, we may consider that option once we look into this improvement in a follow up.

rpl:
@Rob--W as we agreed I've rebased and adapted this patch on top of the changes introduced in #1645 (and also looked into making the change to the package-lock.json file small and limited to just the new dependencies).

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,286 @@
...
+   * to exit).
+   */
+  registerCleanup(fn: Function): void {
+    this.cleanupCallbacks.add(fn);
+  }
+
+  /**
+   * Exits the runner, by closing the managed Chromium instance.
+   */
+  async exit(): Promise<void> {
```
Rob--W: Note: The whole implementation is not prepared to handle `.exit()` calls while `run` is still pending. If `exit()` is called while the helper (extension) directories are being set up, or before Chrome has finished launching, then the cleanup won't finish properly.

Is this intentional?

tests/functional/test.cli.run.js
```diff
@@ -53,8 +53,9 @@ describe('web-ext run', () => {
     const argv = [
       'run',
       '--target', 'firefox-desktop',
-      '--target', 'not-supported',
```
Rob--W: (minor)
Why did this `--target not-supported` flag move? I asked for it to be put in between valid arguments (in the original PR), because an invalid argument amidst valid arguments covers more scenarios than an invalid parameter in the end.

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,286 @@
...
+        },
+      })
+    );
+
+    const wssInfo = this.wss.address();
+
+    const bgPage = `(function bgPage(webExtParams) {
+      async function getAllDevExtensions() {
+        const [managerExtension, allExtensions] = await Promise.all([
+          new Promise(r => chrome.management.getSelf(r)),
```
Rob--W: There is no need to use `chrome.management.getSelf()` if all you're interested in is the extension ID - use `chrome.runtime.id` instead.

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,286 @@
...
+        return allExtensions.filter((extension) => {
+          return extension.installType === "development" &&
+            extension.id !== managerExtension.id;
+        });
+      }
+
+      const setEnabled = (extensionId, value) =>
+        new Promise(r => chrome.management.setEnabled(extensionId, value, r));
+
+      async function reloadExtension(extensionId) {
```
Rob--W: Consider throwing or returning early if `extensionId === chrome.runtime.id`

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,286 @@
...
+        permissions: ['management', 'tabs'],
+        background: {
+          scripts: ['bg.js'],
+        },
+      })
+    );
+
+    const wssInfo = this.wss.address();
+
+    const bgPage = `(function bgPage(webExtParams) {
```
Rob--W: Unused `webExtParams`. Let's change to `(() => {`

src/extension-runners/chromium.js
Rob--W: Let's add a comment to my comment above in the code, so we don't forget about it.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -93,6 +98,40 @@ describe('util/extension-runners/chromium', async () => {
...
 
+    // Clear console stream from previous messages and start recording
+    consoleStream.stopCapturing();
+    consoleStream.flushCapturedLogs();
+    consoleStream.startCapturing();
+    // Make verbose to capture debug logs.
+    consoleStream.makeVerbose();
+
+    // Emit a fake socket object as a new wss connection.
+    const socket = await new Promise((resolve) => {
```
Rob--W: The "error" event is synchronously added in the "connection" event. Since there is no asynchronicity, let's not use `await` here. After all, if the "error" event listener was somehow added asynchronously, then there is a chance that the "error" event is triggered before the listener existed, and cause an uncaught error.

```javascript
const fakeSocket = new EventEmitter();
const socketOnSpy = sinon.spy(fakeSocket, 'on');
runnerInstance.wss.emit('connection', fakeSocket);
sinon.assert.calledOnce(socketOnSpy);
// TODO:maybe check that the spy was called with "error"?
socket.emit('error', ...);

```

tests/functional/test.cli.run.js
rpl: I moved it because if the invalid parameter is in the middle then all the targets that follows are never reached.

The test would clearly pass without this change... but there will be no test which ensures that `firefox-android` (and now `chromium`) are considered valid targets.

src/extension-runners/chromium.js
rpl: :+1:

src/extension-runners/chromium.js
rpl: makes sense.

src/extension-runners/chromium.js
rpl: unused parameter removed

src/extension-runners/chromium.js
rpl: I added some additional changes to ensure we wait for the async setup to complete (and additional test to cover that part in the unit tests)

package.json
```diff
@@ -54,6 +54,7 @@
     "addons-linter": "1.10.0",
     "bunyan": "1.8.12",
     "camelcase": "5.3.1",
+    "chrome-launcher": "0.10.7",
```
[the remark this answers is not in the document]
rpl: I update the dependencies (and added some changes needed to fix some incompatibility with the updated chrome-launcher dependency).

rpl:
I'm going to rebase this PR on top of #1685 (because the CI job will fail for the eslint-utils security vulnerability detected in the npm audit step).

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -0,0 +1,365 @@
...
+    consoleStream.startCapturing();
+    // Make verbose to capture debug logs.
+    consoleStream.makeVerbose();
+
+    // Emit a fake socket object as a new wss connection.
+
+    const fakeSocket = new EventEmitter();
+    const onFn = fakeSocket.on;
+    // $FLOW_IGNORE: override read-only prop for testing purpose.
+    fakeSocket.on = sinon.spy((...args) => {
```
Rob--W: Why not `const fakeSocketOn = sinon.spy(fakeSocket, 'on');` ?

src/extension-runners/chromium.js
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * Returns the runner name.
+   */
+  getName() {
+    return 'Chromium';
+  }
+
+  async run(): Promise<void> {
+    if (!this._promiseSetupDone) {
```
Rob--W: Can `run` be called more than once ...? If not, I would throw if it is set. But that would affect line coverage (unless you add a dummy test that calls run twice...), so another alternative is to unconditionally assign to `_promiseSetupDone`.

src/extension-runners/chromium.js
rpl: also filed as #1686

src/extension-runners/chromium.js
rpl: also filed as #1687

PR #1403, opened 2018-12-13
fix: Fixed travis windows failures related to the FakeStdin used in the unit tests

### LINKED ISSUE #1386: Investigate and re-enable test.extension-runners.js unit test skipped on travis windows workers
In #1385 we are going to move from AppVeyor to Travis for the CI windows worker.

Unfortunately there is a single unit test that fails in a Travis windows worker, even if it doesn't actually fail when running locally on Windows 10 or on AppVeyor (which also uses a Windows Server on the windows workers).

We should investigate the failure further and re-enable the test on travis windows workers.

The test case that is failing is the following one:
https://github.com/mozilla/web-ext/blob/7d8b703e3741adc1b134b72ea7b7c40d97ac3678/tests/unit/test-extension-runners/test.extension-runners.js#L507

and the following is the error that triggers the failure:
```
  Uncaught Error: read EISDIR
           at exports._errnoException (util.js:1020:11)
           at TTY.onread (net.js:580:26)
```

(Travis job with the above failure: https://travis-ci.org/rpl/web-ext/jobs/448019223#L1981-L1986)

### LINKED ISSUE #1390: Replace nsp w/ npm audit
https://github.com/mozilla/web-ext/blob/9d28fb8847ef04eefb4676ea8033b44b55c551fd/package.json#L124

Per https://www.npmjs.com/package/nsp

> This package has been deprecated
> Author message: The Node Security Platform service is shutting down 9/30 - https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting

It has been replaced w/ [`npm audit`](https://docs.npmjs.com/getting-started/running-a-security-audit), which isn't an exact replacement and is missing a couple features (like built-in support for .nsprc style exceptions -- https://github.com/mozilla/web-ext/blob/master/.nsprc)

https://docs.npmjs.com/cli/audit

rpl:
This PR applies some changes needed to fix the windows-only failures on travis CI jobs which are triggered by the FakeStdin class used in some of the unit tests.

This travis ci windows-only issue started to trigger failures in most of the recently created pull requests, e.g. https://travis-ci.org/mozilla/web-ext/jobs/459221488#L713

By fixing the above (new issue) we can also fix #1386 (which is another test that uses the FakeStdin class and it has been disabled because it was consistently failing on the travis windows workers).

Quite unfortunately, while testing this PR travis has started to fail consistently (on any platform this time) because `nsp check` wasn't able to reach "api.nodesecurity.io" anymore (e.g. see https://travis-ci.org/rpl/web-ext/jobs/467447950#L2162), and so I had to look into fixing also #1390 as part of this PR (so that I could double-check that the unit tests were now able to pass on all the configured travis workers).

The commit that fixed #1390, replaces nsp check with a small nodejs script which run `npm audit --json` (using npx to run the last available version of npm also on node 6), identifies the advisories to ignore based on the .nsprc file (as it was supported by nsp) and then make the job to fail if there are any advisories that are not explicitly listed in the nsprc's exceptions.

Fixes #1386
Fixes #1390


Rob--W:
> The commit that fixed #1390, replaces nsp check with a small nodejs script which run npm audit --json (using npx to run the last available version of npm also on node 6), identifies the advisories to ignore based on the .nsprc file (as it was supported by nsp) and then make the job to fail if there are any advisories that are not explicitly listed in the nsprc's exceptions.

I don't think that we should care about Node.js 6 with `npm audit`. As long as we run `npm audit` with the latest npm on at least machine, we're good. Also, please consider adding a comment to the audit script and point to the npm bug that I referenced in #1390 ( `https://github.com/npm/npm/issues/20565` ) so that we can remove the script (and new shelljs dependency) if `npm audit` ever starts supporting nsprc.

rpl:
I just updated this PR with the following changes:
- added the inline comment to `script/audit-deps` as suggested
- added nodejs 10 to the travis config (on which `npm audit --json` is already available)
- splitted the travis job into stages (so that part of the stages are only going to run on the desidered combination of platform and nodejs version)

and I have also filed #1404, to remove the currently ignored advisories in a follow up.

PR #1505, opened 2019-02-18
fix: Upgraded project development environment to babel 7 and webpack 4 

rpl:


tests/unit/test-util/test.temp-dir.js
```diff
@@ -58,4 +58,9 @@ describe('util.TempDir', () => {
     assert.throws(() => tmp.path(), /cannot access path.* before.* create/);
   });
 
+  it('does not throw on remove called before a temp dir is created', () => {
+    const tmp = new TempDir();
+    assert.equal(tmp._removeTempDir, undefined);
```
Rob--W: If you're testing this internal, also add a test where `create` is called, and check that the property exist. That will ensure that the test gets the desired attention if the `TempDir` class ever gets refactored.

tests/unit/test-firefox/test.firefox.js
```diff
@@ -629,14 +629,13 @@ describe('firefox', () => {
...
-      const FxProfile = {
-        Finder() {
-          return {
-            readProfiles: fakeReadProfiles,
-            getPath: fakeGetPath,
-            profiles: fakeProfiles,
-          };
-        },
+      const FxProfile = {};
+      FxProfile.Finder = function Finder() {
```
Rob--W: Why this change?

tests/unit/test-cmd/test.sign.js
```diff
@@ -397,6 +397,17 @@ describe('sign', () => {
       }
     ));
 
+    it('throws unexpected errors', async () => {
+      const fakeAsyncFsReadFile = sinon.spy(async () => {
+        throw new Error('Unexpected fs.readFile error');
+      });
+      await assert.isRejected(
+        getIdFromSourceDir('fakeSourceDir', fakeAsyncFsReadFile),
```
Rob--W: If you want to minimize the number of injected dependencies, then you could create a directory or file without the right permissions (since the the implementation doesn't handle the `EACCES` error). This makes the test a bit less artificial than injecting a fake `fs.readFile`.

tests/unit/test-firefox/test.firefox.js
rpl: These tests has started to fail because of Finder wasn't considered a constructor, I think that the upgrade from babel 6 to babel 7 somehow unveiled the issue.

tests/unit/test-cmd/test.sign.js
rpl: yeah, I was initially thinking of going for that strategy, but I would prefer to don't add this test to the ones skipped on windows (I would like to reduce the number of tests skipped on windows, not growing that list), and so I preferred to use this strategy instead.

tests/unit/test-util/test.temp-dir.js
rpl: Main reason for testing that internal was avoiding coverall failure :-)
But given that we are testing it, let's also add the additional assertion ;-)

done in the updated version of this PR.

tests/unit/test-util/test.artifacts.js
```diff
@@ -148,4 +148,17 @@ describe('prepareArtifactsDir', () => {
...
+      () => Promise.reject(new Error('Unexpected fs.access error')));
+    const fakeAsyncMkdirp = sinon.spy(
+      () => Promise.reject(new Error('Unexpected mkdirp error')));
+
+    await assert.isRejected(prepareArtifactsDir(tmpDir.path(), {
+      asyncFsAccess: fakeAsyncFsAccess,
+    }), /Unexpected fs.access error/);
+
+    sinon.assert.called(fakeAsyncFsAccess);
+    sinon.assert.notCalled(fakeAsyncMkdirp);
```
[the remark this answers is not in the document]
rpl: That's right, thanks for catching it. Fixed in the last update on this PR.

rpl:
> Have you thought of using stubs instead of injecting fake implementations? 
> The advantage is that.... The disadvantage...

@Rob--W yeah, obviously I've been also thinking about that, to be honest I would like to rework our unit tests to avoid all the "dependency injection" that we are currently using (almost everywhere in the project) just to make those tests possible, not just the ones added in this PR.

But in this PR I really want to avoid any change that is not strictly necessary to complete the port to babel 7 and webpack 4, and so I just added the minimum set of additional tests needed to make coveralls happy again, using the testing strategy currently used in the rest of the project.

In the meantime I filed it as #1511 to follow up on that.

tests/unit/test-util/test.artifacts.js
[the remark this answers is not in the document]
Rob--W: Looks good. To ensure that the test continues to be useful if `mkdirp` were to be dropped, I checked w hether there is any code that verifies that the `asyncMkdirp` parameter is used. And there is: https://github.com/mozilla/web-ext/blob/a9b9a76b047cbb4b8313bb364f4094788184e859/tests/unit/test-util/test.artifacts.js#L135-L147

So this part of the code is good to go.

PR #1518, opened 2019-02-22
chore: Move from deprecated conventional-changelog-lint to @commitlint

rpl:
This PR contains the changes needed to port the project development environment from the deprecated `conventional-changelog-lint` package to the corresponding `@commitlint` packages (which are actually maintained).

As part of these changes I've been also:
- converting  (used by the travis jobs to lint the pull request subject when needed) from a grunt task to a nodejs script
- removing `conventional-changelog-lint` usage from the `travis-pr-title-lint` script
- removing from `.nsprc` exception list the advisory https://npmjs.com/advisories/612
  (as it is not needed anymore once we have been able to remove `conventional-changelog-lint`)


rpl:
@Rob--W most of `scripts/travis-pr-title-lint` is actually coming from the previous grunt task (which was defined in `tasks/travis-pr-title-lint`), the main differences between the two are:

- the new `scripts/travis-pr-title-lint` is a nodejs script instead of being a grunt task
- `tasks/travis-pr-title-lint` was using conventional-changelog-lint as a library, whereas `scripts/travis-pr-title-lint` is using executing the commitlint bin script and the message that we want to lint is piped to the commitlint script

scripts/travis-pr-title-lint
```diff
@@ -0,0 +1,181 @@
...
+        return;
+      }
+
+      var body = '';
+      response.on('data', function(data) {
+        try {
+          body += data;
+
+          // Once we get the closing title tag, we can read
+          // the pull request title and
```
Rob--W: The sentence is cut off.

And please use the Github API instead of parsing HTML:

Example: https://api.github.com/repos/mozilla/web-ext/pulls/1518
Documentation: https://developer.github.com/v3/pulls/

scripts/travis-pr-title-lint
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      console.log('There is more than one commit in this pull request,',
+                  'we are going to check the pull request title...');
+
+      message = await getPullRequestTitle();
+    }
+
+    lintMessage(message);
+  } catch (err) {
+    var errMessage = err.stack ? err.stack : err;
```
Rob--W: If there is no stack on an error object, then `err` may be an (error) object, and at the next line `errorMessage.trim()` would throw. Consider using the following:
```
var errMessage = `${err.stack || err}`.trim();
```

scripts/travis-pr-title-lint
rpl: > The sentence is cut off.

I'll fix the inline comment. 

> And please use the Github API instead of parsing HTML:

No, unfortunately I can't do this, I already know how it is going to end up:

the first version of this script was already using the github API, but then the travis jobs started to fail because of the API usage limits and I had to workaround it this way (See #731, as I mentioned in that PR I wasn't happy about the workaround neither, but it has definitely fixed the intermittency and it worked without any further changes for more than 2 years now, #731 has been merged on Jan 4 2017).

scripts/travis-pr-title-lint
```diff
@@ -0,0 +1,181 @@
...
+          // close the http request.
+          if (body.includes('</title>')) {
+            response.removeAllListeners('data');
+            response.emit('end');
+
+            var titleStart = body.indexOf('<title>');
+            var titleEnd = body.indexOf('</title>');
+
+            // NOTE: page slice is going to be something like:
+            // "<title> PR title by author · Pull Request #NUM · mozilla/web-ext · GitHub"
```
Rob--W: The logic here does not account for HTML entities. Could that ever cause problems?

It seems that `"`, `'` and `>` are escaped, and I expect that `&` and `<` would also be escaped, at the least. For example, see issues 813, 1310, 1107

scripts/travis-pr-title-lint
```diff
@@ -0,0 +1,181 @@
...
+    });
+  });
+}
+
+function lintMessage(message) {
+  if (!message) {
+    throw new Error('Unable to lint an empty message.');
+  }
+
+  const res = shelljs.exec(`echo ${message} | commitlint`, {silent: true});
```
Rob--W: There is a command injection vulnerability here.

Consider using commitlint as a library instead of calling it via the CLI: https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/lint

scripts/travis-pr-title-lint
rpl: Well, theoretically it may not be causing any problem, but for sure we will be linting something different from the actual message, and so I opted to add [html-entities](https://github.com/mdevils/node-html-entities) to the dev dependencies and use it to decode the message retrieved from the webpage before resolving it.

scripts/travis-pr-title-lint
rpl: ah, yeah, good catch.

I would prefer to don't use commitlint as a library for two reasons:

- I was using the previous changelog-commit-lint as a library, and that part changed a lot and a couple of times between changelog-commit-lint and this new @commitlint/* packages, and because of that we couldn't migrate to @commitlint so easily (because we had to change this script a lot)

- To be able to use @commitlint/* as a library we have to directly depend from something like 3 more @commitlint packages (@commitlint/load, @commitlint/lint and @commitlint/format)

In the updated version of this PR I opted for the following alternative approach:
the arbitrary message is first saved in a temp file, and then the temp file is piped to the commitlint bin script.

This should prevent the message from be able to do any command injection, and we don't have to directly depend from all those @commitlint sub-packages.

scripts/travis-pr-title-lint
```diff
@@ -138,7 +144,12 @@ function lintMessage(message) {
     throw new Error('Unable to lint an empty message.');
   }
 
-  const res = shelljs.exec(`echo ${message} | commitlint`, {silent: true});
+  const tmpFileName = tmp.tmpNameSync();
+  fs.writeFileSync(tmpFileName, message);
+
+  const res = shelljs.exec(`cat ${tmpFileName} | commitlint`, {silent: true});
```
Rob--W: This `cat` won't work on Windows, will it (and do we even care?)?

Consider using [`child_process.execSync`](https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options) directly, without temporary file. Like this:

```javascript
let stdout;
try {
  stdout = require('child_process').execSync('commitlint', {
    input: message,
    windowsHide: true,
    encoding: 'utf-8',
  });
} catch (e) {
  stdout = e.stdout; // How about e.stderr ...?
  if (e.status) { // Non-zero exit code
    throw stdout;
  }
}
return stdout.trim();
```

rpl: `cat` does actually work on travis windows workers, because the travis job runs using git bash shell (See the Windows Build Environment section at https://blog.travis-ci.com/2018-10-11-windows-early-release):

https://travis-ci.org/mozilla/web-ext/jobs/500323517#L488

To be fair, we don't care that much of making this work on windows, as the only important thing is that it runs once on travis.

Nevertheless, I like the `child_process.execSync` strategy, as it spares us from creating a temp file at all, and so I'm going to update this PR to use that approach instead.

scripts/travis-pr-title-lint
```diff
@@ -138,13 +143,25 @@ function lintMessage(message) {
...
+    // execSync failure or timeouts.
+    if (e.error) {
+      throw e.error;
+    }
 
-  if (res.code !== 0) {
-    throw res.stdout;
+    // commitlint non-zero exit
+    if (e.status) {
+      // stderr by default will be output to the parent process' stderr (See
```
Rob--W: How does this comment relate to the `throw stdout`? Should I read this as if you had ended the comment with `// so just throw stdout` ?

rpl: yep,it is, let me add this part to the inline comment ;-)

Issue #1307, opened 2018-04-16
Unknown argument: no-config-discovery


eight04:
### Is this a feature request or a bug?
Bug.

### What is the current behavior?

```
> web-ext run --no-config-discovery
Options:
  --version                        Show version number                 [boolean]
  --source-dir, -s                 Web extension source directory.
                     [string] [required] [default: "D:\\Dev\\webextension-test"]
  --artifacts-dir, -a              Directory where artifacts will be saved.
  [string] [required] [default: "D:\\Dev\\webextension-test\\web-ext-artifacts"]
  --verbose, -v                    Show verbose output      [boolean] [required]
  --ignore-files, -i               A list of glob patterns to define which files
                                   should be ignored. (Example:
                                   --ignore-files=path/to/first.js
                                   path/to/second.js "**/*.log")         [array]
  --no-input                       Disable all features that require standard
                                   input                    [boolean] [required]
  --config, -c                     Path to a CommonJS config file to set option
                                   defaults                             [string]
  --config-discovery               Discover config files in home directory and
                                   working directory. Disable with
                                   --no-config-discovery.
                                                       [boolean] [default: true]
  --target, -t                     The extensions runners to enable (e.g.
                                   firefox-desktop, firefox-android). Specify
                                   this option multiple times to run against
                                   multiple targets.
                                            [array] [default: "firefox-desktop"]
  --firefox, -f, --firefox-binary  Path or alias to a Firefox executable such as
                                   firefox-bin or firefox.exe. If not specified,
                                   the default Firefox will be used. You can
                                   specify the following aliases in lieu of a
                                   path: firefox, beta, nightly,
                                   firefoxdeveloperedition.             [string]
  --firefox-profile, -p            Run Firefox using a copy of this profile. The
                                   profile can be specified as a directory or a
                                   name, such as one you would see in the
                                   Profile Manager. If not specified, a new
                                   temporary profile will be created.   [string]
  --keep-profile-changes           Run Firefox directly in custom profile. Any
                                   changes to the profile will be saved.
                                                                       [boolean]
  --no-reload                      Do not reload the extension when source files
                                   change                              [boolean]
  --pre-install                    Pre-install the extension into the profile
                                   before startup. This is only needed to
                                   support older versions of Firefox.  [boolean]
  --pref                           Launch firefox with a custom preference
                                   (example:
                                   --pref=general.useragent.locale=fr-FR). You
                                   can repeat this option to set more than one
                                   preference.                           [array]
[...]

PR #1520, opened 2019-02-23
fix: Updated yargs to version 13.2.1

rpl:
This PR updates yargs to version 13.2.1 and also fixes the "Unknown argument: no-config-discovery" issue (#1307).

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -122,10 +122,10 @@ export class Program {
     this.options = {...this.options, ...options};
     Object.keys(options).forEach((key) => {
       options[key].global = true;
-      if (options[key].demand === undefined) {
+      if (options[key].demandOption === undefined) {
         // By default, all options should be "demanded" otherwise
         // yargs.strict() will think they are missing when declared.
-        options[key].demand = true;
+        options[key].demandOption = true;
       }
     });
     this.yargs.options(options);
@@ -182,9 +182,7 @@ export class Program {
         throw new UsageError(`Unknown command: ${cmd}`);
       }
       if (globalEnv === 'production') {
-        checkForUpdates ({
-          version: getVersion(this.absolutePackageDir),
-        });
+        checkForUpdates({version});
       }
 
       const configFiles = [];
@@ -339,35 +337,35 @@ Example: $0 --help run.
       alias: 'v',
       describe: 'Show verbose output',
       type: 'boolean',
-      demand: false,
+      demandOption: false,
     },
     'ignore-files': {
       alias: 'i',
       describe: 'A list of glob patterns to define which files should be ' +
                 'ignored. (Example: --ignore-files=path/to/first.js ' +
                 'path/to/second.js "**/*.log")',
-      demand: false,
+      demandOption: false,
       requiresArg: true,
       type: 'array',
     },
     'no-input': {
       describe: 'Disable all features that require standard input',
       type: 'boolean',
-      demand: false,
+      demandOption: false,
     },
     'config': {
       alias: 'c',
       describe: 'Path to a CommonJS config file to set ' +
         'option defaults',
       default: undefined,
-      demand: false,
+      demandOption: false,
       requiresArg: true,
       type: 'string',
     },
     'config-discovery': {
       describe: 'Discover config files in home directory and ' +
         'working directory. Disable with --no-config-discovery.',
-      demand: false,
+      demandOption: false,
       default: true,
       type: 'boolean',
     },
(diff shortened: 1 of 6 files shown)


rpl:
@Rob--W this PR is currently based on top of #1518, I'll rebase it once #1518 is reviewed and merged (but if you want to take a look at this in the meantime, take into account that this PR is actually composed by the single 93ae74d commit).

src/program.js
```diff
@@ -66,13 +67,21 @@ export class Program {
     argv = argv || process.argv.slice(2);
 
     // NOTE: always initialize yargs explicitly with the package dir
-    // so that we are sure that it is going to load the 'boolean-negation: false'
-    // config (See web-ext#469 for rationale).
+    // so that we are sure that it is going to load web-ext package.json
```
Rob--W: The function signature suggests that `absolutePackageDir` is the current working dir, which makes the comment as a whole look a bit odd. I think that the comment would be a bit more clear if you also mention that `bin/web-ext` fixes the `absolutePackageDir` value to the `web-ext` project root.

And in this PR you've also removed the `yargs` field. Is this to ensure that we aren't unexpectedly loading additional rules? If so, modify the comment to emphasize that.

src/program.js
```diff
@@ -174,13 +183,16 @@ export class Program {
       }
       if (globalEnv === 'production') {
         checkForUpdates ({
-          version: getVersion(absolutePackageDir),
+          version: getVersion(this.absolutePackageDir),
```
Rob--W: Can't you use the `version` variable from line 168? Like this: `checkForUpdates({version});`

src/program.js
rpl: > And in this PR you've also removed the `yargs` field.

For yargs field you mean the yargs property in the web-ext package.json file?

(I removed that because it is deprecated: https://github.com/yargs/yargs/blob/14920d13a0807e77fbbce9aa2370ec44b8a0e8d1/yargs.js#L1026)

Rob--W: The comment refers to "web-ext package.json". What is its significance?

rpl: That means that we want to be sure that yargs looks for the web-ext package.json file, instead of trying to discover a package.json file on its own by looking elsewhere (e.g. by looking for it in the current work dir or the package.json related to the currently running nodejs script).

While working on this I noticed that yargs is actually not always using the directory we pass it (e.g. it is not using that parameter [here](https://github.com/yargs/yargs/blob/14920d13a0807e77fbbce9aa2370ec44b8a0e8d1/yargs.js#L824) and [here](https://github.com/yargs/yargs/blob/14920d13a0807e77fbbce9aa2370ec44b8a0e8d1/yargs.js#L1024)), and so in some cases it is still looking for a package.json that could not be the one that we want to point yargs to (and this is making yargs to emit the warning log I linked above while running the unit tests, because yargs is looking to the "yargs" section in the mocha package.json).

The above issue doesn't seem to be triggering any problems right now (because during the regular web-ext usage and functional tests, the current nodejs script is `bin/web-ext`, and so yargs is looking for the right package.json file in that case, and the issue is not currently triggering any unexpected behavior in the unit tests), but I'm going to open an issue / pull request on yargs to discuss this unexpected yargs behavior.

Rob--W: As currently written, it seems that the web-ext package.json has some special config. With your explanation, it's clearer, but please update the comment to emphasize that our `package.json` has no config, and that this logic here is meant to avoid having side effects that would occur when loaded from an arbitrary directory, such as when run from the `mocha` directory, referencing 469 for rationale.

rpl:
> The comment about version was not addressed. 

@Rob--W You mean this one: https://github.com/mozilla/web-ext/pull/1520#discussion_r260426508?

Because that is addressed: https://github.com/mozilla/web-ext/blob/76ece188e053f950b8b6d5cc8dd7842aa8d12520/src/program.js#L185

Rob--W:
I must have been looking at an old commit at the time of the review. It has indeed been addressed.

src/program.js
```diff
@@ -67,9 +67,12 @@ export class Program {
     argv = argv || process.argv.slice(2);
 
     // NOTE: always initialize yargs explicitly with the package dir
-    // so that we are sure that it is going to load web-ext package.json
+    // to avoid side-effects due to yargs looking for its configuration
+    // section from a package.json file stored in an arbitrary directory
     // (e.g. in tests yargs would end up loading yargs config from the
-    // mocha package.json), See web-ext#469 for rationale.
+    // mocha package.json), web-ext package.json doesn't contain any yargs
```
Rob--W: The comma here should be a period.

rpl: :+1:

PR #1521, opened 2019-02-24
fix: Removed grunt and cleared npm audit advisories exception list

rpl:


rpl:
@Rob--W (similarly to #1520) this PR is currently based on top of both #1518 and #1520, I'll rebase it once those two PRs are being reviewed and merged (but if you want to take a look at this in the meantime, take into account that this PR is actually composed by the single 93ee34e commit).

scripts/develop
```diff
@@ -0,0 +1,71 @@
...
+let changed = new Set();
+
+wp.on('change', (changedFile) => {
+  changed.add(changedFile);
+});
+
+wp.on('aggregated', () => {
+  // Filter out temp files from detected changes.
+  const changes = Array.from(changed).filter((filePath) => {
+    return !path.basename(filePath).startsWith('.#');
```
Rob--W: What is so special about files that start with `.#`? There are also many other kinds of temp files (e.g. ending with `~`), so I'm a bit curious here.

scripts/build
```diff
@@ -0,0 +1,16 @@
...
+const config = require('./lib/config');
+const webpack = require('./lib/webpack');
+
+shell.set('-e');
+
+shell.echo('Clean dist files...');
+shell.rm('-rf', config.clean);
+
+shell.echo('Running webpack...');
+webpack().catch(() => {
```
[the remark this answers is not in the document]
rpl: That error has been already logged to the stderr from inside the webpack method, which is imported from `./scripts/lib/webpack.js`:

```js
const webpack = require('webpack');

module.exports = () => {
    webpack(require('../../webpack.config.js')).run((err, stats) => {
      if (err) {
        console.error(err);
        reject(err);
      } else {
        ...
```

and so the promise catch here is just to ensure that the build script exits the process with an exit code 1 right after that.

scripts/develop
rpl: The temp files created by my usual editor (emacs) starts with `.#`, and so this part was an unfinished (as the actual goal was to filter out files that may be changed whereas we may don't want them to trigger a run of all these tasks.

As agreed during our last triage meeting, I changed it to filter out any file that starts with a dot instead.

scripts/develop
```diff
@@ -61,11 +48,42 @@ wp.on('aggregated', () => {
...
+
   console.log('\nRunning functional tests');
   if (!mochaFunctional()) {
     notify('mocha functional tests errors');
     return;
   }
+}
+
+wp.on('change', (changedFile, mtime) => {
+  if (mtime == null) {
```
Rob--W: Let's use `=== null` to avoid accepting `undefined`. The documentation at https://github.com/webpack/watchpack/tree/9e8152b850a2492cb27f624e8a898ea23fde199f#api only says something about `null`, not `undefined`.

rpl: sounds good, tweak applied in e3fc21e

PR #1552, opened 2019-03-06
chore: Replaced es6-promisify with nodejs 8 util.promisify helper

rpl:
This PR replaces es6-promisify with a small util module included in-tree.

The new es6-promisify release has deprecated the multiArgs option, that is basically the only es6-promisify option that we use, and so given that we only have a small number of places where we were using es6-promisify I opted to replace it with a small module that should provide just "what we actually use" (and in the end "one less npm dependency" is nice side-effect ;-)).

This PR also unvealed some mistakes we made on some of the mocks defined in the unit test suite (same issues were also triggered by upgrading the project to the API of the last es6-promisify release, but I didn't digged too much into why with the previous version of es6-promisify they did seem to work just fine).

### DIFF
--- a/src/util/promisify.js
+++ b/src/util/promisify.js
@@ -1,14 +1,16 @@
 /* @flow */
 
-// A small promisify helper to make it easier to customize a
-// function promisified (using the 'util' module available in
-// nodejs >= 8) to resolve to an array of results:
-//
-//    import {promisify} from 'util';
-//    import {multiArgsPromisedFn} from '../util/promisify';
-//
-//    aCallbackBasedFn[promisify.custom] = multiArgsPromisedFn(tmp.dir);
-//    ...
+/*
+ * A small promisify helper to make it easier to customize a
+ * function promisified (using the 'util' module available in
+ * nodejs >= 8) to resolve to an array of results:
+ *
+ *    import {promisify} from 'util';
+ *    import {multiArgsPromisedFn} from '../util/promisify';
+ *
+ *    aCallbackBasedFn[promisify.custom] = multiArgsPromisedFn(tmp.dir);
+ *    ...
+ */
 export function multiArgsPromisedFn(fn: Function): Function {
   return (...callerArgs: Array<any>): Promise<any> => {
     return new Promise((resolve, reject) => {
--- a/src/util/promisify.js
+++ b/src/util/promisify.js
@@ -1,38 +1,25 @@
 /* @flow */
 
-// promisify helper types and implementation
-
-type PromisifyOptions = {|
-  multiArgs: boolean,
-|};
-
-export default function(
-  fn: Function,
-  bindObject: ?Object = null,
-  options: PromisifyOptions = {multiArgs: false}
-) {
-  let fnWithCb = fn;
-  if (bindObject != null) {
-    fnWithCb = fn.bind(bindObject);
-  }
-
-  return function(...callerArgs: Array<any>): Promise<any> {
+// A small promisify helper to make it easier to customize a
+// function promisified (using the 'util' module available in
+// nodejs >= 8) to resolve to an array of results:
+//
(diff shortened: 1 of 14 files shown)


src/util/promisify.js
```diff
@@ -0,0 +1,41 @@
...
+
+        if (options && options.multiArgs) {
+          resolve(rest);
+        } else {
+          resolve(rest[0]);
+        }
+      }
+
+      try {
+        fnWithCb(...(callerArgs.concat(cb)));
```
Rob--W: `fnWIthCb(...callerArgs, cb);`

src/util/temp-dir.js
```diff
@@ -63,7 +63,7 @@ export class TempDir {
    * been created.
    */
   create(): Promise<TempDir> {
-    const createTempDir = promisify(tmp.dir, {multiArgs: true});
+    const createTempDir = promisify(tmp.dir, tmp, {multiArgs: true});
```
Rob--W: This seems to be the only user of `multiArgs`.

Since we only support Node >= 8, have you considered using the [built-in `util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) and define a method using `util.promisify.custom` to have callback args?

rpl: ah, that sounds good to me too, I didn't know they added this helper in nodejs 8.

I looked into that and update this PR accordingly.

For most of the usage `util.promisify` is already providing the behaviors we need, in some cases we need to explicitly bind the methods on their related object, but it isn't a big deal.

I repurposed the primisify module to provide a `multiArgsPromisedFn` helper, which is now used with promisify.custom to customize the `tmp.dir` behavior (and we can reuse it further if we met other cases that may need that), and so I also kept the unit test (to test the `multiArgsPromisedFn` helper, and also to make sure that the behaviors of the functions wrapped by `util.promisify` are going to be the expected one, e.g. when the original function throws).

src/util/temp-dir.js
```diff
@@ -63,7 +66,9 @@ export class TempDir {
    * been created.
    */
   create(): Promise<TempDir> {
-    const createTempDir = promisify(tmp.dir, tmp, {multiArgs: true});
+    //const createTempDir = promisify(tmp.dir, tmp, {multiArgs: true});
```
Rob--W: Remove line.

src/util/temp-dir.js
```diff
@@ -63,7 +66,9 @@ export class TempDir {
...
+    const createTempDir = promisify(tmp.dir);
```
Rob--W: This whole spread-out method (separate utility file, overriding `promisify.custom` of an *external* library at the top of the file and additional `promisify` invocation here before finally calling the method looks a bit ugly.

Since this is the only user of `util/promisify`, how about just putting the non-generic logic here? Like this:

```javascript
  tmp.dir({
    ...
  }, (err, tmpPath, removeTempDir) => {
    if (err) {
      reject(err);
      return;
    }
    this._path = tmpPath;
    this._removeTempDir = removeTempDir;
    log.debug(`Created temporary directory: ${this.path()}`);
    resolve(this);
  });
});
````

Comment (rpl) [reply src/util/temp-dir.js]: Well, we could move the definition of `createTempFile` near its overriding with `promisify.custom`,
that way we would keep most of it in one place.

I just added c275812 to apply the change described above, let me know how that looks to you.

Comment (Rob--W) [reply src/util/temp-dir.js]: That's looking better, and I was about to suggest that before I pondered about just using promises directly. Using promises directly doesn't look less readable (if any, it looks more readable to me).

If you like keeping `util/promisify.js`, then I would suggest one more change:

```
.then((args) => {
        const [tmpPath, removeTempDir] = args;
```

to

```
.then(([tmpPath, removeTempDir]) => {
```

Issue #1509, opened 2019-02-19
Can't connect to firefox debugger


afk-mcz:
### Is this a feature request or a bug?
Bug

### What is the current behavior?

I wan't to debug my addon using
```web-ext run```
I get error messages and after a while firefox crashes.

### What is the expected or desired behavior?
Being able to connect to the debugger.

### Version information (for bug reports)

* **Firefox version**: 66.0b8 (64-bit)
* **Your OS and version**:  Arch Linux - Kernel: 4.20.8-arch1-1-ARCH
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version

v11.10.0
6.8.0
2.9.3
````
Source code available [here](https://github.com/afk-mcz/busca)

Output using --verbose output: 

````
$ web-ext run
Applying config files: ./package.json, ./web-ext-config.js
[program.js][info] Version: 2.9.3
[cmd/run.js][info] Running web extension from /home/afk/projects/web-extensions/busca/dist
[util/manifest.js][debug] Validating manifest at /home/afk/projects/web-extensions/busca/dist/manifest.json
[extension-runners/firefox-desktop.js][debug] Creating new Firefox profile
[firefox/index.js][debug] Running Firefox with profile at /tmp/cbdc954a-e939-4708-b9fa-11cf8449812e
[firefox/index.js][debug] Checking if remote Firefox port 6005 is available
[firefox/remote.js][debug] Connecting to Firefox on port 6005
[firefox/index.js][debug] Executing Firefox binary: /usr/bin/firefox-developer-edition
[firefox/index.js][debug] Firefox args: -start-debugger-server 6005 -foreground -no-remote -profile /tmp/cbdc954a-e939-4708-b9fa-11cf8449812e -jsconsole --url about:debugging --url https://www.youtube.com/watch?v=C4Uc-cztsJo
[firefox/index.js][info] Use --verbose or open Tools > Web Developer > Browser Console to see logging
[firefox/remote.js][debug] Connecting to the remote Firefox debugger
[firefox/remote.js][debug] Connecting to Firefox on port 6005
[firefox/remote.js][debug] Retrying Firefox (0); connection error: Error: connect ECONNREFUSED ::1:6005
[firefox/remote.js][debug] Connecting to Firefox on port 6005
[firefox/remote.js][debug] Retrying Firefox (1); connection error: Error: connect ECONNREFUSED ::1:6005
[firefox/remote.js][debug] Connecting to Firefox on port 6005
[...]

PR #1563, opened 2019-03-15
feat: Adds binaryArgs to pass additional flags to Firefox binary

denar90:
Hi,
Can we add one more option to run method?

Motivation: https://github.com/GoogleChrome/puppeteer/issues/4162
Partly solves #1122 

Example of usage

```
    const CDPPort = 6006
    await webExt.cmd.run({
      sourceDir: extPath,
      firefox: puppeteerFirefox.executablePath(),
      binaryArgs: [
        `-juggler=${CDPPort}`
      ]
    }, {
      shouldExitProgram: false,
    })

    const browserWSEndpoint = `ws://127.0.0.1:${CDPPort}`
    browser = await puppeteerFirefox.connect({
      browserWSEndpoint,
    })
```

Thanks.

### DIFF
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -42,6 +42,19 @@ import type {
 
 const log = createLogger(__filename);
 
+const ignoredParams = {
+  profilePath: '--profile-path',
+  keepProfileChanges: '--keep-profile-changes',
+  browserConsole: '--browser-console',
+  preInstall: '--pre-install',
+  startUrl: '--start-url',
+  binaryArgs: '--binary-args',
+};
+
+const getIgnoredParamsWarningsMessage = (optionName) => {
+  return `Firefox for Android target does not support ${optionName} option`;
+};
+
 export type FirefoxAndroidExtensionRunnerParams = {|
   ...ExtensionRunnerParams,
 
@@ -261,42 +274,13 @@ export class FirefoxAndroidExtensionRunner {
   }
 
   printIgnoredParamsWarnings() {
-
-    if (this.params.profilePath) {
-      log.warn(
-        'Firefox for Android target does not support custom profile paths.'
-      );
-    }
-
-    if (this.params.keepProfileChanges) {
-      log.warn(
-        'Firefox for Android target does not support --keep-profile-changes.'
-      );
-    }
-
-    if (this.params.browserConsole) {
-      log.warn(
-        'Firefox for Android target does not support --browser-console option.'
-      );
-    }
-
-    if (this.params.preInstall) {
-      log.warn(
-        'Firefox for Android target does not support --pre-install option.'
-      );
-    }
-
-    if (this.params.startUrl) {
-      log.warn(
-        'Firefox for Android target does not support --start-url option.'
-      );
-    }
-
-    if (this.params.binaryArgs) {
-      log.warn(
-        'Firefox for Android target does not support --binary-args option.'
-      );
-    }
+    Object.keys(ignoredParams).forEach((ignoredParam) => {
+      if (this.params[ignoredParam]) {
+        log.warn(
+          getIgnoredParamsWarningsMessage(ignoredParams[ignoredParam])
+        );
+      }
+    });
   }
 
   async adbDevicesDiscoveryAndSelect() {
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -48,7 +48,7 @@ const ignoredParams = {
   browserConsole: '--browser-console',
   preInstall: '--pre-install',
   startUrl: '--start-url',
-  args: '--args',
+  binaryArgs: '--binary-args',
 };
 
 const getIgnoredParamsWarningsMessage = (optionName) => {
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -261,6 +261,7 @@ export class FirefoxAndroidExtensionRunner {
   }
 
   printIgnoredParamsWarnings() {
+
     if (this.params.profilePath) {
       log.warn(
         'Firefox for Android target does not support custom profile paths.'
@@ -288,6 +289,12 @@ export class FirefoxAndroidExtensionRunner {
     if (this.params.startUrl) {
       log.warn(
         'Firefox for Android target does not support --start-url option.'
+      );
+    }
+
+    if (this.params.binaryArgs) {
+      log.warn(
+        'Firefox for Android target does not support --binary-args option.'
       );
     }
   }
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -52,7 +52,7 @@ const ignoredParams = {
 };
 
 const getIgnoredParamsWarningsMessage = (optionName) => {
-  return `Firefox for Android target does not support ${optionName} option`;
+  return `The Firefox for Android target does not support ${optionName}`;
 };
 
 export type FirefoxAndroidExtensionRunnerParams = {|
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -48,7 +48,7 @@ const ignoredParams = {
   browserConsole: '--browser-console',
   preInstall: '--pre-install',
   startUrl: '--start-url',
-  binaryArgs: '--binary-args',
+  args: '--args',
 };
 
 const getIgnoredParamsWarningsMessage = (optionName) => {
--- a/src/extension-runners/firefox-desktop.js
+++ b/src/extension-runners/firefox-desktop.js
@@ -217,9 +217,8 @@ export class FirefoxDesktopExtensionRunner {
       startUrl,
       firefoxApp,
       firefoxClient,
+      binaryArgs = [],
     } = this.params;
-
-    const binaryArgs = [];
 
     if (browserConsole) {
       binaryArgs.push('-jsconsole');
--- a/src/extension-runners/firefox-desktop.js
+++ b/src/extension-runners/firefox-desktop.js
@@ -237,7 +237,7 @@ export class FirefoxDesktopExtensionRunner {
     }
 
     this.runningInfo = await firefoxApp.run(this.profile, {
-      firefoxBinary, args: binaryArgs,
+      firefoxBinary, binaryArgs,
     });
 
     this.runningInfo.firefox.on('close', () => {
--- a/src/extension-runners/firefox-desktop.js
+++ b/src/extension-runners/firefox-desktop.js
@@ -217,21 +217,27 @@ export class FirefoxDesktopExtensionRunner {
       startUrl,
       firefoxApp,
       firefoxClient,
-      args = [],
+      args,
     } = this.params;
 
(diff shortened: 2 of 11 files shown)


denar90:
@rpl thank you so much for the review 👍 
I've update PR. Only thing I'm not sure about tests needed. Please, take a look and specify if more cases need to be covered :)

Thanks :)

src/extension-runners/firefox-android.js
```diff
@@ -290,6 +291,12 @@ export class FirefoxAndroidExtensionRunner {
         'Firefox for Android target does not support --start-url option.'
       );
     }
+
+    if (this.params.binaryArgs) {
```
denar90: @rpl I was thinking about smth like

```
const ignoredParams = {
  profilePath: '--profile-path',
  keepProfileChanges: '--keep-profile-changes',
  browserConsole: '--browser-console',
  preInstall: '--pre-install',
  startUrl: '--start-url',
  binaryArgs: '--binary-args',
};

const getIgnoredParamsWarningsMessage = (optionName) => {
  return `Firefox for Android target does not support ${optionName} option`;
};
```

so function `printIgnoredParamsWarnings` can be a bit cleaner

```
printIgnoredParamsWarnings() {
  Object.entries(ignoredParams).forEach((ignoredParam) => {
      if (this.params[ignoredParam]) {
        log.warn(
          getIgnoredParamsWarningsMessage(ignoredParams[ignoredParam])
        );
      }
    });
}
```

Only thing I stuck is `flow` with some errors, but I can dig in if you approve this approach.

denar90:
btw, should I exclude `package-lock.json` changes from the PR?

rpl:
> btw, should I exclude `package-lock.json` changes from the PR?

yes please (I briefly looked to the diff and it seems that it is changing the exact versions into non exact versions, and so let's omit it from this PR, thanks for pointing it out!)

src/extension-runners/firefox-android.js
rpl: :+1: I like the idea

rpl: (let me know if you get stuck on issues with flow errors, I'll be happy to help you to figure out how to properly handle those)

src/extension-runners/firefox-android.js
```diff
@@ -261,6 +261,7 @@ export class FirefoxAndroidExtensionRunner {
   }
 
   printIgnoredParamsWarnings() {
+
```
rpl: Nit, let's remove this empty line.

src/extension-runners/firefox-desktop.js
```diff
@@ -217,10 +217,9 @@ export class FirefoxDesktopExtensionRunner {
       startUrl,
       firefoxApp,
       firefoxClient,
+      binaryArgs = [],
     } = this.params;
 
-    const binaryArgs = [];
-
     if (browserConsole) {
       binaryArgs.push('-jsconsole');
```
rpl: uhm... I'm thinking that it may be better to append the additional arguments that the user may be passing to a `web-ext run` command (instead of prepending them before the ones added internally by web-ext itself).

@Rob--W what do you think about it?

src/program.js
```diff
@@ -503,6 +503,12 @@ Example: $0 --help run.
         demandOption: false,
         type: 'boolean',
       },
+      'binary-args': {
+        describe: 'Browser binary options. More at - ' +
+        'https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options',
```
rpl: Nit, let's indent this line as the other "multi-line" `describe` values in this file.

tests/unit/test-extension-runners/test.firefox-desktop.js
```diff
@@ -202,6 +202,14 @@ describe('util/extension-runners/firefox-desktop', () => {
     ]);
   });
 
+  it('passes binaryArgs tp Firefox', async () => {
```
rpl: Nit, typo: tp -> to

src/program.js
denar90: wdyt about adding eslint ignore for this line? URL is too long and linter doesn't like it.

src/extension-runners/firefox-android.js
denar90: whoops 😊

tests/unit/test-extension-runners/test.firefox-desktop.js
denar90: whoops 😊

denar90:
>yes please (I briefly looked to the diff and it seems that it is changing the exact versions into non exact versions, and so let's omit it from this PR, thanks for pointing it out!)

had to do force-push, hope it isn't a problem :)

src/program.js
rpl: @denar90 How about omitting the link and give it a more generic description? e.g. something like the following:

```
'binary-args': {
  describe: "Additional command line options passed to the Browser binary',
  ...  
}
``` 

In my opinion, it may be a reasonable choice, e.g. as part of #1392 we may use the `--binary-args` also to pass additional CLI options to Chromium browsers.

Otherwise we may opt to include a shortened link instead (e.g. you can use https://mzl.la/2CEvSV5  which point to the above link).

denar90: >How about omitting the link and give it a more generic description?

SGTM 👍

src/program.js
```diff
@@ -503,6 +503,11 @@ Example: $0 --help run.
         demandOption: false,
         type: 'boolean',
       },
+      'binary-args': {
```
Rob--W: Using `binary-args` here implies that the command line becomes `--binary-args`. That is a bit too long to my taste. Could you rename this to `args`
and add an `alias: ['arg'],`

I would also have liked the `'a'` shorthand, but unfortunately the letter is already taken by the `--artifacts-dir` command.

src/extension-runners/firefox-desktop.js
Rob--W: Appending rather than prepending sounds good to me.

src/extension-runners/firefox-android.js
```diff
@@ -42,6 +42,19 @@ import type {
...
+  profilePath: '--profile-path',
+  keepProfileChanges: '--keep-profile-changes',
+  browserConsole: '--browser-console',
+  preInstall: '--pre-install',
+  startUrl: '--start-url',
+  binaryArgs: '--binary-args',
+};
+
+const getIgnoredParamsWarningsMessage = (optionName) => {
+  return `Firefox for Android target does not support ${optionName} option`;
```
Rob--W: While you're at it, let's improve the English:

`The Firefox for Android target does not support the ${optionName} option`
or
`The Firefox for Android target does not support ${optionName}`.

src/extension-runners/firefox-desktop.js
```diff
@@ -217,22 +217,21 @@ export class FirefoxDesktopExtensionRunner {
       startUrl,
       firefoxApp,
       firefoxClient,
+      args = [],
     } = this.params;
 
-    const binaryArgs = [];
```
Rob--W: In this version of the PR the arguments are still prepended instead of appended. Restore the `binaryArgs` variable and use `binaryArgs.push(...args);` before passing it to `firefoxApp.run`.

denar90: Oh, looks like I misunderstood you. Does it look fine now - https://github.com/mozilla/web-ext/pull/1563/commits/28cb262c13bdf627b5f3076eca9fc705149c5f37?

src/firefox/index.js
```diff
@@ -88,7 +88,7 @@ export type FirefoxRunnerParams = {|
   'no-remote'?: boolean,
   'foreground'?: boolean,
   'listen': number,
-  'binary-args'?: Array<string> | string,
+  'args'?: Array<string> | string,
```
Rob--W: You cannot simply rename `binary-args` to `args` here, because this is the API to an external library: https://github.com/mozilla-jetpack/node-fx-runner/blob/71e3848b6adac1054829391c826fd88cfb28f621/lib/run.js#L42-L48

Revert all changes in this file (and use `binaryArgs` instead of `args: binaryArgs` in the other file).

denar90: Ok, so just keep args only for cli? all other code will remain as is?

Rob--W: Yes. The external interface of `web-ext run` will use `args`, and `binary-args` is only used internally, imposed by the contract of the fx-runner interface.

tests/unit/test-extension-runners/test.firefox-desktop.js
```diff
@@ -172,36 +172,44 @@ describe('util/extension-runners/firefox-desktop', () => {
       params.firefoxApp.run,
       sinon.match.any,
       sinon.match.has(
-        'binaryArgs',
-        sinon.match.array.deepEquals(expectedBinaryArgs)
+        'args',
```
Rob--W: This one should also be `binaryArgs`.

rpl:
@denar90 Thanks! Looks great to me too, Let's merge it!

caitmuenster:
Thanks, @denar90! Your contribution has been added to our [contribution wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

denar90:
@rpl @Rob--W thank you for the review and your comments. Really aprriciate your effort 👍 

>Thanks, @denar90! Your contribution has been added to our contribution wiki.

@caitmuenster thanks 😄 

>Would you be interested in creating an account on mozillians.org? I'd be happy to vouch for you!

Wow, I'll be more than happy to become part of the community 🙌 
Here's my profile - https://mozillians.org/en-US/u/denar90/

denar90:
btw, about documentation for the new option, do you need some help with it to make release v3.1 faster?

Rob--W:
@denar90 The documentation that you added at `--help` should be sufficient.

The list of open tasks for 3.1.0 is at https://github.com/mozilla/web-ext/projects/3

caitmuenster:
Thanks, @denar90! You're vouched. 

Welcome onboard! 🎉

zhl463:
@denar90 I use this to launch firefox with extension, but shows 
Error
      connect ECONNREFUSED 127.0.0.1:6006
I can launch firefox, and load extension.

denar90:
@zhl463 thanks for trying this one :)
Can you follow a [guide](https://github.com/GoogleChrome/puppeteer/blob/f26bb7f4c44d9b7db5cc73c4af32db6fa5bcd3a2/experimental/puppeteer-firefox/README.md#add-ons) for puppeteer I've made and give feedback if it works for you?

Rob--W:
@zhl463 That is likely this bug: https://github.com/mozilla/web-ext/issues/1509

zhl463:
> @zhl463 thanks for trying this one :)
> Can you follow a [guide](https://github.com/GoogleChrome/puppeteer/blob/f26bb7f4c44d9b7db5cc73c4af32db6fa5bcd3a2/experimental/puppeteer-firefox/README.md#add-ons) for puppeteer I've made and give feedback if it works for you?

Yes, I had followed this guide  for puppeteer-firefox . but having this error

zhl463:
> @zhl463 That is likely this bug: #1509

Thanks for replying. I will try it.

PR #1575, opened 2019-04-14
profile in opening example

janat08:


Rob--W:
The README only explains briefly how to use `web-ext` as a library and a command-line. It does not aim to describe all options of `web-ext`.

The full reference is on MDN:

- Getting started is linked from the README: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Getting_started_with_web-ext
- The `web-ext run` reference is linked from Getting started: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run

janat08:
oh well.

PR #1583, opened 2019-04-21
Allow comma as separator when checking permissions via adb

M-Reimer:
The separator between permission name and "granted=true" is a comma on older Android versions. Verified on Samsung Galaxy S5 (original ROM).

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -160,7 +160,8 @@ export default class ADBUtils {
     // Set to true the required permissions that have been granted.
     for (const line of pmDumpLogs) {
       for (const perm of permissions) {
-        if (line.includes(`${perm}: granted=true`)) {
+        if (line.includes(`${perm}: granted=true`) ||
+            line.includes(`${perm}, granted=true`)) {
           permissionsMap[perm] = true;
         }
       }


rpl:
@M-Reimer Thanks a lot for digging into the underlying reason for the issue and creating a PR to fix it!
The change is quite small and it looks good to me (it may be nice to add an additional test case for this fix, but I can add it when I land this if you prefer).

M-Reimer:
I have not much knowledge about how you implement tests. Especially as every project on the internet does it differently.
If it's easy and you can point me into the right direction (is there actually a test for the current status?), then I can give it a try.
If it's just 5 or 10 minutes work for you, then maybe it's better if you do it when merging.

rpl:
> I have not much knowledge about how you implement tests. Especially as every project on the internet does it differently.

Eh, I know the feeling :-)

When I need to fix a bug in a third party project that I don't know too much about I usually use a kind of "reverse engineering" strategy:
I basically apply a change that should break a test case related to the part/component/feature that I'm going to change, if such test case exists, and then I run the existing tests for the project. This is usually enough to point me to the part that I'm interested in.

As an example in this case I would have temporarily changed the code that detects the granted permissions to only look for the "comma separator" and not for the "colon separator" and run `npm run test`, which produces the following failure:

```
 1) utils/adb
       ensureRequiredAPKRuntimePermissions
         does not reject when all the required permissions have been granted:
     AssertionError: expected promise to be fulfilled but it was rejected with 'UsageError: Required android.permission.READ_EXTERNAL_STORAGE has not be granted for org.mozilla.firefox. Please grant them using the Android Settings or using the following adb command:\n\t adb shell pm grant org.mozilla.firefox android.permission.READ_EXTERNAL_STORAGE\n'
``` 

which points to the "does not reject when all the required permissions have been granted" test case, defined in "tests/unit/test-utils/test.adb.js", exactly the test case that we need to change to cover the fix applied in this PR.

> If it's easy and you can point me into the right direction (is there actually a test for the current status?), then I can give it a try.
> If it's just 5 or 10 minutes work for you, then maybe it's better if you do it when merging.

Sure thing, I'm already glad that you looked into what was the underlying issue, this kind of issues are often pretty tricky to reproduce because they requires to be able to run a test on a device or emulator which presents the same behavior that the affected users are experiencing. I owe you ;-)

For the record this would be more or less the kind of change I'm going to apply to the adb utils test cases:

```diff
diff --git a/tests/unit/test-util/test.adb.js b/tests/unit/test-util/test.adb.js
index 0c732ea..a318c64 100644
--- a/tests/unit/test-util/test.adb.js
+++ b/tests/unit/test-util/test.adb.js
@@ -37,11 +37,18 @@ ${fakeSocketFilePrefix}  ${fakeRDPUnixSocketFile}
 
 // NOTE: fake 'pm dump <APK>' output related to the granted permissions for an
 // android application.
-const fakeAndroidGrantedPermissions = `
+const fakeAndroidGrantedPermissionsColon = `
 android.permission.READ_EXTERNAL_STORAGE: granted=true
 android.permission.WRITE_EXTERNAL_STORAGE: granted=true
 `;
 
+// Some Android device uses a comma instead of a double colon
+// (See #1583).
+const fakeAndroidGrantedPermissionsComma = `
+android.permission.READ_EXTERNAL_STORAGE, granted=true
+android.permission.WRITE_EXTERNAL_STORAGE, granted=true
+`;
+
 const {assert} = chai;
 
 function getFakeADBKit(
@@ -392,36 +399,43 @@ describe('utils/adb', () => {
          );
        });
 
-    it('does not reject when all the required permissions have been granted',
-       async () => {
-         const adb = getFakeADBKit({
-           adbClient: {
-             shell: sinon.spy(() => Promise.resolve('')),
-           },
-           adbkitUtil: {
-             readAll: sinon.spy(() => {
-               return Promise.resolve(
-                 new Buffer(fakeAndroidGrantedPermissions)
-               );
-             }),
-           },
-         });
-         const adbUtils = new ADBUtils({adb});
+    describe('Android Granted Permissions discovery', async () => {
+      async function assertPermissions(fakeAndroidPmDump) {
+        const adb = getFakeADBKit({
+          adbClient: {
+            shell: sinon.spy(() => Promise.resolve('')),
+          },
+          adbkitUtil: {
+            readAll: sinon.spy(() => {
+              return Promise.resolve(
+                new Buffer(fakeAndroidPmDump)
+              );
+            }),
+          },
+        });
+        const adbUtils = new ADBUtils({adb});
+
+        const promise = adbUtils.ensureRequiredAPKRuntimePermissions(
+          'device1', 'org.mozilla.firefox', [
+            'android.permission.READ_EXTERNAL_STORAGE',
+            'android.permission.WRITE_EXTERNAL_STORAGE',
+          ]
+        );
 
-         const promise = adbUtils.ensureRequiredAPKRuntimePermissions(
-           'device1', 'org.mozilla.firefox', [
-             'android.permission.READ_EXTERNAL_STORAGE',
-             'android.permission.WRITE_EXTERNAL_STORAGE',
-           ]
-         );
+        await assert.isFulfilled(promise);
+        sinon.assert.calledOnce(adb.fakeADBClient.shell);
+        sinon.assert.calledWith(
+          adb.fakeADBClient.shell,
+          'device1', ['pm', 'dump', 'org.mozilla.firefox']
+        );
+      }
 
-         await assert.isFulfilled(promise);
-         sinon.assert.calledOnce(adb.fakeADBClient.shell);
-         sinon.assert.calledWith(
-           adb.fakeADBClient.shell,
-           'device1', ['pm', 'dump', 'org.mozilla.firefox']
-         );
-       });
+      it('detects permissions on pm versions that uses colon separator',
+         () => assertPermissions(fakeAndroidGrantedPermissionsColon));
+
+      it('detects permissions on pm versions that uses comma separator',
+         () => assertPermissions(fakeAndroidGrantedPermissionsComma));
+    });
   });
 
   describe('amForceStopAPK', () => {
```

rpl:
@M-Reimer I've just updated my previous comment to include a more readable diff generated using "git diff --ignore-space-change".

rpl:
Closing this PR because it is now superseded by #1590 (which contains the commit from this PR + a separate commit which contains the test case mentioned in https://github.com/mozilla/web-ext/pull/1583#issuecomment-487001759.

PR #1590, opened 2019-04-29
fix: Allow comma as separator when checking permissions via adb

rpl:
This PR contains the changes introduced by #1583, to ensure that web-ext will be able to detect the granted permissions on android versions where `pm dump` uses a comma as separator between permission name and "granted=true" (e.g. as in the Samsung Galaxy S5 original ROM), with the addition of a test case to cover the fix in the automated unit tests.

### DIFF
--- a/tests/unit/test-util/test.adb.js
+++ b/tests/unit/test-util/test.adb.js
@@ -57,7 +57,7 @@ function getFakeADBKit(
   const fakeTransfer = new EventEmitter();
   const adbUtilReadAllStub = sinon.stub();
 
-  adbUtilReadAllStub.onCall(0).returns(Promise.resolve(new Buffer('')));
+  adbUtilReadAllStub.onCall(0).returns(Promise.resolve(Buffer.from('')));
 
   const fakeADBClient = {
     listDevices: sinon.spy(() => {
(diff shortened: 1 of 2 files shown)


M-Reimer:
Thanks for pushing this forward! 👍

tests/unit/test-util/test.adb.js
```diff
@@ -392,36 +399,41 @@ describe('utils/adb', () => {
...
-         const adbUtils = new ADBUtils({adb});
+    describe('Android Granted Permissions discovery', async () => {
+      async function assertPermissions(fakeAndroidPmDump) {
+        const adb = getFakeADBKit({
+          adbClient: {
+            shell: sinon.spy(() => Promise.resolve('')),
+          },
+          adbkitUtil: {
+            readAll: sinon.spy(() => {
+              return Promise.resolve(new Buffer(fakeAndroidPmDump));
```
Rob--W: `new Buffer(string)` is deprecated. It was already pre-existing, but since you're editing anyway, could you change it to `Buffer.from(fakeAndroidPmDump)` ?

rpl: yeah, I just added a separate commit that replace it also for other usages of "new Buffer(...)" in the same test file.

Issue #1137, opened 2017-11-26
File watcher too eager for changes


freaktechnik:
### Is this a feature request or a bug?
This is a bug

### What is the current behavior?
- Compile files with webpack
- Launch web-ext run on the compiled sources
- web-ext loads initial extension
- web-ext detects "changes" and reloads the extension

Extension this happens with: https://github.com/freaktechnik/justintv-stream-notifications
Full log of webpack + verbose web-ext run: https://gist.github.com/freaktechnik/c129f478277d676e9d6f2e346f8eb9e2

### What is the expected or desired behavior?
There should be no change detected, since the files have not been touched since launching web-ext run.

### Version information (for bug reports)

* **Firefox version**: 59.0a1
* **Your OS and version**: elementary OS 0.3 (Ubuntu 16.04 derivative)
* Paste the output of these commands:

````
v9.0.0
5.5.1
1.10.1
````

**Comment by kumar303** [comment]
Thanks for the verbose log.

This suggests to me that web-ext is looking in the correct directory (this contains only your built files, right?)
```
[watcher.js][debug] Watching for file changes in /home/martin/dev/justintv-stream-notifications/webextension
```

The first change to trigger a reload happens in that directory:

```
[watcher.js][debug] Changed: /home/martin/dev/justintv-stream-notifications/webextension/background/index.js
[watcher.js][debug] Last change detection: 00:51:57 GMT+0100 (CET)
```

I don't see any changes in the `src/` directory so the watcher is working as expected.

Is it unexpected that these files were changed?

**Comment by freaktechnik** [comment]
It is unexpected that these files were changed, exactly. As the only thing actually writing to that director is webpack that ran just before launching web-ext run. This happens even if the only thing running is web-ext (of course there's a DM etc. but that still shouldn't touch the files for all I know). It also only happens for this extension, so it must be something related to the files being built by webpack or similar.

**Comment by kumar303** [comment]
web-ext uses webpack's [watcher implementation](https://github.com/mozilla/web-ext/blob/master/src/watcher.js#L2) with nothing fancy added except maybe the debouncer.

**Comment by freaktechnik** [comment]
I've added some logging to the watcher on change:
```js
console.log("CHANGE", Date.now(), mtime, start);
```
mtime is the modification time that the change event is getting passed and start is the timestamp when the watcher is started (so the Date.now() passed to the watcher when it starts).

This produces the following output:
```
CHANGE 1512081644234 1512081651821 1512081644206
[...]

PR #1625, opened 2019-06-02
Fixes #1573: Added --delay param to CLI

### LINKED ISSUE #1573: Request - Add parameter "--delay" to allow web-ext to wait until a WebExtension build has completed the building process

This is a feature.

The current behavior is, if `web-ext` is launched normally, it will immediately reload any files that were changed. If an extension is in the process of compiling/building, `web-ext` will constantly reload and reload and reload, and may not fully reload completely due to many files being changed in a short period of time.

The other behavior is, if `web-exit` is launched with the parameter, `--no-reload`, then the user needs to manually reload the extension, whenever the compilation process is completed. However, it is inconvenient to have to manually reload the finished extension every time, and this cannot be automated.

The desired behavior is by setting a new parameter, `--delay` and specify a given amount of milliseconds (for accurate timing, which some extensions and production builds may benefit from). This delay will allow `web-ext` to wait for a certain period of time duration that was specified, before executing a `reload` command.

Allowing such process would benefit automated production builds, while allowing the flexibility to wait until the compilation process is completed, and can be independently tweaked to the user's content.

### Example command syntax:

`web-ext reload --delay 5000`

`web-ext` will reload after 5 seconds has elapsed.

suhailsinghbains:
Fixes #1573: Added --delay param to CLI

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -242,6 +242,7 @@ export default async function build(
           throw error;
         });
       },
+      reloadDelay: new Number(1000),
       shouldWatchFile: (...args) => fileFilter.wantFile(...args),
     });
   }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -32,6 +32,7 @@ export type CmdRunParams = {|
   keepProfileChanges: boolean,
   noInput?: boolean,
   noReload: boolean,
+  reloadDelay: Number,
   preInstall: boolean,
   sourceDir: string,
   startUrl?: Array<string>,
@@ -68,6 +69,7 @@ export default async function run(
     ignoreFiles,
     noInput = false,
     noReload = false,
+    reloadDelay = new Number(1000),
     preInstall = false,
     sourceDir,
     startUrl,
@@ -195,6 +197,7 @@ export default async function run(
       artifactsDir,
       ignoreFiles,
       noInput,
+      reloadDelay,
     });
   }
 
--- a/src/extension-runners/index.js
+++ b/src/extension-runners/index.js
@@ -232,13 +232,14 @@ export type WatcherCreatorParams = {|
   onSourceChange?: OnSourceChangeFn,
   ignoreFiles?: Array<string>,
   createFileFilter?: FileFilterCreatorFn,
+  reloadDelay: Number
 |};
 
 export type WatcherCreatorFn = (params: WatcherCreatorParams) => Watchpack;
 
 export function defaultWatcherCreator(
   {
-    reloadExtension, sourceDir, artifactsDir, ignoreFiles,
+    reloadExtension, sourceDir, artifactsDir, ignoreFiles, reloadDelay,
     onSourceChange = defaultSourceWatcher,
     createFileFilter = defaultFileFilterCreator,
   }: WatcherCreatorParams
@@ -251,6 +252,7 @@ export function defaultWatcherCreator(
     artifactsDir,
     onChange: () => reloadExtension(sourceDir),
     shouldWatchFile: (file) => fileFilter.wantFile(file),
+    reloadDelay,
   });
 }
 
@@ -263,6 +265,8 @@ export type ReloadStrategyParams = {|
   artifactsDir: string,
   ignoreFiles?: Array<string>,
   noInput?: boolean,
+  reloadDelay: Number
+
 |};
 
 export type ReloadStrategyOptions = {|
@@ -297,6 +301,7 @@ export function defaultReloadStrategy(
     sourceDir,
     artifactsDir,
     ignoreFiles,
+    reloadDelay: new Number(1000),
   });
 
   extensionRunner.registerCleanup(() => {
--- a/src/program.js
+++ b/src/program.js
@@ -473,6 +473,11 @@ Example: $0 --help run.
         demandOption: false,
         type: 'boolean',
       },
+      'reload-delay': {
+        describe: 'Delay the reload of extension when source files change',
+        demandOption: false,
+        type: 'number',
+      },
       'pre-install': {
         describe: 'Pre-install the extension into the profile before ' +
                   'startup. This is only needed to support older versions ' +
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -19,6 +19,7 @@ export type OnSourceChangeParams = {|
   artifactsDir: string,
   onChange: OnChangeFn,
   shouldWatchFile: ShouldWatchFn,
+  reloadDelay: Number
 |};
 
 // NOTE: this fix an issue with flow and default exports (which currently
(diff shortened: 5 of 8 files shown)


src/cmd/build.js
```diff
@@ -242,6 +242,7 @@ export default async function build(
           throw error;
         });
       },
+      reloadDelay: new Number(1000),
```
Rob--W: Why are you using the `Number` constructor here?

The constructors of [primitive values](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) are rarely used because it is not necessary. The code (in this whole PR) would be more idiomatic if you use the primitive value directly, without wrapping it as an object.

src/cmd/run.js
```diff
@@ -32,6 +32,7 @@ export type CmdRunParams = {|
   keepProfileChanges: boolean,
   noInput?: boolean,
   noReload: boolean,
+  reloadDelay: Number,
```
Rob--W: Use the primitive `number` type instead of `Number`. Similarly in the rest of this PR.

src/extension-runners/index.js
```diff
@@ -232,13 +232,14 @@ export type WatcherCreatorParams = {|
   onSourceChange?: OnSourceChangeFn,
   ignoreFiles?: Array<string>,
   createFileFilter?: FileFilterCreatorFn,
+  reloadDelay: Number
```
Rob--W: Add a trailing comma (also in the rest of this PR where applicable).

The comma is optional for the last property of an object literal, but we still use it anyway, because doing so makes it possible to add a new property without having to change the previous line.

src/program.js
```diff
@@ -473,6 +473,11 @@ Example: $0 --help run.
         demandOption: false,
         type: 'boolean',
       },
+      'reload-delay': {
+        describe: 'Delay the reload of extension when source files change',
```
Rob--W: This does not accurately describe the effect of the parameter.

src/watcher.js
```diff
@@ -29,14 +30,20 @@ declare function exports(params: OnSourceChangeParams): Watchpack;
...
+}: OnSourceChangeParams
 ): Watchpack {
   // TODO: For network disks, we would need to add {poll: true}.
   const watcher = new Watchpack();
 
   const executeImmediately = true;
-  onChange = debounce(onChange, 1000, executeImmediately);
+
+  if (reloadDelay) {
+    onChange = debounce(onChange, reloadDelay, executeImmediately);
```
Rob--W: Please read step 4 of https://github.com/mozilla/web-ext/issues/1573#issuecomment-496253303 again. If you don't understand the difference, then https://github.com/mozilla/web-ext/issues/1137#issuecomment-364833409 might provide more context. If it's still not clear after reading that, feel free to ask for clarification here.

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -301,6 +301,7 @@ describe('util/extension-runners', () => {
         onSourceChange: sinon.spy(() => {}),
         ignoreFiles: ['path/to/file', 'path/to/file2'],
         reloadExtension: sinon.spy(() => Promise.resolve()),
+        reloadDelay: new Number(1000),
```
Rob--W: Why was this added?

`reloadDelay` should be an optional parameter. The `reloadDelay` option should not be set if the default value works for the test.

tests/unit/test-cmd/test.run.js
```diff
@@ -187,6 +188,13 @@ describe('run', () => {
     assert.equal(reloadStrategy.called, false);
   });
 
+  it('Reload Delay', async () => {
```
Rob--W: A code base with a consistent coding style is easier to work with. So when you introduce new code to an existing code base, try to follow the existing code style.

Although we have no strict requirements on the format of the test description, all other tests appear to start with a lowercase word, and are a bit more descriptive than just "Reload Delay".

tests/unit/test-cmd/test.run.js
```diff
@@ -187,6 +188,13 @@ describe('run', () => {
...
+    const cmd = prepareRun();
+    const {reloadDelay} = cmd.argv;
+
+    assert.equal(typeof(reloadDelay), typeof(new Number()));
```
Rob--W: This proposed test provides little value, because the type system already ensures that the `reloadDelay` parameter has the right type.

The unit test should verify that the functionality that you've added works as expected. A prerequisite to writing good unit tests is an understanding of the feature itself. A basic test should show that the feature implements the required functionality.
A good test is able to catch bugs in the implementation. To write a good test, try to think what could go wrong with the implementation, and keep that in mind while writing a test.

src/program.js
```diff
@@ -473,6 +473,12 @@ Example: $0 --help run.
         demandOption: false,
         type: 'boolean',
       },
+      'reload-delay': {
+        describe: 'Allow web-ext to wait until a WebExtension build has ' +
+                  'completed the building process',
```
Rob--W: While this option can be used to support that scenario, this is not an accuate description of this option.

suhailsinghbains:
> The test is still lacking. Please read my previous comment for more guidance, and run the test and run the linter before updating the PR again. If anything is unclear, just ask.

Yeah, I need some more help. I was thinking of a valid input i.e. number. But, now I think a number greater than 1000 should also be there, otherwiese, the delay would be too less for the reloading.
Thanks

Rob--W:
> > The test is still lacking. Please read my previous comment for more guidance, and run the test and run the linter before updating the PR again. If anything is unclear, just ask.
> 
> Yeah, I need some more help. I was thinking of a valid input i.e. number. But, now I think a number greater than 1000 should also be there, otherwiese, the delay would be too less for the reloading.

To be clear, the problem is not just the choice of a number. It's that your test is not testing functionality, but types.

Before you can implement a test, take a step back and try to understand what functionality you are supposedly adding. Write it down, maybe just in English. Then write (in English or in code) a test scenario to verify the behavior, and ask me to review it.

caitmuenster:
Hey @suhailsinghbains, are you will working on this PR?

suhailsinghbains:
Hi @caitmuenster 
Yeah, I wanted to complete this test case. I am still thinking what test case should I write? I don't understand apart from the number type and the minimum value check test cases, what could I go for?

tommai78101:
@suhailsinghbains What about invalid arguments such as parsing a character as an ASCII decimal number? Or you could try testing for extremely edge cases like negative values, floating point numbers representing seconds and not milliseconds, or maximum values?

Rob--W:
@suhailsinghbains The feature that you're implementing is the ability to throttle reloads with the `--reload-delay` flag.

To test that the feature works, you need to change (watched) files multiple times, and verify that the `onChange` callback is only called once within the interval as specified by the `reloadDelay` option.

PR #1634, opened 2019-06-21
fix: Stop bundling dependencies with slashes

### LINKED ISSUE #1629: 'run' command fails due to absent 'es6-promise'

bug

`web-ext … run` directly fails with an error: `Error: Cannot find module 'es6-promise'`.
 
The error appears related to one of web-ext’s dependencies, `@cliqz-oss/node-firefox-connect`. That dependency does however neatly install es6-promise as its dependency, so I have no idea why it is not found there. Perhaps the problem is somehow in the module resolver (or is it node or webpack doing the resolving here?).

The problem could perhaps disappear by cleaning up some of the forked dependencies; [supposedly](https://github.com/cliqz-oss/node-firefox-connect/pull/2) es6-promise is not even necessary anymore for the node-firefox-connect module. But it would be nice to figure out the source of the problem.

It is common to not encounter this error because npm often 'lifts' the es6-promise module for deduplication, with the side-effect that it will no longer fail to find it. Hence to reliably reproduce this error, install web-ext without dependency deduplication using `--legacy-bundling`. For example:

```
git clone https://github.com/mdn/webextensions-examples
cd webextensions-examples
npm install --legacy-bundling web-ext
npx web-ext -v -s beastify run
```

Relevant console output:

```
[program.js][error]
Error: Cannot find module 'es6-promise'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/external "es6-promise":1:1)
    at n (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/node_modules/@cliqz-oss/node-firefox-connect/index.js:5:15)
    at n (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/src/util/stdin.js:5:1)
    at Object.<anonymous> (/home/user/webextensions-examples/node_modules/web-ext/dist/webpack:/src/firefox/remote.js:250:10)

[program.js][error] Error code: MODULE_NOT_FOUND
```

Firefox: 67
OS: Ubuntu
node: v8.16.0
npm: 6.9.0
web-ext: 3.0.0 (and earlier versions, since 2.5.0, which [introduces](https://github.com/mozilla/web-ext/pull/1284/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R52) the @cliqz-oss/node-firefox-connect dependency)

Rob--W:
See commit message.

Fixes #1629

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,17 +1,14 @@
 /*eslint prefer-template: 0*/
 var path = require('path');
-var fs = require('fs');
 
 var webpack = require('webpack');
 
 var nodeModules = {};
 
-// This is to filter out node_modules as we don't want them
-// to be made part of any bundles.
-fs.readdirSync('node_modules')
-  .filter(function(x) {
-    return ['.bin'].indexOf(x) === -1;
-  })
+// Do not bundle any external module, because those are explicitly added as
+// "dependencies" in package.json. Bundling them anyway could result in bugs
+// like https://github.com/mozilla/web-ext/issues/1629
+Object.keys(require('./package.json').dependencies)
   .forEach(function(mod) {
     nodeModules[mod] = 'commonjs ' + mod;
   });


Rob--W:
`--legacy-bundling` is legacy behavior. To match production, we should try to run with the default / most common configuration as much as possible.

rpl:
> `--legacy-bundling` is legacy behavior. To match production, we should try to run with the default / most common configuration as much as possible.

I agree, but it does still concern me that we are not able to catch this kind of dependency loading issues.

I'm fine with proceeding with merging the fix as is, but it would be great (at least in a follow up) to look into trigger this kind of issue (that's the reason why we have a "test-functional running in production mode" as part of our CI jobs, "having to test this manually" kills its purpose) at least while there is still a way (e.g. using an additional npm option) to trigger it.

Rob--W:
> > `--legacy-bundling` is legacy behavior. To match production, we should try to run with the default / most common configuration as much as possible.
> 
> I agree, but it does still concern me that we are not able to catch this kind of dependency loading issues.

My concern with using `--legacy-bundling` only is that we won't catch problems if we ever have an issue related to the non-legacy way of bundling dependencies.

> I'm fine with proceeding with merging the fix as is, but it would be great (at least in a follow up) to look into trigger this kind of issue (that's the reason why we have a "test-functional running in production mode" as part of our CI jobs, "having to test this manually" kills its purpose) at least while there is still a way (e.g. using an additional npm option) to trigger it.

Could you open a new PR with your proposed patch (and try to address my review feedback if you agree that it's a potential issue)?

rpl:
> > > `--legacy-bundling` is legacy behavior. To match production, we should try to run with the default / most common configuration as much as possible.
> > 
> > 
> > I agree, but it does still concern me that we are not able to catch this kind of dependency loading issues.
> 
> My concern with using `--legacy-bundling` only is that we won't catch problems if we ever have an issue related to the non-legacy way of bundling dependencies.
> 
> > I'm fine with proceeding with merging the fix as is, but it would be great (at least in a follow up) to look into trigger this kind of issue (that's the reason why we have a "test-functional running in production mode" as part of our CI jobs, "having to test this manually" kills its purpose) at least while there is still a way (e.g. using an additional npm option) to trigger it.
> 
> Could you open a new PR with your proposed patch (and try to address my review feedback if you agree that it's a potential issue)?

What I was proposing is to start from the set of changes linked in my previous comments to prepare an additional 'legacy-bundling production mode" test step in the CI job, but if you prefer I can take a look into it as part of a follow up.

PR #1638, opened 2019-06-25
fix: Fixed linting error related to eslint v6

rpl:
This PR contains two fixes needed to prevent linting errors on eslint v6 (e.g. See https://travis-ci.org/mozilla/web-ext/jobs/550038386#L227).

Unfortunately these fixes are not yet enough to be able to merge #1635: 
(once these linting errors are fixed) there are test failures related to the "addons-linter not being yet adapted to run successfully on eslint 6" (even if the addons-linter has eslint 5 has its own exact dependency, the presence of eslint 6 as a web-ext dependency doesn't seem to be forcing npm to esplicitly include eslint 5 in the addons-linter node_modules dir as we would expect).

### DIFF
--- a/tests/unit/helpers.js
+++ b/tests/unit/helpers.js
@@ -158,22 +158,15 @@ export function makeSureItFails(): Function {
 export function fake<T>(
   original: Object, methods: Object = {}, skipProperties: Array<string> = []
 ): T {
-  var stub = {};
+  const stub = {};
 
   // Provide stubs for all original members:
-  var props = [];
-  var obj = original;
-  while (obj) {
-    props = props.concat(Object.getOwnPropertyNames(obj));
-    obj = Object.getPrototypeOf(obj);
-  }
+  const proto = Object.getPrototypeOf(original);
+  const props = Object.getOwnPropertyNames(original)
+    .concat(Object.getOwnPropertyNames(proto))
+    .filter((key) => !skipProperties.includes(key));
 
-  var proto = Object.getPrototypeOf(original);
   for (const key of props) {
-    if (skipProperties.indexOf(key) >= 0 ||
-       (!original.hasOwnProperty(key) && !proto.hasOwnProperty(key))) {
-      continue;
-    }
     const definition = original[key] || proto[key];
     if (typeof definition === 'function') {
       stub[key] = () => {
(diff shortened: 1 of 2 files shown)


tests/unit/helpers.js
```diff
@@ -171,7 +171,8 @@ export function fake<T>(
   var proto = Object.getPrototypeOf(original);
   for (const key of props) {
```
Rob--W: This whole implementation looks somewhat flawed.

The first `while (obj) {` loop collects all property names from the whole prototype chain.
Then the `for (const of props) {` loop iterates over these properties, but skips any property that is not part of the original object or its immediate prototype.

If the logic is correct, then it should suffice to just iterate over the entries of the object and its immediate parent, i.e. remove above from `var props = [];` until the end of the first loop, and change this loop to:

```javascript
for (const key of Object.keys(original).concat(Object.keys(proto))) {
```

Or maybe even `for (const [key, definition] of Object.entries(...)) {`

If you follow this suggestion, then there is no need for the `hasOwnProperty` check any more, and you can thus also remove the eslint suppression rule.

rpl: `Object.keys` only returns enumerable properties, and so it doesn't work as a replacement for `Object.getOwnPropertyNames`.

Nevertheless, we can still avoid to call `hasOwnProperty` (and the eslint suppression) by just collecting the results from calling `Object.getOwnPropertyNames` from the original object (and its prototype if any).

I've just updated the patch accordingly.

tests/unit/helpers.js
```diff
@@ -158,22 +158,15 @@ export function makeSureItFails(): Function {
...
   // Provide stubs for all original members:
-  var props = [];
-  var obj = original;
-  while (obj) {
-    props = props.concat(Object.getOwnPropertyNames(obj));
-    obj = Object.getPrototypeOf(obj);
-  }
+  const proto = Object.getPrototypeOf(original);
+  const props = Object.getOwnPropertyNames(original).concat(
+    proto ? Object.getOwnPropertyNames(proto) : []
```
Rob--W: According to the current implementation, `proto` is expected to be non-void.
It's not necessary to check for emptiness of `proto`. If you do, and want to be resilient against arbitrary values, then for completeness you should also add `proto && ` to the loop below, in case the property has a falsey value (e.g. `{__proto__: nul, propname: false}`).

rpl: Thanks for pointing it out. 

I opted to simplify it by considering the `proto` to always be non-void (we don't have any such case, better to keep it simple).

PR #1639, opened 2019-06-25
test: Run functional tests in legacy bundling mode as part of the travis CI jobs

rpl:
This PR adds an additional step to the travis CI jobs, to verify that web-ext is able to load the expected npm dependencies if installed with the "npm legacy bundling mode".

Follow up for #1634

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,10 +56,13 @@ script:
 ## (See #1082 for rationale).
 - TEST_PRODUCTION_MODE=1 npm run test-functional
 
+## run functional test suite in "npm legacy bundling mode"
+## (See #1639 for rationale).
+- TEST_PRODUCTION_MODE=1 TEST_LEGACY_BUNDLING=1 npm run test-functional
+
 ## NOTE: remove the custom python path from the npm config, and also
 ## remove production package.json.
 - npm config delete python
-- rm artifacts/production/package.json
 
 after_script: npm run publish-coverage
 
--- a/scripts/test-functional
+++ b/scripts/test-functional
@@ -1,6 +1,9 @@
 #!/usr/bin/env node
 
+const path = require('path');
+
 const shell = require('shelljs');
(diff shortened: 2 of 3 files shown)


rpl:
The test fails as expected in "npm legacy bundling mode" without the fix from #1634 (https://travis-ci.org/mozilla/web-ext/jobs/550293587#L1332), but still pass whle running in "npm production mode" (https://travis-ci.org/mozilla/web-ext/jobs/550293587#L1109).

.travis.yml
```diff
@@ -56,10 +56,13 @@ script:
 ## (See #1082 for rationale).
 - TEST_PRODUCTION_MODE=1 npm run test-functional
 
+## run functional test suite in "npm legacy bundling mode"
+## (See #1634 for rationale).
+- TEST_PRODUCTION_MODE=1 TEST_LEGACY_BUNDLING=1 npm run test-functional
+
 ## NOTE: remove the custom python path from the npm config, and also
 ## remove production package.json.
```
Rob--W: Nit: update this comment.

scripts/test-functional
```diff
@@ -1,44 +1,50 @@
...
+  if (testLegacyBundling) {
+    shell.echo('\nTest in "npm legacy bundling mode"');
+    npmInstallOptions.push('--legacy-bundling');
+  }
 
   execMochaOptions = {
     env: {
       ...process.env,
-      TEST_WEB_EXT_BIN: './artifacts/production/bin/web-ext',
+      TEST_WEB_EXT_BIN: path.join(destDir, 'node_modules', 'web-ext', 'bin', 'web-ext'),
```
Rob--W: Real production would be `path.join(destDir, 'node_modules', '.bin', 'web-ext')`.

rpl: yeah, but I had to change it because it is not a nodejs script on windows, I'll leave it as it is.

PR #1644, opened 2019-06-30
fix: Add supported choices to --target option

### LINKED ISSUE #1642: --target / -t flag with an invalid value does not result in any meaningful error message
Bug

When an unrecognized value is passed to the `--target` flag (anything besides `firefox-desktop` or `firefox-android`), `web-ext` starts up and exits immediately, without errors.

```
$ web-ext run -t someinvalidvalue
Running web extension from /tmp/testextension
Press R to reload (and Ctrl-C to quit)
```

There should be a warning or error message when the 

* **Firefox version**: not relevant
* **Your OS and version**: not relevant

````
v11.15.0
6.9.0
3.0.0
````

### PR DESCRIPTION
**Pull request by onlywicked**
Fixes #1642

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -436,9 +436,8 @@ Example: $0 --help run.
     .command('run', 'Run the extension', commands.run, {
       'target': {
         alias: 't',
-        describe: 'The extensions runners to enable. ' +
-                  'Specify this option multiple times to ' +
-                  'run against multiple targets.',
+        describe: 'The extensions runners to enable. Specify this option ' +
+                  'multiple times to run against multiple targets.',
         default: 'firefox-desktop',
         demandOption: false,
         type: 'array',
--- a/src/program.js
+++ b/src/program.js
@@ -436,8 +436,8 @@ Example: $0 --help run.
     .command('run', 'Run the extension', commands.run, {
       'target': {
         alias: 't',
-        describe: 'The extensions runners to enable (e.g. firefox-desktop, ' +
-                  'firefox-android). Specify this option multiple times to ' +
+        describe: 'The extensions runners to enable. ' +
+                  'Specify this option multiple times to ' +
                   'run against multiple targets.',
         default: 'firefox-desktop',
         demandOption: false,
--- a/src/program.js
+++ b/src/program.js
@@ -442,6 +442,7 @@ Example: $0 --help run.
         default: 'firefox-desktop',
         demandOption: false,
         type: 'array',
+        choices: ['firefox-desktop', 'firefox-android'],
       },
       'firefox': {
         alias: ['f', 'firefox-binary'],
--- a/tests/functional/test.cli.run.js
+++ b/tests/functional/test.cli.run.js
@@ -1,5 +1,6 @@
 /* @flow */
 import {describe, it} from 'mocha';
+import {assert} from 'chai';
 
 import {
   minimalAddonPath, fakeFirefoxPath,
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: tests/unit/test-cmd/test.run.js [hunk 1]
```diff
@@ -234,6 +234,14 @@ describe('run', () => {
...
   });
 
+  it('will not run extension runner if target is not supported', async () => {
+    const cmd = prepareRun();
+    await cmd.run({target: ['not-supported']});
+
+    sinon.assert.notCalled(androidRunnerStub);
+    sinon.assert.notCalled(desktopRunnerStub);
+  });
+
```
Rob--W: This test doesn't help much, because this tests the internal implementation, and not the CLI frontend. If you add `sinon.assert.notCalled(cmd.options.MultiExtensionRunner)`, then you would see that the test fails, because the runner is run anyway.

There are two relevant conditions to check:
1. `-t firefox-desktop` works.
2. `-t firefox-desktop -t not-supported -t firefox-desktop` results in an error message.

To test (1), you can add the parameter after this line: https://github.com/mozilla/web-ext/blob/72bb28ca8e29e18129fc24bb26c85d9ce5f10b82/tests/functional/test.cli.run.js#L20

To test (2), you could copy that test, cut the parts that we don't need, and add it to the same file.

https://github.com/mozilla/web-ext/blob/master/tests/functional/test.cli.run.js#L11

src/program.js
```diff
@@ -442,6 +442,7 @@ Example: $0 --help run.
         default: 'firefox-desktop',
         demandOption: false,
         type: 'array',
+        choices: ['firefox-desktop', 'firefox-android'],
```
Rob--W: Adding `choices` here automatically causes them to be added to the help test, when `web-ext run --help` is used. Now the `description` (a few lines back) can also be simplified. Could you remove the text between parentheses to shorten the description?

tests/unit/test-cmd/test.run.js
onlywicked: Would it be fine if I use `assert` to validate the `exitCode` and `stderr`?

Rob--W: Yes, that works. The `reportCommandErrors` in the existing test mainly exists to help with debugging if there is an unexpected test failure.

src/program.js
```diff
@@ -436,8 +436,8 @@ Example: $0 --help run.
     .command('run', 'Run the extension', commands.run, {
       'target': {
         alias: 't',
-        describe: 'The extensions runners to enable (e.g. firefox-desktop, ' +
-                  'firefox-android). Specify this option multiple times to ' +
+        describe: 'The extensions runners to enable. ' +
```
Rob--W: Could you combine this line with the next two lines so that the string is only cut when the 80-character line limit has been reached?

tests/functional/test.cli.run.js
```diff
@@ -47,4 +48,31 @@ describe('web-ext run', () => {
            }
          });
        }));
+
+  it('should not accept: --target INVALIDTARGET',
+     () => withTempAddonDir(
```
Rob--W: In this test, the program is supposed to exit early due to invalid command line parameters. Therefore it is not necessary to create a temporary add-on. You can simplify the test:

```
it('should not accept: --target INVALIDTARGET', async () => {
  const argv = [ ... ] // 'run' and the --target options only
  return execWebExt(argv, {}).then(({exitCode, stderr}) => {
    ...
  });
});

src/program.js
onlywicked: OK

caitmuenster:
Hey @onlywicked, thanks so much for the patch! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition).

onlywicked:
I have signed up at mozillians.org.
My username is same.

onlywicked:
> 
> 
> Hey @onlywicked, thanks so much for the patch! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition).
> 
> Would you be interested in creating an account on mozillians.org? I'd be happy to vouch for you!

I have signed up for mozillians.org.
So can you please vouch for me.

caitmuenster:
Sorry for the delay, @onlywicked! Your Mozillians profile is now vouched. ✨

Welcome onboard! We look forward to seeing you around the project.

Issue #1656, opened 2019-07-12
Remove set-value and mixed-deep exceptions from .nsprc 


rpl:
As part of #1655 we are adding the recent  advisories related to `set-value` and `mixed-deep` in the .nsprc ignore list, because it is making the CI jobs for every pull request to fail (on the "Npm audit and lint github pr title" step, [e.g. as in this job](https://travis-ci.org/mozilla/web-ext/jobs/557924886)).

But #1655 is just a temporary measure to avoid unrelated failures in the other PR, nevertheless we want to remove this exceptions from the .nsprc file as soon as possible.

rpl:
Fixed as part of #1663 (35d07cd1).

PR #1663, opened 2019-07-18
fix: updated fx-runner and commitlint dependency to fix npm audit failure on lodash dep

### LINKED ISSUE #1659: upgrade to lodash required according to 'npm audit'
https://www.npmjs.com/advisories/1065

currently at head we are pinned to 4.17.11, https://github.com/mozilla/web-ext/blob/master/package-lock.json#L31

rpl:
Fixes #1659 and #1656 
Also supersedes #1660 and #1657


rpl:
@Rob--W to fix the npm audit failure I had to rebuild the package-lock.json from scratch (by removing the existing package-lock.json and the node_modules dir, otherwise the resulting package-lock.json was still including the vulnerable lodash version).

By rebuilding the package-lock.json from scratch I've been also able to remove all the exceptions in the .nsprc file.

PR #1679, opened 2019-08-10
chore(deps): switch to open

jamesgeorge007:
[opn](npmjs.com/package/opn) was deprecated and renamed to [open](npmjs.com/package/open)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11581,7 +11581,6 @@
         "inquirer": "^6.2.2",
         "lodash": "^4.17.14",
         "needle": "^2.2.4",
-        "opn": "^5.5.0",
         "os-name": "^3.0.0",
         "proxy-agent": "^3.1.0",
         "proxy-from-env": "^1.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -11581,6 +11581,7 @@
         "inquirer": "^6.2.2",
         "lodash": "^4.17.14",
         "needle": "^2.2.4",
+        "opn": "^5.5.0",
         "os-name": "^3.0.0",
         "proxy-agent": "^3.1.0",
         "proxy-from-env": "^1.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
       "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
-      "dev": true,
       "requires": {
         "@babel/highlight": "^7.0.0"
       }
@@ -306,7 +305,6 @@
       "version": "7.5.0",
       "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.5.0.tgz",
       "integrity": "sha512-7dV4eu9gBxoM0dAnj/BCFDW9LFU0zvTrkq0ugM7pnHEgguOEeOz1so2ZghEdzviYzQEED0r4EAgpsBChKy1TRQ==",
-      "dev": true,
       "requires": {
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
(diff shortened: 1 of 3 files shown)


jamesgeorge007:
cc @rpl

rpl:
Hey @jamesgeorge007 sorry for let you waiting for this small PR, I totally missed it until now.

The actual change looks reasonable (I didn't tried it yet, or looked at the `open` package's changelog for possible backward incompatibilities introduced in the new major release),
but the `package-lock.json` diff looks pretty "crowded" ("crowded" as in "the diff is way bigger than I was expecting" :-P)

The diff on the `packge-lock.json` file for this change should look like:

```diff
         "mimic-fn": "^1.0.0"
       }
     },
+    "open": {
+      "version": "6.4.0",
+      "resolved": "https://registry.npmjs.org/open/-/open-6.4.0.tgz",
+      "integrity": "sha512-IFenVPgF70fSm1keSd2iDBIDIBZkroLeuffXq+wKTzTJlBpesFWojV9lb8mzOfaAzM1sr7HQHuO0vtV0zYekGg==",
+      "requires": {
+        "is-wsl": "^1.1.0"
+      }
+    },
     "opn": {
       "version": "5.5.0",
       "resolved": "https://registry.npmjs.org/opn/-/opn-5.5.0.tgz",
```

Quite unfortunately npm seems to be generating a crazy big package-lock.json diff if the lock file is being updated when the node_modules directory is already populated by a previous npm install.

The trick I use to force npm to generate a clean package-lock.json diff is:

- I make sure that the package-lock.json is unchanged (as it was before adding or changing or removing a dependency from the package.json file)
- then I remove the node_modules directory completely (otherwise it is going to affect the resulting package-lock.json file)
- and finally I run npm install using the last available npm version (using `npx npm@latest install`)

would you mind to look into updating this PR to make the package-lock.json diff small as it should?

Let me know if you are experiencing any issue while trying to achieve it
(in the worst case, I can apply the change locally and preserve the metadata from your patch).

package-lock.json
```diff
@@ -11581,7 +11581,6 @@
         "inquirer": "^6.2.2",
         "lodash": "^4.17.14",
         "needle": "^2.2.4",
-        "opn": "^5.5.0",
```
Rob--W: You cannot remove a dependency of an external package; Undo this line removal.

[`snyk`](https://www.npmjs.com/package/snyk) appears to be referenced through [`dispensary`](https://www.npmjs.com/package/dispensary) (as an entry in [`devDependencies`](https://github.com/mozilla/dispensary/blob/da269f902ca456d1402623a40370462bb791b1df/package.json#L54)). This should therefore not affect the `web-ext` package when you install `web-ext`.

If you want to remove this `opn` dependency, open a pull request at snyk - https://github.com/snyk/snyk
, and once a version with the updated package has been published, open a pull request at dispensary - https://github.com/mozilla/dispensary
.

jamesgeorge007:
we're good to go after [#746](https://github.com/snyk/snyk/pull/746) gets merged :rocket:

Rob--W:
> we're good to go after [#746](https://github.com/snyk/snyk/pull/746) gets merged rocket

Even after merging that, the package-lock here should be updated to reference the latest version of snyk, not?

It helps that `dispensary`'s package.json specifies `"snyk": "^1.192.3"`, because that means that any 1.x.x version can be used instead.

rpl:
@jamesgeorge007 the package-lock.json file is machine generated and so it should never be changed manually (and the CI jobs are failing for that reason).

In https://github.com/mozilla/web-ext/pull/1679#issuecomment-523178996 I described how to make sure that npm itself will generate the expected minimal diff for you (and as I mentioned in that comment if you are experiencing issues even with the strategy described, let us know and we can apply the change locally and preserve the metadata from your patch).

jamesgeorge007:
`npm i` seems to generate the massive `diff` for `package-lock.json` each & every time (even after removing the `node_modules` directory and performing it all over again). It seems that my local environment is all messed up :thinking:

rpl:
@jamesgeorge007 I'm not sure why removing `node_modules` and running npm install using the last npm version (`npx npm@latest install`) is not enough to get the expected `package-lock.json` diff in your local development environment.

I've been using this strategy over the last few months and it seems to be consistently producing the expected diff for me.

Let me know if you figured out why is not working for you (as I'm curious about which are the underlying issues), otherwise don't worry we will deal with it (I'm going to apply your patch locally, generate the package-lock.json, and then I'll create a new PR with the metadata from the commit in this PR unchanged).

rpl:
I'm closing this PR, as it is now superseded by #1715 (which includes the changes from this PR and fixes the diff applied to the package-lock.json file).

PR #1690, opened 2019-08-31
fix: typo

ccarruitero:


Rob--W:
```diff
- Most WebExtension functionality is baked into the browsers themselves
+ Most WebExtension functionality is backed into the browsers themselves
```

This is not a typo. [Most WebExtension functionality] is baked into [the browsers themselves]" means that the functionality is already part of the browsers. A synonym (alternative phrase with the same meaning) for "baked into" is "ingrained in".

PR #1695, opened 2019-09-03
fix: make wssBroadcast an async function

### LINKED ISSUE #1686: chromium extension-runner reloadAllExtensions should wait wssBroadcast to be processed by the connected clients
As a followup to #1392, the `chromium` extension-runner should wait the `"webExtReloadAllExtensions"` message (broadcasted from the websocket server part of the `chromium` extension runner) to have been received and handled by all the connected clients [1] (See https://github.com/mozilla/web-ext/pull/1392/files#r231533242).

[1]:where the connected client is the "extension manager extension" used to remotely control a chromium instance, in particular to reload the extension passed to the `web-ext run` command automatically when one of its file has been changed).

ccarruitero:
fixes #1686

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -151,12 +151,43 @@ export class ChromiumExtensionRunner {
     });
   }
 
-  wssBroadcast(data: Object) {
-    for (const client of this.wss.clients) {
-      if (client.readyState === WebSocket.OPEN) {
-        client.send(JSON.stringify(data));
+  async wssBroadcast(data: Object) {
+    return new Promise((resolve) => {
+      const clients = this.wss ? new Set(this.wss.clients) : new Set();
+
+      function cleanWebExtReloadComplete() {
+        const client = this;
+        client.removeEventListener('message', webExtReloadComplete);
+        client.removeEventListener('close', cleanWebExtReloadComplete);
+        clients.delete(client);
       }
-    }
+
+      const webExtReloadComplete = async (message) => {
+        const msg = JSON.parse(message.data);
+
+        if (msg.type === 'webExtReloadExtensionComplete') {
+          for (const client of clients) {
+            cleanWebExtReloadComplete.call(client);
+          }
+          resolve();
+        }
+      };
+
+      for (const client of clients) {
+        if (client.readyState === WebSocket.OPEN) {
+          client.addEventListener('message', webExtReloadComplete);
+          client.addEventListener('close', cleanWebExtReloadComplete);
+
+          client.send(JSON.stringify(data));
+        } else {
+          clients.delete(client);
+        }
+      }
+
+      if (clients.size === 0) {
+        resolve();
+      }
+    });
   }
 
   async createReloadManagerExtension() {
@@ -216,9 +247,8 @@ export class ChromiumExtensionRunner {
         const msg = JSON.parse(evt.data);
         if (msg.type === 'webExtReloadAllExtensions') {
           const devExtensions = await getAllDevExtensions();
-          for (var ext of devExtensions) {
-            reloadExtension(ext.id);
-          }
+          await Promise.all(devExtensions.map(ext => reloadExtension(ext.id)));
+          ws.send(JSON.stringify({ type: 'webExtReloadExtensionComplete' }));
         }
       };
     })()`;
@@ -234,9 +264,7 @@ export class ChromiumExtensionRunner {
   async reloadAllExtensions(): Promise<Array<ExtensionRunnerReloadResult>> {
     const runnerName = this.getName();
 
-    // TODO(rpl): wait for the wssBroadcast to be processed by the connected
-    // client (https://github.com/mozilla/web-ext/issues/1686).
-    this.wssBroadcast({
+    await this.wssBroadcast({
       type: 'webExtReloadAllExtensions',
     });
 
--- a/tests/unit/test-extension-runners/test.chromium.js
+++ b/tests/unit/test-extension-runners/test.chromium.js
@@ -3,7 +3,7 @@
 import EventEmitter from 'events';
 
 import {assert} from 'chai';
-import {describe, it} from 'mocha';
+import {describe, it, beforeEach, afterEach} from 'mocha';
 import deepcopy from 'deepcopy';
 import fs from 'mz/fs';
 import sinon from 'sinon';
@@ -130,8 +130,13 @@ describe('util/extension-runners/chromium', async () => {
         message.match('Fake wss socket ERROR')
       )));
 
+    const reload = (client, resolve, data) => {
+      client.send(JSON.stringify({ type: 'webExtReloadExtensionComplete' }));
+      resolve(data);
+    };
+
     const waitForReloadAll = new Promise((resolve) =>
-      wsClient.on('message', resolve));
+      wsClient.on('message', (data) => reload(wsClient, resolve, data)));
     await runnerInstance.reloadAllExtensions();
     assert.deepEqual(JSON.parse(await waitForReloadAll),
                      {type: 'webExtReloadAllExtensions'});
@@ -139,7 +144,7 @@ describe('util/extension-runners/chromium', async () => {
     // TODO(rpl): change this once we improve the manager extension to be able
     // to reload a single extension.
     const waitForReloadOne = new Promise((resolve) =>
-      wsClient.on('message', resolve));
+      wsClient.on('message', (data) => reload(wsClient, resolve, data)));
     await runnerInstance.reloadExtensionBySourceDir('/fake/sourceDir');
     assert.deepEqual(JSON.parse(await waitForReloadOne),
                      {type: 'webExtReloadAllExtensions'});
@@ -150,9 +155,15 @@ describe('util/extension-runners/chromium', async () => {
     await new Promise((resolve) => wsClient2.on('open', resolve));
     wsClient.close();
 
+    const waitForReloadClient2 = new Promise((resolve) =>
+      wsClient2.on('message', (data) => reload(wsClient2, resolve, data)));
+
     await runnerInstance.reloadAllExtensions();
+    assert.deepEqual(JSON.parse(await waitForReloadClient2),
+                     {type: 'webExtReloadAllExtensions'});
+
     const waitForReloadAllAgain = new Promise((resolve) =>
-      wsClient2.on('message', resolve));
+      wsClient2.on('message', (data) => reload(wsClient2, resolve, data)));
     await runnerInstance.reloadAllExtensions();
     assert.deepEqual(JSON.parse(await waitForReloadAllAgain),
                      {type: 'webExtReloadAllExtensions'});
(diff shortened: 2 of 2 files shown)


ccarruitero:
@Rob--W thanks for the explanation. I'll work on complete the fix

ccarruitero:
@Rob--W I added the changes you mentioned, but I have some doubts about tests.

First, I planned to test that the event listener was added, I found how to get the number of events listeners, but not sure why is failing

Then I think to test that the method wait for the message back for return, but really not sure how to test this. 

Do you have any thought?

src/extension-runners/chromium.js
```diff
@@ -217,7 +231,8 @@ export class ChromiumExtensionRunner {
         if (msg.type === 'webExtReloadAllExtensions') {
           const devExtensions = await getAllDevExtensions();
           for (var ext of devExtensions) {
-            reloadExtension(ext.id);
+            await reloadExtension(ext.id);
+            ws.send(JSON.stringify({ type: 'webExtReloadExtensionComplete' }));
```
Rob--W: This should be done after the loop.

And although it does currently not matter, I hope that you are aware of the fact that previously the `reloadExtension` was run in parallel, but after the change they are run sequentially. In general, if you want to keep running them in parallel, use use an `Array`'s `map` method to generate an array of promises, and then use `Promise.all`.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,26 @@ export class ChromiumExtensionRunner {
...
   }
 
-  wssBroadcast(data: Object) {
+  async wssBroadcast(data: Object) {
+    const promisesArray = [];
     for (const client of this.wss.clients) {
       if (client.readyState === WebSocket.OPEN) {
-        client.send(JSON.stringify(data));
+        promisesArray.push(new Promise((resolve) => {
+          client.on('message', async (message) => {
```
Rob--W: With this implementation, the number of `message` listeners will increase whenever reload is called.

I suggest to store the list of clients in an array, create one function, and then use that function as a listener for every `client`. When any of the clients return a message, all listeners should be unregistered, and the promise should be resolved. When a client is disconnected, the listener should be unregistered and the client be removed from the list (and if the list is empty, the promise should be resolved).

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+          resolve();
+        }
+      };
+
+      clients.forEach((client) => {
+        if (client.readyState === WebSocket.OPEN) {
+          client.on('message', webExtReloadComplete);
+          client.on('close', () => {
+            client.removeListener('message', webExtReloadComplete);
+            clients.splice(clients.indexOf(client), 1);
```
Rob--W: Let's use a `Set` instead of an `Array`, then you can use `clients.delete(client)` instead of having to use `splice` / `indexOf`.

Whenever you remove an element from the set, also check whether it may be necessary to check whether the set is empty, i.e. `if (clients.size === 0) { resolve(); }`.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+            client.removeListener('message', webExtReloadComplete)
+          );
+          resolve();
+        }
+      };
+
+      clients.forEach((client) => {
+        if (client.readyState === WebSocket.OPEN) {
+          client.on('message', webExtReloadComplete);
+          client.on('close', () => {
```
Rob--W: The close listeners should also be removed eventually.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+      if (!clients || clients.length === 0) {
+        resolve();
       }
-    }
+
+      const webExtReloadComplete = async (message) => {
+        const msg = JSON.parse(message);
+
+        if (msg.type === 'webExtReloadExtensionComplete') {
+          clients.forEach((client) =>
```
Rob--W: `for (let client of clients) {`

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+            client.removeListener('message', webExtReloadComplete)
```
Rob--W: Let's be consistent - use `off` because you used `on` before.

(the behavior is identical, but for consistency use on/off (preferred) or addListener/removeListener).

src/extension-runners/chromium.js
```diff
@@ -217,8 +242,10 @@ export class ChromiumExtensionRunner {
         if (msg.type === 'webExtReloadAllExtensions') {
           const devExtensions = await getAllDevExtensions();
           for (var ext of devExtensions) {
+            // just removing await .. fix and keep feature?
```
Rob--W: The `await` is necessary. Otherwise there is no guarantee that all extensions have been reloaded.

Assuming that `reloadExtension` never rejects, the logic would be

```
await Promise.all(devExtensions.map(ext => reloadExtension(ext.id)));
```

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,58 @@ describe('util/extension-runners/chromium', async () => {
...
+      });
+      assert(await runnerInstance.reloadAllExtensions());
+    });
+
+    it('resolve when all client disconnect', async (done) => {
+      await connectClient();
+      await wsClient.close();
+
+      try {
+        await runnerInstance.reloadAllExtensions().catch(e => {
```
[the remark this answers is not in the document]
Rob--W: To know when a client has closed, tou need to await its "close".

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,58 @@ describe('util/extension-runners/chromium', async () => {
...
+
+    afterEach(async () => {
+      wsClient.close();
+      await runnerInstance.exit();
+    });
+
+    it('does not resolve before complete message from client', async () => {
+      await connectClient();
+      await runnerInstance.reloadAllExtensions();
+      assert.fail();
```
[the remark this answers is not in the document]
Rob--W: First, to comment on your code snippet:
Never use 

```
try {
  await somePromise.catch(handleError);
} catch (e) {
  handleError(e);
}
```

... because the first `.catch` will ensure that the promise never rejects.

Instead, do

```
try {
  await somePromise;
} catch (e) {
  handleError(e);
}
```

And if you are expecting a test, then you shouldn't use the above pattern. Because if `somePromise` were to resolve successfully, then `handleError` is not called. The code should cause a test failure if the error is not thrown, and otherwise (if an error is thrown), check the error and proceed to the next part of the test.

Secondly, it seems that you are attempting to await a promise, and you want to verify that it doesn't resolve until an explicit message has been sent. That logic is flawed, because the test would then wait forever.
Instead of waiting forever, what you should do is to have the following logic:

1. Initialize flag, e.g. to false.
2. create client and register message listener on client.
3. await `reloadAllExtensions()`.
4. Set flag to true.

The message listener should then wait a little bit (e.g. 333ms), assert that the flag is still false (i.e. that `reloadAllExtensions` has not returned yet), and then return a message.

ccarruitero:
@Rob--W I just updated the PR. Sorry for delay.

Take a look when you have a chance.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,71 @@ describe('util/extension-runners/chromium', async () => {
...
+      wsClient = new WebSocket(wsURL);
+      await new Promise((resolve) => wsClient.on('open', resolve));
+    };
+
+    afterEach(async () => {
+      wsClient.close();
+      await runnerInstance.exit();
+    });
+
+    it('does not resolve before complete message from client', async (done) => {
```
ccarruitero: I modified the test based on https://github.com/mozilla/web-ext/pull/1695#discussion_r324148564, but couldn't make this pass yet.

I'm setting a variable (`reloadMessage`), creating a client and registering a message listener inside a promise (`waitCompleteMessage`), that change the variable value and resolve when receive complete message. Add a timeout and await `reloadAllExtensions`.

The timeout is asserting that variable isn't change, send complete message, await `waitCompleteMessage` and assert variable was changed.

But `waitCompleteMessage` isn't resolving. I was adding some logs and notice that the listener doesn't called after sent complete message, even when is added before `reloadAllExtensions`.

Do you have any advice?

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,71 @@ describe('util/extension-runners/chromium', async () => {
...
+        const msg = JSON.stringify({type: 'webExtReloadExtensionComplete'});
+        wsClient.send(msg);
+      });
+      assert.isOk(await runnerInstance.reloadAllExtensions());
+    });
+
+    it('resolve when all client disconnect', async () => {
+      await connectClient();
+      wsClient.on('close', async () => {
+        assert.isOk(await runnerInstance.reloadAllExtensions());
```
Rob--W: This implementation is incorrect. The test will finish as soon as `wsClient.close();` has been called. You should call `reloadAllExtensions()`, save its return value in a variable, call `wsClient.close()` and then wait for the resolution of the promise that you've saved in the variable.

tests/unit/test-extension-runners/test.chromium.js
ccarruitero: I finally fix this test. I'm using just one flag that check is false before and become true after `reloadAllExtensions`, but let me know if you prefer to use two flags instead.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+            client.removeListener('message', webExtReloadComplete);
+          }
+          resolve();
+        }
+      };
+
+      for (const client of clients) {
+        if (client.readyState === WebSocket.OPEN) {
+          client.on('message', webExtReloadComplete);
+          client.on('close', () => {
```
Rob--W: The `close` listeners are still here. They need to be removed eventually, just like the `'message'` listener.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,37 @@ export class ChromiumExtensionRunner {
...
+            client.off('message', webExtReloadComplete);
+            clients.delete(client);
+          });
+
+          client.send(JSON.stringify(data));
+        }
```
Rob--W: Add an `else` block that calls `clients.delete(client);`, and move the `if (clients.size === 0) {` check from above to after the loop.

ccarruitero:
@Rob--W thanks for all your comments, I just updated the PR. Take a look when you have a chance

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,75 @@ describe('util/extension-runners/chromium', async () => {
...
+    const connectClient = async () => {
+      const wssInfo = runnerInstance.wss.address();
+      const wsURL = `ws://${wssInfo.address}:${wssInfo.port}`;
+      wsClient = new WebSocket(wsURL);
+      await new Promise((resolve) => wsClient.on('open', resolve));
+    };
+
+    afterEach(async () => {
+      if (wsClient) {
+        wsClient.close();
```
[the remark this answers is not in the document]
Rob--W: ^ `wsClient = null;` is still missing.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -357,4 +368,75 @@ describe('util/extension-runners/chromium', async () => {
...
+      await runnerInstance.reloadAllExtensions();
+    });
+
+    it('resolve when all client disconnect', async () => {
+      await connectClient();
+      await new Promise((resolve) => {
+        wsClient.on('close', () => {
+          resolve(runnerInstance.reloadAllExtensions());
+        });
+        wsClient.close();
```
Rob--W: Need `wsClient = null;` as well.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,41 @@ export class ChromiumExtensionRunner {
...
+  async wssBroadcast(data: Object) {
+    return new Promise((resolve) => {
+      const clients = this.wss ? new Set(this.wss.clients) : new Set();
+
+      const webExtReloadComplete = async (message) => {
+        const msg = JSON.parse(message);
+
+        if (msg.type === 'webExtReloadExtensionComplete') {
+          for (const client of clients) {
+            client.off('message', webExtReloadComplete);
```
Rob--W: I noticed that the tests fail on Node 8. That is because EventEmitter's `off` method was only introduced in Node 10.
We're going to drop support for Node 8 soon, because Node 8 is EOL in 2 months from now (https://github.com/nodejs/Release#release-schedule ).

If you want to land your patch *before* we drop Node 8 support, then you need to change `.off` to `removeListener` (3x)

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,41 @@ export class ChromiumExtensionRunner {
...
+          }
+          resolve();
+        }
+      };
+
+      for (const client of clients) {
+        if (client.readyState === WebSocket.OPEN) {
+          client.on('message', webExtReloadComplete);
+          const cleanWebExtReloadComplete = () => {
```
Rob--W: Move this to before `if (msg.type === 'webExtReloadExtensionComplete') {` and use a function expression instead of an arrow funtion. Then you can use `let client = this;`.

By doing so, you will be able to remove the `'close'` listener inside the `if (msg.type === 'webExtReloadExtensionComplete') {` handler as well (which is the reason why I wrote this comment in the first place).

src/extension-runners/chromium.js
Rob--W: We have dropped for Node 8:  https://github.com/mozilla/web-ext/commit/884106c65f5e89a9e3c36b2dbb0609a1af27501d
... so you can use `.off` again.

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,42 @@ export class ChromiumExtensionRunner {
...
+        clients.delete(client);
       }
-    }
+
+      const webExtReloadComplete = async (message) => {
+        const msg = JSON.parse(message.data);
+
+        if (msg.type === 'webExtReloadExtensionComplete') {
+          for (const client of clients) {
+            client.removeEventListener('message', webExtReloadComplete);
```
Rob--W: This should be replaced with `cleanWebExtReloadComplete.call(client);`

src/extension-runners/chromium.js
```diff
@@ -151,12 +151,42 @@ export class ChromiumExtensionRunner {
...
-    for (const client of this.wss.clients) {
-      if (client.readyState === WebSocket.OPEN) {
-        client.send(JSON.stringify(data));
+  async wssBroadcast(data: Object) {
+    return new Promise((resolve) => {
+      const clients = this.wss ? new Set(this.wss.clients) : new Set();
+
+      function cleanWebExtReloadComplete() {
+        const client = this;
+        client.removeEventListener('message', webExtReloadComplete);
```
Rob--W: You should also remove the `close` listener.

ccarruitero:
@Rob--W Thanks a lot for all your comments. I just updated the PR.

src/extension-runners/chromium.js
ccarruitero: I'm getting some timeout errors when change to on/off. I think we could keep add/remove listener keywords.

PR #1725, opened 2019-10-05
feat: Enabled by default visibility of extension pages and content scripts logs in the BrowserConsole

### LINKED ISSUE #1702: Enable "devtools.browserconsole.contentMessages" before running Firefox
Maybe both. The bug is that "console.log" no longer is visible in a browser console launched via "--bc" parameter. The feature request is that the mentioned pref should be set to "true" so "console.log" is visible again.

If I have "console.log" in my Addon, then these messages aren't shown in browser console.

I'm currently "working around" this "bug" in all my Addons:
https://github.com/M-Reimer/togglewebsitecolors/commit/b2a4b7130e8d72857e28935cd628b674cf529505

But I think this shouldn't be needed if "--bc" is on the command line. In this case the developer wants to see messages caused by his Addon.

Messages should be visible

* **Firefox version**: 69
* **Your OS and version**: Arch Linux (rolling release)

````
$ node --version && web-ext --version
v12.10.0
3.1.1

No "npm" installed on my system
````

### PR DESCRIPTION
**Pull request by rbrishabh**
Fixes #1702 

Hey @rpl!

I hope you're doing well!

I believe this is the only change that is needed to fix this issue. Is that right?

Also, do I need to use a conditional statement to only set ```extensions.logging.enabled``` to ```true``` when ```--browser-console``` is sent as a parameter when running web-ext?

Thanks!

### DIFF
--- a/src/firefox/preferences.js
+++ b/src/firefox/preferences.js
@@ -28,8 +28,8 @@ const prefsCommon: FirefoxPreferences = {
   // Disable the prompt for allowing connections.
   'devtools.debugger.prompt-connection': false,
 
-  // Turn off platform logging because it is a lot of info.
-  'extensions.logging.enabled': false,
+  // Allow extensions to log messages on browser's console
+  'extensions.logging.enabled': true,
 
   // Disable extension updates and notifications.
   'extensions.checkCompatibility.nightly': false,
--- a/src/firefox/preferences.js
+++ b/src/firefox/preferences.js
@@ -27,9 +27,11 @@ const prefsCommon: FirefoxPreferences = {
   'devtools.debugger.remote-enabled': true,
   // Disable the prompt for allowing connections.
   'devtools.debugger.prompt-connection': false,
+  // Allow extensions to log messages on browser's console.
+  'devtools.browserconsole.contentMessages': true,
 
-  // Allow extensions to log messages on browser's console
-  'extensions.logging.enabled': true,
+  // Turn off platform logging because it is a lot of info.
+  'extensions.logging.enabled': false,
 
   // Disable extension updates and notifications.
   'extensions.checkCompatibility.nightly': false,

### CODE REVIEW

File: src/firefox/preferences.js [hunk 1]
```diff
@@ -28,8 +28,8 @@ const prefsCommon: FirefoxPreferences = {
   // Disable the prompt for allowing connections.
   'devtools.debugger.prompt-connection': false,
 
-  // Turn off platform logging because it is a lot of info.
-  'extensions.logging.enabled': false,
+  // Allow extensions to log messages on browser's console
+  'extensions.logging.enabled': true,
```
rpl: "extensions.logging.enabled" is not the right about:config preference (as it enables verbose logs for some Firefox internals related to the extensions, e.g. the AddonManager is going to log more verbosely).

The one that we want is **devtools.browserconsole.contentMessages**, as also mentioned in #1702 issue summary, which makes the Browser Console to list messages coming from Firefox child processes (by default only the logs emitted from the main Firefox process are going to be visible in the Browser Console).

rbrishabh:
Ahh! Since I didn't find ```devtools.browserconsole.contentMessages```, I assumed I was supposed to look for something similar for logging considering the issue summary of #1702 as just a pointer to the kind of change you might want to make.

Anyways, I made the change in ```preferences.js``` now. 

Please review this again and let me know what you think.
Thanks!

@rpl

rbrishabh:
Hey @rpl, could you find any time to look into this?

Thanks!

rpl:
@rbrishabh Thanks, it looks good and works as expected.

> Also, do I need to use a conditional statement to only set extensions.logging.enabled to true when --browser-console is sent as a parameter when running web-ext?

I wouldn't bother to "only enabling it when --browser-console option is also passed to web-ext run", as an extension developer I would prefer it to be enabled by default to ensure that if I open the Browser Console manually (without using the --browser-console option) I'm still going to see the extension logs without having to find out the option and enabled it manually (the option can be toggle from the gear icon in the top-right corner of the Browser Console, but I think that it is quite easy to miss to notice it).

If we experience issues related to that preference we may re-evaluate this choice in a separate follow up.

In the meantime, thanks for the PR! I'm going to merge right now \o/

PR #1731, opened 2019-10-15
chore: Upgrade to node 10

### LINKED ISSUE #1556: web-ext requires Node >= 10

Bug

When you try to install web-ext on Node 8, you receive an error that you need Node 10. The reason for this is the `addons-linter` dependency, which depends on `dispensary`, which [requires Node 10](https://github.com/mozilla/dispensary/blob/master/package.json#L10).

It doesn't really matter for me as I can use 10 just fine, but since the release notes explicitly mention bumping the required version to 8, you might want to fix this.

web-ext to be able to be installed in Node 8 just fine.

````
````

This was in CI, but using web-ext 3.0.0.

wagnerand:
Fix #1556.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,7 @@
 {
   "presets": [
     ["@babel/env", {
-      "targets": {"node": "8"}
+      "targets": {"node": "10"}
     }],
     "@babel/flow"
   ],
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,29 +6,29 @@ os:
 - windows
 
 node_js:
-- '8'
 - '10'
+- '12'
 
 # Skip node_js 6 on travis windows workers (as it often fails because
 # travis fails to init the node_js 6 environment).
 matrix:
   fast_finish: true
   exclude:
-  - node_js: '8'
+  - node_js: '10'
     os: windows
 
 jobs:
   include:
   - stage: npm audit and lint github PR title
     ## Keep this in sync with the last version listed in the node_js property.
-    node_js: '10'
+    node_js: '12'
     os: linux
     script:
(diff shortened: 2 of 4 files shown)


kumar303:
The node 12 failures look like they might be related to a problem in `coveralls`. You could make the node 12 tests non-blocking and file a follow-up issue to get CI working in node 12. Or maybe use an env var to skip coverage on node 12 tests (with a to-do bug) if that's easier.

rpl:
> The node 12 failures look like they might be related to a problem in `coveralls`. You could make the node 12 tests non-blocking and file a follow-up issue to get CI working in node 12. Or maybe use an env var to skip coverage on node 12 tests (with a to-do bug) if that's easier.

@kumar303 yeah, "not using nyc on the travis windows workers" seems to the simplest workaround to this issue and I'm going to apply that workaround as part of #1738, which is also blocked on the same failure (I investigated it a bit and filed #1739 to track it and deal with it later). 

The underlying problem seems to be actually a new one of those odd npm-related issues (I've come to the conclusion that it is a hidden npm feature, "npm provides you one new free puzzle to solve per month", this month I've been particularly lucky and I got two: #1738 and https://github.com/mozilla/web-ext/pull/1726#issuecomment-546094200 :-P).

kumar303:
> npm provides you one new free puzzle to solve per month

Ha!

rpl:
@wagnerand we have fixed the nyc-related failure in #1742, would you mind to rebase this PR on top of the updated master branch?

wagnerand:
@rpl All green now.

rpl:
> @rpl All green now.

Great! thanks for rebasing the PR

Issue #1739, opened 2019-10-25
Investigate "nyc"-related failure while running test-coverage npm script on travis windows workers


rpl:
In some recent Pull Requests, the travis CI job running on windows is failing with the following error message:

```
Running mocha unit tests...
----------|----------|----------|----------|----------|-------------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------|----------|----------|----------|----------|-------------------|
All files |        0 |        0 |        0 |        0 |                   |
----------|----------|----------|----------|----------|-------------------|
The command "npm run test-coverage" exited with 1.
internal/modules/cjs/loader.js:638
    throw err;
    ^
Error: Cannot find module 'C:\Users\travis\build\mozilla\web-ext\node'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
```

After some tests to be able to bisect what combination of node, npm and nyc is able to trigger the issue, I've been able to reproduce the same failure locally consistently and it looks related to something going wrong between `nyc` and the particular `npm` version being used (sigh :°-| ).

To identify the npm version that starts to trigger this issue I've executed the following command on windows 10 and git bash on multiple npm versions:

```
$ export NPMV=NN.N.N; rm -rf node_modules/; npx npm@$NPMV ci && npx npm@$NPMV run test-coverage
``` 

with the following results:

```
- npm@6.4.1 => OK
- npm@6.5.0 => OK
- npm@6.6.0 => OK
- npm@6.10.3 => OK
- npm@6.11.0 => BROKEN but different error
- npm@6.11.3 => BROKEN
- npm@6.12.0 => BROKEN
```

On npm@6.11.0 the error is:

```
> web-ext@3.2.0 test-coverage D:\MOZLAB\web-ext
> nyc npm run test

'find_dp0' is not recognized as an internal or external command,
operable program or batch file.
internal/modules/cjs/loader.js:573
    throw err;
    ^

Error: Cannot find module 'D:\nyc\bin\nyc.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:571:15)
[...]

PR #1741, opened 2019-10-27
Fixes #1335 Added parameter to build command to set custom filename

### LINKED ISSUE #1335: Add option to web-ext build: --filename="%{applications.gecko.id}-%{version}.zip

Feature request

Built files are named according to [safeFileName(`${extensionName}-${manifestData.version}.zip`)](https://github.com/mozilla/web-ext/blob/master/src/cmd/build.js#L143-L144) without any ability to customize the filename.

Add a new option like build --filename="some-file.zip". However, to make it useful, it should also accept parameters for substitution, preferably corresponding to manifest contents. Example: `web-ext build --filename="%{applications.gecko.id}-%{version}.zip`

Originally mentioned here: https://github.com/mozilla/web-ext/issues/855#issuecomment-375214750

web-ext 2.7.0

birtony:
Fixes #1335 
Implemented logic for the custom filename option parameter `--filename` for the `build` command.

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -43,7 +43,8 @@ export type PackageCreatorParams = {|
   fileFilter: FileFilter,
   artifactsDir: string,
   overwriteDest: boolean,
-  showReadyMessage: boolean
+  showReadyMessage: boolean,
+  filename?: string,
 |};
 
 export type LocalizedNameParams = {|
@@ -115,6 +116,7 @@ export async function defaultPackageCreator(
     artifactsDir,
     overwriteDest,
     showReadyMessage,
+    filename,
   }: PackageCreatorParams,
   {
     eventToPromise = defaultEventToPromise,
@@ -132,22 +134,27 @@ export async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
-  let extensionName: string = manifestData.name;
+  let packageName;
+  if (filename) {
+    packageName = `${filename}.zip`;
+  } else {
+    let extensionName: string = manifestData.name;
 
-  let {default_locale} = manifestData;
-  if (default_locale) {
-    default_locale = default_locale.replace(/-/g, '_');
-    const messageFile = path.join(
-      sourceDir, '_locales',
-      default_locale, 'messages.json'
-    );
-    log.debug('Manifest declared default_locale, localizing extension name');
-    extensionName = await getDefaultLocalizedName({
-      messageFile, manifestData,
-    });
+    let {default_locale} = manifestData;
+    if (default_locale) {
+      default_locale = default_locale.replace(/-/g, '_');
+      const messageFile = path.join(
+        sourceDir, '_locales',
+        default_locale, 'messages.json'
+      );
+      log.debug('Manifest declared default_locale, localizing extension name');
+      extensionName = await getDefaultLocalizedName({
+        messageFile, manifestData,
+      });
+    }
+    packageName = safeFileName(
+      `${extensionName}-${manifestData.version}.zip`);
   }
-  const packageName = safeFileName(
-    `${extensionName}-${manifestData.version}.zip`);
   const extensionPath = path.join(artifactsDir, packageName);
 
   // Added 'wx' flags to avoid overwriting of existing package.
@@ -187,6 +194,7 @@ export type BuildCmdParams = {|
   asNeeded?: boolean,
   overwriteDest?: boolean,
   ignoreFiles?: Array<string>,
+  filename?: string,
 |};
 
 export type BuildCmdOptions = {|
@@ -206,6 +214,7 @@ export default async function build(
     asNeeded = false,
     overwriteDest = false,
     ignoreFiles = [],
+    filename,
   }: BuildCmdParams,
   {
     manifestData,
@@ -214,6 +223,7 @@ export default async function build(
       sourceDir,
       artifactsDir,
       ignoreFiles,
+      filename,
     }),
     onSourceChange = defaultSourceWatcher,
     packageCreator = defaultPackageCreator,
@@ -231,6 +241,7 @@ export default async function build(
     artifactsDir,
     overwriteDest,
     showReadyMessage,
+    filename,
   });
 
   await prepareArtifactsDir(artifactsDir);
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -223,7 +223,6 @@ export default async function build(
       sourceDir,
       artifactsDir,
       ignoreFiles,
-      filename,
     }),
     onSourceChange = defaultSourceWatcher,
     packageCreator = defaultPackageCreator,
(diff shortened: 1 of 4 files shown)


src/cmd/build.js
```diff
@@ -214,6 +223,7 @@ export default async function build(
       sourceDir,
       artifactsDir,
       ignoreFiles,
+      filename,
```
rpl: @birtony it looks that this part of the changes in this PR is still triggering a flow check error: https://travis-ci.org/mozilla/web-ext/jobs/603639960#L242

I haven't looked too deeply into the PR yet, but it looks that Flow is complaining that the type signature of the options passed to `createFileFilter` (which is named `FileFilterOptions`) has not been updated:
https://github.com/mozilla/web-ext/blob/065188db35a4f63c2d17ff744cc1de2601a62222/src/util/file-filter.js#L25-L30

(as a side note, don't worry about the failure logged in the jobs executed by the travis windows workers, it is unrelated to you change and we fixed it as part of #1742, if you are curious #1739 provides some additional details about that issue).

birtony: @rpl thank you for your suggestions! I have truly included `filename` in the wrong spot, as it seems to be not needed in `createFileFilter`. I have also provided required tests to cover the new feature in a2286ec.

src/cmd/build.js
```diff
@@ -132,22 +134,27 @@ export async function defaultPackageCreator(
...
-      sourceDir, '_locales',
-      default_locale, 'messages.json'
-    );
-    log.debug('Manifest declared default_locale, localizing extension name');
-    extensionName = await getDefaultLocalizedName({
-      messageFile, manifestData,
-    });
+  let packageName;
+  if (filename) {
+    packageName = `${filename}.zip`;
```
rpl: I was expecting that filename would be the full filename including the extension, at least if the filename we got has already an extension like ".zip" or ".xpi" (and this would also have the nice side-effect of making web-ext build able to create files already using the .xpi extension).

rpl: I'm also thinking that we should likely validate a bit the filename we got. 

e.g. we may at least check that it is a simple file name and not a relative or absolute path, so that the file will still be stored in the artifacts dir (as the build command currently does).

From a quick look to the nodejs API docs, it seems to me that [nodejs path.parse]() method may be useful to achieve that.

src/program.js
```diff
@@ -381,6 +381,15 @@ Example: $0 --help run.
       default: true,
       type: 'boolean',
     },
+    'filename': {
+      alias: 'f',
```
rpl: "f" is already been used in the `web-ext run` command as a shortcut for "--firefox-binary", even if this is technically only available for the build command I would still prefer to possibly using a different single char for this option or alternatively we may also just don't specify a single char alias.

Also, I'm wondering if this is the reason for the failure triggered on the travis CI windows workers: https://travis-ci.org/mozilla/web-ext/jobs/604599391#L677
But it is surprising that it is only failing on windows and not on the travis CI linux workers as well.

src/cmd/build.js
```diff
@@ -132,22 +134,27 @@ export async function defaultPackageCreator(
...
-    const messageFile = path.join(
-      sourceDir, '_locales',
-      default_locale, 'messages.json'
-    );
-    log.debug('Manifest declared default_locale, localizing extension name');
-    extensionName = await getDefaultLocalizedName({
-      messageFile, manifestData,
-    });
+  let packageName;
+  if (filename) {
```
Rob--W: The proposed implementation does not support variables as requested in #1335.

Let's undo the current change to this file, and do something like this:

```
filename = filename.replace( ... , ... );
const packageName = safeFilename(filename);
```

If you don't know exactly how the replacement should be implemented, look at the `getDefaultLocalizedName` method for an example. To start simple, you could support the variable names "name" and "version".

src/cmd/build.js
birtony: Hey @rpl, do you think something like this would be sufficient?
```js
try {
    const valFilename = path.parse(filename);
    if (valFilename.dir !== '') {
      throw new UsageError(
        'Error validating the filename: filename cannot contain any path'
      );
    } else {
      packageName = `${filename}`;
    }
} catch (error) {
    throw new UsageError(
      `Error validating the filename: ${error}`);
}
if (!packageName || packageName === '') {
    // set default filename
    // ...
}
```

birtony: Should we also limit the user on the number of extension types he/she can give to the filename?

src/program.js
birtony: That is an interesting error, yet I am not sure what causes it on Windows. It seems to me that leaving a single char alias empty would be a better choice in this case . I just can't think of any other intuitive alias for the _filename_ option.

src/cmd/build.js
rpl: > Should we also limit the user on the number of extension types he/she can give to the filename?

You mean to only allow some kind of extensions (like .zip and .xpi) to be used as the extension in the final filename, right?
yeah that sounds reasonable to me.

> Hey @rpl, do you think something like this would be sufficient?

I was thinking that we may first replace the part to interpolate and then do something like:

```
export function getPackageNameFromTemplate(
  filenameTemplate: string,
  manifestData: ExtensionManifest
): string {
  ...
  const packageName = ...;

  // Validate the resulting packageName string, after interpolating the manifest property
  // specified in the template string.
  const parsed = path.parse(packageName);
  if (parsed.dir) {
    throw new UsageError(
      `Invalid filename template "${filenameTemplate}". ` +
      `Filename "${packageName}" should not contain a path`
    );
  }
  if (!['.zip', '.xpi'].includes(parsed.ext)) {
    throw new UsageError(
      `Invalid filename template "${filenameTemplate}". ` +
      `Filename "${packageName}" should have a zip or xpi extension`
    );
  }

  return packageName;
```

src/program.js
rpl: sure, omitting a single char alias for this option sounds good to me too.

tests/unit/test-cmd/test.build.js
```diff
@@ -72,6 +72,21 @@ describe('build', () => {
     });
   });
 
+  it('gives the correct custom name to an extension', () => {
```
rpl: Let's also unit test the new helper method that does just interpolate the template string given the manifest data object (so that we can easily cover more of the other scenarios, like the error scenarios where the helper method is meant to throw).

rpl:
I'm closing this PR because it is not yet complete and in the meantime there are some conflicts (with changes landed in master in the meantime) that have to be solved.

If you still want to work on this and the issue has not been fixed yet, feel free to update this PR and mention me in a comment and I'll be happy to re-open this PR.

escapewindow:
@birtony are you planning on picking this back up? If not, do you mind if I try addressing these review comments?

birtony:
@escapewindow things got busy for me, so go ahead! Let me know if I could be of any help

PR #1766, opened 2019-11-24
fix: Reset browser.link.open_newwindow on desktop Firefox too.

emilio:
This confused me quite a lot, see
https://bugzilla.mozilla.org/show_bug.cgi?id=1598947

Changing the default behavior of target="_blank" is quite arbitrary and
unexpected, IMO.


emilio:
`npm run audit-deps` is failing, but that doesn't seem related to my change.

PR #1775, opened 2019-12-06
fix: Removed default '--mute-audio' flag for chromium

### LINKED ISSUE #1774: The Chromium launcher mutes audio by default
It depends... But I'll go with **request**.

The default launch flags for Chromium/Chrome currently include [`--mute-audio`](https://github.com/GoogleChrome/chrome-launcher/blob/v0.11.2/src/flags.ts#L22-L23) (which mutes the whole browser), but that flag [`is not filtered out by web-ext`](https://github.com/mozilla/web-ext/blob/3.2.1/src/extension-runners/chromium.js#L114-L115). 

Whilst I do agree that such flag might be useful for a lot of people, I believe that the default behavior shouldn't be to mute the entire browser, from a consistency standpoint, at least. That's because the Firefox launcher doesn't do that by default.

````
node --version && yarn --version && yarn web-ext --version
v13.2.0
1.21.0
3.2.1
````

### PR DESCRIPTION
**Pull request by jezhou**
Fixes #1774

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -112,7 +112,9 @@ export class ChromiumExtensionRunner {
     }
 
     const chromeFlags = ChromeLauncher.defaultFlags()
-      .filter((flag) => flag !== '--disable-extensions');
+      .filter((flag) => {
+        return flag !== '--disable-extensions' && flag !== '--mute-audio';
+      });
 
     chromeFlags.push(`--load-extension=${extensions}`);
 
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -40,10 +40,13 @@ const log = createLogger(__filename);
 
 const asyncMkdirp = promisify(mkdirp);
 
+const EXCLUDED_CHROME_FLAGS = [
+  '--disable-extensions',
+  '--mute-audio',
+];
+
 export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags()
-  .filter((flag) => {
-    return flag !== '--disable-extensions' && flag !== '--mute-audio';
-  });
+  .filter((flag) => !EXCLUDED_CHROME_FLAGS.includes(flag));
 
 /**
  * Implements an IExtensionRunner which manages a Chromium instance.
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -40,6 +40,11 @@ const log = createLogger(__filename);
 
 const asyncMkdirp = promisify(mkdirp);
 
+export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags()
+  .filter((flag) => {
+    return flag !== '--disable-extensions' && flag !== '--mute-audio';
+  });
+
 /**
  * Implements an IExtensionRunner which manages a Chromium instance.
  */
@@ -111,10 +116,7 @@ export class ChromiumExtensionRunner {
       log.debug(`(chromiumBinary: ${chromiumBinary})`);
     }
 
-    const chromeFlags = ChromeLauncher.defaultFlags()
-      .filter((flag) => {
-        return flag !== '--disable-extensions' && flag !== '--mute-audio';
-      });
+    const chromeFlags = [...DEFAULT_CHROME_FLAGS];
 
     chromeFlags.push(`--load-extension=${extensions}`);
 
--- a/tests/unit/test-extension-runners/test.chromium.js
+++ b/tests/unit/test-extension-runners/test.chromium.js
@@ -26,7 +26,9 @@ import {
 } from '../../../src/util/logger';
 
 const defaultChromeFlags = ChromeLauncher.defaultFlags()
-  .filter((flag) => flag !== '--disable-extensions');
+  .filter((flag) => {
+    return flag !== '--disable-extensions' && flag !== '--mute-audio';
+  });
 
 function prepareExtensionRunnerParams({params} = {}) {
   const fakeChromeInstance = {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: tests/unit/test-extension-runners/test.chromium.js [hunk 1]
```diff
@@ -26,7 +26,9 @@ import {
 } from '../../../src/util/logger';
 
 const defaultChromeFlags = ChromeLauncher.defaultFlags()
-  .filter((flag) => flag !== '--disable-extensions');
+  .filter((flag) => {
+    return flag !== '--disable-extensions' && flag !== '--mute-audio';
+  });
```
rpl: I've been briefly discussing the related issue with @Rob--W yesterday and we agreed that, to avoid "surprises" when we upgrade the chromium-launcher dependency, it may be good to add an additional test case that fails if chromium-launcher adds additional default flags.

You can get the list of the currently expected default flags here: https://github.com/GoogleChrome/chrome-launcher/blob/master/src/flags.ts

would you mind to add an additional test case that compares this "filtered out array of default flags" with a list of expected ones?

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -26,7 +26,9 @@ import {
 } from '../../../src/util/logger';
 
 const defaultChromeFlags = ChromeLauncher.defaultFlags()
```
rpl: I was curious about which are the other default flags in that list and if we should filter out more of them, and so I took a quick look.

The following ones may be worth a look:
 
- `--disable-component-extensions-with-background-pages` seems to have been added because of a puppeteer bug (See https://github.com/puppeteer/puppeteer/issues/4300), but I'm for leave it for now (until we eventually get a report, if it breaks some use case that we are not currently aware of)
- `--disable-sync` may be another flag that it may be reasonable to filter out (e.g. in case an extension developer wants to test something related to the storage.sync API)

@Rob--W do you have any opinion on these two?

Rob--W: The `--disable-component-extensions-with-background-pages  flag should be kept. It disables component extensions, i.e. built-in extensions that provide some functionality. Extensions are probably not relying on this.

I don't have a strong opinion on `--disable-sync`. It is not needed for `chrome.storage.sync`, because that API just stores the data locally if there is no synced account. I would be surprised if extensions need sync for testing, as it only makes the test more unpredictable. I'd keep sync disabled for now, until someone requests it to be enabled.

rpl: @Rob--W great, that is also my current opinion on both. Let's not filter them out yet.

tests/unit/test-extension-runners/test.chromium.js
jezhou: I think the link you shared is from `12.0`, while the version `web-ext` currently uses is `11.2` – There's at least 4 extra flags between the two versions.

I've pulled the expected flags from [`11.2/flags.ts`](https://github.com/GoogleChrome/chrome-launcher/blob/v0.11.2/src/flags.ts) for now. Let me know if this is okay 👍

jezhou:
Added the test; I also slightly refactored so that there's only one source of truth for the expected default flags. I mentioned it briefly above, but I believe the flags we want until we update to `12.0` are in `11.2`, which can be found [here](https://github.com/GoogleChrome/chrome-launcher/blob/v0.11.2/src/flags.ts).

jezhou:
Something I also want to ask about is that sometimes when I run the tests, `wssInfo.address()` seems to return null. I'm working on a macOS, with node `v13.2.0` and pm `6.13.1`.

image

This is from an extension I'm playing around with:

image


I'm not sure if there's something I'm missing in the setup. If this is a separate issue I can file it if needed

tests/unit/test-extension-runners/test.chromium.js
rpl: :+1: yep, you are right, I forgot that we didn't merged #1745 yet and so we are still on the 11.2 release.

rpl:
> Something I also want to ask about is that sometimes when I run the tests, `this.wss.address()` seems to return null. I'm working on a macOS, with node `v13.2.0` and npm `6.13.1`.
> ...
> This is from an extension I'm playing around with:
> ...
> I'm not sure if there's something I'm missing in the setup. 

@jezhou does it also happen on other node version in your system, or just with nodejs v13?

> If this is a separate issue I can file it if needed

yes, would you mind to file a separate issue, it looks unexpected and we should take a look into it (but I've briefly tried and I can't reproduce on my system by just running the tests on nodejs v13, I have to come back to it for further investigations)

src/extension-runners/chromium.js
```diff
@@ -40,6 +40,11 @@ const log = createLogger(__filename);
 
 const asyncMkdirp = promisify(mkdirp);
 
+export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags()
+  .filter((flag) => {
+    return flag !== '--disable-extensions' && flag !== '--mute-audio';
```
rpl: Nit, we may define another const for the chrome flags that we want to filter out, so that we can just add or remove flags from that list based on the issue we may notice (without having to change this filter manually).

@jezhou how that sounds to you?

jezhou: @rpl Sounds good – added a const in b77f064

jezhou:
> @jezhou does it also happen on other node version in your system, or just with nodejs v13?

I also tried it on nodejs v10 with no success. I'll probably try a few combinations to see if I can get it to work, and file an issue sometime tonight 🙂

src/extension-runners/chromium.js
rpl: Thanks!

caitmuenster:
Thanks for the patch, @jezhou! 🙌🏼 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

jezhou:
@caitmuenster Sure 🙂 I just made an account here https://mozillians.org/en-US/u/jezhou/

caitmuenster:
Awesome, you're vouched. Welcome onboard! I look forward to seeing you around the project. :)

PR #1782, opened 2019-12-11
feat: Disabled the default browser check for Chromium (#1781)

### LINKED ISSUE #1781: Make Chromium not prompt the user to set it as the default browser
This is a feature request.

When starting Chromium using `web-ext run -t chromium` it checks whether it is the default browser and prompts the user to set it as one when it isn't.

Chromium should not prompt the user to set it as the default browser.

mariyan-borisov:
Essentially closes #1781.

I didn't want to touch _DEFAULT\_CHROME\_FLAGS_, as it only contains the default flags of `chromium-launcher`, so that's why I added _EXTRA\_CHROME\_FLAGS_. The new list can also be used to add more command line flags in the future, there's plenty to pick from: https://peter.sh/experiments/chromium-command-line-switches/.
The thing I'm not sure about is the correctness of my test.

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -48,6 +48,10 @@ const EXCLUDED_CHROME_FLAGS = [
 export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags()
   .filter((flag) => !EXCLUDED_CHROME_FLAGS.includes(flag));
 
+export const EXTRA_CHROME_FLAGS = [
+  '--no-default-browser-check',
+];
+
 /**
  * Implements an IExtensionRunner which manages a Chromium instance.
  */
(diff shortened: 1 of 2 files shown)


Rob--W:
I think that it's fine to add the flag to `DEFAULT_CHROME_FLAGS`, because there is no real need to distinguish between default flags from upstream, and what we consider to be default flags. We already filter out flags that we don't want anyway.

This flag could be a useful default for testing. Have you tried to get this flag in the upstream package?

mariyan-borisov:
> This flag could be a useful default for testing. Have you tried to get this flag in the upstream package?

No, I haven't. I'm not sure about the goal of that project and whether they are interested in adding this flag as a default.

mariyan-borisov:
Okay, I created [an issue](https://github.com/GoogleChrome/chrome-launcher/issues/181) in the project of `chrome-launcher`. Lets see what they think about my suggestion.

rpl:
@mariyan-borisov thanks for fixing the issue on the upstream project!

I'm closing this in favor of updating the chrome-launcher dependency once released.

PR #1784, opened 2019-12-14
feat: Added option to `--watch-file` to address eager reloading

### LINKED ISSUE #1102: Add option to delay auto-reload (support webpack better)

Feature request

See https://github.com/mozilla/web-ext/issues/1102#issuecomment-339143339

Right now, as soon as a file is modified in the extension directory, the reload is initiated.

This means that if you use a build system, like `webpack`, that takes a separate source code directory to generate the extension files into the target directory, as soon as the first file is modified in the target directory, the extension starts reloading, without the entire extension code being ready. So you have to force another reload with `CTRL+R` to get the correct extension files loaded.

We should have an option like `--reload-delay <seconds>` (or `<milliseconds>` maybe) to delay the reload after the watch system detects the last file change. This would allow the entire build to finish before the extension is reloaded.

### LINKED ISSUE #1137: File watcher too eager for changes

### LINKED ISSUE #1626: Add parameter "--watch-file" to allow reloading after build is done
Feature

`web-ext` will try to reload add-on too soon. If you are using some builder, this will often result in partially loaded add-on as your sources are only being build.
See the related bug that tries to solve the save problem with a delay parameter: https://github.com/mozilla/web-ext/issues/1573

As @Rob--W suggested: https://github.com/mozilla/web-ext/issues/1573#issuecomment-482526896 one solution would be to watch only single file, that would be modified last when build is done, for example some "build.version".
Even though it requires changes in your bundler, it's still easy to do and it's reactive - so it will save time and your CPU as `web-ext` will not try to reload not fully build add-on.

jezhou:
Fixes #1626 
Fixes #1102 
Fixes #1137

This PR adds a **`--watch-file`** parameter, which should resolve the problem of `web-ext` reloading changes too eagerly when using a custom builder/bundler system like Webpack. It helps with this issue by only reloading the extension when the developer touches a specific file _after_ building their extension (this require some configuration on the developer's part).

I got confused in the implementation details since the issue describe the feature as watching a particular file, but some of the finer details seem to focus on watching a new directory.

I wasn't sure which one to implement – file or directory – so I did file because a) it provides a finer grain of control that I think solves the problem and b) the assertions in `test.watcher.js` were easier to write using a file.

I think this isn't too bad for a first pass, but `test.watcher.js` ended up being a little tricker than I thought, so definitely need some second eyes on that. I'll annotate why I thought it was tricky below.


tests/unit/test.watcher.js
```diff
@@ -38,18 +50,47 @@ describe('watcher', () => {
...
+              }),
+            // Time out if no files are changed
+            new Promise((resolve) => setTimeout(() => {
               watcher.close();
-              sinon.assert.calledOnce(onChange);
-              // This delay seems to avoid stat errors from the watcher
-              // which can happen when the temp dir is deleted (presumably
-              // before watcher.close() has removed all listeners).
-              return new Promise((resolve) => setTimeout(resolve, 2));
-            });
```
jezhou: The way this was written before, `whenFilesChanged` will not resolve unless WatchPack actually detects the changes. This leads to a problem with the new feature because with the new option, if the detected file isn't modified, this will never resolve, which fails the test.

I added a "timeout" to the promise using `Promise.race()` – which after experimenting with seems to correctly fail the assertions if I want them to. The only bad thing about this I guess is that one test will always take 500ms (the timeout in that second promise). I don't think that's too bad, but if there's a better way to do all of this I'd love to learn 😄

jezhou:
I think the build is failing because of `npm audit`. Not sure if I should fix the vulnerability in this PR or leave it to another PR?

Rob--W:
> I think the build is failing because of `npm audit`. Not sure if I should fix the vulnerability in this PR or leave it to another PR?

The error is unrelated to your change, so please create a new PR (or open a new issue).

src/watcher.js
```diff
@@ -42,9 +49,14 @@ export default function onSourceChange(
     proxyFileChanges({artifactsDir, onChange, filePath, shouldWatchFile});
   });
 
-  log.debug(`Watching for file changes in ${sourceDir}`);
+  log.debug(`Watching for file changes in ${watchFile || sourceDir}`);
+
   watcher.watch([], [sourceDir], Date.now());
 
+  if (watchFile) {
```
Rob--W: To avoid reloading too often, when `watchFile` is specified, `sourceDir` should not be watched any more. Could you update the implementation and tests with this expectation?

jezhou:
@Rob--W I added a few more assertions to make sure the Watchers in WatchPack were only watching either the `watchFile` or the `sourceDir`.

Implementation wise, I'm not sure if I need to add anything else. It looks like the WatchPack README says that calling `watch()` twice will override existing files and directories. [It looks like that's the case when I look at the source too.](https://github.com/webpack/watchpack/blob/8d839e8b6edb87f9cd8a1ff97028060c4ecc9bee/lib/watchpack.js#L33-L49 )

tests/functional/test.cli.run.js
```diff
@@ -11,13 +11,14 @@ const EXPECTED_MESSAGE = 'Fake Firefox binary executed correctly.';
...
 
-  it('should accept: --no-reload --source-dir SRCDIR --firefox FXPATH',
+  it('accepts: --no-reload --watch-file --source-dir SRCDIR --firefox FXPATH',
      () => withTempAddonDir(
        {addonPath: minimalAddonPath},
        (srcDir) => {
          const argv = [
            'run', '--verbose', '--no-reload',
            '--source-dir', srcDir,
+           '--watch-file', srcDir,
```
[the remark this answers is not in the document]
jezhou: Done – I threw a `UsageError` here if a directory is passed in. Let me know if there's a better error to throw.

src/cmd/run.js
```diff
@@ -104,6 +109,15 @@ export default async function run(
     noReload = true;
   }
 
+  if (
+    watchFile &&
+    fs.existsSync(watchFile) &&
+    fs.lstatSync(watchFile).isDirectory()
+  ) {
+    throw new UsageError(`The directory "${watchFile}" cannot be ` +
```
Rob--W: This check is not strictly necessary, but it will probably help against common mistakes.

Let's move this whole check to the `src/watcher.js`, right before `watchedFiles.push(watchFile);`.
And use `isFile()` instead of `isDirectory`. That would also defend against cases where the file path refers to non-files that are not directories, such as symlinks.

src/watcher.js
```diff
@@ -42,8 +51,23 @@ export default function onSourceChange(
...
-  log.debug(`Watching for file changes in ${sourceDir}`);
-  watcher.watch([], [sourceDir], Date.now());
+  log.debug(`Watching for file changes in ${watchFile || sourceDir}`);
+
+  const watchedDirs = [];
+  const watchedFiles = [];
+
+  if (watchFile) {
+    if (fs.existsSync(watchFile) && !fs.lstatSync(watchFile).isFile()) {
+      throw new UsageError(`"${watchFile}" cannot be passed` +
```
Rob--W: "cannot be passed" is vague. Could you rephrase the error message so that it is more actionable? E.g. to:

```
throw new UsageError(`"Invalid --watch-file value: ${watchFile}" is not a file`)
```

jezhou:
@Rob--W do you think you can land this? I don't have write permissions for this repo 🙂

Rob--W:
Thanks, merged!

PR #1790, opened 2019-12-17
docs: Fixed link to signing addon

axlwaii:
While going through the README I discovered that the link `http://olympia.readthedocs.org/en/latest/topics/api/signing.html` was not working. I figured similiar to the `linting`, the addon should be linked here. 

If this is not the case please feel free to close the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Here is a partial list of examples:
 * Integrating with services.
   * Mozilla offers some useful services such as
     [linting](https://github.com/mozilla/addons-linter) and
-    [signing](https://github.com/mozilla/sign-addon)
+    [signing](https://addons-server.readthedocs.io/en/latest/topics/api/signing.html)
     extensions.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -199,7 +199,7 @@ Here is a partial list of examples:
 * Integrating with services.
   * Mozilla offers some useful services such as
     [linting](https://github.com/mozilla/addons-linter) and
-    [signing](http://olympia.readthedocs.org/en/latest/topics/api/signing.html)
+    [signing](https://github.com/mozilla/sign-addon)
```
rpl: @axlwaii you are right, the link is outdated, the new one should be https://addons-server.readthedocs.io/en/latest/topics/api/signing.html.

would you mind to update this PR to link this one instead of the sign-addon repository? (the doc page also mentions both `web-ext sign` and `sign-addon`, as well as describing the underlying signing API).

axlwaii: @rpl Thanks for reviewing the changes. I replaced the link with https://addons-server.readthedocs.io/en/latest/topics/api/signing.html as suggested.

PR #1791, opened 2019-12-17
Add support for browser_specific_settings (as noted in docs)

andyrichardson:
## About
The [official documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings) solely states that an addon can specify its `id` in the following manner.

```json
  "gecko": {
    "id": "addon@example.com",
    "strict_min_version": "42.0"
  }
}
```

This ID attribute is not being considered by web-ext. This PR adds support for this feature.


src/util/manifest.js
```diff
@@ -85,6 +86,20 @@ export default async function getValidatedManifest(
 
 
 export function getManifestId(manifestData: ExtensionManifest): string | void {
-  return manifestData.applications ?
-    manifestData.applications.gecko.id : undefined;
+  if (
+    manifestData.applications &&
+    manifestData.applications.gecko.id
```
rpl: it seems reasonable to also check that the `applications.gecko` property actually exists here, as you are doing for the `browser_specific_settings.gecko` one below.

rpl:
As a side note, this was also failing on npm audit for reasons unrelated to this PR (I already merged #1787 to update the dependency that was triggerring the npm audit failure, and so you should also rebase this PR on top of a recent master to avoid that failure on this pull request)

src/util/manifest.js
andyrichardson: I initially added this but then realised the existing code didn't have this check [look like for this reason](https://github.com/mozilla/web-ext/blob/995ccf5b8e651cdfec336be257e908d9c2ee4a03/src/util/manifest.js#L72)

andyrichardson:
Thanks for the feedback. Sounds like this is something that could be merged so I will add some tests around it :+1:

src/util/manifest.js
andyrichardson: Updated the PR + types to cover this use case

andyrichardson:
@rpl Will run these through the linter when I get back to my laptop 👍

Rob--W:
@andyrichardson Do you need help to finish this PR?

tests/unit/test-util/test.manifest.js
```diff
@@ -205,29 +215,55 @@ describe('util/manifest', () => {
...
+    const id = 'basic-manifest@web-ext-test-suite';
+
+    it('returns application gecko ID (application.gecko.id)', () => {
+      assert.equal(getManifestId(basicManifest), id);
+    });
 
-    it('returns a gecko ID', () => {
-      assert.equal(getManifestId(basicManifest),
-                   'basic-manifest@web-ext-test-suite');
+    it('returns browser specific gecko ID (browser_specific_settings.gecko.id)', () => {
```
andyrichardson:
```suggestion
    it('returns browser_specific_settings.gecko.id', () => {
```

rpl:
Closing in favor of #1944

PR #1794, opened 2019-12-19
fix: npm release should only include production assets

rpl:
This PR does exclude from the released npm package any asset that isn't needed in production (like we did for the sign-addon dependency in mozilla/sign-addon#402).

Besides that, this PR also change the test-functional script to use `npm pack` to run the production mode functional tests on the same package we would release on npm (without the need to keep a separate list of files to include).

### DIFF
--- a/scripts/test-functional
+++ b/scripts/test-functional
@@ -5,11 +5,12 @@ const path = require('path');
 const shell = require('shelljs');
 const tmp = require('tmp');
 
-const config = require('./lib/config');
+const pkg = require('../package.json');
 const {mochaFunctional} = require('./lib/mocha');
 
 shell.set('-e');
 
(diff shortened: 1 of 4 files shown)


scripts/test-functional
```diff
@@ -42,9 +44,12 @@ if (testProductionMode) {
   shell.echo('\nPreparing web-ext production mode environment...\n');
   shell.rm('-rf', destDir, packageDir);
   shell.mkdir('-p', destDir, packageDir);
-  shell.cp('-rf', config.copy.productionModeAssets.src, packageDir);
+  shell.pushd(packageDir);
+  shell.exec(`npm pack ${srcDir}`);
```
Rob--W: If srcDir contains a space, then this may fail unexpectedly.

Issue #793, opened 2017-02-07
Environment variables get applied as options to any command


wagnerand:
### Is this a bug or feature request?

This is a bug.

### What is the current behavior?

I have set `WEB_EXT_API_KEY` and `WEB_EXT_API_SECRET` in my `~/.bash_profile`.
Running `web-ext build` for any webextension project fails with the following error:
`Unknown arguments: apiSecret, apiKey`
When I remove the two environment variables, `web-ext` works fine.

### What is the expected or desired behavior?

`web-ext build` (and other commands) should ignore environment variables they don't need.

### Version information (for bug reports)

* **Firefox version**: irrelevant
* **Your OS and version**: MacOS 1.12.3
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````

v7.5.0
4.1.2
1.6.0

**Comment by kumar303** [comment]
oof, good catch. I'm pretty sure this has existed ever since https://github.com/mozilla/web-ext/pull/226 but no one else noticed. It's because those env variables are only recognized by the `sign` command. For other commands, our parser thinks they are env vars for unknown options.

**Comment by eight04** [comment]
This make environment variable totally unusable... and it seems that web-ext use yargs to parse env variable (https://github.com/mozilla/web-ext/blob/master/src/program.js#L88). Should we move this to yargs or any other ideas to fix this?

**Comment by kumar303** [comment]
Oops, I should have filed an upstream bug originally. Here it is: https://github.com/yargs/yargs/issues/873

**Comment by stephenmathieson** [comment]
Any updates here? I just killed over an hour trying to figure this out.

----

Hack I'm using to work-around this for now:

```bash
APIKEY="$WEB_EXT_API_KEY"
APISECRET="$WEB_EXT_API_SECRET"
unset WEB_EXT_API_KEY
unset WEB_EXT_API_SECRET

# Zip the Firefox extension.
npx web-ext build --config-discovery=false

# Zip+sign (XPI) the Firefox extension.
npx web-ext sign \
  --api-key=$APIKEY \
  --api-secret=$APISECRET \
  --config-discovery=false
```

Rob--W:
No updates yet. I have planned a patch to upgrade our yargs dependency to resolve some issues ( #1307 and #1327 in particular). That might or might not fix this issue.
Regardless, you should be able to set the key and secret in the config file once #1327 is resolved.

stephenmathieson:
[...]

PR #1810, opened 2020-01-09
Temporary work-around to #793 by turning off strict argument checking

hildjj:
Marked all failing tests as 'pending'.  Put TODOs on each change, so that when the upstream bug is resolved, this can be backed out easily.

Note to maintainers: this is a *suggestion* with my developer hat on.  Please treat this as a normal request from an external user, with no special authority.

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -83,7 +83,8 @@ export class Program {
     this.yargs.parserConfiguration({
       'boolean-negation': true,
     });
-    this.yargs.strict();
+    // TODO: remove false when https://github.com/yargs/yargs/issues/873 is fixed
+    this.yargs.strict(false);
 
     this.commands = {};
     this.options = {};
@@ -105,7 +106,8 @@ export class Program {
         // web-ext lint ./src/path/to/file.js
         .demandCommand(0, 0, undefined,
                        'This command does not take any arguments')
-        .strict()
+        // TODO: remove false when https://github.com/yargs/yargs/issues/873 is fixed
+        .strict(false)
         .exitProcess(this.shouldExitProgram)
         // Calling env() will be unnecessary after
         // https://github.com/yargs/yargs/issues/486 is fixed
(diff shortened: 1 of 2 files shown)


rpl:
It would be great if we could avoid disabling strict (especially given that the upstream bug is also open for a long time, I'm not sure that a fix will come from a new yargs version anytime soon).

Another option could be to check if we can clear the environment variables that triggers the yargs failure before yargs did the parsing.

If we want to fix the issue with the credential environment variables in particular, that may be doable and not too complex.

rpl:
(As a side note, the failure on the travis job is not related to test failures, the relevant ones are skipped in the same patch, but to the commit message, because it doesn't currently follow the commitlint conventions used in this repository)

Rob--W:
> It would be great if we could avoid disabling strict (especially given that the upstream bug is also open for a long time, I'm not sure that a fix will come from a new yargs version anytime soon).

Disabling strict checking would resolve #793, at the cost of losing the ability to provide feedback about typos or other unrecognized flags. The cost outweighs the benefit in my opinion.

> Another option could be to check if we can clear the environment variables that triggers the yargs failure before yargs did the parsing.

I was also thinking of this. It may be possible to create a general implementation that iterates over all registered flags and then stripping the environment variables (and maybe even warn about the ignored ones), but if specific examples are provided, we can try to look into specific, less general ways to address the issue at hand.

@hildjj Which environment variables encouraged you to create this PR?

hildjj:
I have WEB_EXT_FIREFOX_PROFILE, WEB_EXT_FIREFOX, and WEB_EXT_API_KEY set in the environment I'm using.  It means I can't use either `web-ext run` *or* `web-ext sign`.

hildjj:
I was about to suggest that another work-around would be to finish #176 including docs, but that approach also does not work with `strict` on for the same reason.

I am not really convinced by the argument that having basic functionality work is less important than better first-time user experience in the face of typos.  However, I'll repeat that it's not my call and I'll defer to folks who have dealt with this community for years.

rpl:
I just wanted to double-check how complex was going to be to implement the strategy I was suggesting, and given that it turned out to be a pretty small change I pushed it in #1812 so that we can take a look at it and agree if it is a reasonable approach as I think.

(and I also just added a unit test and an integration test to the PR, and so if the strategy looks a compelling one the pull request is also ready for review).

rpl:
Closing as superseded by #1812 (and has been just merged on master).

PR #1812, opened 2020-01-10
fix: Remove WEB_EXT_* env vars not supported by the selected web-ext command

### LINKED ISSUE #793: Environment variables get applied as options to any command

rpl:
This PR contains another possible approach to fix #793 (as an alternative to disabling the yargs strict parsing as in #1810 

- [x] add new unit tests to cover the changes (and fix coveralls failure due to decreased code coverage)
- [x] add an integration test (to make sure the fix does work as expected when the options and environment vars are actually parsed and validated by yargs).

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -4,8 +4,10 @@ import path from 'path';
 import {readFileSync} from 'fs';
 
 import camelCase from 'camelcase';
+import decamelize from 'decamelize';
 import git from 'git-rev-sync';
 import yargs from 'yargs';
+import { Parser as yargsParser } from 'yargs/yargs';
 
 import defaultCommands from './cmd';
 import {UsageError} from './errors';
@@ -53,6 +55,7 @@ export class Program {
   shouldExitProgram: boolean;
   verboseEnabled: boolean;
   options: Object;
+  programArgv: Array<string>;
 
   constructor(
     argv: ?Array<string>,
@@ -65,6 +68,7 @@ export class Program {
     // NOTE: process.argv.slice(2) removes the path to node and web-ext
     // executables from the process.argv array.
     argv = argv || process.argv.slice(2);
+    this.programArgv = argv;
 
     // NOTE: always initialize yargs explicitly with the package dir
     // to avoid side-effects due to yargs looking for its configuration
@@ -174,6 +178,31 @@ export class Program {
     return argv;
   }
 
+  // Remove WEB_EXT_* environment vars that are not a global cli options
+  // or an option supported by the current command (See #793).
+  cleanupProcessEnvConfigs(systemProcess: typeof process) {
+    const cmd = yargsParser(this.programArgv)._[0];
+    const env = systemProcess.env || {};
+    const toOptionKey = (k) => decamelize(
+      camelCase(k.replace(envPrefix, '')), '-'
+    );
+
+    if (cmd) {
+      Object.keys(env)
+        .filter((k) => k.startsWith(envPrefix))
+        .forEach((k) => {
+          const optKey = toOptionKey(k);
+          const globalOpt = this.options[optKey];
+          const cmdOpt = this.options[cmd] && this.options[cmd][optKey];
+
+          if (!globalOpt && !cmdOpt) {
+            log.debug(`Environment ${k} not supported by web-ext ${cmd}`);
+            delete env[k];
+          }
+        });
+    }
+  }
+
   async execute(
     {
       checkForUpdates = defaultUpdateChecker,
@@ -190,6 +219,7 @@ export class Program {
     this.shouldExitProgram = shouldExitProgram;
     this.yargs.exitProcess(this.shouldExitProgram);
 
+    this.cleanupProcessEnvConfigs(systemProcess);
     const argv = this.getArguments();
 
     const cmd = argv._[0];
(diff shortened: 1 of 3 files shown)


src/program.js
```diff
@@ -174,6 +178,32 @@ export class Program {
...
+    const cmd = yargsParser(this.programArgv)._[0];
+    const env = systemProcess.env || {};
+    const toOptionKey = (k) => decamelize(
+      camelCase(k.replace(`${envPrefix}_`, '')),
+      '-'
+    );
+
+    if (cmd) {
+      Object.keys(env)
+        .filter((k) => k.startsWith(envPrefix))
```
Rob--W: You're looking for keys starting with `envPrefix`, but processing keys with the assumption that they start with `${envPrefix}_`. Let's be consistent.

```javascript
let prefix = `${envPrefix}_`;
...
      Object.keys(env)
        .filter((k) => k.startsWith(prefix))
        .map((k) => k.slice(prefix.length))
        .map((k) => decamelize(camelCase(k), '-'))
        .forEach((optKey) => {
```

Side note: In our documentation, we are only suggesting the use of `WEB_EXT_*`, but practice it seems that `WEB_EXT*` (without the second underscore) is also supported. Since we haven't documented that option, I'm fine with only considering the `WEB_EXT_*` case.

package.json
```diff
@@ -84,6 +84,7 @@
     "watchpack": "1.6.0",
     "ws": "7.2.1",
     "yargs": "15.1.0",
+    "yargs-parser": "16.1.0",
```
Rob--W: Use `yargs.Parser` to import the parser used by `yargs`, instead of adding an extra dependency.

https://github.com/yargs/yargs/blob/7293ad50d20ea0fb7dd1ac9b925e90e1bd95dea8/yargs.js#L1242-L1243

rpl: That was my preference too, but unfortunately it doesn't play well with the webpack bundling.
When importing 'yargs' we do import yargs' `index.js` and [it doesn't export the yargs object exported by yargs.js](https://github.com/yargs/yargs/blob/7293ad50d20ea0fb7dd1ac9b925e90e1bd95dea8/index.js#L9).

I also tried to import it as "yargs/yargs" but webpack didn't liked it (it may be a recent validation added by webpack to avoid issues like [mozilla/sign-addon/pull/383](https://github.com/mozilla/sign-addon/pull/383), but I haven't double-checked if that it is actually the case).

Given that we are using the yargs-parser just to detect the command without reimplementing the parsing that yargs does internally from scratch, but we do not use it to parse any actual option, I think we may import `yargs-parser` directly (at least if I don't manage to find another way in the meantime).

src/program.js
rpl: > You're looking for keys starting with `envPrefix`, but processing keys with the assumption that they start with `${envPrefix}_`. Let's be consistent.

yeah, that's fair :+1:

> Side note: In our documentation, we are only suggesting the use of `WEB_EXT_*`, but practice it seems that `WEB_EXT*` (without the second underscore) is also supported. Since we haven't documented that option, I'm fine with only considering the `WEB_EXT_*` case.

uhm... let me check first what yargs "thinks" about the `WEB_EXT*`, I haven't checked if it is parsing those as well (but from your comment I guess it does, and so it may be more reasonable to make sure that they are not going to trigger the same issue if someone is using them "by mistake", e.g. because of a typo).

package.json
Rob--W: Adding `import foo from 'yargs/yargs';` results in the following error:

```
WARNING in ./node_modules/yargs/yargs.js 378:33-40
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./src/program.js
 @ ./src/main.js
```

The error implies that webpack is attempting to bundle everything, which we obviously don't want.

To fix that, add `yargs/yargs` to `webpack.config.js`, e.g. like this:

```diff
diff --git a/webpack.config.js b/webpack.config.js
index eeb549b..dc39dba 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,6 +13,8 @@ Object.keys(require('./package.json').dependencies)
     nodeModules[mod] = 'commonjs ' + mod;
   });
 
+nodeModules['yargs/yargs'] = 'commonjs yargs/yargs';
+
 var rules = [
   {
     exclude: /(node_modules|bower_components)/,
```
With the above change, I can import and use `yargs/yargs` as expected.

rpl: \o/ looks great to me! Thanks for looking into that part in the meantime!!!

I did applied the change needed to import the yargs-parser from yargs/yargs (and I did also run the functional tests with TEST_LEGACY_BUNDLING=1 (as we do on travis) to double-check that we are also not triggering bugs like mozilla/sign-addon/pull/383 by configuring webpack properly as in the above snippet, and the bundle is generated correctly as we thought).

rpl:
@Rob--W this PR has rebased on top of #1813 and the tests still pass as expected.

should be ready for one last look

apiraino:
thanks @rpl ! :+1:

PR #1813, opened 2020-01-10
fix: Allow import of external modules without bundling

Rob--W:
This is to support use of `import ... from 'yargs/yargs'` in #1812

This would also have been the proper fix instead of #383 .

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,6 +13,16 @@ Object.keys(require('./package.json').dependencies)
     nodeModules[mod] = 'commonjs ' + mod;
   });
 
+// Allow use of importing parts of an external module, without bundling them.
+function nodeModulesExternalsHandler(context, request, callback) {
+  var mod = request.split('/', 1)[0];
+  if (Object.prototype.hasOwnProperty.call(nodeModules, mod)) {
+    callback(null, 'commonjs ' + request);
+    return;
+  }
+  callback();
+}
+
 var rules = [
   {
     exclude: /(node_modules|bower_components)/,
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -13,6 +13,16 @@ Object.keys(require('./package.json').dependencies)
     nodeModules[mod] = 'commonjs ' + mod;
   });
 
+// Allow use of importing parts of an external module, without bundling them.
+function nodeModulesExternalsHandler(context, request, callback) {
+  var mod = request.split('/', 1)[0];
+  if (Object.prototype.hasOwnProperty.call(nodeModules, mod)) {
+    callback(null, 'commonjs ' + request);
```
rpl: :+1: I really like this more generic solution

I'm tempted to say that I wouldn't mind this helper to log something during the webpack build, even just a console.log given that this is only used during development.

e.g. adding something like:
```
    // eslint-disable-next-line no-console
    console.log('Added external handler for', {mod, request});
```

Produce the following output:

```sh
$ npm run build
Clean dist files...
Running webpack...
Added modules external handler for { mod: 'yargs', request: 'yargs/yargs' }
Hash: ...
Version: webpack 4.41.5
Time: ...
```

Knowing from the build logs what additional "nested module" has been dynamically added to the webpack config (based on out imports) may be useful to investigate issue.

Anyway, it would be just "a nice to have", there isn't really no need to block this for an additional log (that we may not even need any time soon, hopefully never :-)).

I'm going to merge this PR right now and then rebase #1812 on top of this changes.

PR #1833, opened 2020-02-07
fix: Use local port to avoid triggering the firewall

Rob--W:
When `web-ext run` starts, the following prompt shows briefly on macOS:

> Do you want the application "node" to
> accept incoming network connections?

This is because we use `listen(0)` to find a port, and by default, this
method listens on any network interface.

Since we are only looking for free ports and not network connections,
limit `listen()` to localhost only.

### DIFF
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -251,8 +251,8 @@ export async function connectWithMaxRetries(
 export function findFreeTcpPort(): Promise<number> {
   return new Promise((resolve) => {
     const srv = net.createServer();
-    // $FLOW_FIXME: flow has his own opinions on this method signature.
-    srv.listen(0, () => {
+    // $FLOW_FIXME: signature for listen() is missing - see https://github.com/facebook/flow/pull/8290
+    srv.listen(0, '127.0.0.1', () => {
       const freeTcpPort = srv.address().port;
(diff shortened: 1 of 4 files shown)


rpl:
@Rob--W it looks that the CI job is currently failing because flow expects a different function signature:
https://travis-ci.org/mozilla/web-ext/jobs/647323742#L241-L257

I haven't tried to apply the change locally, but it may be enough to pass an explicit `undefined` as the third parameter that flow expects (`backlog`) and pass the callback as the forth parameter.

Rob--W:
This is a common signature, so I'll submit a PR against flow. Then we can merge it in #1746.

Rob--W:
Patch is up at https://github.com/facebook/flow/pull/8290

Rob--W:
I added a `FLOW_FIXME` comment. flow is supposed to support us in making our code better, not block us from landing otherwise good patches.

Once my upstream patch is merged, I'll follow up here.

rpl:
> I added a `FLOW_FIXME` comment. flow is supposed to support us in making our code better, not block us from landing otherwise good patches.
> 
> Once my upstream patch is merged, I'll follow up here.

Sure, I was going to suggest that too, no need to wait for the upstream fix for just the two places we had to mark with the flow fixme annotation.

PR #1836, opened 2020-02-11
fix: update reloadExtensionBySourceDir to reload only target extension

### LINKED ISSUE #1687: chromium extension-runner reloadExtensionBySourceDir should be able to identify and reload the single target extension
As a followup to #1392, the `chromium` extension-runner should be able to map the extension's source dir with the assigned extension id, so that `reloadExtensionBySourceDir` can identify and reload only the extension installed from the given source dir (currently it reload all of installed extensions by internally calling `reloadAllExtensions`).

See https://github.com/mozilla/web-ext/pull/1392/files#r231535200 for additional details about how we may achieve the `sourceDir <-> extensionId` mapping for the `chromium` extension-runner.

ccarruitero:
fix #1687

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -23,6 +23,9 @@ import type {
   ExtensionRunnerParams,
   ExtensionRunnerReloadResult,
 } from './base';
+import {
+  WebExtError,
+} from '../errors';
 
 type ChromiumSpecificRunnerParams = {|
    chromiumBinary?: string,
@@ -30,10 +33,15 @@ type ChromiumSpecificRunnerParams = {|
    chromiumLaunch?: typeof defaultChromiumLaunch,
 |};
 
+type ReloadableExtensionParams = {|
+  reloadableExtensions?: Map<string, string>,
+|};
+
 export type ChromiumExtensionRunnerParams = {|
   ...ExtensionRunnerParams,
   // Chromium desktop CLI params.
   ...ChromiumSpecificRunnerParams,
+  ...ReloadableExtensionParams,
 |};
 
 const log = createLogger(__filename);
@@ -60,14 +68,17 @@ export class ChromiumExtensionRunner {
   wss: WebSocket.Server;
   exiting: boolean;
   _promiseSetupDone: ?Promise<void>;
+  reloadableExtensions: Map<string, string>;
 
   constructor(params: ChromiumExtensionRunnerParams) {
     const {
       chromiumLaunch = defaultChromiumLaunch,
+      reloadableExtensions,
     } = params;
     this.params = params;
     this.chromiumLaunch = chromiumLaunch;
     this.cleanupCallbacks = new Set();
+    this.reloadableExtensions = reloadableExtensions || new Map();
   }
 
   // Method exported from the IExtensionRunner interface.
@@ -159,11 +170,15 @@ export class ChromiumExtensionRunner {
         this.exit();
       }
     });
+
+    await this.setupExtensionIds();
   }
 
   async wssBroadcast(data: Object) {
     return new Promise((resolve) => {
       const clients = this.wss ? new Set(this.wss.clients) : new Set();
+      const reloadableExtensions = this.reloadableExtensions;
+      const paramsExtensions = this.params.extensions;
 
       function cleanWebExtReloadComplete() {
         const client = this;
@@ -178,6 +193,15 @@ export class ChromiumExtensionRunner {
         if (msg.type === 'webExtReloadExtensionComplete') {
           for (const client of clients) {
             cleanWebExtReloadComplete.call(client);
+          }
+          resolve();
+        }
+
+        if (msg.type === 'extensionsInfoResult') {
+          for (let i = 0; i < msg.extensions.length; i++) {
+            const { sourceDir } = paramsExtensions[i];
+            const { id } = msg.extensions[i];
+            reloadableExtensions.set(sourceDir, id);
           }
           resolve();
         }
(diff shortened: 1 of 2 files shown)


src/extension-runners/chromium.js
```diff
@@ -292,10 +298,30 @@ export class ChromiumExtensionRunner {
...
+    });
+
+    return [{runnerName}];
+  }
+
+  /**
+   * Generate extensionId from extension sourceDir
+  **/
+  generateId(sourceDir: string): string {
+    const hexHash = crypto.createHmac('sha256', sourceDir).digest('hex');
```
Rob--W: You need to normalize the path. The `sourceDir` value needs to be in the same format as Chrome's internal `FilePath`, and on Windows the paths need to be normalized as well: https://cs.chromium.org/chromium/src/components/crx_file/id_util.cc?l=61&rcl=9e60666b0869fb07b4b6f92947d54f31fda1fb70&targetos=win

This directory-based ID-generation is difficult to get right. Are you sure that you want to go down this road?

src/extension-runners/chromium.js
```diff
@@ -292,10 +298,30 @@ export class ChromiumExtensionRunner {
   async reloadExtensionBySourceDir(
     extensionSourceDir: string // eslint-disable-line no-unused-vars
   ): Promise<Array<ExtensionRunnerReloadResult>> {
-    // TODO(rpl): detect the extension ids assigned to the
-    // target extensions and map it to the extensions source dir
-    // (https://github.com/mozilla/web-ext/issues/1687).
-    return this.reloadAllExtensions();
+    const runnerName = this.getName();
+    const extensionId = this.generateId(extensionSourceDir);
```
Rob--W: Extensin ID does not only depend on the directory location; the manifest is used too, and takes precedence.

For an example, search for `publicKeyToExtensionId` in https://github.com/Rob--W/crxviewer/blob/30ae3c4e7acd495f4bb87d2f74af4496aa237847/src/crxviewer.js

ccarruitero:
@Rob--W, I'm not totally sure how to get the `sourceDir` <-> `extensionId` reference without calculating the `extensionId`. If I'm missing something please let me know.

After the extension is loaded we can communicate with the `reload-manager-extension` via web sockets and get the extensionIds, but we don't have any reference to map the `extensionId` to a specific `sourceDir` directly.

I think this could works If we have only one extension, but we could have problems when have more than one.

Maybe we could use an extra field, like `name`, and have two maps one to reference the `sourceDir` and another with the `extensionId`? We have to consider something if the `name` is localized?

Rob--W:
@ccarruitero The `chrome.management` API provides the following metadata for every extension: https://developer.chrome.com/extensions/management#type-ExtensionInfo

You can compare the available metadata (e.g. `name`) with the known metadata to match the sourceDir and extensionId. Localization must be supported, and doing so shouldn't be too difficult since `web-ext` already has utilities that does that.

Now if you only use `name`, then there is a chance of conflicts when two extensions have the same name. If you want to rule out this issue, then you could try other approaches, such as:
- including more fields (e.g. `version`, `version_name`).
- relying on the order of `management.getAll` (if it is reliable).

... or generate the ID based on the file name. If you are willing to put in the effort to create a reliable path-to-ID generator, then I would gladly review and merge it.

Rob--W:
Do ping me when you believe that the patch is ready for review again.

ccarruitero:
@Rob--W the PR is ready for a new review.

The fail in travis is for an `addons-linter`'s dependency.
https://github.com/mozilla/addons-linter/pull/3075 should fix the dependency problem in `addons-linter`.

src/extension-runners/chromium.js
```diff
@@ -114,6 +129,11 @@ export class ChromiumExtensionRunner {
       this.params.extensions.map(({sourceDir}) => sourceDir)
     ).join(',');
 
+    this.params.extensions.forEach(({sourceDir, manifestData}) => {
+      const { name, version } = manifestData;
+      this.extensionsSourceDirs.set(sourceDir, `${name}-${version}`);
```
Rob--W: name+version is just one of the ways to identify an extension, and not fully reliable either.
When the developer updates one of those fields in manifest.json, the mapping is going to become obsolete. (another way that the logic could break is when the extension manifest is localized, with different default languages for the browser compared to `web-ext`)

Rather than storing two maps, I would prefer that you maintain one map in a more reliable way.

Have you tried to find all unpacked extensions (`management.getAll`), filtered by `installType == "development"`, and see if the order matches with the order of `--load-extension`? If that works, it would be an easier and a reliable way to map extension IDs.

If you wish, you can also add additional logic to verify that the ID of the installed extension matches with the started-up extension.

ccarruitero:
@Rob--W I just noticed that isn't possible specify more than one source dir with web-ext commands, so I think we will have only one extension to reload.

I updated the code to use one map. Take a look when you have a chance

src/extension-runners/chromium.js
```diff
@@ -181,6 +196,17 @@ export class ChromiumExtensionRunner {
...
         }
+
+        if (msg.type === 'extensionsInfoResult') {
+          paramsExtensions.forEach((paramExtension) => {
+            msg.extensions.forEach((extension) => {
+              const { sourceDir } = paramExtension;
+              const { id } = extension;
+              reloadableExtensions.set(sourceDir, id);
+            });
+          });
```
Rob--W: This logic is equivalent to:

```
if (msg.extensions.length) {
  const { id } = msg.extensions[msg.extensions.length - 1];
  paramsExtensions.forEach((paramExtension) => {
    const { sourceDir } = paramExtension;
    reloadableExtensions.set(sourceDir, id);
  });
}
```

... and that logic is incorrect.

ccarruitero:
> Side note: out of curiosity, what sparkled your interest in this issue? I thought that you wanted the ability to reload individual extensions through the library, but from your most recent comment, it seems that you aren't aware of ability to load multiple extensions through the library.

I just find this issue more interesting than others :D.

ccarruitero:
@Rob--W I updated the logic for set the reloadableExtensions map. Take a look when you have a chance.

willdurand:
No activity for years, I am closing this PR. If you want to work on this issue in the future, just submit a new PR!

PR #1855, opened 2020-03-07
fix: Added cleaning up old files In android

Vishalghyv:
fix: Added cleaning up old files In android
Calling cleanupCallbacks.
Also used
if (selectedArtifactsDir)
{log.debug('Cleaning up artifacts directory on the Android device...');
await adbUtils.clearArtifactsDir(selectedAdbDevice);}

But this caused in decrease of coverage.
So just this.cleanupCallbacks functions is called.


Vishalghyv:
This dependency problem is happening while installing web-ext also.
This wasn't the case 2 days ago. I did npm audit fix but 3 dependencies require manual review.

PR #1856, opened 2020-03-07
fix: Cleanup old web-ext run artifacts dir when we are connecting to an android device

### LINKED ISSUE #1591: Android development is not cleaning up old files
Bug

Each execution of:
```
web-ext run --target=firefox-android --android-device=emulator-5554 --firefox-apk=org.mozilla.firefox
```
will create new directory on the SD card and not deleting it afterwards.
After few days of development this consumed 400MB of space on my device.

There should be some clean-up.

* **Firefox version**: 66.02
* **Your OS and version**: Win 10x64 1809

````
````
v12.0.0
6.8.0
3.0.0

https://www.dropbox.com/s/9hvb6d6rms0fspi/Screenshot%202019-04-29%2015.50.26.png?dl=0

Vishalghyv:
fix: Added cleaning up old files In android
Calling cleanupCallbacks.
Also used
if (selectedArtifactsDir)
{log.debug('Cleaning up artifacts directory on the Android device...');
await adbUtils.clearArtifactsDir(selectedAdbDevice);}

But this caused in decrease of coverage.
So just this.cleanupCallbacks functions is called.

(Fixes #1591)

### DIFF
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -128,6 +128,15 @@ export class FirefoxAndroidExtensionRunner {
       adbBin, adbHost, adbPort,
     });
 
+    // Call all the registered cleanup callbacks.
+    for (const fn of this.cleanupCallbacks) {
+      try {
+        fn();
+      } catch (error) {
+        log.error(error);
+      }
+    }
+
     await this.adbDevicesDiscoveryAndSelect();
     await this.apkPackagesDiscoveryAndSelect();
     await this.adbCheckRuntimePermissions();


rpl:
Hi @Vishalghyv, the changes in this pull requests cannot cleaning up the old files stored by a previous `web-ext` run command, just calling the `cleanupCallbacks` is not going to do that.

The approach suggested by @Rob--W in https://github.com/mozilla/web-ext/issues/1591#issuecomment-490904127 is to make web-ext run to remove the directories from the target device sdcard that starts with the `web-ext-artifacts-` prefix, which is what we use for generating the unique name for the temporary directory on the target device.

In the src/extension-runners/firefox-android.js module that you are changing in this PR there is already some code that is interacting with the target device over adb, take a look at that code and think about how you could achieve what Rob is suggesting.

As a side note, you don't need to close the pull requests to apply changes on them, you should create a branch (e.g. using something like `git checkout -b branchname`, see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/proposing-changes-to-your-work-with-pull-requests for more docs about that), create a pull request for that branch and then keep updating that branch with more commits based on the review and feedback comments (or to fix test failures in the continuous integration jobs).

Vishalghyv:
Hey, Thanks for reviewing my pull request. 
Firstly, Yeah I created lot of pull request for small changes, Will certainly use the guide.
I looked through the code as suggested  Just wanted to ask the way of solution.
Since.selectedArtifactsDir is where artifacts are stored because in 
So will the below code work

if (this.selectedArtifactsDir) {
      await this.adbUtils.clearArtifactsDir(this.selectedAdbDevice);
    }
After calling check adbCheckRuntimePermession

If this is not the way, I was thinking of using runShellCommands and directly removing directory using the relative path but checking few conditions before it like directory and permissions

rpl:
> Hey @rpl , Thanks for reviewing my pull request.
> Firstly, Yeah I created lot of pull request for small changes, Will certainly use the guide.
> I looked through the code as suggested Just wanted to ask the way of solution.
> Since.selectedArtifactsDir is where artifacts are stored
> So will the below code work
> 
> if (this.selectedArtifactsDir) {
> await this.adbUtils.clearArtifactsDir(this.selectedAdbDevice);
> }
> After calling check adbCheckRuntimePermession
> 
> If this is not the way, I was thinking of using runShellCommands and directly removing directory using the relative path but checking few conditions before it like directory and permissions

We are already trying to clear the artifacts dir on the device when web-ext is shutting down, but there are still chances that we may have not been able to remove them (e.g. because we got disconnected from the device or emulator we were connected to over adb, as also mentioned by Rob in this comment https://github.com/mozilla/web-ext/issues/1591#issuecomment-490904127).

`this.selectedArtifactsDir` would not be set to the same path of a previous run, because it has a timestamp as a suffix as you can see here:

https://github.com/mozilla/web-ext/blob/9735a1f48696cd2dce719688fc4ce9f7cb3883f2/src/util/adb.js#L197

It would be good to add to the [ADBUtils class](https://github.com/mozilla/web-ext/blob/master/src/util/adb.js#L44) a new method that does what I described in https://github.com/mozilla/web-ext/pull/1856#issuecomment-598221877 

runShellCommands may definitely be useful, but I would also suggest to look if the underlying npm dependency (adbkit) has any additional API method that may be useful to achieve what we need (take a look to the API doc here: https://github.com/openstf/adbkit )

rpl:
superseded by #1866

PR #1866, opened 2020-03-21
fix: Cleanup old web-ext run artifacts dir when we are connecting to an android device

### LINKED ISSUE #1591: Android development is not cleaning up old files

Vishalghyv:
(Fixes #1591)

### DIFF
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -71,6 +71,7 @@ export type FirefoxAndroidExtensionRunnerParams = {|
   adbPort?: string,
   adbDevice?: string,
   adbDiscoveryTimeout?: number,
+  adbCleanArtifacts?: boolean,
   firefoxApk?: string,
   firefoxApkComponent?: string,
 
@@ -384,6 +385,30 @@ export class FirefoxAndroidExtensionRunner {
     await adbUtils.amForceStopAPK(selectedAdbDevice, selectedFirefoxApk);
   }
 
+  async adbOldArtifactsDir(removeArtifacts?: boolean) {
+    const {
+      adbUtils,
+      selectedAdbDevice,
+    } = this;
+
+    const foundDirectories = await adbUtils.checkOrCleanArtifacts(
+      selectedAdbDevice, removeArtifacts
+    );
+
+    if (!foundDirectories) {
+      return;
+    }
+    if (removeArtifacts) {
+      log.info('Old web-ext artifacts have been found and removed ' +
+              `from ${selectedAdbDevice} device`);
+    } else {
+      log.warn(
+        `Old artifacts directories have been found on ${selectedAdbDevice} ` +
+        'device. Use --adb-clean-artifacts to remove them automatically.'
+      );
+    }
+  }
+
   async adbCheckRuntimePermissions() {
     const {
       adbUtils,
@@ -440,6 +465,9 @@ export class FirefoxAndroidExtensionRunner {
       app: 'fennec',
       customPrefs,
     });
+
+    //Checking for older artifacts
+    await this.adbOldArtifactsDir(this.params.adbCleanArtifacts);
 
     // Choose a artifacts dir name for the assets pushed to the
     // Android device.
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -8,6 +8,8 @@ import {
 } from '../errors';
 import {createLogger} from '../util/logger';
 
+const DEVICE_DIR_BASE = '/sdcard';
+const ARTIFACTS_DIR_PREFIX = '/web-ext-artifacts';
 const log = createLogger(__filename);
 
 export type ADBUtilsParams = {|
@@ -194,7 +196,7 @@ export default class ADBUtils {
       return artifactsDir;
     }
 
-    artifactsDir = `/sdcard/web-ext-artifacts-${Date.now()}`;
+    artifactsDir = `${DEVICE_DIR_BASE}${ARTIFACTS_DIR_PREFIX}-${Date.now()}`;
 
     const testDirOut = (await this.runShellCommand(
       deviceId, `test -d ${artifactsDir} ; echo $?`
@@ -212,6 +214,44 @@ export default class ADBUtils {
     this.artifactsDirMap.set(deviceId, artifactsDir);
 
     return artifactsDir;
+  }
+
+  async checkOrCleanArtifacts(
+    deviceId: string, remove?: boolean
+  ): Promise<boolean> {
+    const {adbClient} = this;
+
+    let found = false;
+    log.debug('Finding older artifacts');
+
+    return wrapADBCall(async () => {
+      const files = await adbClient.readdir(deviceId, DEVICE_DIR_BASE);
+
+      for (const file of files) {
+        if (!file.isDirectory() ||
+              !file.name.startsWith('web-ext-artifacts-')) {
+          continue;
+        }
+
+        if (!remove) {
+          return true;
+        }
+
+        found = true;
+
+        const artifactsDir = `${DEVICE_DIR_BASE}/${file.name}`;
+
+        log.debug(
+          `Removing ${artifactsDir} artifacts directory on ${deviceId} device`
+        );
+
+        await this.runShellCommand(deviceId, [
+          'rm', '-rf', artifactsDir,
+        ]);
+      }
+
+      return found;
+    });
   }
 
   async clearArtifactsDir(deviceId: string): Promise<void> {
--- a/tests/unit/test-extension-runners/test.firefox-android.js
+++ b/tests/unit/test-extension-runners/test.firefox-android.js
@@ -128,6 +128,9 @@ function prepareSelectedDeviceAndAPKParams(
     startFirefoxAPK: sinon.spy(() => Promise.resolve()),
     setupForward: sinon.spy(() => Promise.resolve()),
     clearArtifactsDir: sinon.spy(() => Promise.resolve()),
+    checkOrCleanArtifacts: sinon.spy(
+      () => Promise.resolve(true)
+    ),
     setUserAbortDiscovery: sinon.spy(() => {}),
     ensureRequiredAPKRuntimePermissions: sinon.spy(() => Promise.resolve()),
     ...adbOverrides,
@@ -354,6 +357,79 @@ describe('util/extension-runners/firefox-android', () => {
            fakeADBUtils.startFirefoxAPK
          );
        });
+
+    it('check for older artifacts', async () => {
+      const adbOverrides = {
+        getOrCreateArtifactsDir: sinon.spy(
+          () => Promise.resolve('/sdcard/web-ext-dir')
+        ),
+      };
+      const overriddenProperties = {
+        params: {
+          adbDevice: 'emulator-1',
+          firefoxApk: 'org.mozilla.firefox',
+          buildSourceDir: sinon.spy(() => Promise.resolve({
+            extensionPath: fakeBuiltExtensionPath,
+          })),
+          adbCleanArtifacts: false,
+        },
+      };
+      const {
+        params, fakeADBUtils,
+      } = prepareSelectedDeviceAndAPKParams(
+        overriddenProperties, adbOverrides
+      );
+
+      const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+      await runnerInstance.run();
+
+      sinon.assert.calledWithMatch(
+        fakeADBUtils.checkOrCleanArtifacts,
+        'emulator-1',
+        false,
+      );
+
+      sinon.assert.callOrder(
+        fakeADBUtils.amForceStopAPK,
+        fakeADBUtils.checkOrCleanArtifacts
+      );
+    });
+    it('remove plausible older artifacts', async () => {
+      const adbOverrides = {
+        getOrCreateArtifactsDir: sinon.spy(
+          () => Promise.resolve('/sdcard/web-ext-dir')
+        ),
+      };
+      const overriddenProperties = {
+        params: {
+          adbDevice: 'emulator-1',
+          firefoxApk: 'org.mozilla.firefox',
+          buildSourceDir: sinon.spy(() => Promise.resolve({
+            extensionPath: fakeBuiltExtensionPath,
+          })),
+          adbCleanArtifacts: true,
+        },
+      };
+      const {
+        params, fakeADBUtils,
+      } = prepareSelectedDeviceAndAPKParams(
+        overriddenProperties, adbOverrides
+      );
+
+      const runnerInstance = new FirefoxAndroidExtensionRunner(params);
+      await runnerInstance.run();
+
+      sinon.assert.calledWithMatch(
+        fakeADBUtils.checkOrCleanArtifacts,
+        'emulator-1',
+        true,
+      );
+
+      sinon.assert.callOrder(
+        fakeADBUtils.amForceStopAPK,
+        fakeADBUtils.checkOrCleanArtifacts
+      );
+    });
 
     it('does run a specific apk component if specific', async () => {
       const {
--- a/tests/unit/test-util/test.adb.js
+++ b/tests/unit/test-util/test.adb.js
@@ -64,6 +64,7 @@ function getFakeADBKit(
       return [];
     }),
     shell: sinon.spy(() => Promise.resolve('')),
+    readdir: sinon.spy(() => Promise.resolve([])),
     startActivity: sinon.spy(() => {}),
     forward: sinon.spy(() => {}),
     push: sinon.spy(() => {
@@ -620,6 +621,75 @@ describe('utils/adb', () => {
 
          sinon.assert.notCalled(adb.fakeADBClient.shell);
        });
+  });
+
+  describe('checkOrCleanArtifacts', () => {
+    const artifactDir = `web-ext-artifacts-${Date.now()}`;
+    function makeFakeArtifact(artifactName, isDirectory) {
+      return {
+        name: artifactName,
+        isDirectory: () => {
+          return isDirectory;
+        },
+      };
+    }
+    const files = [
+      makeFakeArtifact(artifactDir.concat('fake-dir1'), true),
+      makeFakeArtifact(artifactDir.concat('fake-dir2'), true),
+      makeFakeArtifact(artifactDir.concat('fake-file1'), false),
+      makeFakeArtifact('not-web-ext-artifacts-dir', true),
+    ];
+
+    it('checks old artifacts', async () => {
+      const adb = getFakeADBKit({
+        adbClient: {
+          readdir: sinon.spy(() => Promise.resolve(files)),
+          shell: sinon.spy(() => Promise.resolve('')),
+        },
+        adbkitUtil: {
+          readAll: sinon.spy(() => Promise.resolve(Buffer.from('1\n'))),
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+
+      const promise = adbUtils.checkOrCleanArtifacts('device1', false);
+      const result = await assert.isFulfilled(promise);
+      assert.equal(result, true);
+
+      sinon.assert.calledOnce(adb.fakeADBClient.readdir);
+      //While checking of files shell shoudln't be called
+      sinon.assert.notCalled(adb.fakeADBClient.shell);
+    });
+
+    it('removes plausible artifacts directory', async () => {
+      const adb = getFakeADBKit({
+        adbClient: {
+          readdir: sinon.spy(() => Promise.resolve(files)),
+          shell: sinon.spy(() => Promise.resolve('')),
+        },
+        adbkitUtil: {
+          readAll: sinon.spy(() => Promise.resolve(Buffer.from('1\n'))),
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+      const artifactDirFullPath1 = `/sdcard/${artifactDir}fake-dir1`;
+      const artifactDirFullPath2 = `/sdcard/${artifactDir}fake-dir2`;
+
+      const promise = adbUtils.checkOrCleanArtifacts('device1', true);
+      const result = await assert.isFulfilled(promise);
+      assert.equal(result, true);
+
+      sinon.assert.calledOnce(adb.fakeADBClient.readdir);
+      sinon.assert.calledTwice(adb.fakeADBClient.shell);
+      sinon.assert.calledWithMatch(
+        adb.fakeADBClient.shell, 'device1',
+        ['rm', '-rf', artifactDirFullPath1]
+      );
+      sinon.assert.calledWithMatch(
+        adb.fakeADBClient.shell, 'device1',
+        ['rm', '-rf', artifactDirFullPath2]
+      );
+    });
   });
 
   describe('pushFile', () => {
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -45,6 +45,7 @@ export type CmdRunParams = {|
   adbPort?: string,
   adbDevice?: string,
   adbDiscoveryTimeout?: number,
+  adbCleanArtifacts?: boolean,
   firefoxApk?: string,
   firefoxApkComponent?: string,
 
@@ -87,6 +88,7 @@ export default async function run(
     adbPort,
     adbDevice,
     adbDiscoveryTimeout,
+    adbCleanArtifacts,
     firefoxApk,
     firefoxApkComponent,
     // Chromium CLI options.
@@ -165,6 +167,7 @@ export default async function run(
       adbPort,
       adbBin,
       adbDiscoveryTimeout,
+      adbCleanArtifacts,
 
       // Injected dependencies.
       firefoxApp,
--- a/src/program.js
+++ b/src/program.js
@@ -618,6 +618,11 @@ Example: $0 --help run.
         type: 'number',
         requiresArg: true,
       },
+      'adb-clean-artifacts': {
+        describe: 'Remove old artifact directories from the adb device',
+        demandOption: false,
+        type: 'boolean',
+      },
       'firefox-apk': {
         describe: (
           'Run a specific Firefox for Android APK. ' +


Vishalghyv:
@rpl Sir the main thing that I have done is  adbClient.readdir(deviceId, '/sdcard') and then for their result checked if there is directory and if it then its starting is "web-ext-artifacts-" if so then delete it.
Still checkartifact parameter as mentioned by @Rob--W in #1591 (comment) 
"https://github.com/mozilla/web-ext/issues/1591#issuecomment-490904127"
 is not done at all, I will try to figure it out if the current logic could work.

I couldn't get git to commit in the previous one but in this one I have used branch, so probably future commits won't require new pull request.

src/cmd/build.js
```diff
@@ -220,7 +220,6 @@ export default async function build(
     showReadyMessage = true,
   }: BuildCmdOptions = {}
 ): Promise<ExtensionBuildResult> {
-
```
rpl: Nit, not a big deal, but this change to src/cmd/build.js is unrelated and it would be better to not be part of this pull request.

src/extension-runners/firefox-android.js
```diff
@@ -122,15 +123,29 @@ export class FirefoxAndroidExtensionRunner {
       adbHost,
       adbPort,
       ADBUtils = DefaultADBUtils,
+      cleanArtifacts,
     } = this.params;
 
+
```
rpl: Nit, the additional empty line here is not really needed.

src/extension-runners/firefox-android.js
```diff
@@ -122,15 +123,29 @@ export class FirefoxAndroidExtensionRunner {
       adbHost,
       adbPort,
       ADBUtils = DefaultADBUtils,
+      cleanArtifacts,
```
rpl: It isn't clear to me who should be passing this cleanArtifacts option (it seems that it is not actually being used in the current version of this patch), was it meant to be a new command line option?

src/extension-runners/firefox-android.js
```diff
@@ -122,15 +123,29 @@ export class FirefoxAndroidExtensionRunner {
...
       adbBin, adbHost, adbPort,
     });
+    // Call all the registered cleanup callbacks.
+    for (const fn of this.cleanupCallbacks) {
+      try {
+        fn();
+      } catch (error) {
+        log.error(error);
+      }
+    }
```
rpl: what is the purpose of calling the cleanup callbacks when web-ext is just starting the firefox-android extension runner?
it looks wrong, this isn't going to cleanup old artifacts dir on the target device and it is not going to do anything useful for that purpose.

src/extension-runners/firefox-android.js
```diff
@@ -122,15 +123,29 @@ export class FirefoxAndroidExtensionRunner {
...
 
     await this.adbDevicesDiscoveryAndSelect();
     await this.apkPackagesDiscoveryAndSelect();
     await this.adbCheckRuntimePermissions();
+    if (cleanArtifacts) {
+      await this.adbClearArtifactsDir();
+    }
+    await this.adbCheckArtifactsDir();
```
rpl: could you explain me the rationale of this?
e.g.
- who is going to pass this `cleanArtifacts` set to `true`
- what are the roles of `adbClearArtifactsDir` and `adbCheckArtifactsDir` in the approach you are thinking of?

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +399,58 @@ export class FirefoxAndroidExtensionRunner {
...
+
+      if (del) {
+        log.debug('Cleared Older Artifacts');
+      } else {
+        log.debug('Coudnot Clear Older Artifacts: Fail!!!');
+      }
+
+    }
+    return;
+  }
```
rpl: why `adbClearArtifactsDir` has to call `adbUtils.clearArtifacts` 3 times?

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +399,58 @@ export class FirefoxAndroidExtensionRunner {
...
+      throw new UsageError(
+        'Artifacts directory found please check' +
+        'Run something like web-ext --cleanup-artifacts to remove those files '
+      );
+    } else {
+      return ;
+    }
+
+
+  }
```
rpl: Based on the logs emitted by this method, I guess that you would like to implement something like this:

- the user runs `web-ext run -t firefox-android ...` and `web-ext` detects some old artifacts dir on the target device and logs an **info** (or a **warn**) message to the user to warn the user about it (a **debug** one is not good enough because it would only be visible when web-ext is running in verbose mode, which isn't the default behavior)
- If the user wants to clear the old artifacts automatically, it does run the `web-ext` command again with the addition of the `--clean-artifacts` option (which we may actually rename to `--adb-clean-artifacts` to make it clear that it would remove the artifacts on the target device connected over adb)

I don't think that `web-ext run -t firefox-android` should fail if there are some old artifacts directories on the target device, I'll emit a warning log and leave to the user the choice to when those old artifacts should be cleared.

src/util/adb.js
```diff
@@ -213,6 +213,54 @@ export default class ADBUtils {
 
     return artifactsDir;
   }
+  fun(x: string) {
+    return x;
+  }
```
rpl: This fun doesn't have any purpose here.

src/util/adb.js
```diff
@@ -213,6 +213,54 @@ export default class ADBUtils {
...
+          // right time
+          dirs.forEach(function(dir) {
+            if (dir.isDir()) {
+              if (dir.name.startsWith('web-ext-artifacts-')) {
+                return false;
+              }
+            }
+          });
+        });
+    });
```
rpl: When you `await` on a call to a function that return a promise you don't need to use a callback passed to `then` to get the resolved value, in other words instead of:

```
callToFunctionThatReturnAPromise().then(function (result) { ... });
```

**in an async function** you can just write it as:

```
const result = await callToFunctionThatReturnAPromise();
...
```

See https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await to learn more about how async/await allows you to rewrite Promise-based async code.

src/util/adb.js
```diff
@@ -213,6 +213,54 @@ export default class ADBUtils {
...
+    for (const names of directory) {
+      await this.runShellCommand(deviceId, [
+        'rm', '-rf', names,
+      ]);
+    }
+    return true;
+  }
```
rpl: this `checkArtifacts` and `clearArtifacts` methods are pertty similar to each other, it looks like they could be merged into a single method (which takes an additional boolean parameter that would make it **"check and clear"** the artifacts directories vs. just **checking** them)

Vishalghyv:
@rpl , Thanks for reviewing the pull request and giving such brief and informative comments.
I have gone through them and will certainly work on better solution since, the basic logic required seems to be working, but still CLI and Automated tests are left in the next commits these will follow with completely testing the bug :)

src/util/adb.js
Vishalghyv: Thanks for suggesting. Certainly wouldn't require await here.

src/extension-runners/firefox-android.js
Vishalghyv: This for cli comment if user sends it then first clear artifact otherwise if user doesn't sends cli option then just check them and give info or warning to them as suggested below

package-lock.json
```diff
@@ -3214,8 +3214,7 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
```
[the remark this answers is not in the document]
Vishalghyv: Certainly, changes in pacakage .json will be fixed. Yes, firstly I will try these because they cause problem.

src/extension-runners/firefox-android.js
Vishalghyv: Yes, This needs to be correctly called and used. Will use implementation of current cli options used.

Vishalghyv:
Tests are left;
Added cli option
tested my device 
results : 
does give a log info to use something like --adb-clean-artifacts if older artifacts are found
and using the cli option does remove it

Vishalghyv:
Hey, @rpl  I have tried to implement the changes you requested.
Details of the implementation:
checkOrCleanArtifacts is the main function which creates array of directories starting with web-ext-artifacts which is send to remover function which removes the directories.
And this all is controlled by a check boolean type variable which depends on --adb-clean-artifacts cli option
In first round of tests (just to make sure moving in right direction) first remover is tested with two directories and then just the basic part of checkOrCleanArtifacts.
Please provide your feedback on this.
Query: If I use runShellCommand to remove folder '/sdcard/' what would happen?

src/extension-runners/firefox-android.js
```diff
@@ -131,6 +133,10 @@ export class FirefoxAndroidExtensionRunner {
     await this.adbDevicesDiscoveryAndSelect();
     await this.apkPackagesDiscoveryAndSelect();
     await this.adbCheckRuntimePermissions();
+    if (adbCleanArtifacts) {
+      await this.adbArtifactsDir(true);
+    }
+    await this.adbArtifactsDir(false);
```
rpl, line 139: In my opinion we should only do this once, we clean all the old artifacts dirs or we just check if there are any.

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +390,23 @@ export class FirefoxAndroidExtensionRunner {
...
+    const check = await adbUtils.checkOrCleanArtifacts(selectedAdbDevice,
+                                                       clear);
+    if (check) {
+      log.warn(
+        'Artifacts directory found ' +
+        'run something like web-ext --adb-clean-artifacts' +
+        ' to remove older Artifacts'
+      );
+    }
+  }
```
rpl, line 408: I think that here we should only log the warning if the adbUtils method is only checking if any old artifacts directory is present (and it was not actually removing them in this run).

src/util/adb.js
```diff
@@ -214,6 +214,41 @@ export default class ADBUtils {
...
+    if (directories.length !== 0) {
+      //If directories found and condition is to clean them
+      if (clean) {
+        return false;
+      }
+      //If directories found and condition is to only check
+      return true;
+    } else {
+      return false;
+    }
```
rpl, line 250: it looks that this method can be made a bit simpler.

e.g. (as I mentioned in my previous comment) it doesn't seem that we really need to collect the directories into an array and then iterate again over it.

Besides that, all these if/else blocks between line 241-250 could be replaced with something a bit cleaner (and less nested).

Try to ask yourself some questions to help you get a better picture of what are options you may have here, e.g.

- do we always need to iterate over all the results resolved by `adb.readdir` or could we exit earlier in certain cases? (e.g. if a call to this method is meant to only check if any old artifact dir is still on the device, in which conditions this would be are ready to return to the caller the outcome of that check?)

Vishalghyv:
> 
> 
> Thanks @Vishalghyv, this version looks already cleaner and better than the previous ones.
> Follows another round of review comments for this new version.

I mean first I thought my code was fine but following your suggestion was amazing. I couldn't work yesterday because of some errors in os. But will do today

Vishalghyv:
Sir, Yes I tried to go through all four condition which might occur in this 
if directories found or not and if clean is true or not -- Tried to optimize all four
And since dirs returned for readdir are in reversed order as in folder sdcard and artifacts are created at top thus reverse looping.
Please provide your feedback on it is helping a lot. Thanks by the way of such brief feedback

Vishalghyv:
> And since dirs returned for readdir are in reversed order as in folder sdcard and artifacts are created at top thus reverse looping.
This was required as in my case there were 30 folders in sdcard thus in some might have much more, Maybe this is not required?

src/util/adb.js
```diff
@@ -210,10 +210,45 @@ export default class ADBUtils {
...
+    deviceId: string, clean: boolean): Promise<boolean> {
+    const {adbClient} = this;
+
+    var directories = false;
+
+    await wrapADBCall(async () => {
+      log.debug('Finding Older Artifacts');
+
+      const dirs = await adbClient.readdir(deviceId, '/sdcard');
+      for (var dir = dirs.length - 1; dir >= 0; --dir ) {
```
rpl, line 226: I'm not really sure that we care about the order of the directories that much, if we really do we could still sort them before iterating over them.

I would just iterate over the list we got in the order we got it, with the "lighter" js syntax possible:
```
for (let file of files) {

}
```

src/util/adb.js
```diff
@@ -210,10 +210,45 @@ export default class ADBUtils {
     await this.runShellCommand(deviceId, ['mkdir', '-p', artifactsDir]);
 
     this.artifactsDirMap.set(deviceId, artifactsDir);
-
     return artifactsDir;
   }
 
+  async checkOrCleanArtifacts(
+    deviceId: string, clean: boolean): Promise<boolean> {
```
rpl, line 217: I think that `remove` would be a better name than `clean`, as it makes immediately clear that it means that files are going to be **removed** from the target device.

src/util/adb.js
```diff
@@ -210,10 +210,45 @@ export default class ADBUtils {
...
+    const {adbClient} = this;
+
+    var directories = false;
+
+    await wrapADBCall(async () => {
+      log.debug('Finding Older Artifacts');
+
+      const dirs = await adbClient.readdir(deviceId, '/sdcard');
```
rpl, line 225: lets not call this `dirs`, as it is misleading, this is not going to contains only directories. `files` may be a better name.

src/util/adb.js
```diff
@@ -210,10 +210,45 @@ export default class ADBUtils {
...
+      for (var dir = dirs.length - 1; dir >= 0; --dir ) {
+        if (!dirs[dir].isFile() &&
```
rpl, line 227: I thin I would prefer if we would use `file.isDirectory()`, because it is a bit more explicit, and we are only interested in directories (everything else, like symbolic links, device or socket files etc. does not interest us here)

Vishalghyv:
Sir, I have tried to implement the requested changes.

And will change var directories to let directories in the next commit

Vishalghyv:
Sorry, For so many commits and clean ups left
I tried to find a way to mock creation of dir during, so as to have some basic logic for testing, 
This is the most useful approach I was able to use, by this approach almost all conditions for the function can be tested like with no directories or with remove condition false
But I am sure there must be a better way
And thanks for the great explanation!s!!

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +387,29 @@ export class FirefoxAndroidExtensionRunner {
...
+      adbUtils,
+      selectedAdbDevice,
+    } = this;
+
+    const foundDirectories = await adbUtils.checkOrCleanArtifacts(
+      selectedAdbDevice, removeArtifacts
+    );
+
+    if (foundDirectories) {
+      if (removeArtifacts) {
```
rpl, line 401: we can avoid this nested if/else blocks, if I recall correctly in my previous review comment there were some pointers about how to achieve that.

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +387,29 @@ export class FirefoxAndroidExtensionRunner {
...
+        log.info('Old artifacts directories starting with web-ext-artifacts ' +
+                `removed on ${selectedAdbDevice} device`);
```
rpl, line 403: I think that this message could just say saomething like `Old web-ext artifacts have been found and removed from ${selectedAdbDevice}`.

Also, you need to make this a template string (See [doc page on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Expression_interpolation)) so that the `${...}` parts are going to be replaced with the related value.

src/extension-runners/firefox-android.js
```diff
@@ -384,6 +387,29 @@ export class FirefoxAndroidExtensionRunner {
...
+      } else {
+        log.warn(
+          'Old artifacts directories have been found of the target ' +
+          'android device.Use --adb-clean-artifacts to force their removals.'
```
rpl, line 407: Let's also tweak this message to be more consistent with the other one, e.g. `Old web-ext artifacts have been found on ${selectedAdbDevice}. Use --adb-clean-artifacts to remove them automatically.`

(also, please make sure to put a space between '.' and the message that follows it)

src/util/adb.js
```diff
@@ -210,10 +210,45 @@ export default class ADBUtils {
...
+      const files = await adbClient.readdir(deviceId, '/sdcard');
+
+      for (const file of files) {
+        if (file.isDirectory() &&
+              file.name.startsWith('web-ext-artifacts-')) {
+          directories = true;
+
+          if (!remove) {
+            break;
+          }
```
rpl, line 233: we can avoid this nested if/else blocks, in my previous review comment there should be some pointers about how to achieve that.

tests/unit/test-util/test.adb.js
```diff
@@ -59,11 +67,13 @@ function getFakeADBKit(
 
   adbUtilReadAllStub.onCall(0).returns(Promise.resolve(Buffer.from('')));
 
+
```
rpl, line 70: Nit, remove this unnecessary empty line (there was already one at line 69 to separate the two statements).

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +632,35 @@ describe('utils/adb', () => {
        });
   });
 
+
```
rpl, line 635: Nit, remove this unnecessary empty line (there was already one at line 634 to separate the two statements).

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +632,35 @@ describe('utils/adb', () => {
...
+  describe('checkOrCleanArtifacts', () => {
+    it('cleans the previously artifacts directory when condition true' +
+      ' but not removed earlier', async () => {
```
rpl, line 638: splitting the description into two strings and composing them with + looks a bit bad, I think we can just make the description shorter (e.g. "it(`removes old artifacts directories`, ...)")

also, we need to test the helper function on at least two scenarios: when it does only check if any old artifacts exists and when it does also remove them.

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +632,35 @@ describe('utils/adb', () => {
...
+      const adb = getFakeADBKit({
+        adbClient: {
+          readdir: sinon.spy(() => Promise.resolve([file])),
```
rpl, line 641: maybe let's use more than one file

tests/unit/test-util/test.adb.js
```diff
@@ -30,6 +30,7 @@ const fakeRDPUnixSocketFile = (
   '/data/data/org.mozilla.firefox/firefox-debugger-socket'
 );
 
+const artifactDir = `web-ext-artifacts-${Date.now()}`;
```
rpl, line 33: Move this into the related test case, it doesn't seem that we need this to be global.

tests/unit/test-util/test.adb.js
```diff
@@ -49,6 +50,13 @@ android.permission.READ_EXTERNAL_STORAGE, granted=true
 android.permission.WRITE_EXTERNAL_STORAGE, granted=true
 `;
 
+const file = {
+  name: artifactDir,
+  isDirectory: () => {
+    return true;
+  },
+};
```
rpl, line 58: Same here, we could create the fake file objects into the related test case.
Especially because we should test the helper function on mode than just one file that is a directory (we should also be sure that the behavior will be the expected one when there are files that are not directories or directories that do not match the expected prefix in their name)

tests/unit/test-util/test.adb.js
```diff
@@ -59,11 +67,13 @@ function getFakeADBKit(
 
   adbUtilReadAllStub.onCall(0).returns(Promise.resolve(Buffer.from('')));
 
+
   const fakeADBClient = {
     listDevices: sinon.spy(() => {
       return [];
     }),
     shell: sinon.spy(() => Promise.resolve('')),
+    readdir: sinon.spy(() => Promise.resolve([file])),
```
rpl, line 76: I think that this spy could return an empty array by default (because for most of the test cases we don't care about this and an empty dir represents the scenario where no old artifacts dir have been found, which is a completely valid base scenario)

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +632,35 @@ describe('utils/adb', () => {
...
+        adbClient: {
+          readdir: sinon.spy(() => Promise.resolve([file])),
+          shell: sinon.spy(() => Promise.resolve('')),
+        },
+        adbkitUtil: {
+          readAll: sinon.spy(() => Promise.resolve(Buffer.from('1\n'))),
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+      const temp = '/sdcard/';
```
rpl, line 649: temp doesn't seem to be used anywhere else bug in the next line, doesn't seem necessary to give it a name.

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +632,35 @@ describe('utils/adb', () => {
...
+      const artifactsDir = temp + artifactDir;
```
rpl, line 650: calling this `artifactsDir` doesn't explain what is different from `artifactDir` and this one, `artifactDirFullPath` would have been a better name (as it immediately suggests that the difference is that this one is a full absolute path).

src/util/adb.js
Vishalghyv: In the meta code that you provided,
for (let file of filenames) {
  if (file is not a directory or
      its name doesn't start with the artifacts name prefix) {
     continue; // no need to do anything else on this item, move to the next
  }

  hasOldArtifacts = true;

  if (this should delete the old artifacts) {
     remove the artifact dir
  }
}

Have the same level of nested if/else blocks and in this as you pointed that if we have to only check the condition for old artifacts then if one is found we should break out of the loop, which wasn't possible in the meta code

tests/unit/test-util/test.adb.js
Vishalghyv: Can't directly concatenate the strings thus this was required, and can't ise concat function as it will add sdcard in end of artifactDir

src/extension-runners/firefox-android.js
Vishalghyv: I am not sure but this is a template string and it is replacing the ${...} part

Vishalghyv:
Maybe four files are not necessary but this allowed to make sure that the test would break if the function did not worked properly especially when it's the case of deleting a folder from the users phone

src/util/adb.js
```diff
@@ -210,10 +210,46 @@ export default class ADBUtils {
...
     this.artifactsDirMap.set(deviceId, artifactsDir);
-
     return artifactsDir;
   }
 
+  async checkOrCleanArtifacts(
+    deviceId: string, remove?: boolean): Promise<boolean> {
+    const {adbClient} = this;
+
+    let directories = false;
```
rpl, line 220: This `directories` is now a boolean, but its name kind of suggests that it is an array (and so the name is misleading).
Also, there is no need to define a state variable outside of the wrapped function, if you look at the definition of wrapADBCall you'll see that it does return a promise that resolves to the value returned by the wrapped function (see https://github.com/mozilla/web-ext/blob/e60411282655cbb23426544b8b5658f435505394/src/util/adb.js#L30) and so we can instead:
- change `await wrapADBCall` to `return wrapADBCall(...)`
- rename `directories` to `found` (as it is easier to recognize it as a boolean) and move it into the wrapped function
- then `return true` in `if (!remove) (because we don't need to look further in that case and we can just return earlier
- then set found to true after the `if (!remove) { return true; }` and return the value set to `found` at the end on the wrapped function

src/util/adb.js
```diff
@@ -210,10 +210,46 @@ export default class ADBUtils {
...
+          continue;
+        }
+        directories = true;
+
+        if (!remove) {
+          break;
+        }
+
+        const temp = '/sdcard/';
+        const artifactsDir = temp + file.name;
```
rpl, line 238: As I mentioned in a previous review comment, we should use template strings instead of composing strings with the `+` operator (to be even more clear about it this time: something like `temp + file.name` should become `${temp}${file.name}` enclosed inside two ` characters).

Besides that lets define a `const DEVICE_DIR_BASE = '/sdcard/'` and a `const ARTIFACTS_DIR_PREFIX = 'web-ext-artifacts'` at the start of this file (e.g. right before the `const log = ...`)
and then use those const in both `getOrCreateArtifactsDir` and `checkOrCleanArtifacts` to define the string paths we need.

tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -128,6 +128,7 @@ function prepareSelectedDeviceAndAPKParams(
     startFirefoxAPK: sinon.spy(() => Promise.resolve()),
     setupForward: sinon.spy(() => Promise.resolve()),
     clearArtifactsDir: sinon.spy(() => Promise.resolve()),
+    checkOrCleanArtifacts: sinon.spy(() => Promise.resolve()),
```
rpl, line 131: In this test file we should actually add a test case to assert that this spy is being called as expected (when we expect it to be called, and with the parameters we expect) to actually cover the changes applied to the firefox-android test runner (and also make sure to properly cover the possible scenarios, e.g. when it should have been called to only check for the old artifacts and when it is supposed to be called to remove those old artifacts).

tests/unit/test-util/test.adb.js
rpl: not sure what you mean with tou can't use concat function? there is no real need to use different ways to concatenate strings (the String concat method and the + operator).

you can just compose all of them together using the template string syntax I mentioned in another review comment:
```
const part1 = "part1";
const part2 = "part2";
const part3 = "part3";
const aComposedString = `${part1}${part2}${part3}`;
```

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +623,75 @@ describe('utils/adb', () => {
...
+        isDirectory: () => {
+          return isDirectory;
+        },
+      };
+    }
+    const files = [
+      makeFakeArtifact(artifactDir.concat('fake1'), true),
+      makeFakeArtifact(artifactDir.concat('fake2'), true),
+      makeFakeArtifact(artifactDir.concat('fake2'), false),
+      makeFakeArtifact('not-web-ext-artifacts', true),
```
rpl, line 640: nit, lets rename these fake files as `fake-dir1`, `fake-dir2`, `fake-file3`, and `not-web-ext-artifacts`
(fake2 is currently repeated twice, also fake1 and fake2 doesn't tell us much, making clear the role of that fake file by including `dir` and `file` in their name would make it almost immediately clear).

tests/unit/test-util/test.adb.js
```diff
@@ -622,6 +623,75 @@ describe('utils/adb', () => {
...
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+
+      const promise = adbUtils.checkOrCleanArtifacts('device1', false);
+      const result = await assert.isFulfilled(promise);
+      assert.equal(result, true);
+
+      sinon.assert.calledOnce(adb.fakeADBClient.readdir);
+      sinon.assert.notCalled(adb.fakeADBClient.shell);
```
rpl, line 660: Nit, let's add an inline comment right above this line to explain what notCalled is meant to assert (just something like `// No file should be removed and so shell shouldn't be called in this test.`)

Vishalghyv:
It is not done,
I am having trouble to call checkOrCleanArtifacts with different parameters in one test, and the test written is not really useful except for checking of order, @rpl could you suggest a way to approach it. Thanks

Vishalghyv:
Should I use two tests for different params, two running instance or is there another way?
And except for checking expected params call and order of the order what else can I check

Vishalghyv:
If the test might look like the current version, then this test can be added to the one above it i.e in 'stops any running instances of the selected Firefox apk'

And should adbCleanArtifacts directly retrieved from params

rpl:
@Vishalghyv this pull request still container the unrelated changes to package.json and package-lock.json, removing the file in a new commit isn't the right way to remove those changes.

I did cleaned up and rebased your branch locally and pushed here in my github fork here (in the branch named pr-1866-Vishalghyv-REBASE-DONE):
- https://github.com/rpl/web-ext/commit/38abec61d2a3db23a7dbc0edc263c540d653b149

you can get the commit from there and overwrite this PR with it (it has to be in a branch with the same name of your and and then pushed into your github fork using --force to overwrite the branch, otherwise git will refuse to push the commit on the existing branch because it is not a new commit on top of the existing one)

If you are curious about how I achieved that, here is what I did locally:

- started an interactive git rebase of the entire branch on the parent commit of your first commit:
  ```git rebase -i 2f9456b^1```
  where `2f9456b` is the sha1 of your first commit, and ```^1``` means "the parent commit of the one specified on the left"

- git opens my editor with a list of the commit to rebase (the first column is the operation to run on the commit as part of the rebase, the second column is the sha1 of the commit), I marked the first commit as pick and all the others as fixup (to squash all the commits into the first one and discards their commit message):

```
pick 2f9456b fix: Cleanup old web-ext run artifacts dir when we are connecting to an android device
fixup <SHA1> next commit message
fixup <SHA1> other commit message
...
fixup <SHA1> last commit message
```

- at this point there is only a single commit with all the changes in the pull request, but the changes to package.json and package-lock.json are still there

- to remove those unrelated changes I asked git to restore the content of those file as "they where on the commit right above the current one" (because that is exactly your original branching point):
```
git checkout HEAD^1 -- package.json
git checkout HEAD^1 -- package-lock.json
```

- at this point there are changes to the two files (with their original content) pending to be committed, by amending the current commit we can get rid of the unrelated changes:
```
git commit --amend
```

- now the last step is to rebase the entire branch to the current master
```
git remote add upstream https://github.com/mozilla/web-ext
git fetch upstream
git rebase upstream/master
```

- at this point the branch should be right on top of the most recent commit in the master branch for this repository, which can also be verified by using git log --graph:
```
* commit 99797560a54e62e4166b0a83bf69501d971ee5d6 (HEAD -> Vishalghyv-new)
| Author: Vishhalghyv <vishal38785@gmail.com>
| Date:   Sat Mar 21 19:56:50 2020 +0530
| 
|     fix: Cleanup old web-ext run artifacts dir when we are connecting to an android device
| 
|  src/cmd/run.js                                            |  3 +++
|  src/extension-runners/firefox-android.js                  | 28 ++++++++++++++++++++++++++++
|  src/program.js                                            |  5 +++++
|  src/util/adb.js                                           | 42 +++++++++++++++++++++++++++++++++++++++++-
|  tests/unit/test-extension-runners/test.firefox-android.js | 76 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
|  tests/unit/test-util/test.adb.js                          | 70 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
|  6 files changed, 223 insertions(+), 1 deletion(-)
| 
* commit 7d315770b7064eaa445e1e171cdcedf9e4656c67 (upstream/master, upstream/HEAD, master)
| Author: Luca Greco <lgreco@mozilla.com>
| Date:   Wed May 6 00:25:50 2020 +0200
| 
|     chore(audit): Temporarily add yargs-parser adv url to nsprc
| 
|  .nsprc | 2 ++
|  1 file changed, 2 insertions(+)
| 
...
```

Mission accomplished! yay \o/

Using the revision system can be tricky some times, but the more you practice it (and then learn about it as a side effect), the easier and faster it gets to work with it.

I hope that this will help you to get the cleanup on this branch done.

Let me know if you have still issues and you are unable to get this branch cleaned up as needed.

Vishalghyv:
Mission accomplished! yay \o/

Vishalghyv:
Thanks Sir for the help :)

Vishalghyv:
Tried the original version by you, Comment was really good, I saved it  as a code snippet for squashing commits and rebase.

rpl:
Closing as superseded by #1965 (which included a rebased version of the changes in this PR, with the author metadata preserved, plus some small tweaks and cleanups as described in #1965)

Issue #1870, opened 2020-03-27
Add support for Reference Browser in ADB discovery


jonalmeida:
### Is this a feature request or a bug?
Feature

### What is the current behavior?

[Reference Browser](https://github.com/mozilla-mobile/reference-browser/) is what many teams use to prototype mobile browser features. Adding support for webext debugging would make it easier.

### What is the expected or desired behavior?
To have 

### Version information (for bug reports)

n/a

rpl:
Fixed in #1871

PR #1871, opened 2020-03-27
feat: Add reference browser support to the firefox-android extension runner (#1870)

jonalmeida:
In addition to Fenix support (#1834), we'd like to add Reference Browser as well. 🙂

### DIFF
--- a/tests/unit/test-extension-runners/test.firefox-android.js
+++ b/tests/unit/test-extension-runners/test.firefox-android.js
@@ -114,6 +114,7 @@ function prepareSelectedDeviceAndAPKParams(
     discoverInstalledFirefoxAPKs: sinon.spy(() => Promise.resolve([
       'org.mozilla.fennec', 'org.mozilla.firefox',
       'org.mozilla.fenix', 'org.mozilla.fenix.nightly',
+      'org.mozilla.reference.browser',
     ])),
     getAndroidVersionNumber: sinon.spy(() => Promise.resolve(20)),
     amForceStopAPK: sinon.spy(() => Promise.resolve()),
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -120,7 +120,8 @@ export default class ADBUtils {
           line.startsWith('org.mozilla.fennec') ||
           line.startsWith('org.mozilla.fenix') ||
           line.startsWith('org.mozilla.geckoview') ||
-          line.startsWith('org.mozilla.firefox')
+          line.startsWith('org.mozilla.firefox') ||
+          line.startsWith('org.mozilla.reference.browser')
         );
       });
   }


tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -114,6 +114,7 @@ function prepareSelectedDeviceAndAPKParams(
     discoverInstalledFirefoxAPKs: sinon.spy(() => Promise.resolve([
       'org.mozilla.fennec', 'org.mozilla.firefox',
       'org.mozilla.fenix', 'org.mozilla.fenix.nightly',
+      'org.mozilla.reference.browser'
```
Rob--W: Please add a trailing comma.

jonalmeida:
> Did you verify that you're able to run extensions in the Reference browser after this change?

Yes, I did. It requires a minor change in r-b to allow web-ext to launch the browser which I'll be landing soon. 🙂

Rob--W:
I see that you've opened a new pull request at the Reference Browser's repository as well.

Did you know that even without any changes to `web-ext` and the reference browser, that you are already able to run `web-ext` with the Reference browser? For example:

`web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.reference.browser --firefox-apk-component=BrowserActivity`

To avoid the need for https://github.com/mozilla-mobile/reference-browser/pull/1096 , you could add
```javascript
if (!apkComponent && apk === 'org.mozilla.reference.browser') {
   apkComponent = 'BrowserActivity';
}
```

before https://github.com/mozilla/web-ext/blob/ffe57268cd946104b0213f001c712ad2cb2ef78e/src/util/adb.js#L273-L274

jonalmeida:
It seemed more in line with our other browser apps to follow the `activity-alias` for .App since we have to add it for other production browsers as well. Keeping R-B in sync with those makes testing reproducible.

jonalmeida:
> Did you know that even without any changes to web-ext and the reference browser, that you are already able to run web-ext with the Reference browser?


Did this work for you? In my testing it wasn't able to launch the activity. So I opted for adding the alias as mentioned above.

Rob--W:
I did the following:

1. Create minimal extension that opens a new tab:

manifest.json
```json
{
    "name": "Test ext",
    "version": "1",
    "manifest_version": 2,
    "background": {
        "scripts": ["background.js"]
    }
}
```

background.js
```
browser.tabs.create({url:"https://example.com"});
```
2. Launch `web-ext` with R-B using the command from https://github.com/mozilla/web-ext/pull/1871#issuecomment-605310726
3. Verified that R-B launched, with example.com

(I installed R-B from the Play store just today)

jonalmeida:
Interesting, this command seems to work for release, but not debug builds. I get this error instead (without the activity-alias):

```
Error: Activity class {org.mozilla.reference.browser.debug/org.mozilla.reference.browser.debug.BrowserActivity} does not exist.
    at /Users/jalmeida/src/web-ext/node_modules/adbkit/lib/adb/command/host-transport/startactivity.js:57:21
    at tryCatcher (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/promise.js:577:18)
    at Async._drainQueue (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:15:14)
```

That's probably because it's sending the intent with the variant (debug) in the package name:

What it's sending: `org.mozilla.reference.browser.debug/org.mozilla.reference.browser.debug.BrowserActivity`
What it should be sending: `org.mozilla.reference.browser.debug/org.mozilla.reference.browser.BrowserActivity`

I don't believe that's a bug with web-ext though as we've run into this in Fenix where the IDE would make the same error (https://github.com/mozilla-mobile/fenix/pull/9462#issuecomment-605083758).

jonalmeida:
Ping @Rob--W . What do you think about this? The reason it isn't a bug on Fenix variants is because we use the [same activity-alias](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/AndroidManifest.xml#L39-L41) there too.

Rob--W:
The value of `component` ends up being used as the value of `-n` in `am start ... -n org.mozilla.reference.browser.debug/.BrowserActivity`.

I think that the best course of action is to pass a fully-qualified name instead, without the `.debug` suffix.

Rob--W:
@jonalmeida Could you update the PR to implement the above idea, and test whether it also works with other mobile products such as Fenix (and maybe even geckoview_example)?

jonalmeida:
> @jonalmeida Could you update the PR to implement the above idea, and test whether it also works with other mobile products such as Fenix (and maybe even geckoview_example)?

This seems outside the scope of this issue - I can file a follow-up for it if you'd like. In R-B, we have everything we need to land this patch as is.

Rob--W:
> This seems outside the scope of this issue

That's OK, let's not block the merge on my next question.

> I can file a follow-up for it if you'd like. In R-B, we have everything we need to land this patch as is.

Has the `.debug` issue also been solved (https://github.com/mozilla/web-ext/pull/1871#issuecomment-605400070)? If not, please file a follow-up.

jonalmeida:
> Has the `.debug` issue also been solved ([#1871 (comment)](https://github.com/mozilla/web-ext/pull/1871#issuecomment-605400070))? If not, please file a follow-up.

I've filed https://github.com/mozilla/web-ext/issues/1891

Issue #1365, opened 2018-09-02
ignore-files no longer works with "!(...)" glob pattern


Lusito:
### Is this a feature request or a bug?

bug

### What is the current behavior?
`web-ext build -i "!(manifest).json"`
This includes all json files.

### What is the expected or desired behavior?
Should only include all json files except manifest.json.
This worked in an older version.. not sure when this changed.

### Version information (for bug reports)

* **Firefox version**: 61
* **Your OS and version**: Windows 7
* **web-ext** 2.8.0
* **Node** 9.8.0

Rob--W:
This behavior changed in #1348 .

Does this address your needs? `-i "*.json" -i "!manifest.json"`

Lusito:
Yes, it does, however I think that at least this description has to be updated:

> A list of glob patterns to define which files should be ignored. (Example: --ignore-files=path/to/first.js path/to/second.js "**/*.log")

Since `!(...)` is a valid glob pattern and the above statement suggests that it should work.

rpl:
@Rob--W I'm going to close PR #1874 because I don't think that the change in that PR is really going to be enough to make it clearer why a glob pattern like `!(manifest).json` doesn't exclude all json files besides manifest.json.

The reason for the change in behavior is that before #1348 `!(manifest).json` was resolved as `/path/to/srcdir/!(manifest).json`, after #1348 that pattern is turned into `!/path/to/srcdir/(manifest).json)`.

I'm not sure if the command line options description is a good enough place to explain this properly, it may be more reasonable to try document this option in a bit more detail in the web-ext docs part of https://github.com/mozilla/extension-workshop.

PR #1874, opened 2020-04-01
docs: Added description for Issue 1365

Vishalghyv:


rpl:
I'm closing this PR because (as described in a bit more detail in https://github.com/mozilla/web-ext/issues/1365#issuecomment-663707246) the change itself doesn't look enough to make the issue and the expected behavior that much more clear, and I'm not even sure that the cli option description is going to be a good place to describe this properly.

PR #1886, opened 2020-04-20
fix: use getRoot request to support Firefox 77+

### LINKED ISSUE #1884: `web-ext run` cannot install a temporary add-on with FF Nightly (build >= 2020-04-14)
`web-ext run` cannot install a temporary add-on when we use it with FF Nightly (since yesterday).

The error message is:

```
$ ./bin/web-ext run --firefox "/Users/williamdurand/projects/mozilla/gecko/objdir-frontend-debug-artifact/dist/NightlyDebug.app/Contents/MacOS/firefox" -s ~/Downloads/web-extension-permission-example-master
Running web extension from /Users/williamdurand/Downloads/web-extension-permission-example-master

WebExtError: Temporary add-on installation is not supported in this version of Firefox (you need Firefox 49 or higher). For older Firefox versions, use --pre-install
    at FirefoxDesktopExtensionRunner.startFirefoxInstance (/Users/williamdurand/projects/mozilla/web-ext/dist/webpack:/src/extension-runners/firefox-desktop.js:278:19)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

```

I used `mozregression` to find the root cause of this error because it comes from some RDP/FF-related code (thanks for the suggestion @rpl). The error is caused by this patch: https://phabricator.services.mozilla.com/D70174

The `listTabs` method does not return global actors anymore and we cannot retrieve `addonsActor` (if I understand correctly). It looks like we should be using `getRoot` instead and I confirm that this sketch of patch fixes the issue locally:

```diff
diff --git a/src/firefox/remote.js b/src/firefox/remote.js
index 2a16e09..a67a5e0 100644
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -90,7 +90,7 @@ export class RemoteFirefox {
     addonPath: string
   ): Promise<FirefoxRDPResponseAddon> {
     return new Promise((resolve, reject) => {
-      this.client.request('listTabs', (error, tabsResponse) => {
+      this.client.request('getRoot', (error, tabsResponse) => {
         if (error) {
           return reject(new WebExtError(
             `Remote Firefox: listTabs() error: ${error}`));
```

That being said, we need to guarantee backward compatibility so we have to make sure the `getRoot` method is (1) available and (2) return the `addonsActor` in all FF versions we support. In addition, we should have a more future-proof and robust solution so that it won't break again.

* **Firefox version**: `central/default`
* **Your OS and version**: MacOS 10.13.6

````
v13.4.0
6.13.4
4.1.0
````

### PR DESCRIPTION
**Pull request by Rob--W**
Fixes #1884.
 
- Firefox 77+ requires `getRoot` because `listTabs` was removed in https://bugzilla.mozilla.org/show_bug.cgi?id=1618691
- `getRoot` is supported since Firefox 55, with https://bugzilla.mozilla.org/show_bug.cgi?id=1352157
- I still added a `getList` fallback to not unnecessarily breaking backwards compatibility, so this is a safe patch for a minor release.

### DIFF
--- a/src/firefox/remote.js
+++ b/src/firefox/remote.js
@@ -86,27 +86,50 @@ export class RemoteFirefox {
     });
   }
 
+  getAddonsActor(): Promise<string> {
+    return new Promise((resolve, reject) => {
+      // getRoot should work since Firefox 55 (bug 1352157).
+      this.client.request('getRoot', (err, actors) => {
+        if (!err) {
+          if (actors.addonsActor) {
+            resolve(actors.addonsActor);
+          } else {
+            reject(new RemoteTempInstallNotSupported(
+              'This version of Firefox does not provide an add-ons actor for ' +
+              'remote installation.'));
+          }
+          return;
+        }
+        log.debug(`Falling back to listTabs because getRoot failed: ${err}`);
+        // Fallback to listTabs otherwise, Firefox 49 - 77 (bug 1618691).
+        this.client.request('listTabs', (error, tabsResponse) => {
+          if (error) {
+            return reject(new WebExtError(
+              `Remote Firefox: listTabs() error: ${error}`));
+          }
+          // addonsActor was added to listTabs in Firefox 49 (bug 1273183).
+          if (!tabsResponse.addonsActor) {
+            log.debug(
+              'listTabs returned a falsey addonsActor: ' +
+              `${tabsResponse.addonsActor}`);
+            return reject(new RemoteTempInstallNotSupported(
+              'This is an older version of Firefox that does not provide an ' +
+              'add-ons actor for remote installation. Try Firefox 49 or ' +
+              'higher.'));
+          }
+          resolve(tabsResponse.addonsActor);
+        });
+      });
+    });
+  }
+
   installTemporaryAddon(
     addonPath: string
   ): Promise<FirefoxRDPResponseAddon> {
     return new Promise((resolve, reject) => {
-      this.client.request('listTabs', (error, tabsResponse) => {
-        if (error) {
-          return reject(new WebExtError(
-            `Remote Firefox: listTabs() error: ${error}`));
-        }
-        if (!tabsResponse.addonsActor) {
-          log.debug(
-            'listTabs returned a falsey addonsActor: ' +
-            `${tabsResponse.addonsActor}`);
-          return reject(new RemoteTempInstallNotSupported(
-            'This is an older version of Firefox that does not provide an ' +
-            'add-ons actor for remote installation. Try Firefox 49 or ' +
-            'higher.'));
-        }
-
+      this.getAddonsActor().then((addonsActor) => {
         this.client.client.makeRequest({
-          to: tabsResponse.addonsActor,
+          to: addonsActor,
           type: 'installTemporaryAddon',
           addonPath,
         }, (installResponse) => {
@@ -120,7 +143,7 @@ export class RemoteFirefox {
           log.info(`Installed ${addonPath} as a temporary add-on`);
           resolve(installResponse);
         });
-      });
+      }).catch(reject);
     });
   }
 
--- a/tests/functional/fake-firefox-binary.js
+++ b/tests/functional/fake-firefox-binary.js
@@ -3,8 +3,8 @@
 const net = require('net');
 
 const REPLY_INITIAL = {from: 'root'};
-const REQUEST_LIST_TABS = {to: 'root', type: 'listTabs'};
-const REPLY_LIST_TABS = {from: 'root', addonsActor: 'fakeAddonsActor'};
+const REQUEST_ACTORS = {to: 'root', type: 'getRoot'};
+const REPLY_ACTORS = {from: 'root', addonsActor: 'fakeAddonsActor'};
 const REQUEST_INSTALL_ADDON = {
   to: 'fakeAddonsActor',
   type: 'installTemporaryAddon',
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by willdurand** [comment]
ha well, I was also working on that issue.. I guess I can move on something else now :)

File: src/firefox/remote.js [hunk 1]
```diff
@@ -86,27 +86,43 @@ export class RemoteFirefox {
...
 
+  getAddonsActor(): Promise<string> {
+    return new Promise((resolve, reject) => {
+      // getRoot should work since Firefox 55 (bug 1352157).
+      this.client.request('getRoot', (ignoredError, actors) => {
+        if (actors && actors.addonsActor) {
+          resolve(actors.addonsActor);
+          return;
+        }
+        log.debug('getRoot failed, falling back to listTabs');
```
rpl, line 97: Nit: would you mind to include the `ignoredError` in this debug message?
just to make investigate issues easier in case the `getRoot` request failed for unrelated reasons to Firefox version being older than 55.

src/firefox/remote.js
```diff
@@ -86,27 +86,43 @@ export class RemoteFirefox {
...
-          log.debug(
-            'listTabs returned a falsey addonsActor: ' +
-            `${tabsResponse.addonsActor}`);
-          return reject(new RemoteTempInstallNotSupported(
-            'This is an older version of Firefox that does not provide an ' +
-            'add-ons actor for remote installation. Try Firefox 49 or ' +
-            'higher.'));
-        }
-
+      this.getAddonsActor().then((addonsActor) => {
```
rpl, line 123: Nit,  I see that we do reject explicitly if the `'installTemporaryAddon'` request rejects, but I'm wondering if the error we catch and reject there makes it clear if the request did fails because `addonsActor` was not defined.
Maybe `getAddonsActor` could still check that `addonsActor` it is going to resolve is actually defined and reject with a more clear error message if it is not (and so without the misleading error `This is an older version of Firefox...` in this case).

rpl:
> ha well, I was also working on that issue.. I guess I can move on something else now :)

ouch, I'm so sorry @willdurand. I'm the one to blame, there was an assigned comment in our shared triaging meeting document but (to make sure that it couldn't be missed) I should have explicitly assigned you the issue and added a comment to mention that you were going to prepare a pull request for it by today or tomorrow.

src/firefox/remote.js
Rob--W: > Nit, I see that we do reject explicitly if the `'installTemporaryAddon'` request rejects, but I'm wondering if the error we catch and reject there makes it clear if the request did fails because `addonsActor` was not defined.

Do you mean that it is not obvious from `installTemporaryAddon` that the error propagated through `.catch(reject)` is from `getAddonsActor`? Would this readability issue be resolved by using `.then(...., reject)` instead of `.then(....).catch(...)`?

> Maybe `getAddonsActor` could still check that `addonsActor` it is going to resolve is actually defined and reject with a more clear error message if it is not (and so without the misleading error `This is an older version of Firefox...` in this case).

`getAddonsActor` already checks whether `addonsActor` is defined before resolving. Could you clarify what you're requesting here?

rpl: I see what you mean, but this comment wasn't about readability but about the fallback logic.

The current version does check if `getRoot` resolved but `addonsActor` is undefined, but at the moment it does fallback to `listTabs` in that case:

```
      this.client.request('getRoot', (ignoredError, actors) => {
        if (actors && actors.addonsActor) {
          resolve(actors.addonsActor);
          return;
        }
        ...
        // Fallback to listTabs otherwise, Firefox 49 - 77 (bug 1618691)
        this.client.request('listTabs', (error, tabsResponse) => {
          ...
```

if `getRoot` used to already include `addonsActor` when we introduced it for the first time in Firefox 55, then I think that we may instead reject if `ignoredError` and `actors.addonsActor` are both undefined, and fallback only if `getRoot` did explicitly fail.

How that sounds?

src/firefox/remote.js
```diff
@@ -86,27 +86,50 @@ export class RemoteFirefox {
...
+          } else {
+            reject(new RemoteTempInstallNotSupported(
+              'This version of Firefox does not provide an add-ons actor for ' +
+              'remote installation.'));
+          }
+          return;
+        }
+        log.debug(`Falling back to listTabs because getRoot failed: ${err}`);
+        // Fallback to listTabs otherwise, Firefox 49 - 77 (bug 1618691).
+        this.client.request('listTabs', (error, tabsResponse) => {
```
rpl, line 105: Nit, how about renaming `err` to `getRootErr` and `error` to `listTabsErr`?

src/firefox/remote.js
```diff
@@ -86,27 +86,50 @@ export class RemoteFirefox {
...
+    return new Promise((resolve, reject) => {
+      // getRoot should work since Firefox 55 (bug 1352157).
+      this.client.request('getRoot', (err, actors) => {
+        if (!err) {
+          if (actors.addonsActor) {
+            resolve(actors.addonsActor);
+          } else {
+            reject(new RemoteTempInstallNotSupported(
+              'This version of Firefox does not provide an add-ons actor for ' +
+              'remote installation.'));
```
rpl, line 99: If this is an unexpected condition as I think it is, this error message sounds a bit misleading for that purpose.

Should we at least suggest to the users that are experiencing this error to report an issue in the web-ext repo?

(and maybe use just `WebExtError` or a new subclass, instead of `RemoteTempInstallNotSupported`).

0c0w3:
Can this be merged soon? It's making it hard for me to develop an add-on against an m-c build with patches applied. (I'm on the urlbar team and I'm developing a urlbar WebExtensions API.)

rehandalal:
> Can this be merged soon?

+1 - it's interfering with our work on the experiments infrastructure team as well.

rpl:
> Can this be merged soon? It's making it hard for me to develop an add-on against an m-c build with patches applied. (I'm on the urlbar team and I'm developing a urlbar WebExtensions API.)

@0c0w3 @rehandalal sure, I plan to merge this today and we should also be able to release it on npm as well.

0c0w3:
Thanks! (Thanks for web-ext too, it's nice.)

rpl:
@0c0w3 @rehandalal this has been just released on npm as v4.2.0 (See [release notes](https://github.com/mozilla/web-ext/releases/tag/4.2.0) to know which other changes are included in the new version along with this fix).

Let us know if any other issue in the newly released version is still blocking you.

rehandalal:
@rpl thanks for the release! it's working great for me on Ubuntu but my teammate is having an unrelated issue with this release and Nightly on MacOS that is a complete blocker.

PR #1900, opened 2020-05-05
add --filename option to specify a filename template

### LINKED ISSUE #1335: Add option to web-ext build: --filename="%{applications.gecko.id}-%{version}.zip

escapewindow:
Fixes #1335. Based off #1741 and its review comments.

Questions:

- Does `-n` make sense as a single letter filename option? (`n` for name) I'm happy to add that or keep it as-is.
- Is there a reason why we [can't specify a directory path](https://github.com/mozilla/web-ext/pull/1741/files#r348005562)? I could see dying if the path specified doesn't exist, but I'm not sure why we can't write to a different directory.
- The code, as written, changes the `@` in the name to an underscore `_` (`minimal-example_web-ext-test-suite`. Is that expected? Should we allow for people to name the file what they want, if they pass in a flag (or warn that their name doesn't follow certain conventions)?

JavaScript isn't my strong suit; please review accordingly :)

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -24,6 +24,7 @@ import type {ExtensionManifest} from '../util/manifest';
 import type {FileFilterCreatorFn} from '../util/file-filter';
 
 const log = createLogger(__filename);
+const DEFAULT_FILENAME_TEMPLATE = '{name}-{version}.zip';
 
 
 export function safeFileName(name: string): string {
@@ -43,7 +44,8 @@ export type PackageCreatorParams = {|
   fileFilter: FileFilter,
   artifactsDir: string,
   overwriteDest: boolean,
-  showReadyMessage: boolean
+  showReadyMessage: boolean,
+  filename?: string,
 |};
 
 export type LocalizedNameParams = {|
@@ -104,6 +106,55 @@ export async function getDefaultLocalizedName(
   return Promise.resolve(extensionName);
 }
 
+// https://stackoverflow.com/a/22129960
+export function getStringPropertyValue(
+  prop: string,
+  obj: Object,
+): string {
+  const properties = prop.split('.');
+  const value = properties.reduce((prev, curr) => prev && prev[curr], obj);
+  if (!['string', 'number'].includes(typeof value)) {
+    throw new UsageError(
+      `Manifest key "${prop}" is missing or has an invalid type: ${value}`
+    );
+  }
+  const stringValue = `${value}`;
+  if (!stringValue.length) {
+    throw new UsageError(`Manifest key "${prop}" value is an empty string`);
+  }
+  return stringValue;
+}
+
+function getPackageNameFromTemplate(
+  filenameTemplate: string,
+  manifestData: ExtensionManifest
+): string {
+  const packageName = filenameTemplate.replace(
+    /{([A-Za-z0-9._]+?)}/g,
+    (match, manifestProperty) => {
+      return getStringPropertyValue(manifestProperty, manifestData);
+    }
+  );
+
+  // Validate the resulting packageName string, after interpolating the manifest property
+  // specified in the template string.
+  const parsed = path.parse(packageName);
+  if (parsed.dir) {
+    throw new UsageError(
+      `Invalid filename template "${filenameTemplate}". ` +
+      `Filename "${packageName}" should not contain a path`
+    );
+  }
+  if (!['.zip', '.xpi'].includes(parsed.ext)) {
+    throw new UsageError(
+      `Invalid filename template "${filenameTemplate}". ` +
+      `Filename "${packageName}" should have a zip or xpi extension`
+    );
+  }
+
+  return packageName;
+}
+
 export type PackageCreatorFn =
     (params: PackageCreatorParams) => Promise<ExtensionBuildResult>;
 
@@ -115,6 +166,7 @@ export async function defaultPackageCreator(
     artifactsDir,
     overwriteDest,
     showReadyMessage,
+    filename = DEFAULT_FILENAME_TEMPLATE,
   }: PackageCreatorParams,
   {
     eventToPromise = defaultEventToPromise,
@@ -132,7 +184,7 @@ export async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
-  let extensionName: string = manifestData.name;
+  let filenameTemplate = filename;
 
   let {default_locale} = manifestData;
   if (default_locale) {
@@ -142,12 +194,16 @@ export async function defaultPackageCreator(
       default_locale, 'messages.json'
     );
     log.debug('Manifest declared default_locale, localizing extension name');
-    extensionName = await getDefaultLocalizedName({
+    const extensionName = await getDefaultLocalizedName({
       messageFile, manifestData,
     });
+    // allow for a localized `{name}`, without mutating `manifestData`
+    filenameTemplate = filenameTemplate.replace(/{name}/g, extensionName);
   }
+
   const packageName = safeFileName(
-    `${extensionName}-${manifestData.version}.zip`);
+    getPackageNameFromTemplate(filenameTemplate, manifestData)
+  );
   const extensionPath = path.join(artifactsDir, packageName);
 
   // Added 'wx' flags to avoid overwriting of existing package.
@@ -187,6 +243,7 @@ export type BuildCmdParams = {|
   asNeeded?: boolean,
   overwriteDest?: boolean,
   ignoreFiles?: Array<string>,
+  filename?: string,
 |};
 
 export type BuildCmdOptions = {|
@@ -206,6 +263,7 @@ export default async function build(
     asNeeded = false,
     overwriteDest = false,
     ignoreFiles = [],
+    filename = DEFAULT_FILENAME_TEMPLATE,
   }: BuildCmdParams,
   {
     manifestData,
@@ -231,6 +289,7 @@ export default async function build(
     artifactsDir,
     overwriteDest,
     showReadyMessage,
+    filename,
   });
 
   await prepareArtifactsDir(artifactsDir);
--- a/src/program.js
+++ b/src/program.js
@@ -342,6 +342,15 @@ type MainParams = {
   runOptions?: Object,
 }
 
+export function throwUsageErrorIfArray(errorMessage: string): any {
+  return (value: any): any => {
+    if (Array.isArray(value)) {
+      throw new UsageError(errorMessage);
+    }
+    return value;
+  };
+}
+
 export function main(
   absolutePackageDir: string,
   {
@@ -428,6 +437,18 @@ Example: $0 --help run.
       demandOption: false,
       default: true,
       type: 'boolean',
+    },
+    'filename': {
+      alias: 'n',
+      describe: 'Name of the created extension package file.',
+      default: undefined,
+      normalize: false,
+      demandOption: false,
+      requiresArg: true,
+      type: 'string',
+      coerce: throwUsageErrorIfArray(
+        'Multiple --filename/-n option are not allowed'
+      ),
     },
   });
 
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -10,6 +10,7 @@ import defaultEventToPromise from 'event-to-promise';
 import build, {
   safeFileName,
   getDefaultLocalizedName,
+  getStringPropertyValue,
   defaultPackageCreator,
 } from '../../../src/cmd/build';
 import {FileFilter} from '../../../src/util/file-filter';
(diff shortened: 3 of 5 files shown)


escapewindow:
I'm not sure if the `audit-deps` error is actionable. I may be misreading, but it looks like the security advisories are for older versions of the dependencies?

Am I on the right track @rpl ? I'm happy to add more coverage, but we may first want to decide if this is an acceptable approach.

src/cmd/build.js
```diff
@@ -104,6 +105,46 @@ export async function getDefaultLocalizedName(
   return Promise.resolve(extensionName);
 }
 
+// https://stackoverflow.com/a/22129960
+function resolve(prop, obj, separator = '.'): string {
+  var properties = Array.isArray(prop) ? prop : prop.split(separator);
+  var value = properties.reduce((prev, curr) => prev && prev[curr], obj);
+  if (typeof value !== 'string') {
```
rpl, line 112: Nit, `manifest_version` is a number, maybe we can be a bit more permissive and allow "number" as well.

src/cmd/build.js
```diff
@@ -104,6 +105,46 @@ export async function getDefaultLocalizedName(
...
+    throw new UsageError(`Manifest is missing key: ${prop}`);
```
rpl, line 113: `Manifest key "${prop}" is missing or has an invalid type: ${value}`?

src/cmd/build.js
```diff
@@ -104,6 +105,46 @@ export async function getDefaultLocalizedName(
   return Promise.resolve(extensionName);
 }
 
+// https://stackoverflow.com/a/22129960
+function resolve(prop, obj, separator = '.'): string {
```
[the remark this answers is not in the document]
escapewindow: I chose `getStringPropertyValue`, but I'm not tied to that name.

src/program.js
```diff
@@ -429,6 +429,14 @@ Example: $0 --help run.
       default: true,
       type: 'boolean',
     },
+    'filename': {
+      describe: 'Name of the created extension package.',
```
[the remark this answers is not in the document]
escapewindow: I wasn't aware there was a directory option. I think that's sufficient.

src/program.js
```diff
@@ -429,6 +429,14 @@ Example: $0 --help run.
...
+      default: undefined,
+      normalize: true,
```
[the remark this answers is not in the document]
escapewindow: Hm.

I built `normandy-devtools`, then ran

`../web-ext/bin/web-ext build -s dist -n "aki-{name}.xpi" -a .`, where `../web-ext` contains a build of my current patch in this PR. That resulted in a `./aki-normandy-devtools.xpi`.

Does that answer this question?

src/cmd/build.js
```diff
@@ -132,7 +174,10 @@ export async function defaultPackageCreator(
     filter: (...args) => fileFilter.wantFile(...args),
   });
 
-  let extensionName: string = manifestData.name;
+  let filenameTemplate = '{name}-{version}.zip';
```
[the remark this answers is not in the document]
escapewindow: This ended up causing issues in the `defaultPackageCreator` test coverage.
Rather than adjust that test, I set `filename = DEFAULT_FILENAME_TEMPLATE` in `defaultPackageCreator` as well. Does that work?

src/program.js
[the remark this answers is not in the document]
escapewindow: Once I allowed for a `number` as well, I was able to `../web-ext/bin/web-ext build -s dist -n "aki-{name}-{manifest_version}.xpi" -a .`, which gave me `aki-normandy-devtools-2.xpi`.

escapewindow:
> @escapewindow Thanks a lot for taking care of reviving the pull request for this enhancement!
> 
> > I'm not sure if the `audit-deps` error is actionable. I may be misreading, but it looks like the security advisories are for older versions of the dependencies?
> 
> yeah, you are not misreading it. Don't worry about that, I'll take care of it in #1902, then you'll just need to rebase this once we merged that pull request and the audit-deps failure should go away.

Thanks! I rebased against master. Between that, and the commit lint, I had to rebase the entire PR. I'm hoping it's still easy to re-review?

> > Am I on the right track @rpl ?
> 
> This looks definitely on the right track, follows a round of review comments related to this version.

Thanks for your review! I think it's ready for re-review.

> > Questions:
> > 
> > * Does `-n` make sense as a single letter filename option? (`n` for name) I'm happy to add that or keep it as-is.
> 
> 👍 `n` isn't used for any other option, and so I think that I would be ok with it.

Added. I don't have a strong opinion here.

> > * Is there a reason why we [can't specify a directory path](https://github.com/mozilla/web-ext/pull/1741/files#r348005562)? I could see dying if the path specified doesn't exist, but I'm not sure why we can't write to a different directory.
> 
> well, we could probably allow it, but I have the feeling that we may then need to also check if any of the interpolated strings contains something that may be handled like a path to don't make that surprising (eg. "../../").
> 
> The developer can already customize the location of the artifacts dir with another one of the command line parameters, wouldn't that be enough for most of the use cases?
> 
> I would keep it simple in this PR and eventually re-evaluate that separately in a follow up if we agreed that there may be some reasonable use case that isn't already covered by customizing the artifacts dir location.

I think the `-a` option to specify the artifacts dir is sufficient. Thanks!

> > * The code, as written, changes the `@` in the name to an underscore `_` (`minimal-example_web-ext-test-suite`. Is that expected? Should we allow for people to name the file what they want, if they pass in a flag (or warn that their name doesn't follow certain conventions)?
> 
> That should be due to `safeFileName`, I can't currently recall if there was any reason to not include `@` into the safe characters, at a first glance it seems that `@` should be actually fine, let me think about it, we can eventually add it to the regexp used inside `safeFileName` (we can as well do it in a follow up)

I'm still a bit concerned that the name isn't 100% deterministic -- we'll get different filenames if `safeFileName` munges it, or if it's localized. But in my real life tests, I was able to specify a name and directory, and got the expected output, so I think this is an overall improvement.

Thanks for your help!

src/program.js
```diff
@@ -429,6 +429,15 @@ Example: $0 --help run.
       default: true,
       type: 'boolean',
     },
+    'filename': {
+      alias: 'n',
+      describe: 'Name of the created extension package file.',
+      default: undefined,
+      normalize: true,
+      demandOption: false,
+      requiresArg: false,
```
rpl, line 438: #### requiresArg: true

This should be `requiresArg: true`, otherwise the user would be allowed to call it without any argument which doesn't make sense (it defaults to the default template, but still not something we need to support).

#### Explicitly disallow multiple --filename option in the same command

Unfortunately yargs would still allow to call `-n` more than once and the parsed filename property would become an array.
In the current version of this patch, that is going to make `getPackageNameFromTemplate` to throw because `filenameTemplate` will be an Array:
```
TypeError: filenameTemplate.replace is not a function
    at getPackageNameFromTemplate (/home/rpl/mozlab/projects/webext-tools/web-ext/dist/webpack:/src/cmd/build.js:132:40)
    at defaultPackageCreator (/home/rpl/mozlab/projects/webext-tools/web-ext/dist/webpack:/src/cmd/build.js:205:5)
    at build (/home/rpl/mozlab/projects/webext-tools/web-ext/dist/webpack:/src/cmd/build.js:296:18)
```

to be fair there are other options that we should explicitly check to be sure that they are not called more than once, but we could take care of just this one in this pull request (and we will take care of the others in some follow ups).
Unfortunately it doesn't seem that yargs provide a configuration option specifically for this, but we could use the `coerce` option to do it in a way that can be then easily reused (in follow ups) for the other options.
How about something like this:
```js

function throwUsageErrorIfArray(errorMessage: string): any {
  return (value: any): any => {
    if (Array.isArray(value)) {
      throw new UsageError(errorMessage);
    }
    return value;
  };
}

...

    'filename': {
      alias: 'n',
      describe: 'Name of the created extension package file.',
      default: undefined,
      demandOption: false,
      requiresArg: true,
      type: 'string',
      coerce: throwUsageErrorIfArray(
        'Multiple --filename/-n option are not allowed'
      ),
    },
  });
....
```

The unfortunate part of this change is that I guess that it would make coveralls unhappy ;-)
To "make coveralls happy again", we could add an integration test for this as part of `tests/functional/test.cli.build.js`.

src/cmd/build.js
```diff
@@ -104,6 +106,51 @@ export async function getDefaultLocalizedName(
   return Promise.resolve(extensionName);
 }
 
+// https://stackoverflow.com/a/22129960
+function getStringPropertyValue(
```
rpl, line 110: Nit, we could export this function and unit test a couple of the other cases by calling it directly (e.g. to quickly test its behavior on a number property set to `0`, or an empty string etc.).

tests/unit/test-cmd/test.build.js
```diff
@@ -72,6 +72,40 @@ describe('build', () => {
     });
   });
 
+  it('throws an error if the filename manifest value is not a string', () => {
```
rpl, line 75: Nit, `throws on missing manifest properties in filename template`? or something like that.

src/program.js
escapewindow: > The unfortunate part of this change is that I guess that it would make coveralls unhappy ;-)
To "make coveralls happy again", we could add an integration test for this as part of tests/functional/test.cli.build.js.

Done, but `test-functional` doesn't seem to be run during `npm run test-coverage` ? Maybe if I push to my branch Github will pick up the test-functional run as well?

escapewindow: Fixed, by exporting the function and adding coverage in the `test.program.js` unit test.

rpl: @escapewindow
My bad, I forgot that we are not collecting the coverage from test-functional, and I recall that to make it do so there were some issues that I didn't have the time to look into at the time (and now I don't even recall what it was in enough detail).

Looking to the coveralls report:

https://coveralls.io/builds/30660260/source?filename=src%2Fprogram.js#L348

it seems that it is only complaining because on the new `throwUsageErrorIfArray`, in particular about the `if (Array.isArray(...)) {...}` branch.

Let's just export the helper function and test that in a unit test added to `tests/unit/test.program.js`.
That should be a small and simple test and enough to make coveralls to stop complaining.

My apologies again for forgetting that the functional test would have not be enough to make coveralls happy with the coverage (but we can keep the test, yargs has the bad habit to be very surprising in subtle changes from version to version, I prefer if we can have a proper integration test for this).

About normalize, it is totally ok to remove it, anyway we did already determine that it was not going to do anything needed or useful.

escapewindow: Yeah, agreed. Looks like we came to the same solution :)

I think this is ready for re-re-review. Thanks!

tests/unit/test-cmd/test.build.js
```diff
@@ -72,6 +73,40 @@ describe('build', () => {
...
+    return withTempDir(
+      (tmpDir) =>
+        build({
+          sourceDir: fixturePath('minimal-web-ext'),
+          artifactsDir: tmpDir.path(),
+          filename: '{missingKey}-{version}.xpi',
+        })
+          .then(makeSureItFails())
+          .catch((error) => {
+            log.info(error);
```
rpl, line 86: Nit, these long.info(error) are not really needed, I just noticed that we did already have one which got merge in a old patch, but not a big deal. We will have to rewrite all these tests into async functions at some point (to make them a bit more readable), and so don't worry about that.

tests/unit/test-cmd/test.build.js
```diff
@@ -86,6 +121,44 @@ describe('build', () => {
     );
   });
 
+  it('throws an error if the filename contains a path', () => {
```
rpl, line 124: Nit, it would probably good to have also one test that verifies that if an interpolated part of the template is a string that looks like a path then we still throw this kind of UsageError (which is actually the case with the version in this patch, it would be mostly an additional test to have more coverage in case of a refactoring), but we can deal with it in a follow up.

escapewindow:
@rpl Is there a release planned anytime soon? We have a number of webextensions that would use this feature, if it were available. Thanks!

rpl:
> @rpl Is there a release planned anytime soon? We have a number of webextensions that would use this feature, if it were available. Thanks!

@escapewindow yeah, we should cut a new release and publish it on npm soon. Sorry for letting you wait.

Issue #1891, opened 2020-04-22
Cannot load extensions on android build variants


jonalmeida:
This is a follow-up from this PR where we learnt that build variants aren't supposed: https://github.com/mozilla/web-ext/pull/1871#issuecomment-605400070
### Is this a feature request or a bug?
Bug

### What is the current behavior?

1. Install a local `debug` build of [Reference Browser](https://github.com/mozilla-mobile/reference-browser/); affects Fenix too.
2. Run the following command:
```
web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.reference.browser.debug --firefox-apk-component=BrowserActivity
```
3. Observe.

### What is the expected or desired behavior?
You should be able to install your web extension.

### What is the actual behaviour?

You see the following output in your terminal:

```
Error: Activity class {org.mozilla.reference.browser.debug/org.mozilla.reference.browser.debug.BrowserActivity} does not exist.
    at /Users/jalmeida/src/web-ext/node_modules/adbkit/lib/adb/command/host-transport/startactivity.js:57:21
    at tryCatcher (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/promise.js:577:18)
    at Async._drainQueue (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/jalmeida/src/web-ext/node_modules/bluebird/js/main/async.js:15:14)

```

### Version information (for bug reports)

* **Firefox version**: Desktop: 77.0a1 (2020-04-21) (64-bit), Android: GV 77.0.20200422093542
* **Your OS and version**: Mac OS 10.15.4 (19E287) / Android 10
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````
v13.11.0
6.13.7
4.1.0

Rob--W:
Thanks for filing.

Fixing this is relatively simple, see

https://github.com/mozilla/web-ext/pull/1871#issuecomment-610472381
[...]

PR #1918, opened 2020-06-03
fix: Enable extensions to load on android build variants

MikeM711:
This PR fixes https://github.com/mozilla/web-ext/issues/1891 to add support for extensions loaded on android build variants.

Tested using:

```
web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.reference.browser.debug --firefox-apk-component=org.mozilla.reference.browser.BrowserActivity

web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=org.mozilla.fenix.HomeActivity
```

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -277,10 +277,8 @@ export default class ADBUtils {
       component = `${apk}/.App`;
     } else if (apkComponent.includes('.')) {
       component = `${apk}/${apkComponent}`;
-    } else if (apkComponent) {
-      component = `${apk}/.${apkComponent}`;
     } else {
-      component = `${apk}/.App`;
+      component = `${apk}/.${apkComponent}`;
     }
 
     await wrapADBCall(async () => {
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -271,15 +271,14 @@ export default class ADBUtils {
       value: `-profile ${deviceProfileDir}`,
     }];
 
-    let component;
-
     if (!apkComponent) {
-      component = `${apk}/.App`;
-    } else if (apkComponent.includes('.')) {
-      component = `${apk}/${apkComponent}`;
-    } else {
-      component = `${apk}/.${apkComponent}`;
+      apkComponent = '.App';
+    } else if (!apkComponent.includes('.')) {
+      apkComponent = `.${apkComponent}`;
     }
+    // if `apkComponent` starts with a '.', then adb will expand
+    // the following to: `${apk}/${apk}.${apkComponent}`
+    const component = `${apk}/${apkComponent}`;
 
     await wrapADBCall(async () => {
       await adbClient.startActivity(deviceId, {
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -271,8 +271,17 @@ export default class ADBUtils {
       value: `-profile ${deviceProfileDir}`,
     }];
 
-    const component = apkComponent ?
-      `${apk}/.${apkComponent}` : `${apk}/.App`;
+    let component;
+
+    if (!apkComponent) {
+      component = `${apk}/.App`;
+    } else if (apkComponent.includes(".")) {
+      component = `${apk}/${apkComponent}`;
+    } else if (apkComponent) {
+      component = `${apk}/.${apkComponent}`;
+    } else {
+      component = `${apk}/.App`;
+    }
 
     await wrapADBCall(async () => {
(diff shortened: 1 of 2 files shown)


src/util/adb.js
```diff
@@ -271,8 +271,15 @@ export default class ADBUtils {
...
-    const component = apkComponent ?
-      `${apk}/.${apkComponent}` : `${apk}/.App`;
+    let component;
+
+    if (!apkComponent) {
+      component = `${apk}/.App`;
+    } else if (apkComponent.includes('.')) {
+      component = `${apk}/${apkComponent}`;
+    } else {
+      component = `${apk}/.${apkComponent}`;
```
Rob--W, line 281: Prepend a comment:

```
// adb expands the following to: `${apk}/${apk}.${apkComponent}`
```

MikeM711: Just to be sure, would you like this comment after the line 278  `else if` statement?

MikeM711:
Sure! We can take a look at whitelists. 

If I'm understanding correctly, would whitelisting be similar to writing:

```
web-ext run -t firefox-android --android-device=<device ID here>
```

and receiving a list of packages to choose from? And using that idea to show a list of `--firefox-apk-component` options?

src/util/adb.js
Rob--W: It kind of applies to every line indeed. Let's modify the apkComponent variable if needed and add the comment before `component`, like this:

```javascript
if (!apkComponent) {
  apkComponent = ".App";
} else if (!apkComponent.includes(".")) {
  apkComponent = '.' + apkComponent;
}
// if `apkComponent` starts with a '.', then adb will expand
// the following to: `${apk}/${apk}.${apkComponent}`
const component = `${apk}/${apkComponent}`;
```

Rob--W:
> Sure! We can take a look at whitelists.
> 
> If I'm understanding correctly, would whitelisting be similar to writing:
> 
> ```
> web-ext run -t firefox-android --android-device=<device ID here>
> ```
> 
> and receiving a list of packages to choose from? And using that idea to show a list of `--firefox-apk-component` options?

No, `.App` should still be used by default (unless that is somehow not providing the default activity for testing). `web-ext run` should ideally not require parameters if it can find reasonable defaults on its own. In #1891 there was some discussion on `performancetest` and how it doesn't work on debug builds; the hardcoded list could help with resolving that. But that's something to discuss in a separate issue, not this PR. Otherwise the discussion is going to be scattered over three issues (this issue, #1981 and the new issue to file), which makes it difficult to follow.

MikeM711:
Hi Rob, apologies for the wait.
I will be modifying that `apkComponent`, testing that out, and adding the extra comments after work (5:00PM EST), definitely expect an update today.

src/util/adb.js
MikeM711: That looks really good - define the `component` after the `apkComponent` check, I like it

MikeM711:
No problem! Thanks @Rob--W and also @jonalmeida for the help.

Rob - is exploring whitelists in a separate issue still something you would like to do? I'd be happy to lend a hand if you still think there is some merit to the idea.

Rob--W:
> Rob - is exploring whitelists in a separate issue still something you would like to do? I'd be happy to lend a hand if you still think there is some merit to the idea.

I'm positive towards a patch to support debug builds without requiring a fully qualified value in `--apk-component`. We already have some hardcoded app IDs in https://github.com/mozilla/web-ext/blob/e60411282655cbb23426544b8b5658f435505394/src/util/adb.js#L120-L124 , and I think that the patch would probably refactor the code by moving that hardcoded list elsewhere, and then be reused to improve the app detection and generate a good candidate for debug build detection.

MikeM711:
Great, I'll be filing a new issue for that under "feature request".

Before I begin a new issue, I think it would be best to make sure it starts off on the right foot. Just making sure, we would like the expected behavior to go from something like this

```
--firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=org.mozilla.fenix.HomeActivity
```
to this?
```
--firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=HomeActivity
```
Just making sure I'm understanding the vocab properly.

Rob--W:
@MikeM711 Yes, your understanding is correct.

Issue #1909, opened 2020-05-19
Unable to use custom chromium profile. `--chromium-profile` flag is not getting respected.


deepTier5:
### Is this a feature request or a bug?
Bug

### What is the current behavior?
As per the doc --chromium-profile takes the current path of the Chrome/Chromium profile, but every time Chrome is staring with a temporary profile instead of supplied one.

I am running
```bash
web-ext run --source-dir={source_dir} --target chromium --chromium-profile={chrome_profile_path} --start-url={start_url} --verbose
```
The verbose log on stdout
```bash
[program.js][info] Version: 4.2.0
[program.js][debug] Discovering config files. Set --no-config-discovery to disable
[config.js][debug] Discovered config "/home/djmsuman/.web-ext-config.js" does not exist or is not readable
[config.js][debug] Discovered config "/home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/web-ext-config.js" does not exist or is not readable
[program.js][info] Applying config file: ./package.json
[config.js][debug] Loading JS config file: "/home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/package.json" (resolved to "/home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/package.json")
[config.js][debug] Looking for webExt key inside package.json file
[config.js][debug] Config file /home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/package.json did not define any options. Did you set module.exports = {...}?
[cmd/run.js][info] Running web extension from /home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/build
[util/manifest.js][debug] Validating manifest at /home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/build/manifest.json
[util/temp-dir.js][debug] Created temporary directory: /tmp/tmp-web-ext-27230BgUeUgsNz2cM
[extension-runners/chromium.js][debug] Creating reload-manager-extension in /tmp/tmp-web-ext-27230BgUeUgsNz2cM/reload-manager-extension-1589925088612
[extension-runners/chromium.js][debug] Starting Chromium instance...
[cmd/run.js][info] The extension will reload if any source file changes
[util/file-filter.js][debug] Resolved path **/*.xpi with sourceDir /home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/build to /home/djmsuman/Workspace/BrowserExtentions/autopilot-admin/build/**/*.xpi
[...]

PR #1920, opened 2020-06-04
fix: issue 1909: call chromiumLaunch with parameter userDataDir instead of --user-data-dir and set of --profile-directory

pearnaly:
fix: issue 1909: call chromiumLaunch with parameter userDataDir instead of --user-data-dir and set of --profile-directory
See discussion https://github.com/mozilla/web-ext/issues/1909
Thank you @Rob--W


src/extension-runners/chromium.js
```diff
@@ -127,8 +127,12 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...this.params.args);
     }
 
+    let userDataDir = false;
     if (this.params.chromiumProfile) {
-      chromeFlags.push(`--user-data-dir=${this.params.chromiumProfile}`);
+      const pathParts = path.parse(this.params.chromiumProfile);
+      const profileDir = pathParts.base;
+      chromeFlags.push(`--profile-directory=${profileDir}`);
```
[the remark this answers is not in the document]
pearnaly: Yes exactly! I changed it this way to have it fully compatible with --firefox-profile, so that the user provides the full path to the profile, and not the folder containing the profiles. This way the user also doesn't have to deal with the undocumented --profile-directory flag. The documentation of --chromium-profile ("Path to a custom Chromium profile") remains correct this way.

If the default --user-data-dir is correct, it is also possible to provide only the name of the profile, like for Firefox.

Do you have concerns?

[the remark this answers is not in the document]
pearnaly: To simplify @Rob--W , `--chromium-profile = --user-data-dir + --profile-directory`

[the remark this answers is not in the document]
Rob--W: For the most common scenarios (using a temporary directory, or even a copy of a single-user profile), just passing `--user-data-dir` is sufficient. I don't want to complicate `web-ext` with `--profile-directory`, because this is a special case AND already possible without special support from `web-ext` (i.e. the `--arg` flag).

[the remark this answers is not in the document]
pearnaly: Okay for me @Rob--W, but it will mean:
- change the [documentation](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/) of `--chromium-profile` from "Path to a custom Chromium profile." to "Path to the directory containing the profiles" (or equivalent). Can you organize this change?
- loose the compatibility with `--firefox-profile`
I don't agree with it, but if you insist I do the change since I would like to have it finally somehow working.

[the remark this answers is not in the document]
Rob--W: There are two bugs here:

* According to your comment at https://github.com/mozilla/web-ext/issues/1909#issuecomment-637594465 , the `--chromium-profile` flag fails to match the specified directory on Windows (even though it works fine on Linux). This is a bug that we should certainly fix.
* Now in this PR and thread, you are advocating for support of a specific "User Account" within a given user data directory, via the `--profile-directory` flag. This is more work than needed to fix the previous bug, and I didn't feel the need to delay the fix of the previous bug for this new feature.

I'm not adamant against supporting User Accounts, but note that it will take some effort. I'll post a separate comment with lots of more details, I hope that it helps.

Rob--W:
Since you [appear to be very interested](https://github.com/mozilla/web-ext/pull/1920#discussion_r435369336) in supporting Chrome's "Profiles" other than `Default`, here are a few notes:

* `--user-data-dir` is the primary mechanism to [create separate data directories for running parallel instances of the Chrome browser](https://www.chromium.org/developers/creating-and-using-profiles).
* `--profile-directory` cannot live on its own without `--user-data-dir` (when unspecified, some default user data directory is chosen).
* When `--profile-directory` is used, the state of the user data directory changes (the default profile then changes to the selected profile directory). Consequently, using a profile within the default user data directory may cause unexpected behavior. For example, when an external application opens a link, a new tab will be opened in the (last opened) testing profile rather than the actual chosen (default) profile from the user.
* `web-ext run` currently supports the selection of Firefox profiles by name rather than path. This is not documented nor supported in `web-ext run` with Chrome. To have the same feature for Chrome (using the system's default user data dir), we would have to detect the location of the default user data dir, and pass `--profile-directory` if needed. There is ambiguity: does `--chromium-profile "Profile 1"` mean "Profile 1" of the default user data dir, or "Profile 1" in the current directory? Chrome will create the user data dir/profile if it does not exist. This ambiguity is bad and is destructive.
* The official `chrome-launcher` package does not support `--profile-directory` ( https://github.com/GoogleChrome/chrome-launcher/issues/172 / https://github.com/GoogleChrome/chrome-launcher/issues/190). Once they figure out a standard way to select a specific directory (e.g. even without `--user-data-dir` support), then it would be easier to support the same here in `web-ext`.
* Changing the meaning of `--chromium-profile` is a backwards-incompatible change for existing users of `web-ext`. With the proposed change, Chrome is going to create extra directories / files in the parent directory of the `--chromium-profile` path, which would be very unexpected and therefore undesirable.
* If you are interested in introducing support for combing `--user-data-dir` and `--profile-directory`, make sure that there is no ambiguity whatsoever before using the parent directory. You need to make sure that the parent directory is really the user data dir, AND that the given directory is NOT a user data dir (AND maybe even that the given directory is a profile directory). If unclear, treat the given path as a user-data-dir.

pearnaly:
1) Hi @Rob--W, okay I see now the conflict between the two chrome mechanisms for profiles. both directories levels can be considered as "profiles", which is confusing. According to [the chromium docu](http://www.chromium.org/user-experience/multi-profiles), the naming is "profile"/"account", but according to the command line params names and to the `chrome://version/` page it is "data dir"/"profile".

2) Yes you can provide only `--profile-directory` (on my Windows at least). in this case Chrome takes the default `user-data-dir` (like for Firefox).

3) Yes it is true, using `--profile-directory` changes the default profil in `user-data-dir`. But is is the same behavious if you provide it using this fix or passing the parameters [as you proposed](https://github.com/mozilla/web-ext/issues/1909#issuecomment-636385559). At least, if the users uses profiles in his everyday `user-data-dir` to debug, it is his problem (actually wanted) that these data are modified (what I were looking for). Users who don't want it will provide a path different from the defaut one or won't use it to use the temp profile.

4) It is fortunately not necessary to detect the position of the default used-data-dir. We just have to not provide it and chrome will use the default. But you are right the documentation has to be anyway extended, either to explain that `--chromium-profile` has the meaning of `user-data-dir` and `profile-directory` is not supported by the lib and can be passe manually OR that `--chromium-profile` has the meaning of the profile path according to `chrome://version/` and that (new!) only the profile directory can be provided, as for Firefox.

5) the chrome launcher doesn't support this flag, and it is probably okay since we can pass it ourselves. But if they start to support it yes we will have to make sure they don't try to set some colliding default.

6) It is indeed the critical point... I was pushing to this change since I considered as a bug that the docu named `chromium-profile` what would be the profiles container according to the `chrome://version/` naming schema. Is actually also the way @deepTier5 understood it in [his comment](https://github.com/mozilla/web-ext/issues/1909#issuecomment-636323174). I think most of the user will instinctively want to provide the full path. Since the chromium support is quite new and the profile flag probably few used (?) and has this windows bug anyway, it could be our last chance to change (fix) its meaning, but I understand your concern and leave it to you.

7) No I think double support in the same param is too complicated and even unpossible to tell apart. It will be either/or. Linux users currently using it would have to add `/Default`. But it is more clear on the long term.

What should I do?

Rob--W:
>     2. Yes you can provide only `--profile-directory` (on my Windows at least). in this case Chrome takes the default `user-data-dir` (like for Firefox).
> 
>     3. Yes it is true, using `--profile-directory` changes the default profil in `user-data-dir`. But is is the same behavious if you provide it using this fix or passing the parameters [as you proposed](https://github.com/mozilla/web-ext/issues/1909#issuecomment-636385559). At least, if the users uses profiles in his everyday `user-data-dir` to debug, it is his problem (actually wanted) that these data are modified (what I were looking for). Users who don't want it will provide a path different from the defaut one or won't use it to use the temp profile.
>
>     4. It is fortunately not necessary to detect the position of the default used-data-dir. We just have to not provide it and chrome will use the default.

`web-ext` intentionally tries to avoid damaging the user profile unless explicitly requested.
When a Firefox profile name is given instead of a path, `web-ext run --keep-profile-changes -p [name here]` will look up the location and copy it: https://github.com/mozilla/web-ext/blob/f2043c70e1823df00676fce6ffa7e9040fc475a6/src/firefox/index.js#L404-L447

In order to support this for Chrome, there should also be something similar.

>     5. the chrome launcher doesn't support this flag, and it is probably okay since we can pass it ourselves. But if they start to support it yes we will have to make sure they don't try to set some colliding default.

I mentioned this not only because of potential compat concerns, but also to point out that the odds of having proper support for this feature increases if the official upstream project properly supports the feature.

>     6. It is indeed the critical point... I was pushing to this change since I considered as a bug that the docu named `chromium-profile` what would be the profiles container according to the `chrome://version/` naming schema. Is actually also the way @deepTier5 understood it in [his comment](https://github.com/mozilla/web-ext/issues/1909#issuecomment-636323174). I think most of the user will instinctively want to provide the full path. Since the chromium support is quite new and the profile flag probably few used (?) and has this windows bug anyway, it could be our last chance to change (fix) its meaning, but I understand your concern and leave it to you.
> 
>     7. No I think double support in the same param is too complicated and even unpossible to tell apart. It will be either/or. Linux users currently using it would have to add `/Default`. But it is more clear on the long term.
> 
> 
> What should I do?

The default behavior should be to treat the flag as `--user-data-dir`. I see your points on the user taking the directory location from `chrome://version` (in the specific scenario of re-using an existing profile). That specific case could be supported by detecting the directories (there are some files/directories that are specific to user-data-dir / profile-dir, it is not as impossible as it seems).

I'll have to discuss this with other project members, but I currently believe that the path forwards is 7 and/or awaiting official upstream support (in chrome-launcher).

Rob--W:
I've discussed this with @rpl and we have agreed that the following approach will provide the requested functionality with reasonable compatibility:

1. Only one parameter, `--chromium-profile`.
2. If the directory exists, check if it is certainly a profile directory and NOT a user-data-dir. If this cannot be decided, keep the current behavior (treat as user-data-dir) (end of steps).
3. If `--keep-profile-changes` is used, check if the parent directory is a user-data-dir. If it is, then use the given profile. If it is not, then we should not use the parent directory as the `user-data-dir` (whether to exit with an error or treating it as `--user-data-dir` is up to you).
4. Otherwise (by default, if `--keep-profile-changes` is not set), create a new directory (to use as user-data-dir) and put the non-Default profile directory in it (and pass it to `--profile-directory`).

Note that this approach favors the use of `--user-data-dir` when in doubt. This is reasonable because users can always use `--arg=" --profile-directory=Profile 1"` if wanted.

To detect whether a directory looks like a user-data-dir: Check if the "Local State" file and "Default" directory exist.
To detect whether a directory looks like a profile-directory: Check if the given directory exists and the "Secure Preferences" file exist.
I looked in a new profile and in a 9-year-old Chromium profile, and confirmed that checking the existence of those files doesn't cause false positives (at first `--user-data-dir` contained the actual profile. Based on timestamps it seems that somewhere in 2011 the profile moved to the Default subdirectory).

pearnaly:
It's fine for me and I started.
I use sofar [fs-extra](https://github.com/jprichardson/node-fs-extra) for the files operations
I will have a hard time with the tests, I will try with [mock-fs](https://github.com/tschaub/mock-fs)...

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
...
+        'Secure Preferences');
+      //Secure Preferences is typical for a profile dir
+      return fs.pathExistsSync(securePreferencesCandidate);
+    }
+
+    let userDataDir = false;
+    const chromiumProfile = this.params.chromiumProfile;
+    if (chromiumProfile) {
+      const isProfileDirAndNotUserData =
+        fs.pathExistsSync(chromiumProfile)
```
Rob--W, line 151: This part of the check is not needed. `isProfileDir` returning true implies that the directory exists.

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
...
+        && isProfileDir(chromiumProfile)
+        && !isUserDataDir(chromiumProfile);
+
+      if (isProfileDirAndNotUserData) {
+        const pathParts = path.parse(chromiumProfile);
```
Rob--W, line 156: Let's use the following:

`const { dir: parentDir, base: profileName } = path.parse(chromiumProfile);`

The `parentDir` variable makes it very obvious that we're working with parent directories, which removes the need for the `// now check if the parent dir is a user-data-dir` comment below.

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
...
+              'remove --keep-profile-changes or use a profile in a ' +
+              'user-data-dir directory');
+          }
+          userDataDir = pathParts.dir;
+
+        } else {
+          // the user provided an existing profile directory but doesn't want
+          // the changes to be kept. we copy this directory to a temporary
+          // user data dir.
+          userDataDir = Launcher.prototype.makeTmpDir();
```
Rob--W, line 172: Don't use undocumented internal implementation details from an external library.

The `web-ext` project itself already has an utility to create temporary directories here: https://github.com/mozilla/web-ext/blob/9315b781ec73eead946c07551a24a413ddb1338a/src/util/temp-dir.js

An example of usage is here: https://github.com/mozilla/web-ext/tree/9315b781ec73eead946c07551a24a413ddb1338a/src/extension-runners

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
...
+          // copy profile dir to this temp user data dir.
+          fs.copySync(chromiumProfile, path.join(
+            userDataDir,
+            profileDirName));
+        }
+        chromeFlags.push(`--profile-directory='${profileDirName}'`);
```
Rob--W, line 178: Is it necessary to quote the profile dir name?

Chrome seems to auto-generate names like `Profile <n>`, but there is nothing that stops Chrome from creating a directory with a quote in its name when it is passed:

```
chromium --user-data-dir=/tmp/dir --profile-directory=\'test\'
```

results in `/tmp/dir/'test'` being created (at least on Linux).

What happens if you try to use a directory with quotes?

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
...
+      } else {
+        userDataDir = chromiumProfile;
```
Rob--W, line 180: Prepend comment to explain what `chromiumProfile` means at this point, so that the reader doesn't have to scroll all the way up.

And this doesn't properly handle the `--keep-profile-changes` case - see my top-level review comment for suggestions on improving this.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -356,10 +358,13 @@ describe('util/extension-runners/chromium', async () => {
     await runnerInstance.exit();
   });
 
-  it('does pass a user-data-dir flag to chrome', async () => {
+  it('does pass user-data-dir flag to chrome', async () => {
+
+    mock({}); // empty file system
```
Rob--W, line 363: Just use the `withTempDir` utility + fs to create the temporary files in the tests, instead of depending on `mock-fs`.

package.json
```diff
@@ -65,6 +65,7 @@
     "es6-error": "4.1.1",
     "event-to-promise": "0.8.0",
     "firefox-profile": "1.3.1",
+    "fs-extra": "9.0.1",
```
Rob--W, line 68: I'm not keen on adding new dependencies because it generally results in more maintenance load (e.g. having to publish new versions when a security issue is found in the dependency). But since we're already indirectly depending on this module, I'm fine with it.

In a follow-up issue/PR we can probably replace `mz` with this, since `mz`'s functionality overlaps with `fs-extra`.

tests/unit/test-extension-runners/test.chromium.js
pearnaly: done. also removed mock-fs dependency

src/extension-runners/chromium.js
pearnaly: I removed the quotes

src/extension-runners/chromium.js
pearnaly: I switched to TempDir, but with TempDir you cannot just get a temp path without creating the directory itself. I have to delete it and rewrite it after, which is quite ugly. My method would like to just get a path, and create the actual dir separately...

src/extension-runners/chromium.js
pearnaly: like it

src/extension-runners/chromium.js
pearnaly: yes indeed

src/extension-runners/chromium.js
```diff
@@ -7,26 +7,26 @@
...
 
-import {fs} from 'mz';
+import fs from 'fs-extra';
 import asyncMkdirp from 'mkdirp';
 import {
-  LaunchedChrome,
-  launch as defaultChromiumLaunch,
   default as ChromeLauncher,
+  launch as defaultChromiumLaunch,
+  LaunchedChrome, Launcher,
```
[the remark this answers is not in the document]
pearnaly: I reverted this import

.eslintrc
```diff
@@ -12,7 +12,7 @@
       "modules": true,
       "restParams": true,
       "spread": true
-    },
+    }
```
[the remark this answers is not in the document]
pearnaly: done

src/extension-runners/chromium.js
```diff
@@ -127,8 +127,58 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...this.params.args);
     }
 
-    if (this.params.chromiumProfile) {
-      chromeFlags.push(`--user-data-dir=${this.params.chromiumProfile}`);
+    function isUserDataDir(dirCandidate: string): boolean {
+      const localStateCandidate = path.join(dirCandidate, 'Local State');
+      const defaultCandidate = path.join(dirCandidate, 'Default');
+      // Local State and Default are typical for the user-data-dir
+      return fs.pathExistsSync(localStateCandidate)
```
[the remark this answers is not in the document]
pearnaly: I switched to isDirectory. The use of fs-extra is now reduced, and we could reevaluate if it is still justified.

src/extension-runners/chromium.js
pearnaly: I made an early return statement for it, so it should be clear now

pearnaly:
I think my 3 tests test all the pathes. Do you see another test case?

pearnaly:
I have an issue with the error test. `npm run test` is passing on my computer but failing on the CI pipeline. I will have a look at it again tomorrow

src/extension-runners/chromium.js
```diff
@@ -127,8 +193,21 @@ export class ChromiumExtensionRunner {
...
-      chromeFlags.push(`--user-data-dir=${this.params.chromiumProfile}`);
+    const profileParts =
+      await ChromiumExtensionRunner.getProfileParts(this.params);
+
+    if (profileParts.userDataDir
+      && profileParts.profileDirName
+      && !this.params.keepProfileChanges
+    ) {
+      // copy profile dir to this temp user data dir.
+      fs.copySync(this.params.chromiumProfile, path.join(
```
Rob--W, line 204: Let's use `await fs.copy`, as there is no real need for this to be synchronous.

> The use of fs-extra is now reduced, and we could reevaluate if it is still justified.

There are multiple (dev) dependencies with overlapping functionality. For example, `mz` and `fs-extra` both offer promisified `fs` methods, and `fs-extra` and `copy-dir` both offer ways to copy. I think that `fs-extra` is the best replacement in the long term.

src/extension-runners/chromium.js
```diff
@@ -127,8 +175,46 @@ export class ChromiumExtensionRunner {
...
+          'user-data-dir. The changes cannot be kept. Please either ' +
+          'remove --keep-profile-changes or use a profile in a ' +
+          'user-data-dir directory');
+      }
+    } else if (userDataDir) {
+      // the user provided an existing profile directory but doesn't want
+      // the changes to be kept. we copy this directory to a temporary
+      // user data dir.
+      const tmpDir = new TempDir();
+      await tmpDir.create();
```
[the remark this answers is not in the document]
rpl: @Rob--W TempDir should be doing that automatically (actually the "tmp" npm dependency should be doing it internally, [even on process exit](https://github.com/raszi/node-tmp/blob/7ae22ed2d56c10d425a66e99fe8bc10c925442e6/lib/tmp.js#L670)). At the moment I can't reproduce #1957 as I did mention in that issue.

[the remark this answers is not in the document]
Rob--W: Thanks for correcting me, I'll mark this request as resolved.

src/extension-runners/chromium.js
```diff
@@ -127,8 +175,46 @@ export class ChromiumExtensionRunner {
...
+    if (userDataDir && this.params.keepProfileChanges) {
+      if (profileDirName
+        && !await ChromiumExtensionRunner.isUserDataDir(userDataDir)) {
+        throw new Error('The profile you provided is not in a ' +
+          'user-data-dir. The changes cannot be kept. Please either ' +
+          'remove --keep-profile-changes or use a profile in a ' +
+          'user-data-dir directory');
+      }
+    } else if (userDataDir) {
+      // the user provided an existing profile directory but doesn't want
```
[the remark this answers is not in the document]
Rob--W: This comment hasn't been addressed yet.

To address it, you need to remove ` if (userDataDir) {` here (to turn this into  `} else {`) and it to the `} else {` below.

```
      } else if (userDataDir) { // <-- Here
        await fs.copy(userDataDir, tmpDirPath);
```

src/extension-runners/chromium.js
```diff
@@ -126,8 +174,44 @@ export class ChromiumExtensionRunner {
...
+      // the changes to be kept. we copy this directory to a temporary
+      // user data dir.
+      const tmpDir = new TempDir();
+      // TODO: Remove tmpDir - https://github.com/mozilla/web-ext/issues/1957
```
Rob--W, line 195: Sorry for the confusion, as Luca mentioned at https://github.com/mozilla/web-ext/pull/1920/files#r454511824 , the `tmp` library already removes the file at exit. So this comment can be removed.

Rob--W:
I locally checked out the PR to prepare for merging, but noticed that the test failure is actually caused by your changes.

Minimal reproduction:

```
$ ./node_modules/.bin/mocha tests/unit/test-extension-runners/test.chromium.js -f 'throws an error on profile in invalid user-data-dir'


  util/extension-runners/chromium
    ✓ throws an error on profile in invalid user-data-dir


  1 passing (40ms)

^C
$ # ^ only exited because I explicitly triggered an exit.
```

In that test, you're not calling `runnerInstance.exit();`. As a result, the internal server that it had started doesn't exit and keeps the process alive.

pearnaly:
We got it! Thank you @Rob--W for accompanying me in my fist open source PR.

Rob--W:
You're welcome, and thanks again for raising attention on issue #1909 and submitting a patch!

caitmuenster:
Thanks for the patch, @pearnaly! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

pearnaly:
Thank you Caitlin!
      i created an account: pearnaly
Have a nice day

Viele Grüße
Paul-Émile Arnaly


Am Mo., 20. Juli 2020 um 22:00 Uhr schrieb Caitlin Neiman <
notifications@github.com>:

caitmuenster:
Excellent! Your profile has been vouched. 🎉

Welcome onboard! We look forward to seeing you around.

PR #1933, opened 2020-06-17
docs: fix examples with modules

lfilho:
Running the examples with node 14 (for example) would return undefined errors.

As mentioned in https://github.com/mozilla/web-ext/issues/1927

PR #1934, opened 2020-06-18
fix: improve and test importing web-ext as an ESM module or requiring as a CommonJS module

rpl:
This pull request contains a small set of proposed tweaks to improve a bit (and cover with tests) how web-ext can be used as a library, imported as an ESM module or required as a CommonJS module.

Given that this change what web-ext is going to export when loaded as a module, it would be likely better to be included in a major release (to avoid to break npm packages that may depend on the current exports).  

Other pull requests related to this:
- #1933


rpl:
@Rob--W no need to rush to review this, as I wrote in the description this isn't something that we should include in the upcoming v. 4.3.0 (I added it in the backlog of the next major release, v5.0.0), but I did pushed it so that we can evaluate this approach (and test coverage for this) in the meantime.

index.mjs
```diff
@@ -0,0 +1,13 @@
+// This module provides a nicer ESM module wrapper around the module generated by webpack.
+
+// NOTE: disabled eslint rules:
+// - import/extensions: we need the explicit .js here to force node to look for the .js file instead of
+//   looking for a .mjs file.
+// - import/no-unresolved: the dist/web-ext.js file is likely not build yet when this runs in the CI jobs
```
Rob--W, line 6: "is likely not build yet" -> "has likely not built yet".

Issue #1935, opened 2020-06-20
 Support for android build variants without requiring a fully qualified value in the `--apk-component` flag


MikeM711:
Relevant discussion: https://github.com/mozilla/web-ext/issues/1891#issuecomment-632423453 and https://github.com/mozilla/web-ext/pull/1918#issuecomment-644250208.

### Is this a feature request or a bug?
Feature

### What is the current behavior?

In order to run an android build variant with an `apkComponent`, we need to pass in a fully qualified value in the `--apk-component` flag.

For example, the below command will successfully install an extension on the `performancetest` build of [Fenix](https://github.com/mozilla-mobile/fenix/) using the `HomeActivity` component.

```
web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=org.mozilla.fenix.HomeActivity
```

### What is the expected or desired behavior?

The desired behavior would be having a successful install by requiring a shorter value in the `--apk-component` flag. Using the above command as an example, the below command would be its equivalent.
```
web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=HomeActivity
```

### Version information (for bug reports)

* **Firefox version**: Desktop: 77.01.1 (64-bit), Android: 68.8.0
* **Your OS and version**: Mac: 10.13.3, Android 10
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````

v12.16.3
6.14.4
master-d61d94cbcf4834ec7613e80e51fc655dce06275c

MikeM711:
@Rob--W Would you like to have a hardcoded list outside of `adb.js` or keep it inside the file? 

Originally I was thinking right inside the `src` directory, in case we would need to reuse the list somewhere else in the future.

Rob--W:
Since it's not really related to ADB, a separate file, e.g. `src/firefox/package-identifiers.js` sounds good to me.

MikeM711:
I have everything set up, just have a quick question:

The `test.adb.js` file uses `geckoview_example` as an android build to test with. I can get the tests and everything working by sticking the `geckoview_example` in the list of browsers. Is this okay to do? 

The good thing is that the application stops you from using packages that aren't installed, before it touches the list to begin with.

I can post a PR if that makes the explanation easier.

Rob--W:
[...]

PR #1941, opened 2020-06-26
feat: Support for android build variants without requiring a fully qualified value in the `--apk-component` flag

MikeM711:
Fixes https://github.com/mozilla/web-ext/issues/1935

Desired behavior from the original issue:

```
web-ext run -t firefox-android --android-device=<device ID here> --firefox-apk=org.mozilla.fenix.performancetest --firefox-apk-component=HomeActivity
```


tests/unit/test-util/test.adb.js
```diff
@@ -780,8 +781,8 @@ describe('utils/adb', () => {
 
       const promise = adbUtils.startFirefoxAPK(
         'device1',
-        'org.mozilla.geckoview_example',
-        'org.mozilla.geckoview_example.GeckoViewActivity', // firefoxApkComponent
+        'org.mozilla.geckoview_example.release',
+        'GeckoViewActivity', // firefoxApkComponent
```
Rob--W, line 785: Those tests should not be modified (because the old syntax should still work).

Please add a new test.

src/firefox/package-identifiers.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+
+export const browserList = [
+  'org.mozilla.fennec',
+  'org.mozilla.fenix',
```
Rob--W, line 5: Could you run an actual test with `org.mozilla.fenix.nightly` to verify that it works, and if it does work, add a unit test for it? It is not entirely obvious that `org.mozilla.fenix.nightly` ought to use `org.mozilla.fenix` as the name.

src/firefox/package-identifiers.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+
+export const browserList = [
```
Rob--W, line 3: Rename to `packageIdBases` ?

or just `export default`. The file name, `packageIdentifiers`, is already a nice name on its own.

src/util/adb.js
```diff
@@ -271,14 +272,17 @@ export default class ADBUtils {
       value: `-profile ${deviceProfileDir}`,
     }];
 
-    if (!apkComponent) {
-      apkComponent = '.App';
-    } else if (!apkComponent.includes('.')) {
-      apkComponent = `.${apkComponent}`;
```
Rob--W, line 277: This original logic should be kept. It allows users to use a custom app that isn't in the hardcoded list. For example if they develop their own GeckoView-based app.

The change that you need is to check if `apk` starts with any of the known prefixes, and if it is, prepend that prefix before `.${apkComponent}`.

tests/unit/test-util/test.adb.js
MikeM711: Sure thing - I'll add a new test for `org.mozilla.fenix.nightly`

src/util/adb.js
```diff
@@ -271,6 +272,17 @@ export default class ADBUtils {
...
+      if (apk.startsWith(browser) && apkComponent) {
+        if (apk !== browser && !apkComponent.includes('.')) {
+          apkComponent = `${browser}.${apkComponent}`;
+        }
+        // if 'geckoview_example' has been found, leave the
+        // loop, before it checks 'geckoview'
+        break;
+      }
+    }
+
```
MikeM711, line 285: Hopefully this isn't too hacky - it's mainly due to wanting to leave the loop if `geckoview_example` has been checked, so it doesn't check again on `geckoview`

src/firefox/package-identifiers.js
MikeM711: Yup, just added the test below

src/util/adb.js
Rob--W: `break` after finding the matching `apk` is not a hack, it looks normal.

src/util/adb.js
```diff
@@ -271,6 +272,17 @@ export default class ADBUtils {
       value: `-profile ${deviceProfileDir}`,
     }];
 
+    for (const browser of packageIdentifiers) {
+      if (apk.startsWith(browser) && apkComponent) {
+        if (apk !== browser && !apkComponent.includes('.')) {
```
Rob--W, line 277: This logic should be after the following block below, to make sure that the `browser` is also prepended by default (`.App`):
```
    if (!apkComponent) {
      apkComponent = '.App';
    } else if (!apkComponent.includes('.')) {
      apkComponent = `.${apkComponent}`;
    }
```

It would be nice to have unit tests to cover these cases.

And then the logic should be like this:

```
if (apkComponent.startsWith(".")) {
  // loop over packageIdentifiers,
  // and if apk is either identical to the package,
  // or apk starts with `${package}.`,
  // prepend the package identifier before `apkComponent`
  // and break. 
}
```

MikeM711: > or apk starts with `${package}.`

Ah! That's a good idea for the whole "geckoview" vs "geckoview_example"

src/util/adb.js
```diff
@@ -311,22 +311,23 @@ export default class ADBUtils {
...
     } else if (!apkComponent.includes('.')) {
       apkComponent = `.${apkComponent}`;
     }
+
+    // if `apk` is a browser package or the `apk` has a
+    // browser package prefix: prepend the package identifier
+    // before `apkComponent`
+    if (apkComponent.startsWith('.')) {
+      for (const browser of packageIdentifiers) {
+        if (apk.startsWith(`${browser}.`)) {
```
MikeM711, line 325: Quick note (before I move on to adding tests):

In the previous comment, there was the condition: 
```
if apk is either identical to the package, or apk starts with `${package}.`
````
For the tests to work, I took out the `apk is identical to the package` part of the condition.

The offended test was: `starts a given APK component without a period`
* Incoming apk: `org.mozilla.geckoview_example`
* Incoming apkComponent: `GeckoViewActivity`
* Expected result: `org.mozilla.geckoview_example/.GeckoViewActivity`
* Result received: `org.mozilla.geckoview_example/org.mozilla.geckoview_example.GeckoViewActivity`

Would you like me to expand out the Expected Result of this test to match?

Rob--W: For ADB the two results are equivalent.

The behavior of the received result is probably more desirable to support the .debug/. performancetest scenarios.

MikeM711:
Hey, sorry I've been spotty on this PR - backlogged with work and class.

On Aug 23rd I should have a nice break - this PR is a simple fix, but I like to have full attention for it before proceeding. Should be good then.

Rob--W:
Thanks for the heads-up @MikeM711 and no worries! Your contributions are appreciated, regardless of how much time it takes.

MikeM711:
Assuming that the previous commit looks good to you,

For APK tests we currently have:

| Test name | Incoming apk |  Incoming apkComponent| 
|--- |---|---|
|rejects an UsageError on adb binary not found' | org.mozilla.firefox_mybuild |  undefined|
|starts Firefox APK on a custom profile (only used by Fennec) | org.mozilla.firefox_mybuild |  undefined | 
|starts a given APK component without a period | org.mozilla.geckoview_example | GeckoViewActivity  |
|starts a given APK component with a period | org.mozilla.geckoview_example | org.mozilla.geckoview_example.GeckoViewActivity |
|starts a given APK component on fenix.nightly | org.mozilla.fenix.nightly | HomeActivity |

To take care of, https://github.com/mozilla/web-ext/pull/1941#discussion_r452216040, what do you think about these few tests?

| Test name | Incoming apk |  Incoming apkComponent| 
|--- |---|---|
|starts an undefined APK component |org.mozilla.geckoview_example | undefined  |
|starts a fully-qualified APK component on the build-variant: fenix.nightly | org.mozilla.fenix.nightly | org.mozilla.fenix.HomeActivity |
|starts a given APK component that begins with a period | org.mozilla.fenix.nightly | .HomeActivity |

Rob--W:
The proposed tests look good. Thanks for the table with the clear summarizing overview.

> starts an undefined APK component

I would make this "starts without specifying an APK component"

MikeM711:
No problem! Let me know how the 3 tests look

Rob--W (approved):
LGTM

PR #1944, opened 2020-06-29
feat: Add support for browser_specific_settings

### LINKED ISSUE #1901: web-ext ignores manifest browser_specific_settings.gecko.id field
Bug

I'm trying to sign a new version of an unlisted add-on, but running
```
web-ext sign --channel unlisted --source-dir dist --api-key $KEY --api-secret $SECRET
```
Results in 
```
$ web-ext sign --channel unlisted --source-dir dist --api-key $KEY --api-secret $SECRET
Building web extension from /code/klaus/mono-ui/packages/extension/dist
No extension ID specified (it will be auto-generated)
Specifying a channel for a new add-on is unsupported. New add-ons are always in the unlisted channel.
Server response: Duplicate add-on ID found. (status: 400)
FAIL

o: The extension could not be signed
    at /code/klaus/mono-ui/node_modules/web-ext/dist/web-ext.js:1:66636
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at E.execute (/code/klaus/mono-ui/node_modules/web-ext/dist/web-ext.js:1:33162)

error Command failed with exit code 1.
```

Whereas I can successfully sign the new version, when I run the same command but append `--id firefox-addon@klausapp.com` to the end of `web-ext sign`.

My `manifest.json` file includes:
```
    "gecko": {
      "id": "firefox-addon@klausapp.com",
      "update_url": "https://extension.com/update-manifest.json"
    }
  }
```

If the `--id` flag is not passed in, it should respect the `browser_specific_settings.gecko.id` field. Actually, the [docs](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#web-ext-sign) even mention that the `--id` flag should do _nothing_ in this scenario and the identifier from the manifest would be used:

> A custom identifier string for the extension. This has no effect *if the extension already declares an identifier in its manifest*.

* **Firefox version**: 76.0b8 (64-bit) (N/A?)
* **Your OS and version**: MacOS Catalina 10.15.3

````

v13.8.0
6.14.4
4.2.0
````

### PR DESCRIPTION
**Pull request by Lusito**
So, as request, I've made a cleaned up version of #1791. It fixes #1901 and resolves #1791.
I've kept the history so that the work of @andyrichardson can still be credited.

Flow and chromium tests don't seem to run on windows, so I'll wait for travis to check these. lint and unit-tests are green.

What I've done:
- Undone the formatting changes
- Replaced the reducer + try/catch version with a simple for of loop. Call me crazy, but I don't like try/catch for things that can be checked easily.
- Made the tests DRY.
- Fixed flow errors

### CODE REVIEW

File: tests/unit/test-extension-runners/test.firefox-desktop.js [hunk 1]
```diff
@@ -285,6 +285,7 @@ describe('util/extension-runners/firefox-desktop', () => {
     assert.equal(install.asProxy, true);
     assert.equal(install.manifestData.applications.gecko.id,
                  manifestData.applications &&
+                 manifestData.applications.gecko &&
```
Lusito, line 288: This seems to be the pattern to silence flow here, but it feels wrong. Should we maybe just test for the exact string here rather than optional chaining these properties? (same for the test.sign.js version)

Lusito:
@rpl Should be green now (it hangs on npm ci for the windows machine).
I only have one remark about the flow fix, since I've never actually used flow before.

tests/unit/test-util/test.manifest.js
```diff
@@ -211,14 +211,39 @@ describe('util/manifest', () => {
...
     });
 
-    it('returns undefined when ID is not specified', () => {
-      assert.strictEqual(getManifestId(manifestWithoutApps), undefined);
+    describe('without applications and browser_specific_settings', () => {
+
+      it('returns undefined when ID is not specified', () => {
+        assert.strictEqual(getManifestId(manifestWithoutApps), undefined);
+      });
+
```
rpl, line 246: On the Firefox side "browser_specific_settings" has an higher priority than "applications" and in this patch it looks that web-ext is going to do the same, it would be nice to add an explicit test case, would you mind add one small test case to cover this in the automated tests?

src/util/manifest.js
```diff
@@ -85,6 +86,15 @@ export default async function getValidatedManifest(
...
+    manifestData.browser_specific_settings,
+    manifestData.applications,
+  ];
+  for (const apps of manifestApps) {
+    if (apps && apps.gecko && apps.gecko.id) {
+      return apps.gecko.id;
+    }
+  }
+
+  return undefined;
```
rpl, line 99: Thanks! I actually like this version.

I took another look to what is being done on the Firefox side and it looks that we are doing a slightly different check here:
https://searchfox.org/mozilla-central/rev/5a4aaccb28665807a6fd49cf48367d47fbb5a19a/toolkit/mozapps/extensions/internal/XPIInstall.jsm#472-479

in particular, on the Firefox side, it looks that if browser_specific_settings.gecko is defined we only try to retrieve the id from that object and we don't look to applications.gecko.id anymore.

It may be reasonable to match that here in web-ext too, how that sounds to you?

rpl:
Closing in favor of #1974 (which include this changes, plus some small tweaks as described in #1974 description)

Lusito:
Weird, I did not get notified about your code review comments, but I got notified by the mention. Thanks for the merge!

PR #2038, opened 2020-10-02
fix: Improve visibility of message to enable "Remote Debugging via USB" with web-ext run -t firefox-android

### LINKED ISSUE #1969: Improve visibility of message to enable "Remote Debugging via USB" with web-ext run -t firefox-android
We have received multiple reports (e.g. #1899,  #1949 ) where the use thought that there was a bug in `web-ext run -t firefox-android`, while the real issue is that they forgot to enable "Remote Debugging via USB". We print a message with the instructions once:

https://github.com/mozilla/web-ext/blob/fc91c367723b8a7fb993c64927829aa66f9746d6/src/extension-runners/firefox-android.js#L585-L591

... but users still don't see this, e.g. because it is buried in the log files. We should improve the visibility of this instruction to help users with reminding that they need to enable this. For example, we could print the same message again when the connection continues to fail repeatedly.

sonalprabhu:
Fixes #1969 
fix: Looped the "Check Remote Debugging is ON" warning and highlighted it
Screenshot (109)


src/util/adb.js
```diff
@@ -347,14 +347,16 @@ export default class ADBUtils {
...
-    deviceId: string, apk: string,
+    deviceId: string, apk: string, msg: string,
     {maxDiscoveryTime, retryInterval}: DiscoveryParams = {}
   ): Promise<string> {
     let rdpUnixSockets = [];
 
     const discoveryStartedAt = Date.now();
 
     while (rdpUnixSockets.length === 0) {
+      log.info(`\n\x1b[93m${msg}\x1b[39m\n`);
```
rpl, line 363: Don't bother to make the message yellow for now, I think that it is a good idea, but this should also be checking that the terminal does actually support colors, take care of potential compatibility issues in different operating system terminal colors support etc.

To deal with styling the output we may want to introduce [chalk](https://github.com/chalk/chalk) as a new dependency, but I prefer to deal with it in a separate PR (e.g. because when we introduce a new dependency we have to also update the package-lock.json file, and I prefer to do it separately).

For now let's just call log.info(msg) here.

rpl, line 363: It would be nice to also add a new test case to cover this in this group of test cases:
https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.adb.js#L1042

we should be able to collect the logs and make an assertion about it using a testing strategy similar to what we are using here: https://github.com/mozilla/web-ext/blob/29798cbc61e78e024911150734081c2f4ddd1be6/tests/unit/test-extension-runners/test.chromium.js#L128-L152

[This other existing test case here](https://github.com/mozilla/web-ext/blob/5492a0bdbbf210348967158aa7c80e9fb032593c/tests/unit/test-util/test.adb.js#L1096-L1130) may be helpful to figure out how you could simulate a unix socket discovery that should have looped a few times and printed the expected message in the new test case.

Take a look to the code linked above and give it a a try, but feel free to ping me in this PR if you need more pointers or help to figure out issues with creating this test (also feel free to push in this PR an incomplete test case that does still fail, that would still be helpful to me because I can more quickly look to what you are trying and how it does fail and provide more helpful pointers).

rpl (changes requested):

src/util/adb.js
sonalprabhu: Sure. I will keep the message the normal color.

rpl:
@sonalprabhu as a side note, don't worry about the audit-deps failure in the travis CI job, it is unrelated to the changes in this PR, we'll take care of that and you'll just need to rebase this PR once we merged a fix for it.

src/util/adb.js
sonalprabhu: `
it('informs the user if remote debugging is not enabled', async () => {
      const adb = getFakeADBKit({
        adbClient: {
          shell: sinon.spy(() => Promise.resolve('')),
        },
        adbkitUtil: {
          readAll: sinon.spy(() => {
            return Promise.resolve(Buffer.from(''));
          }),
        },
      });
      const adbUtils = new ADBUtils({adb});

      const maxDiscoveryTime = 50;
      const retryInterval = 10;
      const msg = (
        'Waiting for org.mozilla.firefox_mybuild Remote Debugging Server...' +
        '\nMake sure to enable "Remote Debugging via USB" ' +
        'from Settings -> Developer Tools if it is not yet enabled.'
      );

      consoleStream.stopCapturing();
      consoleStream.startCapturing();
      consoleStream.makeVerbose();

      const promise = adbUtils.discoverRDPUnixSocket(
        'device1', 'org.mozilla.firefox_mybuild', {
          maxDiscoveryTime,
          retryInterval,
        }
      );
      const {capturedMessages} = consoleStream;
      consoleStream.stopCapturing();

      assert.ok(capturedMessages.some(
        (message) => (
          message.match('[debug]') &&
          message.match(msg)
        )));
      await assert.isRejected(promise, WebExtError);
    });
`
@rpl 
I have tried capturing messages from log and checking them here. Does this test case fulfill the required purpose or am I making some mistakes here?

rpl: @sonalprabhu from a quick look I think that it is a good start, it may require just some tweaks, e.g.:
- it may be better to stop capturing the messages after `await assert.isRejected(...)`, just to be sure that the `adb.discoverRDPUnixSocket(...)` async call did complete
- Nit, it may be enough to match just a fragment of the logged message (e.g. just `Make sure to enable "Remote Debugging via USB"` the part), it should be enough to cover the fix and we may not need to update the test if we do apply some small tweaks to the logged message (which would always be required if we do assert the entire message)

would you mind to update this PR so that I can see it running in the CI job? (and be able to also give this PR a try locally by just pulling your pull request as is, without having to add the test manually into the test files).

would you mind to also rebase the PR to the current main branch? the previous unrelated CI job failure has been fixed in the main branch.

sonalprabhu: Sure, will do that.

tests/unit/test-util/test.adb.js
```diff
@@ -1129,6 +1131,44 @@ describe('utils/adb', () => {
...
+      });
+      const adbUtils = new ADBUtils({adb});
+
+      const maxDiscoveryTime = 50;
+      const retryInterval = 10;
+
+      consoleStream.stopCapturing();
+      consoleStream.flushCapturedLogs();
+      consoleStream.startCapturing();
+      consoleStream.makeVerbose();
```
rpl, line 1153: I think that the following tweaked version should be enough and slightly better:
```
      consoleStream.flushCapturedLogs(); <--- clear the captured logs
      consoleStream.makeVerbose(); <--- make the log verbose (to be able to assert the log level)
      consoleStream.startCapturing(); <--- make sure that we start capturing logs
      ...
      const promise = ... <-- get the promise
      await assert.isRejected(promise, WebExtError);  <--- wait the async function to complete
      ... // get the message that we expect by filtering the captured logs
      consoleStream.stopCapturing(); <--- stop capturing we don't need it anymore
      ... // assertions on the expected message and log level
```

tests/unit/test-util/test.adb.js
```diff
@@ -1129,6 +1131,44 @@ describe('utils/adb', () => {
...
+          readAll: sinon.spy(() => {
+            return Promise.resolve(Buffer.from(''));
+          }),
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+
+      const maxDiscoveryTime = 50;
+      const retryInterval = 10;
+
```
rpl, line 1149: Nit, let's remove this two consts and just pass the value in the call to `adb.discoverRDPUnixSocket` (we are not using them for any assertion and so we can make this test a bit shorter by making this small optional change).

tests/unit/test-util/test.adb.js
```diff
@@ -1129,6 +1131,44 @@ describe('utils/adb', () => {
...
+        'device1', 'org.mozilla.firefox_mybuild', {
+          maxDiscoveryTime,
+          retryInterval,
+        }
+      );
+      const {capturedMessages} = consoleStream;
+
+      assert.ok(capturedMessages.some(
+        (message) => (
+          message.match('[debug]') &&
```
rpl, line 1165: Surprisingly this assertion doesn't seem to be fully working, the log we are expecting is emitted with `log.info` and so this test should actually fail and pass if we use '[info]' instead.

Another thing that we may also take care of is to make it clear when this test fails if the message wasn't found or wasn't emitted with the right log level, at the moment the two conditions are checked by a single assertion and so a failure would make it immediately clear.

Let's change this to get the message that we want using something like `const foundMessage = consoleStream.caputedMessages.find((message) => message.includes(expectedMessage))`, then check the two conditions in separate assertions:

```
      assert.ok(foundMessage);
      assert.ok(foundMessage &&
        foundMessage.includes(expectedLogLevel));
```

(NOTE: the `foundMessage && foundMessage.includes` is mainly needed to make the flow type checker happy ;-))

rpl (changes requested):
@sonalprabhu good job, this looks almost ready, there are just a few more tweaks needed (and a small fix to the test case, which is currently not failing even if it is expecting the wrong log level).

rpl:
@sonalprabhu the last CI jobs did fail because of some trailing spaces in test.adb.js (added when you added back the empty line that was previously there):

- https://travis-ci.org/github/mozilla/web-ext/jobs/734232572#L246

 You can fix that issue manually (by removing the trailing spaces and leave it as just an empty line as it was originally) or by running eslint --fix on that test file.

sonalprabhu:
Can I create a new pull request? I am facing some problems with this branch

rpl:
> Can I create a new pull request? I am facing some problems with this branch

it would be better to don't create an new one, I can help you to figure out how to handle the issue (if you have still any issue on this branch), describe me what is the issue that you are facing and I'll look into it and provide you more hints on how to deal with it.

sonalprabhu:
I cleared the issues with that.

rpl (approved):

rpl:
@sonalprabhu looks great! I just approved this version of the patch :tada: 
Thanks a lot for taking care of this issue!

sonalprabhu:
> @sonalprabhu looks great! I just approved this version of the patch 🎉
> Thanks a lot for taking care of this issue!

Thanks a lot for your detailed guidelines :)

caitmuenster:
Thanks so much for the patch, @sonalprabhu! 🙌 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Welcome onboard! We look forward to seeing you around the project.

PR #2039, opened 2020-10-02
fix(deps): Switch from adbkit to @devicefarmer/adbkit npm package

### LINKED ISSUE #2025: Replace adbkit dependency
Feature request

Currently [adbkit](https://github.com/openstf/adbkit) is a dependency of web-ext. This project is [not maintained any more](https://github.com/openstf/adbkit#this-project-along-with-other-ones-in-openstf-organisation-is-provided-as-is-for-community-without-active-development). Adbkits dependencies will never get updated.

Currently node-forge, one of adbkits dependencies, has a [vulnerability](https://snyk.io/vuln/SNYK-JS-NODEFORGE-598677).

To solve this and future problems, adbkit should be replaced.

rpl:
Fixes #2025

Some notes from some additional checks I did:

- The developer working on the replacement package @devicefarmer/adbkit is also part of the organization under which the adbkit package was developed, and the new @devicefarmer/adbkit seems to be actively maintained as discussed in the issue linked above.

- I also looked to the diff between the 3 adbkit/adbkit-* packages we are currently using and the 3 related package under the new devicefarmer github organization (both in their original coffeescript sources and the resulting transpiled js code) and the changes are minimal, matching what we would expect from a patch release, and nothing that should change the behavior of the code that is already using the adbkit package in web-ext.

- To be totally fair, there is some small changes around the adb connection object that at a first glance may trigger some slightly different behaviors in case of connection errors, but I do have tested `web-ext run` on a real android device and tried the behavior on unexpected disconnection of the device and the behavior seems to be the same (there is some fixes that we should apply around that, but it isn't a blocker for this pull request, I'll file it as a follow up).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3547,6 +3547,40 @@
       "integrity": "sha512-VoNqai1vR5anRF5Tuh/+SWDFk7xi7oMwHrHrbm1BprYXjB2RJsWLhUrStMssDxEl5lW/z3EUdg8RvH/IUBccSQ==",
       "dev": true
     },
+    "@devicefarmer/adbkit": {
+      "version": "2.11.3",
+      "resolved": "https://registry.npmjs.org/@devicefarmer/adbkit/-/adbkit-2.11.3.tgz",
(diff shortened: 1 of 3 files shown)


Rob--W (approved):

Standard8:
@Rob--W would it be possible to get a new release with this in, to help make it easier to get the node-forge update?

rpl:
@Standard8 this fix has been just released on npm as part of web-ext v5.2.0.

Standard8:
Thank you!

PR #2053, opened 2020-10-19
fix: defaultReloadStrategy should catch extension reloading errors

rpl:
This PR does apply a small change to the `defaultReloadStrategy` helper function to make sure that we do catch the rejection that `extensionRunner.reloadAllExtensions may have triggered (and then log the error as a warning, and the error stack as an additional debug level log).

The new behavior is going to:
- fix the new failures triggered by the related test case on mocha 8.2.0 (and unblock #2052 for merging, currently this issue triggers failures on the CI)
- prevent newer nodejs version from exiting the process because of an unhandled rejection (which in the current nodejs versions is just logged on the console, but in future nodejs version this is meant to make the entire nodejs process to exit with error)

### DIFF
--- a/tests/unit/test-extension-runners/test.extension-runners.js
+++ b/tests/unit/test-extension-runners/test.extension-runners.js
@@ -536,7 +536,7 @@ describe('util/extension-runners', () => {
          const {extensionRunner, reloadStrategy} = prepare({
            stubExtensionRunner: {
              reloadAllExtensions: sinon.spy(
-               () => Promise.reject(Error('fake reload error'))
+               () => Promise.reject(new Error('fake reload error'))
              ),
            },
          });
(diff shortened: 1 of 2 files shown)


tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -536,28 +536,46 @@ describe('util/extension-runners', () => {
...
          sinon.spy(fakeStdin, 'setRawMode');
 
+         // Stub the `fakeStdin.once` method to be able to wait
+         // once a promise resolved when the reloadStrategy method
+         // did call `stdin.once('keypress', ...)`.
+         const fakeStdinOnce = fakeStdin.once;
+         sinon.stub(fakeStdin, 'once');
+         fakeStdin.once.callsFake((...args) => {
+           if (args[0] === 'keypress') {
+             fakeStdin.emit('once-keypress-called');
```
Rob--W, line 554: Rather than misusing the event emitter interface as a side effect, it might be more obvious to store the `resolve` function in a temporary variable, and call it from here.

Rob--W (approved):
Looks good, just one nit that you may choose to ignore.

PR #2054, opened 2020-10-19
chore: Fix on flow type signature to pass type checking on flow >= v0.132.0

rpl:
This PR contains a number of changes needed to pass the flowtype checks on flow >= v0.132.0 (see #2003).

Most of the changes are only applied to the flow type signatures, but for some of the flowtype errors I opted to make some small changes on the implementation side (when it did look more appropriate).

Many of the new flowtype checks errors in #2003 are related to the dependencies, it looks that more recent flow versions are detecting when the types used are coming from a dependency that isn't typed using flow and the resulting types are all going to degrate to `any`.

Unfortunately the npm modules that are triggering these flowtype errors do not have their own flowtype declarations, and there isn't a third party one in the flow-types repository neither.

The easiest approach seems to be to include our own flow-typed declarations files in the project repository.

These declaration files are only including the subset of the modules that is actually being used in the web-ext sources, everything else is currently omitted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "README.md",
     "dist/*.js",
     "src/**",
-    "index.mjs"
+    "index.mjs",
+    "flow-typed/*.js"
   ],
   "engines": {
     "node": ">=10.0.0",
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     "README.md",
     "dist/*.js",
     "src/**",
-    "index.mjs",
-    "flow-typed/*.js"
+    "index.mjs"
   ],
   "engines": {
     "node": ">=10.0.0",
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -43,7 +43,7 @@ const EXCLUDED_CHROME_FLAGS = [
   '--mute-audio',
 ];
 
-export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags()
+export const DEFAULT_CHROME_FLAGS: Array<string> = ChromeLauncher.defaultFlags()
   .filter((flag) => !EXCLUDED_CHROME_FLAGS.includes(flag));
 
 /**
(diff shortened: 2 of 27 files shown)


package.json
```diff
@@ -12,7 +12,8 @@
     "README.md",
     "dist/*.js",
     "src/**",
-    "index.mjs"
+    "index.mjs",
+    "flow-typed/*.js"
```
Rob--W, line 16: Why is this exported?

src/extension-runners/chromium.js
```diff
@@ -307,6 +307,10 @@ export class ChromiumExtensionRunner {
       })
     );
 
+    if (!this.wss) {
+      throw new Error('WebSocketServer closed');
```
Rob--W, line 311: What's this doing here? There will be a runtime error at `this.wss.address()` below, anyway.

Removing this would reduce the need for extra test coverage. (but since you have test coverage anyway, perhaps tweak the error message if you care about the scenario)

src/extension-runners/firefox-android.js
```diff
@@ -89,9 +89,9 @@ export type FirefoxAndroidExtensionRunnerParams = {|
  */
 export class FirefoxAndroidExtensionRunner {
   // Wait 3s before the next unix socket discovery loop.
-  static unixSocketDiscoveryRetryInterval = 3 * 1000;
+  static unixSocketDiscoveryRetryInterval: number = 3 * 1000;
```
Rob--W, line 92: Is Flow really that dumb to not be able to infer the type from this declaration?

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -624,6 +632,9 @@ describe('util/extension-runners/chromium', async () => {
     });
 
     const connectClient = async () => {
+      if (!runnerInstance.wss) {
```
Rob--W, line 635: Similarly there is going to be a runtime error anyway at `runnerInstance.wss.address()`. Does this help with anything?

Rob--W (commented):
Does this mean that we have to maintain these type definitions whenever the library updates?
Would Flow complain if the external libraries' actual types change without us updating these types?

What's the benefit of adding these types over typing as any?

package.json
rpl: agreed with @Rob--W to don't export the custom flow definitions for the third party dependencies used internally for now, we can include them if any other package depends from web-ext and would need them.

src/extension-runners/chromium.js
rpl: I opted to replace this with a $FlowIgnore suppress comment.

tests/unit/test-extension-runners/test.chromium.js
rpl: I opted to replace this with a $FlowIgnore suppress comment.

src/extension-runners/firefox-android.js
rpl: as discussed in the triage meeting: I didn't dig too much, but I got the feeling that flow did enforce explicitly type annotations (instead of just trust what type did flow infer) on what it is being exported explicitly from a module (which to me does make kind of sense: "infer the type used internally as much as possible, enforce explicit type annotation of what is exported outside of a module").

Issue #1911, opened 2020-05-20
When R is pressed `web-ext run throws Error: Unexpected packet from actor root{"addons..


Rob--W:
### Is this a feature request or a bug?
Bug.

### What is the current behavior?

Start `web-ext run` with any extension and keep R pressed. Soon `web-ext` throws and exits, with the following log:

```
Last extension reload: 19:53:50 GMT+0200 (Central European Summer Time)
/path/to/node_modules/web-ext/node_modules/@cliqz-oss/firefox-client/lib/client.js:174
      throw new Error("Unexpected packet from actor " +  message.from
            ^
Error: Unexpected packet from actor root{"addons":[{"actor":"server1.conn0.webExtensionDescriptor7","debuggable":true, ...... "actor":"server1.conn0.webExtensionDescriptor24","id":"gmp-gmpopenh264","manifestURL":null,"name":"OpenH264 Video Codec provided by Cisco Systems, Inc.","traits":{"isDescriptor":true},"type":"plugin","warnings":[]}],"from":"root"}
    at Client.handleMessage (/path/to/node_modules/web-ext/node_modules/@cliqz-oss/firefox-client/lib/client.js:174:13)
    at Client.readMessage (/path/to/node_modules/web-ext/node_modules/@cliqz-oss/firefox-client/lib/client.js:221:10)
    at Client.onData (/path/to/node_modules/web-ext/node_modules/@cliqz-oss/firefox-client/lib/client.js:187:16)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:302:12)
    at readableAddChunk (_stream_readable.js:278:9)
    at Socket.Readable.push (_stream_readable.js:217:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23)

```

I can reproduce with a minimal extension, i.e. an empty directory whose `manifest.json` consists of just `{"manifest_version":2,"version":"1","name":"x"}`

### What is the expected or desired behavior?
`web-ext` should not throw / crash.

### Version information (for bug reports)

node v14.2.0
npm 6.14.4
web-ext 4.2.0
Firefox Beta 77.0b6
ArchLinux

Yrkan:
Just tested with the latest version of web-ext and it seems to be already **fixed**.

PR #2061, opened 2020-10-31
Fix R key press cause web-ext to crash

thamimemel:
Fixes :  #1911
I reproduced the error unintentionally and found that there is an open issue for it, this patch fixes the bug but needs review for any side effects, or maybe code refactoring.
The idea is to keep track of the reloading process in `this.reloading` and allow reloading only if it is false.
**IMPORTANT: please check if this works properly in android, i only tested in desktop.**


Rob--W (changes requested):
Instead of adding booleans in multiple places, I suggest to rename `reloadAddon` to something else, like `reloadAddonNow`, and introduce a new `reloadAddon` method that temporarily sets a boolean at the entry of the function, calls `reloadAddonNow` in a try-finally, and resets the boolean at the end.

To make sure that an add-on is actually reloaded when the reload is requested while the add-on is loading, `reloadAddon` should be called recursively if the boolean was set.

And please add unit tests to verify that the functionality works as intended.

willdurand:
No activity for years, I am closing this PR. If you want to work on this issue in the future, just submit a new PR, thanks!

PR #2064, opened 2020-11-03
feat: Added --profile-create-if-missing option

### LINKED ISSUE #2058: Create profile directory if it doesn't exist and `--profile-create-new` has been combined to the `-p` cli option
Feature request

Today, if I pass a profile that does not exist, the `web-ext run` command fails:

```console
> web-ext run --source-dir ./dist/ -p extension_development
Running web extension from /Users/a70e7cc5/code/proj/dist

o: Could not copy Firefox profile from extension_development: Error: cannot find profile extension_development
    at Module.C (/Users/a70e7cc5/code/proj/node_modules/web-ext/dist/web-ext.js:1:22412)
```

It would be nice if there was a flag to create the profile if it doesn't exist. Even better, it would be nice to be able to pass a directory to create the profile in.

I'm able to work around this in my launch script:

```ts
import webExt from 'web-ext';
import { default as fxRunner } from 'fx-runner';

async function launchFirefox(browserProfileDirectory: string) {
  await createFirefoxProfileIfNotExists(browserProfileDirectory);
  await webExt.cmd.run({
    sourceDir: BUILD_DIR,
    firefoxProfile: browserProfileDirectory,
    keepProfileChanges: true,
    startUrl: ['about:debugging#/runtime/this-firefox'],
  });
}

async function createFirefoxProfileIfNotExists(browserProfileDirectory: string): Promise<void> {
  const devProfileName = 'extension_development';

  if (!fs.existsSync(browserProfileDirectory)) {
    console.info(`${devProfileName} Firefox profile does not exist. Creating...`);

    // The `-CreateProfile` flag creates a new Firefox profile stub and immediately exits
    const createProfile = await fxRunner({
      'no-remote': true,
      'binary-args': ['-CreateProfile', `${devProfileName} ${browserProfileDirectory}`],
    });
    const createProfileProc = createProfile.process;
      createProfileProc.on('close', resolve);
      createProfileProc.on('error', reject);
    });
  }
}
```

* **Firefox version**: 83.0b4
* **Your OS and version**: MacOS 10.15.7

````
> node --version && npm --version && web-ext --version
v12.18.4
6.14.6
5.3.0
````

### PR DESCRIPTION
**Pull request by akhilpanchal**
Fixes #2058

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -129,7 +129,7 @@ export default async function run(
       log.info(`Profile directory ${profileDir} already exists`);
     } else {
       log.info(`Profile directory not found. Creating directory ${profileDir}`);
-      fs.mkdirSync(profileDir);
+      await fs.mkdir(profileDir);
     }
   }
 
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -10,7 +10,6 @@ import {
   connectWithMaxRetries as defaultFirefoxClient,
 } from '../firefox/remote';
 import {createLogger} from '../util/logger';
-import {WebExtError} from '../errors';
 import defaultGetValidatedManifest from '../util/manifest';
 import {
   createExtensionRunner,
@@ -31,7 +30,7 @@ export type CmdRunParams = {|
   pref?: FirefoxPreferences,
   firefox: string,
   firefoxProfile?: string,
-  profileCreateNew?: boolean,
+  profileCreateIfMissing?: boolean,
   ignoreFiles?: Array<string>,
   keepProfileChanges: boolean,
   noInput?: boolean,
@@ -76,7 +75,7 @@ export default async function run(
     pref,
     firefox,
     firefoxProfile,
-    profileCreateNew,
+    profileCreateIfMissing,
     keepProfileChanges = false,
     ignoreFiles,
     noInput = false,
@@ -124,10 +123,10 @@ export default async function run(
 
   const profileDir = firefoxProfile || chromiumProfile;
 
-  if (profileCreateNew && profileDir) {
+  if (profileCreateIfMissing && profileDir) {
     const isDir = fs.existsSync(profileDir);
     if (isDir) {
-      throw new WebExtError(`Directory ${profileDir} already exists.`);
+      log.info(`Profile directory ${profileDir} already exists`);
     } else {
       log.info(`Profile directory not found. Creating directory ${profileDir}`);
       fs.mkdirSync(profileDir);
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -1,4 +1,6 @@
 /* @flow */
+import { fs } from 'mz';
+
 import defaultBuildExtension from './build';
 import {
   showDesktopNotification as defaultDesktopNotifications,
@@ -8,6 +10,7 @@ import {
   connectWithMaxRetries as defaultFirefoxClient,
 } from '../firefox/remote';
 import {createLogger} from '../util/logger';
+import {WebExtError} from '../errors';
 import defaultGetValidatedManifest from '../util/manifest';
 import {
   createExtensionRunner,
@@ -28,6 +31,7 @@ export type CmdRunParams = {|
   pref?: FirefoxPreferences,
   firefox: string,
   firefoxProfile?: string,
+  profileCreateNew?: boolean,
   ignoreFiles?: Array<string>,
   keepProfileChanges: boolean,
   noInput?: boolean,
@@ -72,6 +76,7 @@ export default async function run(
     pref,
     firefox,
     firefoxProfile,
+    profileCreateNew,
     keepProfileChanges = false,
     ignoreFiles,
     noInput = false,
@@ -116,6 +121,18 @@ export default async function run(
   // object containing one or more preferences.
   const customPrefs = pref;
   const manifestData = await getValidatedManifest(sourceDir);
+
+  const profileDir = firefoxProfile || chromiumProfile;
+
+  if (profileCreateNew && profileDir) {
+    const isDir = fs.existsSync(profileDir);
+    if (isDir) {
+      throw new WebExtError(`Directory ${profileDir} already exists.`);
+    } else {
+      log.info(`Profile directory not found. Creating directory ${profileDir}`);
+      fs.mkdirSync(profileDir);
+    }
+  }
 
   const runners = [];
 
--- a/src/program.js
+++ b/src/program.js
@@ -581,9 +581,9 @@ Example: $0 --help run.
         type: 'string',
       },
       'profile-create-new': {
-        describe: 'Create a new profile in the directory specified by the ' +
-                  '-p option if it doesn\'t already exist. If it already ' +
-                  'exists, an error will be shown.',
+        describe: 'Create directory if the profile directory specified by ' +
+                  'the -p option doesn\'t already exist. If it already ' +
+                  'exists, an error will be thrown.',
         demandOption: false,
         type: 'boolean',
       },
--- a/src/program.js
+++ b/src/program.js
@@ -580,6 +580,13 @@ Example: $0 --help run.
         demandOption: false,
         type: 'string',
       },
+      'profile-create-new': {
+        describe: 'Create a new profile in the directory specified by the ' +
+                  '-p option if it doesn\'t already exist. If it already ' +
+                  'exists, an error will be shown.',
+        demandOption: false,
+        type: 'boolean',
+      },
       'keep-profile-changes': {
         describe: 'Run Firefox directly in custom profile. Any changes to ' +
                   'the profile will be saved.',
--- a/src/program.js
+++ b/src/program.js
@@ -580,10 +580,8 @@ Example: $0 --help run.
         demandOption: false,
         type: 'string',
       },
-      'profile-create-new': {
-        describe: 'Create directory if the profile directory specified by ' +
-                  'the -p option doesn\'t already exist. If it already ' +
-                  'exists, an error will be thrown.',
+      'profile-create-if-missing': {
+        describe: 'Create the profile directory if it does not already exist',
         demandOption: false,
         type: 'boolean',
       },
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -1,6 +1,7 @@
 /* @flow */
 import path from 'path';
 
+import { fs } from 'mz';
 import {afterEach, beforeEach, describe, it} from 'mocha';
 import {assert} from 'chai';
 import sinon from 'sinon';
@@ -325,5 +326,47 @@ describe('run', () => {
            sourceDir: '/fake/source/dir',
          },
        );
-     });
+     }
+  );
+
+  it('creates dir when profile doesn\'t exist using profile-create-new',
+     async () => {
+       sinon.stub(fs, 'existsSync').returns(false);
+       sinon.stub(fs, 'mkdirSync');
+
+       const firefoxProfile = '/pretend/path/to/Firefox/profile';
+       const cmd = prepareRun();
+
+       await cmd.run({
+         firefoxProfile,
+         profileCreateNew: true,
+       });
+
+       sinon.assert.calledWith(fs.mkdirSync, firefoxProfile);
+
+       fs.existsSync.restore();
+       fs.mkdirSync.restore();
+     }
+  );
+
+  it('throws error when profile dir already exists using profile-create-new',
+     async () => {
+       sinon.stub(fs, 'existsSync').returns(true);
+
+       const firefoxProfile = '/pretend/path/to/Firefox/profile';
+       const cmd = prepareRun();
+
+       const promise = cmd.run({
+         firefoxProfile,
+         profileCreateNew: true,
+       });
+
+       await assert.isRejected(
+         promise,
+         `Directory ${firefoxProfile} already exists.`
+       );
+
+       fs.existsSync.restore();
+     }
+  );
 });
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -339,78 +339,72 @@ describe('run', () => {
       fs.existsSync.restore();
     });
 
-    const fakeFirefoxProfile = '/pretend/path/to/Firefox/profile';
-    const fakeChromiumProfile = '/pretend/path/to/Chromium/profile';
+    const fakeProfile = '/pretend/path/to/profile';
 
     async function testCreateProfileIfMissing(
-      existsSyncReturnValue,
-      runParams,
-      expectedProfile
+      expectProfileExists,
+      runParams
     ) {
-      fs.existsSync.returns(existsSyncReturnValue);
+      fs.existsSync.returns(expectProfileExists);
       const cmd = prepareRun();
 
       await cmd.run(runParams);
 
-      if (existsSyncReturnValue) {
+      if (expectProfileExists) {
         sinon.assert.notCalled(fs.mkdir);
       } else {
-        sinon.assert.calledWith(fs.mkdir, expectedProfile);
+        sinon.assert.calledWith(fs.mkdir, fakeProfile);
       }
 
-      if (expectedProfile === fakeFirefoxProfile) {
+      if (runParams.target === 'chromium') {
+        sinon.assert.calledOnce(chromiumRunnerStub);
+
+        const {chromiumProfile} = chromiumRunnerStub.firstCall.args[0];
+        assert.equal(chromiumProfile, fakeProfile,
+                     'Got the expected chromiumProfile option');
+      } else {
         sinon.assert.calledOnce(desktopRunnerStub);
         const firefoxProfile = desktopRunnerStub
           .firstCall
           .args[0]
           .profilePath;
-        assert.equal(firefoxProfile, fakeFirefoxProfile,
+        assert.equal(firefoxProfile, fakeProfile,
                      'Got the expected firefoxProfile option');
-      } else {
-        sinon.assert.calledOnce(chromiumRunnerStub);
-
-        const {chromiumProfile} = chromiumRunnerStub.firstCall.args[0];
-        assert.equal(chromiumProfile, fakeChromiumProfile,
-                     'Got the expected chromiumProfile option');
       }
     }
 
     it('creates dir when firefox profile does not exist',
        async () => testCreateProfileIfMissing(
          false, {
-           firefoxProfile: fakeFirefoxProfile,
+           firefoxProfile: fakeProfile,
            profileCreateIfMissing: true,
-         }, fakeFirefoxProfile
-       )
+         })
     );
 
     it('creates dir when chromium profile does not exist',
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: tests/unit/test-cmd/test.run.js [hunk 1]
```diff
@@ -325,5 +326,47 @@ describe('run', () => {
...
+       });
+
+       await assert.isRejected(
+         promise,
+         `Directory ${firefoxProfile} already exists.`
+       );
+
+       fs.existsSync.restore();
+     }
+  );
```
akhilpanchal, line 371: Should I also add a test that covers the use case of `--chromium-profile` with `--profile-create-new`?

rpl: @akhilpanchal for the unit test I think that it does make sense to also assert the expected behavior when passing `chromiumProfile` to cmd.run instead of `firefoxProfile`.

I'm actually curious about: what is the current  `web-ext run -t chromium --chromium-profile ...` behavior (without any of these changes) when the given profile directory doesn't exist?

akhilpanchal: @rpl 
Sure. I can add more tests to cover the case for `--chromium-profile` with `--profile-create-new`.

Following is the output for the `web-ext run -t chromium --chromium-profile ...` command without any of these changes:

```sh
$ web-ext run -t chromium --chromium-profile /Users/apanchal/MyCode/extension-borderify/foobar
Running web extension from /Users/apanchal/MyCode/extension-borderify

Error: ENOENT: no such file or directory, stat '/Users/apanchal/MyCode/extension-borderify/foobar'

Error code: ENOENT
```

src/cmd/run.js
```diff
@@ -117,6 +122,18 @@ export default async function run(
   const customPrefs = pref;
   const manifestData = await getValidatedManifest(sourceDir);
 
+  const profileDir = firefoxProfile || chromiumProfile;
+
+  if (profileCreateNew && profileDir) {
+    const isDir = fs.existsSync(profileDir);
+    if (isDir) {
+      throw new WebExtError(`Directory ${profileDir} already exists.`);
```
rpl, line 130: @akhilpanchal I did re-think a bit more about this and this part of the approach that I suggested in my comment https://github.com/mozilla/web-ext/issues/2058#issuecomment-717184560 is actually not covering the use case describe in the [issue description](https://github.com/mozilla/web-ext/issues/2058#issue-729774986) in a nice way, because we will throw an error if the directory exist (while the workaround launch script in the issue description is not throwing in that case).

My apologies for missing to notice this when I wrote the initial guidance comment in the issue.

I think that relaxing this and just log a message (we may opt for a `log.info` if we want the user to see it even if web-ext run isn't in verbose mode) if the directory already exist may be actually reasonable (and it would better cover the enhancement request)

To better match the expected behavior we may also renamed the option as `--profile-create-if-missing`.

src/program.js
```diff
@@ -580,6 +580,13 @@ Example: $0 --help run.
         demandOption: false,
         type: 'string',
       },
+      'profile-create-new': {
+        describe: 'Create directory if the profile directory specified by ' +
+                  'the -p option doesn\'t already exist. If it already ' +
+                  'exists, an error will be thrown.',
```
rpl, line 586: nit: `Create the profile directory if it does not already exist.`

This version does also omit the -p option, mainly because:
- that is only the option for the Firefox profile, and this is going to cover the chromium profile cli option as well
- I don't think we need to describe that here in too much detail (and option name may actually be clear enough for the user to guess)
- we should still add a more detailed description for the option in the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop) repository, as part of the web-ext cli options doc page (and we will be happy if you want to create a followup PR in that repo to take care of that after this is landed and we are about to release it in a new web-ext version)

src/cmd/run.js
```diff
@@ -117,6 +122,18 @@ export default async function run(
   const customPrefs = pref;
   const manifestData = await getValidatedManifest(sourceDir);
 
+  const profileDir = firefoxProfile || chromiumProfile;
+
+  if (profileCreateNew && profileDir) {
+    const isDir = fs.existsSync(profileDir);
```
rpl, line 128: Given that this is already an async function and we are using the fs promise-based API exported by mz, we could easily use the non `*Sync` version of the `fs.exists` and `fs.mkdir` methods and just await on them. 

would you mind to look into making this change?

rpl (changes requested):
Thanks @akhilpanchal this PR looks in a very good shape, good job!

I'm requesting just a couple of small changes, one is actually due to my mistake in the guidance comment (I missed to notice that one of the behaviors I suggested is actually kind of conflicting with the use case described in the enhancement request), the rest should be just small and easy nits.

Let me know if any of my comments below are unclear (or also if you disagree with any of my points) and I'll be happy to discuss about those before you plan the changes, or ping me as soon this PR is ready for another (likely its last) review round.

src/cmd/run.js
akhilpanchal: I completely agree with you about not throwing and just logging an info message that tells the user that the directory already exists. Will update this behavior and also update the option name.

src/program.js
akhilpanchal: Sure. That sounds good! 👍

src/cmd/run.js
akhilpanchal: @rpl 
I spent some time looking into making these calls async.
The async version of `fs.existsSync` (`fs.exists`) is deprecated and to make this call `async`, we will have to either use `fs.stat` or `fs.access` inside of a try catch block (to check for the not found exception and only then create the directory) that might make this code a little more verbose than it needs to be.

We could also use the [path-exists library](https://github.com/sindresorhus/path-exists) that might alleviate the need to have a try catch block.

Do you have any other thoughts?

rpl: @akhilpanchal thanks! that's a very nice research work, I really appreciated that!
That note in the path-exists readme was also a pretty interesting read (the part that warns about the risks of race conditions, in this particular case I don't think we need to worry about that, but it still a good point to keep in mind in general).

Personally I think that we can spare us to add a dependency and just go for `fs.existsSync` for checking if the directory exists and just prefer the async method for the `fs.mkdir` part, does that sound reasonable for you too?

akhilpanchal: Thanks @rpl. Appreciate that :)
Yes, I agree with you that we can spare ourselves from adding a new dependency and just use the async `fs.mkdir` call and keep the sync `fs.existsSync`.
Sending an update to the PR in a few moments.

tests/unit/test-cmd/test.run.js
```diff
@@ -325,5 +326,101 @@ describe('run', () => {
...
+           profileCreateIfMissing: true,
+         });
+
+         sinon.assert.calledWith(fs.mkdir, firefoxProfile);
+
+         fs.existsSync.restore();
+       }
+    );
+
+    it('creates dir when chromium profile doesn\'t exist',
```
rpl, line 358: Nitpick similar to the previous one: `creates a chromium profile directory if it does not exist`

tests/unit/test-cmd/test.run.js
```diff
@@ -325,5 +326,101 @@ describe('run', () => {
...
+           chromiumProfile,
+           target: 'chromium',
+           profileCreateIfMissing: true,
+         });
+
+         sinon.assert.calledWith(fs.mkdir, chromiumProfile);
+
+         fs.existsSync.restore();
+       }
+    );
```
rpl, line 374: These two tests are so similar to each other that it may be reasonable to refactor it into a single parametrized test, e.g. something like:

```
async function testCreateProfileIfMissing(runParams, expectedMkdirParam) {
  sinon.stub(fs, 'existsSync').returns(false);
  const cmd = prepareRun();
  await cmd.run({ ...runParams, promiseCreateIfMissing: true });
  sinon.assert.calledWith(fs.mkdir, fs.mkdir, expectedMkdirParam);
}

const fakeProfilePath = '/pretend/profile/path'

it('creates a Firefox profile directory if it does not exist',
   testCreateProfileIfMissing.bind({ firefoxProfile: fakeProfielPath }, fakeProfilePath));

it('creates a Chromium profile directory if it does not exist',
   testCreateProfileIfMissing.bind(
     { chromiumProfile: fakeProfielPath, target: 'chromium' },
     fakeProfilePath));
```

does that sounds reasonable to you too?

tests/unit/test-cmd/test.run.js
```diff
@@ -325,5 +326,101 @@ describe('run', () => {
...
+
+         await cmd.run({
+           chromiumProfile,
+           target: 'chromium',
+           profileCreateIfMissing: true,
+         });
+
+         sinon.assert.calledWith(fs.mkdir, chromiumProfile);
+
+         fs.existsSync.restore();
```
rpl, line 372: I think that we should restore fs.existsSync in the afterEach (and stub it in beforeEach, eventually changing what it does return in the particular test if necessary) as for the other stub, in case one test case fails and `fs.existsSync.restore()` isn't reached.

tests/unit/test-cmd/test.run.js
```diff
@@ -325,5 +326,101 @@ describe('run', () => {
...
+       }
+    );
+
+    it('Uses profile directory when firefox profile directory already exists',
```
rpl, line 376: Nitpicks:
- `Uses` should be in lowercase, we may also tweak it into `uses the give firefox profile if it does exist`.
- we may want to assert that fs.mkdir was not called in this case
- stub and restore fs.existsSync in beforeEach and afterEach for the same reason described in my previous comment

Another thing that I've been thinking of is:
could these other two tests be refactored in a single parametrized test function as for the other two?

tests/unit/test-cmd/test.run.js
```diff
@@ -325,5 +326,101 @@ describe('run', () => {
...
+
+         const {chromiumProfile} = chromiumRunnerStub.firstCall.args[0];
+         assert.equal(chromiumProfile, fakeChromiumProfile,
+                      'Got the expected chromiumProfile option');
+
+         fs.existsSync.restore();
+       }
+    );
+  });
+});
```
rpl, line 426: Nitpick, add the newline here

rpl (changes requested):
@akhilpanchal this looks almost there, follows just some nitpicks related to the test cases.

Feel free to ping me if there is anything unclear (or if I was misreading something and my comment may be wrong).

Thanks again for working on this enhancement!

tests/unit/test-cmd/test.run.js
akhilpanchal: I definitely agree there is quite a bit of duplication in the tests. I actually feel even the last 2 tests are very similar.
How do you feel about going a step further:

```js
async function testCreateProfileIfMissing(existsSyncReturnValue, runParams, expectedMkdirParam) {
  sinon.stub(fs, 'existsSync').returns(existsSyncReturnValue);
  const cmd = prepareRun();
  await cmd.run({ ...runParams, promiseCreateIfMissing: true });
  sinon.assert.calledWith(fs.mkdir, fs.mkdir, expectedMkdirParam);
}

const fakeProfilePath = '/pretend/profile/path'

it('creates a Firefox profile directory if it does not exist',
   testCreateProfileIfMissing.bind(false, { firefoxProfile: fakeProfielPath }, fakeProfilePath));

it('creates a Chromium profile directory if it does not exist',
   testCreateProfileIfMissing.bind(
     false,
     { chromiumProfile: fakeProfielPath, target: 'chromium' },
     fakeProfilePath));

it('Uses profile directory when firefox profile directory already exists',
   testCreateProfileIfMissing.bind(true, { firefoxProfile: fakeProfielPath }, fakeProfilePath));

it('Uses profile directory when chromium profile directory already exists',
   testCreateProfileIfMissing.bind(
     true,
     { chromiumProfile: fakeProfielPath, target: 'chromium' },
     fakeProfilePath));
```

rpl: @akhilpanchal sure thing, if all four tests can be expressed with a single parametrized function and still be readable enough (so that we will not have to spent to much time to decode a failure, if any of these test would fail at some point in the future).

I think that for the last 2 test cases the assertion `sinon.assert.calledWith(fs.mkdir, ...)` should actually be replaced by one that assert that mkdir has not been called at all.

(I guess that if we may also assert in all 4 tests that we did run the extension runner with the expected profile dir as the last 2 tests seems to be doing in this version).

akhilpanchal: @rpl 
Thanks for all your feedback 👍.
I addressed all your comments in the latest commit.

I did abstract the common test code into a single function.
Does the implementation look reasonable to you?
If not, I can either split them out to 2 different parameterized tests like you suggested earlier, OR any other way you feel that makes the tests more readable and easier to maintain.
Let me know.

akhilpanchal:
@rpl 
Some questions that I had related to the project setup:
I am using VSCode with ESLint and Prettier Plugins and looks like the auto formatting isn't working for me for this project. So I had to manually fix things like indentation errors.
* Which Editor do you use?
* Do you rely on any plugins to do auto-fix formatting and ESLint errors OR do you use the ESLint CLI?

tests/unit/test-cmd/test.run.js
```diff
@@ -326,101 +326,91 @@ describe('run', () => {
...
+
+      await cmd.run(runParams);
+
+      if (existsSyncReturnValue) {
+        sinon.assert.notCalled(fs.mkdir);
+      } else {
+        sinon.assert.calledWith(fs.mkdir, expectedProfile);
+      }
+
+      if (expectedProfile === fakeFirefoxProfile) {
```
rpl, line 361: another option to decide which extension runner we should run the assertion on would be to look to `runParams.target` and based on that (if it is == to `chromium`) run the assertion on the chromiumRunnerStub.

and by doing that I think that we would not need to use a separate fake profile dir for the two of them and we could just use one for both.


how that sounds?

tests/unit/test-cmd/test.run.js
```diff
@@ -326,101 +326,91 @@ describe('run', () => {
...
+           firefoxProfile: fakeFirefoxProfile,
            profileCreateIfMissing: true,
-         });
-
-         sinon.assert.calledWith(fs.mkdir, firefoxProfile);
-
-         fs.existsSync.restore();
-       }
+         }, fakeFirefoxProfile
+       )
```
rpl, line 384: Nitpick:  would you mind to move this `)` to the previous line and keep the parameters on separate lines?
I mean something like:
```js
testCreateProfileIfMissing(
  false,
  {
    ...
  },
  fakeFirefoxProfile);
```

as it is more or less the convention we do use at the moment, same for other calls that follow.

tests/unit/test-cmd/test.run.js
```diff
@@ -326,101 +326,91 @@ describe('run', () => {
...
-         const firefoxProfile = '/pretend/path/to/Firefox/profile';
-         const cmd = prepareRun();
-
-         await cmd.run({
-           firefoxProfile,
+    const fakeFirefoxProfile = '/pretend/path/to/Firefox/profile';
+    const fakeChromiumProfile = '/pretend/path/to/Chromium/profile';
+
+    async function testCreateProfileIfMissing(
+      existsSyncReturnValue,
```
rpl, line 346: Nitpick: we may call this parameter as `expectProfileExists` and then use it for `fs.existsSync` and the if..else at line 355-359 (just because I prefer the name to rapresent the conditions of the tests instead of the name of the stubbed function that will return that value).

src/cmd/run.js
```diff
@@ -117,6 +121,18 @@ export default async function run(
   const customPrefs = pref;
   const manifestData = await getValidatedManifest(sourceDir);
 
+  const profileDir = firefoxProfile || chromiumProfile;
+
+  if (profileCreateIfMissing && profileDir) {
```
rpl, line 126: while I was looking this part one more time I did notice another corner case that may be worth to consider (and eventually handle in a follow up):

if the user does call web-ext run with --profile-create-if-missing but without any profile parameter we currently just ignore the --profile-create-if-missing, it isn't strictly wrong and so it is not really a bit deal, but another option would be to throw an explicit `UsageError` to make the user aware that --profile-create-if-missing should be paired with --firefox-profile or --chromium-profile).

what do you think?

as I did mention it is not a really big deal and so it seems ok to defer it to a separate follow up.

rpl (approved):
@akhilpanchal this looks pretty great, thank you so much for refactoring the 4 test cases into a single parametrized function, I really appreciate that.

The following comments are related to some additional nits (still only on the test case) that it would be nice to apply before landing this (but I'm also approving the PR because they are just small nitpicks nothing really that much critical, and one is a note about a corner case that we may want to cover explicitly but it can be evaluated and then covered as a separate followup).

src/cmd/run.js
akhilpanchal: Sure. That does make sense. I can do that in a follow up PR. Should we be creating a separate issue for that? Or should I just send another PR after this one is merged?

rpl: > Sure. That does make sense. I can do that in a follow up PR. Should we be creating a separate issue for that? Or should I just send another PR after this one is merged?

@akhilpanchal thanks! just a separate PR for that part would be fine

in this one let's just cover the nits on the test, and after that I'll just merge it and you can open the follow up.

rpl:
> @rpl
> Some questions that I had related to the project setup:
> I am using VSCode with ESLint and Prettier Plugins and looks like the auto formatting isn't working for me for this project. So I had to manually fix things like indentation errors.
> 
>     * Which Editor do you use?
> 
>     * Do you rely on any plugins to do auto-fix formatting and ESLint errors OR do you use the ESLint CLI?

I totally missed this comment, my apologies.

Personally I use neovim (VSCode is a pretty nice editor, and way more friendly then vim, but I'm used to have a pretty "terminal + zsh + tmux"-based workflow and I like to use an editor that is running in the tmux session related to that project or task), this project doesn't use prettier at the moment and so some conventions are actually enforced by eslint and others are just preferred by us for consistency (but eslint is not going to complain nor format automatically the one that are not enforced by eslint rules).

tests/unit/test-cmd/test.run.js
akhilpanchal: Sounds perfect. 👍

src/cmd/run.js
akhilpanchal: @rpl 
I have addressed all the nits. Please let me know if I missed something.
Once this is merged, I will create a follow up for this enhancement.
Also, you mentioned making changes to the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop) repo. Should that be also linked to an issue in that repo?

rpl:
> @rpl
> I have addressed all the nits. Please let me know if I missed something.
> Once this is merged, I will create a follow up for this enhancement.

@akhilpanchal 
looks great! I just merged it on master. :tada: 

Thanks again for contributing this enhancement!!!

> Also, you mentioned making changes to the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop) repo. Should that be also linked to an issue in that repo?

Sure, linking the web-ext issue in the extension-workshop PR description comment would be nice :+1:
Add me as a reviewer (or mention me in a comment) once you open it.

PR #2069, opened 2020-11-10
fix: Use .BrowserApp as default reference browser apk component

### LINKED ISSUE #2041: Reference Browser not supported unless --firefox-apk-component is set
Bug

Upon running the following:

```
web-ext run -t firefox-android --android-device=DEVICEID_HERE  --firefox-apk=org.mozilla.reference.browser
```

... the Reference Browser app (https://github.com/mozilla-mobile/reference-browser) is not launched and the following error is displayed:

> Error: Activity class {org.mozilla.reference.browser/org.mozilla.reference.browser.App} does not exist.

The reference browser should be launched. As a work-around, I have to add ` --firefox-apk-component=BrowserActivity` .

To fix this, we would have to recognize the special component at https://github.com/mozilla/web-ext/blob/82c4250935bf7c5954ed2482447669628b3aaaa5/src/util/adb.js#L314-L315

````
````

> v14.13.0
6.14.7
5.2.0

### PR DESCRIPTION
**Pull request by ankushduacodes**
Fixes #2041

### CODE REVIEW

File: src/firefox/package-identifiers.js [hunk 1]
```diff
@@ -8,3 +8,7 @@ export default [
   'org.mozilla.firefox',
   'org.mozilla.reference.browser',
 ];
+
+export const defaultApkComponents = {
+  'org.mozilla.reference.browser': '.BrowserActivity',
+};
```
rpl, line 14: Nitpick: would you mind to add a new line here?

src/util/adb.js
```diff
@@ -7,7 +7,8 @@ import {
   WebExtError,
 } from '../errors';
 import {createLogger} from '../util/logger';
-import packageIdentifiers from '../firefox/package-identifiers';
+import packageIdentifiers, { defaultApkComponents }
+  from '../firefox/package-identifiers';
```
rpl, line 11: Nitpick: would you mind to reformat this as:
```
import packageIdentifiers, {
  defaultApkComponents,
} from '../firefox/package-identifiers';
```

Just to keep it consistent with the convention we used for some other multiline imports already part of the sources.

src/util/adb.js
```diff
@@ -312,7 +313,11 @@ export default class ADBUtils {
     }];
 
     if (!apkComponent) {
-      apkComponent = '.App';
+      if (defaultApkComponents[apk]) {
+        apkComponent = defaultApkComponents[apk];
+      } else {
+        apkComponent = '.App';
+      }
```
rpl, line 320: Nitpick, not that I do look to this again I think that we could unconditionally set '.App' first and then override it if we do have a specific default apk component for that apk. I mean just the following small tweak:
```
if (!apkComponent) {
  apkComponent = '.App';
  if (defaultApkComponents[apk]) {
    apkComponent = defaultApkComponents[apk];
  }
}
```

rpl (changes requested):
Thanks @ankushduacodes this looks to be going in the right direction :+1: .

Besides the small tweaks suggested below, this is also missing additional coverage for the new expected behavior (and that is also pointed out by the coveralls failure).

This can be covered by a new small unit test, a new test case in [tests/unit/test-util/test.adb.js](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.adb.js), inside the group of tests defined here: https://github.com/mozilla/web-ext/blob/6030528b713103ab9038a4ee06ad2737defef604/tests/unit/test-util/test.adb.js#L767

Taking a Look to the other tests in that group should help you to figure out how to write the new test (but feel free to ping me and ask me some additional question if you get stuck after giving it a try).

ankushduacodes:
@rpl I am getting these errors in travis-ci:
```
⧗   input: Fix reference browser apk component issue
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
```
Could you please explain what are subject and type?

rpl:
> @rpl I am getting these errors in travis-ci:
> 
> ```
> ⧗   input: Fix reference browser apk component issue
> ✖   subject may not be empty [subject-empty]
> ✖   type may not be empty [type-empty]
> ```
> 
> Could you please explain what are subject and type?

@ankushduacodes That's because the pull request title isn't using the commit message conventions that are described here (also linked right below the error):
- https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages 

It can be easily fixed by just updating the pull request title (e.g. `fix: Use .BrowserApp as as the default reference browser apk component` would be validate successfully), then the validation should pass when you update the PR again and a new CI job is going to be executed on travis.

----

Follows some more context about the reasons behind the error:

- we do use a convention for the commit message, so that we can generate the changelogs in a semi-automated way
- to prevent landing commits that do not follow the conventions, we do lint them as part of the CI job
- when there is only one commit in the pull request, we only enforce the convention on the commit message
- when there is more then one commit in a pull request, github's "squash and merge" action does use the pull request title as the default message for the generated squashed commit, and so we enforce the convention on the pull request title to avoid merging a commit that doesn't follow the convention into master and missing to notice it.

rpl:
@ankushduacodes I updated the PR title as described in https://github.com/mozilla/web-ext/pull/2069#issuecomment-724783760 (`fix` is the conventional-changelog type, everything after the `:` is the subject, and so the updated PR title does now pass the conventional changelog linting task on the CI).

The only remaining piece for this PR should now be the additional test case to cover the part that coveralls is complaining about, the additional `if (...) {}` branch added in src/utils/adb.js (the same piece briefly described https://github.com/mozilla/web-ext/pull/2069#pullrequestreview-527299408).

Let me know how it is going and if you have further questions about that.

ankushduacodes:
> @ankushduacodes I updated the PR title as described in [#2069 (comment)](https://github.com/mozilla/web-ext/pull/2069#issuecomment-724783760) (`fix` is the conventional-changelog type, everything after the `:` is the subject, and so the updated PR title does now pass the conventional changelog linting task on the CI).
> 
> The only remaining piece for this PR should now be the additional test case to cover the part that coveralls is complaining about, the additional `if (...) {}` branch added in src/utils/adb.js (the same piece briefly described [#2069 (review)](https://github.com/mozilla/web-ext/pull/2069#pullrequestreview-527299408)).
> 
> Let me know how it is going and if you have further questions about that.

Hi @rpl, Thank you for fixing the title for me.
I am looking into sinonjs documentation to understand how to tests are being operated. Once I am done I will try to incorporate a test for the if statement as suggested by you. Thank you again for being patient.

ankushduacodes:
@rpl I have taken a look at the group of tests that are related to the PR ([outlined in this review](https://github.com/mozilla/web-ext/pull/2069#pullrequestreview-527299408))
I am just confused about what are we testing here, are we testing if the reference browser runs without specifying a apk component or are we providing a different apk component to like .App to see it fail?

tests/unit/test-util/test.adb.js
```diff
@@ -935,6 +935,41 @@ describe('utils/adb', () => {
...
+            'org.mozilla.reference.browser.BrowserActivity',
+          extras: [{
+            key: 'args',
+            value: '-profile /fake/custom/profile/path',
+          }],
+          wait: true,
+        }
+      );
+    });
+
```
rpl, line 972: This test case looks pretty good!
I would just like a few more changes to also cover the opposite scenario (when a `firefoxApkComponent` has been passed explicitly as not undefined).

The new test case should be pretty similar to the one you already wrote, they will basically only differ for the `firefoxApkComponent` parameter and the resulting `component` property that we expect to get from the `startActivity`'s `calledWithMatch`, and so it should be possible to just parametrize this test case and run it twice, something like the following should allow us to cover both without duplicating almost the same test case twice:

```
async function testReferenceBrowserApkComponent(
  firefoxApkComponent, expectedApkComponent
) {
  // almost everything exactly like the test case you already wrote, with just
  // the following small changes: 
  ...
  const promise = adbUtils.startFirefoxAPK(..., firefoxApkComponent, ...);
  ...
  const apkName = 'org.mozilla.reference.browser';
  const component =
    `${apkName}/${apkName}.${expectedApkComponent}`
  sinon.assert.calledWithMatch(
    ..., {
     ...
     component,
     ...
    }); 
}

it('starts reference browser with default APK component if not specified', () => {
  return testReferenceBrowserApkComponent(undefined, 'BrowserActivity');
});

it('starts reference browser with the given APK component', () => {
  return testReferenceBrowserApkComponent('CustomActivity', 'CustomActivity');
});
```

Let me know if there is any issue with this approach that you would need help with (I haven't tried the code snippet above, and so there may be some typo or some other small change that I may have missed to mention explicitly.

rpl (changes requested):
@ankushduacodes thanks! the test you created looks pretty great, good job!

would you mind to also add assertions to double-check that if the firefoxApkComponent parameter is passed as not undefined it isn't going to be overridden with the default one, in the inline review comment that follows I described in more detail how I think we can adapt your test case to be reused for testing both the scenario.

Let me know how that sounds to you.

tests/unit/test-util/test.adb.js
ankushduacodes: Sounds good!... I think I know what you mean...

ankushduacodes: I had defined the function on top of the file but then I moved it from top of the file to where the tests are that use the function. Please let me know how it looks

tests/unit/test-util/test.adb.js
```diff
@@ -935,6 +935,52 @@ describe('utils/adb', () => {
...
+          readAll: sinon.spy(() => Promise.resolve(Buffer.from('\n'))),
+        },
+      });
+      const adbUtils = new ADBUtils({adb});
+      const apkName = 'org.mozilla.reference.browser';
+      const component = `${apkName}/${apkName}.${expectedApkComponent}`;
+      const promise = adbUtils.startFirefoxAPK(
+        'device1',
+        apkName,
+        firefoxApkComponent, // firefoxApkComponent
```
rpl, line 955: Nit, no need for the comment now that we are passing an explicit variable with the exact same name.

rpl (approved):
@ankushduacodes looks great! :tada: (and I also prefer the function to be nearby the only two test cases that are using it)
The review comment below is just for a very small nitpick (just an inline comment that we don't need anymore), would you mind to remove that comment before I merge this?

Thanks again for contributing this fix!!!

tests/unit/test-util/test.adb.js
ankushduacodes: Done 🚀

caitmuenster:
Thanks so much for the patch, @ankushduacodes! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Welcome onboard! We look forward to seeing you around.

ankushduacodes:
@caitmuenster Thank you for recognizing my contribution. I just have a small request, I am seeing that on the `Link to profile` section instead of my github profile link, it links to this pull request,  Is it possible to change it?

caitmuenster:
Absolutely! So sorry about that, @ankushduacodes. It's been fixed now! :)

ankushduacodes:
Thank you @caitmuenster, I really appreciate that.

PR #2073, opened 2020-11-15
feat: threw error if profile-create-new-if-missing used without profile

akhilpanchal:
Follow up to #2064 based on @rpl's [comment](https://github.com/mozilla/web-ext/pull/2064#discussion_r518873064).

Throws error if `profile-create-new-if-missing` option used without `firefox-profile` or `chromium-profile`.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -124,7 +124,13 @@ export default async function run(
 
   const profileDir = firefoxProfile || chromiumProfile;
 
-  if (profileCreateIfMissing && profileDir) {
+  if (profileCreateIfMissing) {
+    if (!profileDir) {
+      throw new UsageError(
+        '--profile-create-if-missing should be paired ' +
+        'with --firefox-profile or --chromium-profile'
+      );
+    }
     const isDir = fs.existsSync(profileDir);
     if (isDir) {
       log.info(`Profile directory ${profileDir} already exists`);
@@ -132,11 +138,6 @@ export default async function run(
       log.info(`Profile directory not found. Creating directory ${profileDir}`);
       await fs.mkdir(profileDir);
     }
-  } else if (profileCreateIfMissing) {
-    throw new UsageError(
-      '--profile-create-if-missing should be paired ' +
-      'with --firefox-profile or --chromium-profile'
-    );
   }
 
   const runners = [];
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -11,6 +11,7 @@ import {
 } from '../firefox/remote';
 import {createLogger} from '../util/logger';
 import defaultGetValidatedManifest from '../util/manifest';
+import {UsageError} from '../../src/errors';
 import {
   createExtensionRunner,
   defaultReloadStrategy,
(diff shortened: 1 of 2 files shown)


src/cmd/run.js
```diff
@@ -131,6 +132,11 @@ export default async function run(
       log.info(`Profile directory not found. Creating directory ${profileDir}`);
       await fs.mkdir(profileDir);
     }
+  } else if (profileCreateIfMissing) {
+    throw new UsageError(
+      '--profile-create-if-missing should be paired ' +
+      'with --firefox-profile or --chromium-profile'
+    );
```
rpl, line 139: we could move this into the previous if block by doing something like:

```
if (profileCreateIfMissing) {
  if (!profileDir) {
    throw new UsageError("...);
  }
  // the rest of the if block from line 128 to line 133.
}
```

tests/unit/test-cmd/test.run.js
```diff
@@ -406,5 +407,23 @@ describe('run', () => {
...
+           await cmd.run({ profileCreateIfMissing: true });
+         } catch (error) {
+           actualError = error;
+         }
+
+         assert.instanceOf(actualError, UsageError);
+         assert.match(
+           actualError && actualError.message,
+           /should be paired with --firefox-profile or --chromium-profile/
+         );
```
rpl, line 426: we should use `await assert.isRejected(...)` (provided by the chai-as-promised dev dependency and also used in some of the other tests, e.g. test.adb.js should be one of the test file already using this assertion functions).

rpl (changes requested):
@akhilpanchal Thanks for looking into quickly jumping on covering the followups related to #2064!!! 
Follows a couple of review comments.

akhilpanchal:
> @akhilpanchal Thanks for looking into quickly jumping on covering the followups related to #2064!!!
> Follows a couple of review comments.

Thanks @rpl. I addressed your feedback in the latest commit. 
I also have a [pull request open in the extension-workshop repo](https://github.com/mozilla/extension-workshop/pull/832) for documenting.
Let me know if you have any feedback for me.

Thanks again for all your guidance and comments :)

src/cmd/run.js
```diff
@@ -123,7 +124,13 @@ export default async function run(
 
   const profileDir = firefoxProfile || chromiumProfile;
 
-  if (profileCreateIfMissing && profileDir) {
+  if (profileCreateIfMissing) {
+    if (!profileDir) {
+      throw new UsageError(
+        '--profile-create-if-missing should be paired ' +
```
rpl, line 130: would you mind to replace `should be paired with` with `requires`? (mostly because I'm not really sure if paired is a common why to refer to that in english, "--profile-create-if-missing requires --firefox-profile or --chromium-profile" is shorter and should be clear enough).

Same change has to be applied to the assertion on the test side.

rpl (commented):
@akhilpanchal good job, this looks great. I would just tweak the UsageError message a bit (my apologies I should have noticed it sooner and pointed it out in the previous review round) before merging this on master, would you mind to do that? (the tweak is described below).

src/cmd/run.js
akhilpanchal: > @akhilpanchal good job, this looks great. I would just tweak the UsageError message a bit (my apologies I should have noticed it sooner and pointed it out in the previous review round) before merging this on master, would you mind to do that? (the tweak is described below).

No worries at all! Makes sense to me. It is shorter and clear.
I addressed it.

rpl (approved):

rpl:
@akhilpanchal Thanks a lot! looks great and I'm going to merge it right now! :tada:

PR #2074, opened 2020-11-17
fix: Accept the --no-input global cli option and ensure it is parsed into a noInput boolean option

### LINKED ISSUE #1860: "Unknown argument: input" when specifying `--no-input`
A bug.
Currently `--no-input` is treated as an invalid argument. Passing `--no-input=true` works, but is cumbersome and surprising to users of the CLI.
```
$ web-ext lint --no-input 
web-ext lint

Validate the extension source

Options:
  --version                 Show version number                        [boolean]
  --source-dir, -s          Web extension source directory.
      [string] [required] [default: "/home/abigail/code/firefox-ctrlnumber/src"]
  --artifacts-dir, -a       Directory where artifacts will be saved.
                                                   [string] [required] [default:
                  "/home/abigail/code/firefox-ctrlnumber/src/web-ext-artifacts"]
  --verbose, -v             Show verbose output                        [boolean]
  --ignore-files, -i        A list of glob patterns to define which files should
                            be ignored. (Example:
                            --ignore-files=path/to/first.js path/to/second.js
                            "**/*.log")                                  [array]
  --no-input                Disable all features that require standard input
                                                                       [boolean]
  --config, -c              Path to a CommonJS config file to set option
                            defaults                                    [string]
  --config-discovery        Discover config files in home directory and working
                            directory. Disable with --no-config-discovery.
                                                       [boolean] [default: true]
  --output, -o              The type of output to generate
                            [string] [choices: "json", "text"] [default: "text"]
  --metadata                Output only metadata as JSON
                                                      [boolean] [default: false]
  --warnings-as-errors, -w  Treat warnings as errors by exiting non-zero for
                            warnings                  [boolean] [default: false]
  --pretty                  Prettify JSON output      [boolean] [default: false]
  --self-hosted             Your extension will be self-hosted. This disables
                            messages related to hosting on addons.mozilla.org.
                                                      [boolean] [default: false]
  --boring                  Disables colorful shell output
                                                      [boolean] [default: false]
  -h, --help                Show help                                  [boolean]

Unknown argument: input
$
```

`--no-input` should be treated equivalently to `--no-input=true`.

````
$ node --version
v13.10.1
$ yarn --version
1.22.4
$ web-ext --version
4.1.0
````

### PR DESCRIPTION
**Pull request by rpl**
This PR contains some minor changes to make sure we do accept the --no-input global cli option as documented and that it is parsed into the noInput boolean option the web-ext commands expects (plus a new test case to make sure we do not regress again on future updates of the yargs npm dependency).

Fixes #1860

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -177,6 +177,15 @@ export class Program {
       argv.noReload = !argv.reload;
     }
 
+    // Yargs doesn't accept --no-input as a valid option if there isn't a
+    // --input option defined to be negated, to fix that the --input is
+    // defined and hidden from the yargs help output and we define here
+    // the negated argument name that we expect to be set in the parsed
+    // arguments (and fix https://github.com/mozilla/web-ext/issues/1860).
+    if (argv.input != null) {
+      argv.noInput = !argv.input;
+    }
+
     // Replacement for the "requiresArg: true" parameter until the following bug
     // is fixed: https://github.com/yargs/yargs/issues/1098
     if (argv.ignoreFiles && !argv.ignoreFiles.length) {
@@ -447,6 +456,13 @@ Example: $0 --help run.
     },
     'no-input': {
       describe: 'Disable all features that require standard input',
+      type: 'boolean',
+      demandOption: false,
+    },
+    'input': {
+      // This option is defined to make yargs to accept the --no-input
+      // defined above, but we hide it from the yargs help output.
+      hidden: true,
       type: 'boolean',
       demandOption: false,
     },
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -865,6 +865,16 @@ describe('program.main', () => {
         fakeCommands.run.resetHistory();
       });
     }
+
+    it('does set noInput to undefined if --no-input is not specified',
+       async () => {
+         fakeCommands.run.resetHistory();
+         await execProgram(['run'], { commands: fakeCommands });
+         sinon.assert.calledWithMatch(
+           fakeCommands.run,
+           {noInput: undefined}
+         );
+       });
   });
 
 });
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -841,6 +841,32 @@ describe('program.main', () => {
     }
   );
 
+  describe('--no-input', () => {
+    const fakeCommands = fake(commands, {
+      run: () => Promise.resolve(),
+    });
+
+    const testCases = [
+      ['--no-input', {noInput: true}],
+      ['--no-input=false', {noInput: false}],
+      ['--no-input=true', {noInput: true}],
+      ['--input', {noInput: false}],
+      ['--input=false', {noInput: true}],
+      ['--input=true', {noInput: false}],
+    ];
+
+    for (const [cliArg, expectedParsed] of testCases) {
+      it(`does parse ${cliArg} into the expected command options`, async () => {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/program.js [hunk 1]
```diff
@@ -450,6 +459,13 @@ Example: $0 --help run.
       type: 'boolean',
       demandOption: false,
     },
+    'input': {
+      // This option is defined to make yargs to accept the --no-input
+      // defined above, but we hide it from the yargs help output.
```
Rob--W, line 464: Could you add a unit test to verify that it doesn't appear in the help?

tests/unit/test.program.js
```diff
@@ -841,6 +841,32 @@ describe('program.main', () => {
     }
   );
 
+  describe('--no-input', () => {
+    const fakeCommands = fake(commands, {
+      run: () => Promise.resolve(),
+    });
+
+    const testCases = [
+      ['--no-input', {noInput: true}],
```
Rob--W, line 850: Need test case without any `input` parameters.

tests/unit/test.program.js
```diff
@@ -841,6 +841,32 @@ describe('program.main', () => {
...
+      ['--no-input=false', {noInput: false}],
+      ['--no-input=true', {noInput: true}],
+      ['--input', {noInput: false}],
+      ['--input=false', {noInput: true}],
```
Rob--W, line 854: If we don't support `--input`, then `noInput` should be false here.

tests/unit/test.program.js
rpl: :+1: sgtm

tests/unit/test.program.js
rpl: it is implicitly supported even if hidden, and I'm really not sure that there is an easy way to don't support it without messing with yargs in ways that I would prefer to avoid.

I decided to explicitly included test cases for it because at least we can ensure that it wouldn't regress without being noticed.

src/program.js
rpl: with an integration test for sure, but I'll take a look if we can cover it with a unit test first.

tests/unit/test.program.js
```diff
@@ -841,6 +841,42 @@ describe('program.main', () => {
...
+          fakeCommands.run,
+          expectedParsed
+        );
+        fakeCommands.run.resetHistory();
+      });
+    }
+
+    it('does set noInput to undefined if --no-input is not specified',
+       async () => {
+         fakeCommands.run.resetHistory();
```
Rob--W, line 871: Nit: For consistency with the whole test in this patch, either put `resetHistory` before `execProgram`, or after it. But don't mix them.

Why did you have to create a new `it` though? Can't you use `['-v', {noInput: undefined}],` ?

src/program.js
rpl: added as a new functional test case (it would be possible to unit test if we would have access to the `Program` instance created internally by the `main` function, which would require a bit of refactoring, I opted for the function test as it was straightforward and our small functional tests are still pretty fast to run).

tests/unit/test.program.js
rpl: added as a unit test.

tests/unit/test.program.js
rpl: > Why did you have to create a new `it` though? Can't you use `['-v', {noInput: undefined}],` ?
yeah, that's nicer :+1:

Rob--W (approved):

PR #2077, opened 2020-11-20
feat: Added --watch-ignored new command line option

### LINKED ISSUE #2022: customize file watch path spec to get rid of warnings from chokidar
Bug/feature? Both? A feature to avoid the bug

running `web-ext run` triggers too many watchers and it throws errors:
>Error from chokidar : Error: ENOSPC: System limit for number of file watchers reached

I would like to customize which files are watched so it doesn't output a long list of warnings for every file in the node_modules folder.

* **Firefox version**:
* **Your OS and version**:

````
````
v12.18.4
6.14.6
5.0.0

### PR DESCRIPTION
**Pull request by ankushduacodes**
fixes #2022

### CODE REVIEW

**Comment by ankushduacodes** [comment]
@rpl, I am not sure why onChange function is never being called even if we write to a file that is not being ignored.

**Comment by ankushduacodes** [comment]
@rpl I think I have added all the necessary test, my only two doubt/update are outlined in this [comment](https://github.com/mozilla/web-ext/pull/2077#issuecomment-739105157). Hoping to understand them better :)

Do let me know any other changes are required.

File: tests/unit/test.watcher.js [hunk 1]
```diff
@@ -207,3 +207,48 @@ describe('watcher', () => {
...
+      }
+
+      // Verify foo.txt is being ignored.
+      await fs.writeFile(files[0], '<content>');
+      sinon.assert.notCalled(onChange);
+
+      // Verify that the other two files are not be ignored.
+      await fs.writeFile(files[1], '<content>');
+      await waitDebounce();
+      sinon.assert.calledOnce(onChange);
```
rpl, line 243: > @rpl, I am not sure why onChange function is never being called even if we write to a file that is not being ignored.
> After doing some debugging, I found that `watcher.on('change')` event is never being invoked
> 
> https://github.com/mozilla/web-ext/blob/afd9592579ceb9606f2defe9afa9446c3f8227c5/src/watcher.js#L50-L52
> 
> **Doubt**:
> 
>     1. What difference does it make when we use `watchChange` helper function instead of using `onSourceChange ` directly?   
>, why does it do intial scan when we use `watchChange`?

Totally reasonable doubts, let me try to provide some more details:

The watchChange helper does currently enforce a certain order of the steps, and the file is written before the watcher is going to start monitoring the directory for changes and then updated, that means that the initialScan will happen when the directory being watched already contains the file.

By using the onSourceChange directly we are ensure that we are in control of the steps (without having to further increase the complexity of that helper) and that no files are in the directory being watched and we write them later to cover a couple of expected behaviors around the test case scenario.

Let me know if the details above are answering your questions or if there are other points unclear.

> 
> 
> Thank you!...
> 
> **Update**: I see that tests on travis-ci are passing successfully but on my machine `--watch-ignored` test is failing. Here's a screenshot of that:
> **On linux, tests are passing successfully**

yeah, watching files is one of those things that it is not unlikely to behave differently based on the os (and also the particular hardware)

I'm also afraid that, giving that it does look like a "race condition" kind of issue, it may also start to fail intermittently on the CI over an higher number of builds.

I think that it would be a good idea to look into an alternative strategy that may not be actually subjected to race-related failures that easily.

One possible other strategy could be to use a secondary watcher instance that does not have any watchIgnored parameter, so we can wait for the change event to have been emitted on that additional watcher instance.

Interesting enough, when I tried that locally I did noticethat I had to add a short delay (using `await waitDebounce();`) before writing the file, otherwise none of the two watcher was actually emitting the 'change' event we expect, this match what you did notice by debugging the failure you get on macos.
Nevertheless using the secondary watcher may still be a good idea to make the test as stable as possible (and avoid intermittent failures).

Would you mind to give that a try also on your machine? so that we can double-check that this strategy does make the test to pass consistently also on macos as well as on Linux.

Here is the diff of the changes I did apply locally:

```diff --git a/tests/unit/test.watcher.js b/tests/unit/test.watcher.js
index 03df40b..8f4c833 100644
--- a/tests/unit/test.watcher.js
+++ b/tests/unit/test.watcher.js
@@ -6,6 +6,7 @@ import {it, describe} from 'mocha';
 import {fs} from 'mz';
 import sinon from 'sinon';
 import {assert} from 'chai';
+import Watchpack from 'watchpack';
 
 import {default as onSourceChange, proxyFileChanges} from '../../src/watcher';
 import {withTempDir} from '../../src/util/temp-dir';
@@ -229,24 +230,33 @@ describe('--watch-ignored is passed in', () => {
         debounceTime,
       });
 
+      const watchAll = new Watchpack();
+      watchAll.watch(files, [], 0);
+
       async function waitDebounce() {
         await new Promise((resolve) => setTimeout(resolve, debounceTime * 2));
       }
 
+      async function assertOnChange(filePath, expectedCallCount) {
+        const promiseOnChanged = new Promise((resolve) => watchAll.once(
+          'change',
+          (f) => resolve(f))
+        );
+        await waitDebounce();
+        await fs.writeFile(filePath, '<content>');
+        assert.equal(filePath, await promiseOnChanged);
+        sinon.assert.callCount(onChange, expectedCallCount);
+      }
+
       // Verify foo.txt is being ignored.
-      await fs.writeFile(files[0], '<content>');
-      sinon.assert.notCalled(onChange);
+      await assertOnChange(files[0], 0);
 
       // Verify that the other two files are not be ignored.
-      await fs.writeFile(files[1], '<content>');
-      await waitDebounce();
-      sinon.assert.calledOnce(onChange);
-
-      await fs.writeFile(files[2], '<content>');
-      await waitDebounce();
-      sinon.assert.calledTwice(onChange);
+      await assertOnChange(files[1], 1);
+      await assertOnChange(files[2], 2);
 
       watcher.close();
+      watchAll.close();
       // Leave watcher.close some time to complete its cleanup before withTempDir will remove the
       // test directory.
       await waitDebounce();
```

Let me know how it goes, this test case turned out to be more interesting than expected ;-)

rpl (changes requested):
My apologies again for the delay. Follows a new round of review comments (some nits, and a bigger and more detailed comment related that tricky test case ;-)).

ankushduacodes:
@rpl, I have made the necessary changes outlined [here](https://github.com/mozilla/web-ext/pull/2077#discussion_r540565517).
The tests are passing on both macOS and Linux.
I have also resolved other requested changes.

Please let me know how it looks.
Thank you.

tests/unit/test.watcher.js
```diff
@@ -207,3 +208,57 @@ describe('watcher', () => {
   });
 
 });
+
+describe('--watch-ignored is passed in', () => {
```
rpl, line 212: Nit, would you mind to move this describe inside the `describe('watcher', () => { ... });` block, at the end on it (my apologies I didn't notice this group was outside out that block until now).

tests/unit/test.watcher.js
```diff
@@ -207,3 +208,57 @@ describe('watcher', () => {
...
+  it('does not change if ignored file is touched', () =>
+    withTempDir(async (tmpDir) => {
```
rpl, line 214: > Could I please also ask how was race condition happening earlier?

@ankushduacodes well, I have the feeling that the race is still there, the change we made were mainly aiming to reduce the intermittent failures chances as much as possible.

There may be multiple reasons for the race:
- one is the debounce time we do have (changes of making that trigger a failure reduce by making as able to customize the debounce time to a lower value during the test execution)

- watchpack's watch method takes a third parameter that is the startTime, which is then used to check if the file mtime is >= of the given startTime and not emit the change event if it is not (e.g. see [here](https://github.com/webpack/watchpack/blob/40ddaa9c895e83c5eb490abca6834ae1e71aa8cc/lib/DirectoryWatcher.js#L45-L49) and [here](https://github.com/webpack/watchpack/blob/40ddaa9c895e83c5eb490abca6834ae1e71aa8cc/lib/DirectoryWatcher.js#L224-L236) in the watchpack sources), we do set the startTime to Date.now() in how watcher utils.

- watchpak seems to be internally using a FS_ACCURACY variable that may effect the behavior (in other words if the change event will be emitted or not), initially defined [here](https://github.com/webpack/watchpack/blob/40ddaa9c895e83c5eb490abca6834ae1e71aa8cc/lib/DirectoryWatcher.js#L15).

- the underlying OS feature to watch for changes will be different for each OS and may present slightly different behavior and accuracy 

These are the once that I did notice when I looked into what may be triggering the unexpected failures we were looking into.

Yesterday I did tried a kind of stress test to double-check if the test was going to fail intermittently over a high number of runs, mocha doesn't seem to support any command line option to do that without changing the tests themselves and so I tweaked the test locally as follows:

```
it('does not change if ignored file is touched', async () => {
    for (var i = 0; i < 200; i++) {
      await withTempDir(async (tmpDir) => {
        ...
      });
    }
});
```

and then I execute it by launching mocha manually on this particular test case and with a longer test timeout set using the mocha cli options:

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 800000
```

This was stress test version of the test case was still failing for me from time to time, I would be ok to defer looking more into it and see if those intermittent failures are going to happen for real after we land it (and if it does fail intermittently then we can look into apply the necessary changes to reduce it a bit more as part of a follow up).

src/program.js
```diff
@@ -616,11 +616,19 @@ Example: $0 --help run.
...
+                  ' file changes. This is useful if you use a custom' +
                   ' build process for your extension',
         demandOption: false,
         type: 'string',
       },
+      'watch-ignored': {
+        describe: 'Extension will not reload if anything in this' +
+                  ' directory or file/s changes. This is useful when there' +
+                  ' are a lots of files which are unnecessary to watch' +
+                  ' for example. node_modules folder',
```
rpl, line 628: Nit, how about tweaking the describe message a bit to say something like:

Paths and globs patterns that should not be watched for changes. This is useful if you want to explicitly prevent web-ext from watching part of the extension directory tree, e.g. the node_modules folder.

rpl (changes requested):
@ankushduacodes looks great, follows just two more nits (one small tweak to the option description and moving one of the new test case into the existing group of watcher tests).

In the following comments there is one that provides some more details about the possible reasons for the intermittent failures (and some details about a small stress test experiment I did locally to double-check if that test can still fail from time to time as I was guessing, well... spoiler... it still can :-P, but I'm ok on deferring looking more into that in a follow up if the intermittent failures are going to actually happen after we landed this on master).

tests/unit/test.watcher.js
ankushduacodes: @rpl Thank you for the detailed explanation of why race condition is occuring!..

You are right, the stress test is failing for me as well. Only the first iteration of the test loop is passing for me, Is that the case on your machine as well?

ankushduacodes:
@rpl I am not sure why Circle CI is giving me a build error. Is this internal to web-ext itself?

rpl:
> @rpl I am not sure why Circle CI is giving me a build error. Is this internal to web-ext itself?

@ankushduacodes we migrated yesterday to CircleCI, would you mind to rebase this branch on top of the last changes landed in the repo? that would make the circleci config available in this pull request and CircleCI should stop complaining and start running the 4 CI jobs expected (e.g. like the list of circleci checks in PR #2095).    

----

Some more context in case you are interested in knowing more:

We had to migrate the CI jobs to CircleCI asap to close #2094 before the end of the year and be sure to keep our CI jobs up and running. I landed the changes yesterday on master (#2096) and so now circleci complains because in this branch there is no CircleCI config (the log in the CI failure is `No configuration was found in your project.`) and apparently Travis isn't building neither (I guess that's because the master branch does not have the travis config file anymore, removed in #2096 along with replacing it with the CircleCI config).

tests/unit/test.watcher.js
rpl: > @rpl Thank you for the detailed explanation of why race condition is occurring!..

My pleasure, feel always free to ask us more context when you are curious or have doubts on some details (even they are not blocking issue and seems secondary details), in my opinion discussing about those kind of questions is an important part of a mentored bug (as much as getting the particular pull request done or good first bug fixed) and I'm absolutely happy to answer them with as much context I can. 

> You are right, the stress test is failing for me as well. Only the first iteration of the test loop is passing for me, Is that the case on your machine as well?

Interesting, on my machine it was able to run a bunch of times before triggering one intermittent failure.

The following changes were making the stress test a bit more stable on my machine, would you like to try if they do also on macos? that may be interesting to anticipate how bad the intermittency would be in practice.

- reduced the value of `debounceTime` (I did set it to `0`)
- changed `waitDebounce`'s setTimeout second parameter from `debounceTime * 2` to `10` (or a value a bit higher if based on the behavior of the stress test experiment it seems to make the test a bit more stable)
- added an additional `await waitDebounce();` between `await fs.writeFile(...)` and `sinon.assert.callCount(onChange, ...)` (to leave some time for the onChange spy to actually be called, I had the impression that it was another potential source of intermittency)

If you decide to give it a try, let me know if these tweaks did make the test a bit more stable on MacOS as they seem to do while running on Linux.

rpl (approved):
@ankushduacodes Thanks for rebasing the PR right away (and for covering all the last review comments in the updated version of this patch), you are awesome :heart: 

I'm approving this version, but I'll wait your feedback about that experiment with the stress test (the one related to double-checking that new test case for intermittency issues) before proceeding to merge it.

Ping me in a comment to let me know when you have done with that (either if there are further changes applied to this PR or not).

Thanks again for your work on these contributions!

tests/unit/test.watcher.js
ankushduacodes: @rpl I tried running tests multiple times with `debounceTime` set to zero and different `waitDebounce`'s `setTimeout`'s second param set to 10, 20, 30, 40, 100, 500, 1000.
Added waitDebounce between fs.writeFile and assert statement

<details open>

<summary>In case of 10:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 17:49:46


  watcher
    --watch-ignored is passed in
iteration 0
      1) does not change if ignored file is touched


  0 passing (1m)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     Error: Timeout of 80000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

</details>
<details open>

<summary>In case of 20:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 17:56:42


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
      1) does not change if ignored file is touched


  0 passing (590ms)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)

```

</details>
<details open>

<summary>In case of 30:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 17:58:25


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
      1) does not change if ignored file is touched


  0 passing (593ms)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)
```

</details>

<details open>

<summary>In case of 40:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 18:03:25


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
Passed 1
iteration 2
      1) does not change if ignored file is touched


  0 passing (986ms)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)


```

</details>
<details open>

<summary>In case of 100:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 18:04:15


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
Passed 1
iteration 2
      1) does not change if ignored file is touched


  0 passing (2s)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)


```

</details>
<details open>

<summary>In case of 500:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 18:05:36


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
      1) does not change if ignored file is touched


  0 passing (5s)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)


```

</details>
<details open>

<summary>In case of 1000:</summary>
<br>

```
npx mocha -r @babel/register -r tests/setup.js tests/unit/test.watcher.js -f "ignored file" -t 80000           (base) 18:06:23


  watcher
    --watch-ignored is passed in
iteration 0
Passed 0
iteration 1
      1) does not change if ignored file is touched


  0 passing (9s)
  1 failing

  1) watcher
       --watch-ignored is passed in
         does not change if ignored file is touched:
     expected spy to be called 0 times but was called once
    spy() at onChange (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/debounce/index.js:40:21)
  AssertError: expected spy to be called 0 times but was called once
      spy() at onChange (node_modules/debounce/index.js:40:21)
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:110:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:67:20)
      at Object.callCount (node_modules/sinon/lib/sinon/assert.js:151:17)
      at assertOnChange (tests/unit/test.watcher.js:249:26)
      at /Users/ankushdua/Documents/GitHub/web-ext/tests/unit/test.watcher.js:253:11
      at Context.<anonymous> (tests/unit/test.watcher.js:214:9)


```

</details>


it seems to be highly inconsistent, I also tried running each of these tests above multiple times, and gave me very inconsistent results when it came to asserting the calls to `onChange`, a few of these timed out(Probably because of race condition).

### Doubts:
> one is the debounce time we do have (changes of making that trigger a failure reduce by making as able to customize the debounce time to a lower value during the test execution)

> watchpack's watch method takes a third parameter that is the startTime, which is then used to check if the file mtime is >= of the given startTime and not emit the change event if it is not (e.g. see here and here in the watchpack sources), we do set the startTime to Date.now() in how watcher utils.

> watchpak seems to be internally using a FS_ACCURACY variable that may effect the behavior (in other words if the change event will be emitted or not), initially defined here.

Do you know of any good resource where I can read about these in more detail or Should I look into source code more?

> one is the debounce time we do have (changes of making that trigger a failure reduce by making as able to customize the debounce time to a lower value during the test execution)

I am still unsure about how 👆 this affects race condition.

rpl: > I am still unsure about how :point_up_2: this affects race condition.

@ankushduacodes me too, I really hate when I feel to be missing something going on that would explain an unexpected behavior. Sorry for keeping you waiting a bit, I wanted to give it a try and my only macos machine isn't in a great shape (and wifi not working anymore is one of the most annoying issues it does have :-P)

Anyway I've been able to give it a try to run this test on macos and... not that suprisingly there was one more OS specific behavior that we missed to notice ;-)

To identify what it was happening I opted to add a temporary console.log call inside the `watcher.on('change', filePath => ...);` in `src/watcher.js` and executed the stress test again... well turned out that on macos a `change` event is being fired also for the directory itself from time to time (but pretty often, even on my old macos, I imagine one more recent and in a better shape may trigger it even more often) which is something that wasn't apparently happening on Linux.

To prevent that from happening (we are not interested in that anyway), we can change that `shouldWatchFile: () => true` callback we use in this test and set it to `shouldWatchFile: (filePath) => filePath !== tmpPath`, which will basically filter out that change event and prevent the failures related to this OS-specific difference in behavior.

Besides that I had to add another `await waitDebounce();` right before the `sinon.assert.callCount(onChange, ...)` (as I did mention in one of the previous comments as something that may help to make the test a bit more stable), otherwise the test was still failing pretty soon in the stress test sequence because the onChange callback wasn't called yet by the time we were calling the assertion.

would you mind to try to apply these two small changes and see if it does make the test able to pass the stress test as it did for me? (I kept everything else as is in this version of the PR, `debounceTime` set to `10` and `waitDebounce` using `debounceTime*2`).

if it does make the stress test to pass also on your machine I think that it would be good to update the patch with these two small changes and then land this. how that sounds to you?

ankushduacodes: I think I understand what you mean by the following (Please correct me if I am wrong):
> To prevent that from happening (we are not interested in that anyway), we can change that shouldWatchFile: () => true callback we use in this test and set it to shouldWatchFile: (filePath) => filePath !== tmpPath, which will basically filter out that change event and prevent the failures related to this OS-specific difference in behavior.

We are basically ignoring the `change` event emitted for the directory by returning `false` and basically ignoring the filePath for that event.

rpl: > We are basically ignoring the change event emitted for the directory by returning false and basically ignoring the filePath for that event from inside of proxyFileChanges() function.

@ankushduacodes yep that's exactly what I mean with that :+1:

> I will try out the tests and will update you on it soon!..

:+1: Thanks a lot for helping me with investigating this!!!

ankushduacodes: @rpl The tests ran successfully this time with the changes you mentioned. 
For me the `change` event on the directory was being emitted every 4 iterations on average.

<br>
Tiny reminder/info request:

>>watchpack's watch method takes a third parameter that is the startTime, which is then used to check if the file mtime is >= of the given startTime and not emit the change event if it is not (e.g. see here and here in the watchpack sources), we do set the startTime to Date.now() in how watcher utils.

>>watchpak seems to be internally using a FS_ACCURACY variable that may effect the behavior (in other words if the change event will be emitted or not), initially defined here.

>Do you know of any good resource where I can read about these in more detail or Should I look into source code more?

ankushduacodes:
@rpl According to me the only thing left to do in this PR is an update to the docs. What do you think?

tests/unit/test.watcher.js
rpl: >> watchpak seems to be internally using a FS_ACCURACY variable that may effect the behavior (in other words if the change event will be emitted or not), initially defined here.
>  Do you know of any good resource where I can read about these in more detail or Should I look into source code more?

@ankushduacodes Was this question specifically about that FS_ACCURACY variable? it is an implementation detail of the watchpack package and so their source code is likely the only place where you can get more details from.

I doubt that it would be documented anywhere else, in some cases to get a bit more context about the actual reasons behind a particular line of code I do use the git blame info to find the pull request that a particular change has been merged from and look into the pull request review comments in case there are any more details useful to get a more complete picture.

Honestly I didn't do it in this case, based on how we use the dependency in web-ext those details don't really matter that much, and so I mainly focused my interest on the details that did matter for keeping the test stable over a big enough number of runs.

rpl:
> @rpl According to me the only thing left to do in this PR is an update to the docs. What do you think?

yep, I'm going to give a quick look to the last update and then merge it on master. 
After that, the change has to be documented (in the extension-workshop doc page) and released on npm.

We are not planning to release a new web-ext version before the end of the year and so there is no rush to create the pull request for the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop), we would still merge that extension-workshop PR right after the new option has been released as part of a new web-ext version, but feel free to open it and mention it in a new comment on #2022.

Thanks again for you work on this enhacement! (and for the help investigating the test intermittency on macos! In my opinion that turned out to be a pretty interesting part of this PR ;-)).

ankushduacodes:
@rpl Do you know of any issues that may be are for beginner to intermediate level people? If so, I would love to work on them :-)

PR #2096, opened 2020-12-17
chore(ci): Migrate CI jobs to circleci

rpl:
This pull requests includes the changes needed to migrate our CI jobs from Travis CI to CircleCI:

- new CircleCI workflow config (./circleci/config.yml), the workflow is organized as follows:
  - some notes about the CircleCI environment to set on the project (currently only NPM_TOKEN) and how to validate changes to the workflow file locally using the circleci command
  - `references.nodejs_version` is the set of nodejs version we run the tests on, these are the only properties of the workflow that needs to be updated to change the nodejs versions tested (without having to mess with the job definition and risk to break some by mistake)
  - a set of parametrized commands are used to reuse a set of steps in multiple jobs without duplicating too much stuff
  - a set of pre-configured jobs (one for the steps running on a linux container and one for the windows regression tests)
  - a predefined workflow with 5 job:
    - `build-nodejs-current`: run the full test suite (included additional steps for the functional tests), collect and publish coverage, persist build to reuse in `release_tag` job
   -  `build-nodejs-next` and `build-nodejs-experimental`: run the full test suite, without flow-checks, coverage collection and running the functional tests only once in npm production mode
   - `test-windows`: run unit test and functional tests on windows (to catch windows-only regressions), skip flow-checks, coverage collection. This job does also retry once the functional tests if they fail (they used to be intermittent on travis windows workers)
   - `release_tag`: rebuild in production mode and release the new version on npm

- small changes to scripts used in CI jobs to adapt them to the CircleCI environment (in particular script/travis-pr-title-lint has been adapted to use the CIRCLE_PULL_REQUEST environment var).
  - **NOTE: the project should be configured to only build pull requests, to make sure that the CIRCLE_PULL_REQUEST environment var will be set when the job does run for the first time on a github pull request** (see https://support.circleci.com/hc/en-us/articles/360006357533--CIRCLE-PULL-REQUEST-is-missing-from-my-Pull-Request), this configuration should still make CircleCI to run the workflow on merges to master.

- migrated from coveralls to codecov (because coveralls would require a token which we can't expose to pull requests jobs)

Testing notes:
- the workflow file has been validated (and executed locally on docker to test the expected behaviors, besides the windows job)
- the release_tag job has been tested on a separate isolated test project, to double-check that the job does publish the npm package successfully as expected


.circleci/config.yml
```diff
@@ -0,0 +1,293 @@
...
+      - run_npm_build:
+          node_env: production
+      - run:
+          name: npm registry auth
+          command: echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > .npmrc
+      - run:
+          name: npm registry publish
+          command: npm publish --access=public
+
+commands:
```
muffinresearch, line 185: I'd suggest moving this earlier in the file. It would probably read a bit easier if `references` and `commands` were at the top followed by `jobs` and `workflows` last.

rpl: :+1: yeah, that sounds reasonable

rpl:
All the circleci jobs looks finally good now (included the windows one), it does submit the coverage data to codecov and I've just applied the last small "circleci"-related tweak to the script that lint the pull request message (which does make us sure that it does follows the conventions expected by the changelog generator utility when there is more than one commit in a repo9)

\o/

@muffinresearch @willdurand This should now be good to go, any additional review comments to cover before I'll merge this?

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+        type: enum
+        default: ""
+        enum: ["", "y"]
+      legacy_bundling:
+        type: boolean
+        default: false
+    steps:
+      ## NOTE: by setting the configured python to /bin/false we are forcing
+      ## the production mode tests to fail if any of the dependencies is a
+      ## binary dependency that is build using node-gyp.
```
willdurand, line 110:
```suggestion
      ## binary dependency that is built using node-gyp.
```

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+      - attach_project_repo
+      - run_npm_build:
+          node_env: production
+      - run:
+          name: npm registry auth
+          command: echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > .npmrc
+      - run:
+          name: npm registry publish
+          command: npm publish --access=public
+
```
willdurand, line 287: Do we need those empty lines at the end?

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+    steps:
+      - attach_project_repo
+      - run_npm_build:
+          node_env: production
+      - run:
+          name: npm registry auth
+          command: echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > .npmrc
+      - run:
+          name: npm registry publish
+          command: npm publish --access=public
```
willdurand, line 286: Isn't `public` the default access?

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+  # This is required to avoid a `EACCES` when running `npm link` (which is
+  # executed in the test suite).
+  configure_global_npm:
+    description: create custom directory for global npm installs
+    steps:
+      # NOTE: keep the following steps compatible with both bash and cmd.exe
+      - run: cd .. && mkdir .npm-global
+      - run: npm config set prefix '../.npm-global'
+
+workflows:
```
willdurand, line 152: I'd move the workflows _after_ jobs.

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+  node-tests:
```
willdurand, line 153: We use `default-workflow` elsewhere because it does more than running test suites.

.circleci/config.yml
```diff
@@ -0,0 +1,288 @@
...
+#     circleci local execute -c tmp/process.yml --job build-nodejs-current
+#
+version: 2.1
+
+orbs:
+  # Used in the `test-windows` job to run the test on windows.
+  win: circleci/windows@1.0.0
+
+references:
+  # NOTE: update the nodejs version strings that follows to change th
```
willdurand, line 22:
```suggestion
  # NOTE: update the nodejs version strings that follows to change the
```

willdurand (approved):
r+wc

.circleci/config.yml
rpl: yep, is not for scoped packages, which I used to test this command, not needed in this case, but I guess it shouldn't harm neither

.circleci/config.yml
rpl: sgtm

scripts/lib/mocha.js
```diff
@@ -4,7 +4,7 @@ const shell = require('shelljs');
 
 const config = require('./config');
 
-// Get the explicit path (needed on travis windows workers).
+// Get the explicit path (needed on CI windows workers).
```
muffinresearch, line 7: Good call on generalizing.

muffinresearch (approved):
Thanks for taking the time to get this up and running Luca 👍

PR #2116, opened 2021-01-11
feat: export the adb utilities to be used in third party nodejs scripts

### LINKED ISSUE #1999: Feature request: Add an API to return a list of android devices that web-ext detected

feature

https://github.com/DimensionDev/webpack-web-ext-plugin/blob/master/index.js#L26

Now we need to parse the text. It would be better to have an API for this so it's easier to build a better interface for it.

ankushduacodes:
attempts to resolve #1999


ankushduacodes:
@rpl Am I seem to be on the correct path here?

Also I do have some doubts related to your recent [comment](https://github.com/mozilla/web-ext/issues/1999#issuecomment-754073807)
Doubt:
1. I was not able to run the extension on my mobile because of the following error:
```
UsageError: Select an android device using --android-device=<name>
    at FirefoxAndroidExtensionRunner.adbDevicesDiscoveryAndSelect (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/extension-runners/firefox-android.js:309:13)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at FirefoxAndroidExtensionRunner.run (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/extension-runners/firefox-android.js:132:5)
    at async Promise.all (index 0)
    at MultiExtensionRunner.run (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/extension-runners/index.js:105:5)
    at run (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/cmd/run.js:242:3)
```
How can I set `--android-device=<name>` while using `web-ext` in a third party nodejs script

2. I was trying to make `{adbBin: string}` object as optional param but could not find a way to do it. I tried reading docs for optional params in flowtype but could resolve the issue. I tried to do it another way ie. to define a type:
```js
export type adbParam{| 
    adbBin?: string 
|}
```
but with this I was having to call the function with an empty object something like: `webExt.util.listADBDevices({})`.

3. > util.listADBFirefoxAPKs({adbBin: string}): Promise<Array<string>>: same parameters of the previous one, but returns a Promise that resolves to an array of string that represents the apk packages identified as Firefox/GeckoView-based browser

Shouldn't this function also take a deviceId of which GeckoView packages we may like to see?

ankushduacodes:
@rpl Just a friendly reminder for the above comment. Looking to get some insight :)

src/util/adb.js
```diff
@@ -418,3 +418,16 @@ export default class ADBUtils {
     });
   }
 }
+
+export async function listADBDevices(adbBin?: string): Promise<Array<string>> {
+  const adbClient = new ADBUtils({adbBin});
+  return await adbClient.discoverDevices();
```
rpl, line 424: Nit, no need to `await` before returning, the promise returned by a call to this async function would still only be resolved when `adbClient.discoverDevices()` is resolved.

src/main.js
```diff
@@ -2,9 +2,10 @@
 import {main} from './program';
 import cmd from './cmd';
 import * as logger from './util/logger';
+import {listADBDevices, listADBFirefoxAPKs} from './util/adb';
```
rpl, line 5: As it is, as soon as we load `src/main.js` we will also load `./util/adb` and everything that `adb.js` does import statically, which include `@devicefarmer/adbkit` (and that npm package will likely load a bunch of its own dependencies as well).

As I described in one of my earlier comments on the issue related to this PR, we want to defer this "module loading chain reaction" as much as possible, which means until it is really really needed.

I think that a simple trick to achieve this lazy loading could be to define `adb` as a getter in the `util` object, something like this:
 
```
...
const utils = {
  logger,
  get adb() {
    ...
  },
}
```
and inside the getter retrieve the two methods from `./util/adb` using `require(...)` instead of the static `import` statement.

Give a try and let me know if you have any question/doubts or if it didn't work 
(you can easily double-check locally that the trick is working as expected and `./util/adb.js`is being loaded lazily, e.g. by putting a `console.log(...); // eslint-disable-line` in that file, that way your console.log will tell you when it is being loaded).

rpl (changes requested):
Besides the comments below (the main one is about making sure to load the adb module as lazily as possible), we may also start to think to the testing coverage.

You may take a look to the test that is already covering importing web-ext as a library:
- tests/functional/test.lib.imports.js
- tests/fixtures/require-as-cjs/test-require.js
- tests/fixtures/import-as-esm/test-import.mjs 

As usually, feel free to ask me more questions once you have some questions or doubts about what to test and what strategy to use.

ankushduacodes:
@rpl I have added some basic tests to check if functions are being exposed correctly. How are tests looking so far?

Few questions:
1. There's a lot of code duplication in test-import.mjs and test-require.js. I was thinking of defining a common function that will receive a top-level object `webExt` and tests those asserts. What do you think? If you do think its a good idea, then where should I define that function? in a new file?
2. Could please point me to direction of what else needs to be tested and where? I know we should also test lazy loading of adb but where and how?

Thank you!...

rpl:
> @rpl I have added some basic tests to check if functions are being exposed correctly. How are tests looking so far?
> 
> Few questions:
> 
>     1. There's a lot of code duplication in [test-import.mjs](https://github.com/ankushduacodes/web-ext/blob/export_adb_utilities/tests/fixtures/import-as-esm/test-import.mjs) and [test-require.js](https://github.com/ankushduacodes/web-ext/blob/export_adb_utilities/tests/fixtures/require-as-cjs/test-require.js). I was thinking of defining a common function that will receive a top-level object `webExt` and tests those asserts. What do you think? If you do think its a good idea, then where should I define that function? in a new file?

@ankushduacodes Yeah, this sounds like a good idea, we could just add an helper function in a separate module file and load it in both to avoid repeating the same assertion in both the test files.
I think that it may be reasonable to move both `tests/fixtures/import-as-esm` and `tests/fixtures/require-as-cjs` under the same directory (e.g. one named `tests/fixtures/webext-as-library` or something like that) and then move both `test-import.mjs` and `test-require.js` under the same directory, and put the helper in a new `helpers.js` file in that same directory.   

>     2. Could please point me to direction of what else needs to be tested and where? I know we should also test lazy loading of adb but where and how?

I think that one possible way to explicitly test that the adb module is lazily loaded may be to use the `require.cache` property that nodejs does provide. That cache property is an object that has the loaded module paths as keys.

You may try it interactively by running `npm run build` with your changes applied and then spawn a node repl from the web-ext project root:

```
$ node
Welcome to Node.js v12.16.3.
Type ".help" for more information.
> Object.keys(require.cache)
[]
> require('./dist/web-ext')
...
> Object.keys(require.cache)
[
  '/.../web-ext/dist/web-ext.js',
  '/.../web-ext/node_modules/source-map-support/source-map-support.js',
  ...
]
```

Using this strategy we may do inside the test helper that you are going to create (the one that will be used by both test-import.mjs and test-require.js) something like the following right after requiring/importing 'web-ext':
- before accessing `webExt.util.adb`, assert that `Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))` is `0` (we didn't accessed util.adb and so we expect that the "@devicefarmer/adbkit" dependencies should not have been loaded yet)
- after accessing `webExt.util.adb`, assert that  `Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))` is greater than 0 (because at that point we expect that the ""@devicefarmer/adbkit" dependencies has been loaded)

This last assertion will also make us sure that if we will ever change the dependency the test will fail and that will remind us to update the test with the new dependency (it will be basically a safety net to be sure that the test doesn't turn itself into an "always passing even if broken" kind of test without being noticed ;-)).

Let me know if this pointers were helpful or if you have any further doubts or questions.

ankushduacodes:
@rpl I have a doubt about imports:
When I tried importing helper.js the following way:
```
import _ from './helpers';
```
It gave me an error and the test of importing web-ext as ESM module failed
But when I used:
```
import _ from './helpers.js';
```
the test passes.

Why is that so?

ankushduacodes:
> > ```
> > 2. Could please point me to direction of what else needs to be tested and where? I know we should also test lazy loading of adb but where and how?
> > ```
> 
> I think that one possible way to explicitly test that the adb module is lazily loaded may be to use the `require.cache` property that nodejs does provide. That cache property is an object that has the loaded module paths as keys.
> 
> You may try it interactively by running `npm run build` with your changes applied and then spawn a node repl from the web-ext project root:
> 
> ```
> $ node
> Welcome to Node.js v12.16.3.
> Type ".help" for more information.
> > Object.keys(require.cache)
> []
> > require('./dist/web-ext')
> ...
> > Object.keys(require.cache)
> [
>   '/.../web-ext/dist/web-ext.js',
>   '/.../web-ext/node_modules/source-map-support/source-map-support.js',
>   ...
> ]
> ```
> 
> Using this strategy we may do inside the test helper that you are going to create (the one that will be used by both test-import.mjs and test-require.js) something like the following right after requiring/importing 'web-ext':
> 
> * before accessing `webExt.util.adb`, assert that `Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))` is `0` (we didn't accessed util.adb and so we expect that the "@devicefarmer/adbkit" dependencies should not have been loaded yet)
> * after accessing `webExt.util.adb`, assert that  `Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))` is greater than 0 (because at that point we expect that the ""@devicefarmer/adbkit" dependencies has been loaded)
> 
> This last assertion will also make us sure that if we will ever change the dependency the test will fail and that will remind us to update the test with the new dependency (it will be basically a safety net to be sure that the test doesn't turn itself into an "always passing even if broken" kind of test without being noticed ;-)).
> 
> Let me know if this pointers were helpful or if you have any further doubts or questions.

@rpl I was looking into implementing an interactive node repl to execute above commands and the check for then assert for the statements above. I did look into spawning a child process and processing the data something like following:
```
cp.execFileSync(npm, ['run', 'build']).toString();
  const childProcess = cp.spawn('node', ['-i']);

  childProcess.stdout.setEncoding('utf8')
  childProcess.stdout.on("data", function(data) {
      // process/assert here
  });


  childProcess.stdin.write("Object.keys(require.cache)\n");
  childProcess.stdin.write("const webExt = require('./dist/web-ext')\n");
  childProcess.stdin.write("Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))\n");
  childProcess.stdin.write("webExt.util.adb\n");
  childProcess.stdin.write("Object.keys(require.cache).filter(filePath => filePath.includes('@devicefarmer/adbkit'))\n");
  childProcess.stdin.end()
```

But the issue I am facing, in this case, is that event is triggered every time new data is sent to stdout even if it consists of `>` or a new line or repl welcome message. And I am not able to run asserts on the data being sent as I don't know what data is being sent.

I hope I am making some sense here.

tests/fixtures/webext-as-library/helpers.js
```diff
@@ -0,0 +1,15 @@
...
+  assert.deepEqual(
+    Object.keys(webExt.util.adb).sort(),
+    ['listADBDevices', 'listADBFirefoxAPKs'].sort(),
+  );
+  assert.equal(typeof webExt.cmd.run, 'function');
+}
+
+module.exports = {
+  testWebExtModuleImports,
+};
```
rpl, line 15: > @rpl I was looking into implementing an interactive node repl to execute above commands and the check for then assert for the statements above. I did look into spawning a child process...

We can use that trick directly inside this helper module, no need to spawn a new process (also, [test.lib.imports.js is already spawning this fixture in a child process here](https://github.com/mozilla/web-ext/blob/3ef025b80d83e49b1d20375cd51511e537e4c340/tests/functional/test.lib.imports.js#L39-L41), and so this test is already running into its own nodejs instance and that is already making us sure that the `require.cache` property would contain the `adbkit` npm dependency only if we have loaded the `adb.js` module).

We can just add the assertion here, e.g. something like the following will do:

```js
function testWebExtModuleImports(webExt) {
  ...

  assertImportedADB({expectLoaded: false});
  assert.deepEqual(
    Object.keys(webExt.util.adb).sort(),
    ['listADBDevices', 'listADBFirefoxAPKs'].sort(),
  );
  assertImportedADB({expectLoaded: true});
}

function assertImportedADB({expectLoaded}) {
  const hasAdbDeps = Object.keys(require.cache).filter(
    (filePath) => filePath.includes('@devicefarmer/adbkit')
  ).length > 0;

  const msg = expectLoaded
    ? 'adb module should have been loaded'
    : 'adb module should not be loaded yet';

  assert.equal(hasAdbDeps, expectLoaded, msg);
}
```

This test is expected to pass successfully with the current implementation in `src/main.js`, and it should fail if we change `src/main.js` back to when it wasn't loading the `adb.js` module lazily.

(You may already figured this out, but in case you didn't try yep, you have to set the CI environment var to true to make the test.lib.imports.js test to run:`CI=true npm run test-function`, otherwise it is skipped and only run in the CI jobs, we do this because it does `npm link`/`npm unlink` the web-ext source repository to run that test scenario).

tests/fixtures/webext-as-library/helpers.js
```diff
@@ -0,0 +1,15 @@
+const assert = require('assert');
+
+function testWebExtModuleImports(webExt) {
```
rpl, line 3: Nit, personally I would rename the exported helper function to `testModuleExports` (mainly because we are testing what we do export when web-ext is loaded as a module).

tests/fixtures/webext-as-library/test-import.mjs
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line import/no-unresolved
+import webExt from 'web-ext';
+
+// eslint-disable-next-line import/extensions
+import _ from './helpers.js';
```
rpl, line 5: > @rpl I have a doubt about imports:
> When I tried importing helper.js the following way:
>
> import _ from './helpers';
>
> It gave me an error and the test of importing web-ext as ESM module failed
> But when I used:
>
> import _ from './helpers.js';
>
> the test passes.
> Why is that so?
> Edit: I did some reading on it. So to my understanding first import statement fails because it looks for helpers directory instead of the file. Please correct me if I am wrong.

The reason behind that is likely the nodejs conventions related to ESM modules, in short nodejs looks by default for files with extension `.mjs` when we use `import` instead of `require` (I think you may find some more details about in the original announcement here: https://nodejs.medium.com/announcing-a-new-experimental-modules-1be8d2d6c2ff)

and so:
`import ... from './helper'`
I guess that would be looking for `./helpers.mjs` and `./helpers/index.mjs`, and it fails if neither of them is found.

on the contrary `import ... from './helpers.js'` is explicitly specifying the extension and will override that convention.

tests/fixtures/webext-as-library/test-import.mjs
```diff
@@ -0,0 +1,7 @@
...
+
+_.testWebExtModuleImports(webExt);
```
rpl, line 7: Nit, I would slightly prefer something like:
```
// eslint-disable-next-line import/extensions
import helpers from './helpers.js';

const {testModuleExports} = helpers;

testModuleExports(webExt);
```

tests/fixtures/webext-as-library/test-require.js
```diff
@@ -0,0 +1,5 @@
+const webExt = require('web-ext');
+
+const _ = require('./helpers.js');
+
+_.testWebExtModuleImports(webExt);
```
rpl, line 5: Nit, and here I would slightly prefer:

```
const {testModuleExports} = require('./helpers');

testModuleExports(webExt);
```

tests/functional/common.js
```diff
@@ -21,9 +21,9 @@ export const fixturesDir: string =
 export const minimalAddonPath: string =
   path.join(fixturesDir, 'minimal-web-ext');
 export const fixtureEsmImport: string =
-  path.join(fixturesDir, 'import-as-esm');
+  path.join(fixturesDir, 'webext-as-library');
 export const fixtureCjsRequire: string =
-  path.join(fixturesDir, 'require-as-cjs');
+  path.join(fixturesDir, 'webext-as-library');
```
rpl, line 26: Nit, these two have now both the same value, I guess we can just export one (e.g. named `fixturesUseAsLibrary`).

src/util/adb.js
```diff
@@ -418,3 +418,16 @@ export default class ADBUtils {
...
+  return adbClient.discoverDevices();
+}
+
+export async function listADBFirefoxAPKs(
+  deviceId: string,
+  adbBin?: string
+): Promise<Array<string>> {
+  const adbClient = new ADBUtils({adbBin});
+  return adbClient.discoverInstalledFirefoxAPKs(deviceId);
+}
```
rpl, line 433: Let's add also 2 new unit tests in [test.adb.js](https://github.com/mozilla/web-ext/blob/master/tests/unit/test-util/test.adb.js) to cover their behavior (we may use sinon.stub to stub `ADBUtils.discoverDevices` and `adbclient.discoverInstalledFirefoxAPKs` to cover them with a unit test)

tests/fixtures/webext-as-library/helpers.js
```diff
@@ -0,0 +1,15 @@
...
+  assert.deepEqual(Object.keys(webExt).sort(), ['cmd', 'main', 'util'].sort());
+  assert.deepEqual(Object.keys(webExt.util).sort(), ['logger', 'adb'].sort());
+  assert.deepEqual(
+    Object.keys(webExt.util.adb).sort(),
+    ['listADBDevices', 'listADBFirefoxAPKs'].sort(),
+  );
+  assert.equal(typeof webExt.cmd.run, 'function');
```
rpl, line 10: Nit, I would move this at line 6 (just to "visually" keep the assertions related and the ones unrelated with adb in their separate group of assertions)

rpl (changes requested):
@ankushduacodes follows a new round of review comments (you may also remove the Draft from this PR, it is on the right track, and also actively developed and reviewed) with some more details about how to cover with the functional test the expected lazy loading behavior of the 'adb.js' module (plus some other nits and a comment related to some more test coverage needed in this pull request).

Let me know how it goes (and if you have any further doubts or questions).

tests/fixtures/webext-as-library/helpers.js
ankushduacodes: @rpl I am so sorry for not noticing this. I feel so dumb sometimes. Thank you for guiding me the correct way.

ankushduacodes: > (You may already figured this out, but in case you didn't try yep, you have to set the CI environment var to true to make the test.lib.imports.js test to run:CI=true npm run test-function, otherwise it is skipped and only run in the CI jobs, we do this because it does npm link/npm unlink the web-ext source repository to run that test scenario).

On my machine(mac), even if I set CI to true, `npm unlink` and `npm link` does not run as they require elevated permissions. Even when I first installed web-ext on my machine,  it required me to run `sudo npm link`.

tests/fixtures/webext-as-library/helpers.js
```diff
@@ -1,15 +1,34 @@
...
+    (filePath) => filePath.includes('@devicefarmer/adbkit')
+  ).length > 0;
+
+  const msg = expectLoaded
+    ? 'adb module should have been loaded'
+    : 'adb module should not be loaded yet';
+  // eslint-disable-next-line no-console
+  console.log(Object.keys(require.cache).filter(
+    (filePath) => filePath.includes('@devicefarmer/adbkit')
+  ));
```
rpl, line 27: @ankushduacodes I guess that the console.log here is meant to help investigating why the test is failing when running on windows, right?

The string logged here to stdout is not visible yet in the CI job failure because it is called in the child process that `test.lib.imports.js` is spawning using [`child_process.execFileSync`](https://nodejs.org/api/child_process.html#child_process_child_process_execfilesync_file_args_options) nodejs API method and only the errors are going to be printed by default by the parent process, quoting from the docs linked above:
> stderr by default will be output to the parent process' stderr...

A quick way to make sure that the child process stdout is also printed automatically on the parent process stdout is to change those calls temporarily to include `stdio: 'inherit'` ([see options.stdio doc here to know more about the values allowed for this option](https://nodejs.org/api/child_process.html#child_process_options_stdio)), that should be enough to let us see this debugging logs in the circleci job logs:

```
      execFileSync(node, ['--experimental-modules', 'test-import.mjs'], {
        cwd: tmpDir.path(),
        stdio: 'inherit',
      });
```

would you mind to try this so that we may have some more details to figure out why the test is failing on windows?
(if even this doesn't provide us enough details then I'll reboot one of my machine and try it out on a local windows system)

rpl:
> On my machine(mac), even if I set CI to true, npm unlink and npm link does not run as they require elevated permissions. Even when I first installed web-ext on my machine, it required me to run sudo npm link.

@ankushduacodes ah, yeah, that's likely because you have nodejs installed at system level, on the contrary I use nvm (to move easily switch between nodejs version when I need to) and so my nodejs environment can be changed using my user permission.

You may try to use nvm (it is pretty nice) or you may use the trick we are using in the circleci jobs to solve a similar issue, which is basically configuring the repo to use an npm global directory that can be changed with user permissions, see this part of the circleci job:
https://github.com/mozilla/web-ext/blob/be598cc34772007cdc1177c9c42226a151572ba7/.circleci/config.yml#L151-L158

The first command there is creating the directory where the npm global directory should be, the second command is executed in the repo and points npm to that directory as the global npm directory to use for installing / uninstalling / linking / unlinking npm packages from. 
(don't forget to revert that configuration when you don't need / don't want it anymore)

tests/fixtures/webext-as-library/helpers.js
ankushduacodes: @rpl I was setting up sideloading  windows on my machine to test this out. And I have set it up successfully, I will update you soon on whats going on.

ankushduacodes: @rpl The issue here seems to be with the difference in the way paths are written in UNIX/LINUX ie. using forward slash `/` but on windows we use backward slash `\` which then needs to escaped using another backward slash.
Please check out the following logs:
```
PS C:\Users\ankus\Documents\GitHub\web-ext> node
Welcome to Node.js v14.15.4.
Type ".help" for more information.
> const webExt = require('./dist/web-ext')
undefined
> Object.keys(require.cache)
[
  'C:\\Users\\ankus\\Documents\\GitHub\\web-ext\\dist\\web-ext.js',
  'C:\\Users\\ankus\\Documents\\GitHub\\web-ext\\node_modules\\source-map-support\\source-map-support.js',
  ...
]
> webExt.util.adb
{
  listADBDevices: [AsyncFunction: listADBDevices],
  listADBFirefoxAPKs: [AsyncFunction: listADBFirefoxAPKs]
}
> Object.keys(require.cache).filter(filepath => filepath.includes('@devicefarmer\\adbkit'))
[
  'C:\\Users\\ankus\\Documents\\GitHub\\web-ext\\node_modules\\@devicefarmer\\adbkit\\index.js',
  'C:\\Users\\ankus\\Documents\\GitHub\\web-ext\\node_modules\\@devicefarmer\\adbkit\\lib\\adb.js',
  ...
]
```

src/util/adb.js
ankushduacodes: Hi @rpl, I have added two new tests to [test.adb.js here](https://github.com/ankushduacodes/web-ext/blob/49f1e7175c3fb37b4fae7507afafe6bf108fb8f1/tests/unit/test-util/test.adb.js#L1314-L1340).

Please let me know if I am doing something wrong there.

Thank you !...

ankushduacodes:
@rpl I have tried to update the documentation on `how to run an extension on android` as we talked about on addons matrix channel, but I am not really happy with adding all those comments. Do you have any suggestions on how to write that?

tests/fixtures/webext-as-library/helpers.js
rpl: @ankushduacodes thanks for investigating that windows failure, what you described does definitely make sense and expected.

Instead of conditioning the string we look for in the cache based on the current platform, could we use path.join to get the platform specific format on both the platforms? something like this:

```
const path = require('path');
const adbPathString = path.join('@devidefarmer', 'adbkit');
...
```

would you mind to give that a try?

> (PS: Sorry for delay in response past few days, I got caught up in school work as my finals will be starting soon, I hope you understand :)).

Don't worry at all, school work comes first!
And anyway in my opinion you are doing great (you are always following up to what we discuss and agree on, and you have been pretty responsive from my perspective). I'm really enjoying working with you.

README.md
```diff
@@ -122,9 +122,16 @@ You are able to execute command functions without any argument validation. If yo
 import webExt from 'web-ext';
 
 webExt.cmd.run({
+  // To run an extension on android, uncomment next three lines
```
rpl, line 125: > @rpl I have tried to update the documentation on `how to run an extension on android` as we talked about on addons matrix channel, but I am not really happy with adding all those comments. Do you have any suggestions on how to write that?

@ankushduacodes How about moving everything related to android (this commented fragments and the new parts describing the two adb utility methods) in a "single paragraph + code snipped" at the end of this section?

I think that something like the following may be enough for now:

```
If you would like to run an extension on Firefox for Android:

// Path to adb binary (optional parameter, auto-detected if missing)
const adbBin = ...
// Get an array of device ids (Array<string>)
const deviceIds = await webExt.util.adb.listADBDevices(adbBin);
const adbDevice = ...
// Get an array of Firefox APKs (Array<string>)
const firefoxAPKs = await webExt.util.adb.listADBFirefoxAPKs(
  deviceId, adbBin
);
const firefoxApk = ...

webExt.cmd.run({
  target: 'firefox-android',
  firefoxApk,
  adbDevice,
  sourceDir: ...
}).then((extensionRunner) => ...);
```

How that sounds to you?

tests/fixtures/webext-as-library/helpers.js
ankushduacodes: > Instead of conditioning the string we look for in the cache based on the current platform, could we use path.join to get the platform specific format on both the platforms? something like this:
> 
> ```
> const path = require('path');
> const adbPathString = path.join('@devidefarmer', 'adbkit');
> ...
> ```
> 
> would you mind to give that a try?

For sure!..

> Don't worry at all, school work comes first!

Thank you for your understanding.

> And anyway in my opinion you are doing great (you are always following up to what we discuss and agree on, and you have been pretty responsive from my perspective). I'm really enjoying working with you.

And Thank you for the kind words. I really appreciate that ❤️. I really enjoy working with you as well.

README.md
ankushduacodes: Using await means we will have to add an async function to make it work, no?

rpl: @ankushduacodes if I recall correctly nodejs will support top level `await` starting from nodejs 14, anyway I wouldn't bother to explicitly mention that here (I guess that who is using web-ext as a library is going to be a developer, and we may just assume that they will know how to use `await` in the nodejs version they are targeting).

ankushduacodes: @rpl I have updated the docs, Please have a look and let me know there needs to be any change.

Also, are two new tests looking alright?

README.md
```diff
@@ -121,11 +121,24 @@ You are able to execute command functions without any argument validation. If yo
...
+
 webExt.cmd.run({
-  // To run an extension on android, uncomment next three lines
-  // target: 'firefox-android',
-  // firefoxApk: 'firefox apk package name',
-  // adbDevice: '<deviceId>',
+  // Comment out next 3 lines to run the extension on firefox-desktop
+  target: 'firefox-android',
+  firefoxApk
+  adbDevice
```
rpl: @ankushduacodes I'm sorry, I think I wasn't clear enough in my previous comment about the addition I was thinking of to document this, I meant to add it separately from this fragment, which I think should stay minimal as it was before this change and the new snippet for android at the end of the section, here is a diff that apply the change I had in mind on top of this version of the pull request:

https://gist.github.com/rpl/1e5c11535c686c7e5be6cfead339c865#file-pr-2116-patch-for-changes-to-readme-patch

rpl (changes requested):

tests/unit/test-util/test.adb.js
```diff
@@ -1310,4 +1311,32 @@ describe('utils/adb', () => {
     });
   });
 
+  describe('test exposed adb API functions', () => {
```
rpl, line 1314: Nit, `test exposed ...` -> `exports exposed in util.adb` (or something like that )

tests/unit/test-util/test.adb.js
```diff
@@ -1310,4 +1311,32 @@ describe('utils/adb', () => {
...
+        ADBUtils.prototype, 'discoverInstalledFirefoxAPKs'
+      );
+      stubDiscoverInstalledFirefoxAPKs
+        .resolves(['package1', 'package2', 'package3']);
+      const promise = listADBFirefoxAPKs('device1');
+      const packages = await assert.isFulfilled(promise);
+      sinon.assert.calledWith(
+        new ADBUtils({adb}).discoverInstalledFirefoxAPKs,
+        'device1'
+      );
```
rpl, line 1337: I think this can (and should) be actually asserting on `stubDiscoverInstalledFirefoxAPKs` and so this one should be:
```
sinon.assert.calledWith(
        stubDiscoverInstalledFirefoxAPKs,
        'device1'
      );
```

and we should also be able to remove `const adb = getFakeADBKit({});` that is at line 1326, because it is going to not be needed after this change.

Let me know if for any reason this change doesn't work and the test fails (but I think that it should work just fine).

tests/unit/test.web-ext.js
```diff
@@ -17,6 +18,14 @@ describe('webExt', () => {
...
   });
 
+  it('gives access to listADBDevices', () => {
+    assert.equal(webExt.util.adb.listADBDevices, listADBDevices);
+  });
+
+  it('gives access to listADBFirefoxAPKs', () => {
+    assert.equal(webExt.util.adb.listADBFirefoxAPKs, listADBFirefoxAPKs);
+  });
+
```
rpl, line 28: Nit, Let's wrap these two tests into a `descrive('exposes adb utils', () =>{ ... });`.

rpl (changes requested):
@ankushduacodes this PR looks pretty good, it should be almost ready. Besides the previous command about the changes to the README.md file (https://github.com/mozilla/web-ext/pull/2116#pullrequestreview-574434176) the review comments that follows are related to a few more nits (that shouldn't require any big changes).

src/util/adb.js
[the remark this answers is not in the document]
ankushduacodes: What could be the strategy here, for my own curiosity?

README.md
ankushduacodes: @rpl I am sorry for not understanding prior comment correctly. I have updated the readme with the changes suggests and some additional aesthetic changes.
Please let me know if how it's looking after the change.

README.md
```diff
@@ -162,16 +143,31 @@ webExt.cmd.run({
...
-To get a list of installed firefox APK packages for an ADB client, you may use the following function which takes in deviceId(required) and adbBin(optional) as arguments:
-```js
-webExt.util.adb.listADBFirefoxAPKs(deviceId); // returns a Promise<Array<string>>
+webExt.cmd.run({
+  target: 'firefox-android',
+  firefoxApk,
+  adbDevice,
+  sourceDir: ...
+}).then((extensionRunner) => {...});
 ```
```
rpl: By adding the section title `#### running the extension on Firefox for android` right before the new code snippet, everything after it will look like part of this particular subsection, but the other code snippets are not strictly related to this section.

To avoid reworking the README too much in this PR just for the small addition we care about, I would suggest to avoid the new section title and just put right before the snippet just that small intro paragraph originally suggested.

We may definitely evaluate to make this API doc section a bit easier to navigate, but let's do that in a follow up (mainly to ensure this is ready to merge to the main branch sooner, the other changes in this PR have higher chances to get conflicts with other PRs we may merge in the main time).

README.md
```diff
@@ -162,16 +143,31 @@ webExt.cmd.run({
...
 
+
+
+
+
```
rpl, line 170: One empty line between the end of the code snippet and the start of the paragraph that follows is enough, let's remove the additional empty lines.

tests/unit/test-util/test.adb.js
```diff
@@ -1311,8 +1311,8 @@ describe('utils/adb', () => {
     });
   });
 
-  describe('test exposed adb API functions', () => {
-    it('test listADBDevices', async () => {
+  describe('exports exposed in util.adb', () => {
+    it('lists connected adb devices', async () => {
```
rpl, line 1315: Nit (mainly because this PR would be updated for the changes to the right, and while we are still here let's also do some nits), `it('should export a listADBDevices method', ...)`.

tests/unit/test-util/test.adb.js
```diff
@@ -1322,8 +1322,7 @@ describe('utils/adb', () => {
       assert.deepEqual(devices, ['emulator1', 'device2']);
     });
 
-    it('test listADBFirefoxAPKs', async () => {
-      const adb = getFakeADBKit({});
+    it('lists installed firefox apks on a device', async () => {
```
rpl, line 1325: Nit (mainly because this PR would be updated for the changes to the right, and while we are still here let's also do some nits), `it('should export a listADBFirefoxAPKs method', ...)`.

rpl (changes requested):
@ankushduacodes I'm sorry, this PR does require some more tweaks on the changes to the readme (and the other two nits are just a couple of "nice if tweaked", given that we are going to update this PR for the changes to the readme).

Everything else looks great and ready to be merged.

src/main.js
```diff
@@ -3,8 +3,17 @@ import {main} from './program';
...
+// This only exposes util/logger and a subset of util/adb so far.
 // Do we need anything else?
-const util = {logger};
+const util = {
+  logger,
+  // Lazy load the adb module when util.adb is accessed for the first time, to defer loading
+  // the npm dependencies used by the adb module to when actually needed.
+  // This is being done to continue the courtesy of web-ext issue #1301
+  get adb(): Object {
+    const {listADBDevices, listADBFirefoxAPKs} = require('./util/adb.js');
```
rpl, line 14: @ankushduacodes Nit (sorry I missed this one in my last review pass), I think we can omit `.js` from this `require`, right?

src/util/adb.js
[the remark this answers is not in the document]
rpl: @ankushduacodes I haven't thought about it in too much detail, but I think that the cleanest way would be to change the API we export e.g.

```
const webExt = require('web-ext');
const adbBin = ...;
const adbClient = webExt.util.adb.getClient({adbBin});
...
const adbDevices = await adbClient.listDevices();
const firefoxAPKs = await adbClient.listFirefoxAPKs(adbDevices[0]);
```

(btw I :heart: curiosity ;-), never stop asking question out of curiosity!)

src/main.js
ankushduacodes: Yes, it works. But does mentioning `.js` look more verbose here? as it indicates that it not a directory?

rpl: > Yes, it works. But does mentioning `.js` look more verbose here? as it indicates that it not a directory?

well, we are not including the extension in the imports we have in the same file and so omitting it from the require would be good for consistency (the only case when we want to include the file extension explicitly is when it doesn't work without it, as in `tests/fixtures/webext-as-library/test-import.mjs` file in this patch)

we don't care if `util/adb` turns out to be a directory with an `index.js` file in it, on the contrary if we would ever decide to split the `adb` module into multiple files then we would need to change the `require` here is we don't want to.

README.md
```diff
@@ -122,9 +122,9 @@ You are able to execute command functions without any argument validation. If yo
 import webExt from 'web-ext';
 
 webExt.cmd.run({
+  firefox: '/path/to/Firefox-executable',
```
rpl, line 125: Nit, let's revert this unrelated change, this line was right above `sourceDir: '...'`, at line 127.

README.md
```diff
@@ -142,6 +142,27 @@ webExt.cmd.run({
   });
 ```
 
+If you would like to run an extension on Firefox for Android:
+```js
```
rpl, line 146: Nit, one empty line between the snippet and the paragraph right before it (it doesn't change the markdown rendering, it is just consistent with the other ones that follows)

tests/functional/test.lib.imports.js
```diff
@@ -35,19 +35,21 @@ describe('web-ext imported as a library', () => {
   it('can be imported as an ESM module', async () => {
     await withTempDir(async (tmpDir) => {
       execFileSync(npm, ['link', 'web-ext'], {cwd: tmpDir.path()});
-      shell.cp('-rf', `${fixtureEsmImport}/*`, tmpDir.path());
+      shell.cp('-rf', `${fixturesUseAsLibrary}/*`, tmpDir.path());
       execFileSync(node, ['--experimental-modules', 'test-import.mjs'], {
         cwd: tmpDir.path(),
+        stdio: 'inherit',
```
rpl, line 41: Nit, do we need to keep `stdio: 'inherit'` here (and at line 52)? if I recall correctly the child process stderr would still be echoed by the parent process and so the assertion errors would be visible in case of failure (and we don't log anything to stdout that we really need to see, besides while we were investigating the windows issue and we wanted to see a console.log from the child process as part of the tests output).

If that is the case let's remove `stdio: 'inherit'`.

rpl (approved):
@ankushduacodes thanks a lot for the work on this enhancement (and for sticking with me during the multiple discussions about implementation strategies,  multiple review passes and CI failures to investigate). 

There are 3 small nits described below that I would like to be applied in this patch, but feel free to let me know if you already moved to work on something else and you can't apply them, I'll be happy to make those small changes and merge this (obviously keeping the author metadata as is ;-))

ankushduacodes:
On it. Thank you.

tests/functional/test.lib.imports.js
ankushduacodes: Yep, you are right, removing it.

rpl:
> On it. Thank you.
> 
> (PS: If there is anything left, do let me know. I don't delete branches until after a PR is merged. :))

Thanks @ankushduacodes for handling the last few nits as well, I just merged this on the main branch! :rocket: \o/

ankushduacodes:
Thank you @rpl for the guidance along the way and patiently explaining my (sometimes silly) doubts. I am learning so much. Many more to go.
cheers 🚀 🎉

PR #2121, opened 2021-01-12
fix: Replace AMO API v3 with v4

### LINKED ISSUE #2120: Upgrade web-ext to use AMO API v4
v3 of AMO's API is being deprecated so web-ext should switch to v4.

Rob--W:
Fixes #2120.


rpl (approved):
lgtm. Thanks @Rob--W

PR #2124, opened 2021-01-15
test: Skip unit tests for git commit info in web-ext version when tests are not running from a git repository

yan12125:
For example, from a GitHub source tarball archive

Background: I'm the maintainer of [web-ext](https://archlinux.org/packages/community/any/web-ext/) package on Arch Linux, and I chose GitHub source tarballs (e.g., https://github.com/mozilla/web-ext/archive/5.5.0/web-ext-5.5.0.tar.gz) instead of the full git repository for building that package for saving disk space and a little time when building on Arch Linux's shared build server.

### DIFF
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -911,10 +911,15 @@ describe('program.defaultVersionGetter', () => {
   });
 
   it('returns git commit information in development', function() {
-    const commit = `${git.branch()}-${git.long()}`;
-    const testBuildEnv = {globalEnv: 'development'};
-    assert.equal(defaultVersionGetter(projectRoot, testBuildEnv),
-                 commit);
+    return fs.exists(path.join(projectRoot, '.git')).then((exists) => {
+      if (!exists) {
+        this.skip();
+      }
+      const commit = `${git.branch()}-${git.long()}`;
+      const testBuildEnv = {globalEnv: 'development'};
+      assert.equal(defaultVersionGetter(projectRoot, testBuildEnv),
+                   commit);
+    });
   });
 });
 


tests/unit/test.program.js
```diff
@@ -911,7 +911,15 @@ describe('program.defaultVersionGetter', () => {
   });
 
   it('returns git commit information in development', function() {
-    const commit = `${git.branch()}-${git.long()}`;
+    var commit;
```
rpl, line 914: could we check if the .git directory does actually exist (at the start of this test case) and skip it if the .git directory doesn't exist instead?

rpl (changes requested):

tests/unit/test.program.js
yan12125: Thanks for the suggestion. It sounds better and I use that approach in the latest revision of this patch.

Rob--W:
What guarantees that the tests are run from the repository's root?

If there is none, resolve the directory to an absolute path, e.g. relative to `__dirname`.

yan12125:
> If there is none, resolve the directory to an absolute path, e.g. relative to `__dirname`.

Thanks good suggestion! Updated.

Rob--W (approved):

rpl (approved):
lgtm, thanks!

yan12125:
Thanks for kind reviews!

caitmuenster:
Thanks so much, @yan12125! ❤️ We've added your contribution to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Hope to see you around the project in the future!

PR #2125, opened 2021-01-15
feat: enabled watching multiple files via `--watch-file`

### LINKED ISSUE #2104: Error while specifying multiple --watch-file options
Enhancement

I was trying to test the platform to add some documentation for --watch-file option over at extension-workshop and tried to specify two --watch-file with commands:
```bash
web-ext run -s ~/Documents/GitHub/web-ext/borderify/ -t firefox-desktop --watch-file ~/Documents/GitHub/web-ext/borderify/_.js --watch-file ~/Documents/GitHub/web-ext/borderify/borderify.js
```
and got an error
``` bash
Running web extension from /Users/ankushdua/Documents/GitHub/web-ext/borderify
Installed /Users/ankushdua/Documents/GitHub/web-ext/borderify as a temporary add-on

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
    at validateString (internal/validators.js:124:11)
    at Object.dirname (path.js:1128:5)
    at WatcherManager.watchFile (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/watchpack/lib/watcherManager.js:27:23)
    at Watchpack.<anonymous> (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/watchpack/lib/watchpack.js:38:49)
    at Array.map (<anonymous>)
    at Watchpack.watch (/Users/ankushdua/Documents/GitHub/web-ext/node_modules/watchpack/lib/watchpack.js:37:28)
    at onSourceChange (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/watcher.js:76:11)
    at createWatcher (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/extension-runners/index.js:263:10)
    at reloadStrategy (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/extension-runners/index.js:313:30)
    at run (/Users/ankushdua/Documents/GitHub/web-ext/dist/webpack:/src/cmd/run.js:249:5)

Error code: ERR_INVALID_ARG_TYPE

```

Is it possible to specify multiple files? if not, shouldn't this be good to have that as an option?
Am I doing something wrong here?

* **Firefox version**: Mozilla Firefox (84.0.1)
* **Your OS and version**: MacOS 11.0.1

````
````
v14.15.0
6.14.8
master-b3f590bd1cd9ac8a2de70b289bd7dac6645c89f0

### PR DESCRIPTION
**Pull request by Dexterp37**
Fixes #2104 

This allows passing multiple paths to `--watch-file` in order for `web-ext`. It tweaks the current tests to address the changes and make sure they still pass.

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -18,7 +18,7 @@ export type OnChangeFn = () => any;
 
 export type OnSourceChangeParams = {|
   sourceDir: string,
-  watchFile?: string,
+  watchFile?: Array<string>,
   watchIgnored?: Array<string>,
   artifactsDir: string,
   onChange: OnChangeFn,
@@ -57,18 +57,22 @@ export default function onSourceChange(
     proxyFileChanges({artifactsDir, onChange, filePath, shouldWatchFile});
   });
 
-  log.debug(`Watching for file changes in ${watchFile || sourceDir}`);
+  log.debug(
+    `Watching ${watchFile ? watchFile.join(',') : sourceDir} for changes`
+  );
 
   const watchedDirs = [];
   const watchedFiles = [];
 
   if (watchFile) {
-    if (fs.existsSync(watchFile) && !fs.lstatSync(watchFile).isFile()) {
-      throw new UsageError('Invalid --watch-file value: ' +
-        `"${watchFile}" is not a file.`);
-    }
+    for (const filePath of watchFile) {
+      if (fs.existsSync(filePath) && !fs.lstatSync(filePath).isFile()) {
+        throw new UsageError('Invalid --watch-file value: ' +
+          `"${filePath}" is not a file.`);
+      }
 
-    watchedFiles.push(watchFile);
+      watchedFiles.push(filePath);
+    }
   } else {
     watchedDirs.push(sourceDir);
   }
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -38,7 +38,7 @@ export type CmdRunParams = {|
   noReload: boolean,
   preInstall: boolean,
   sourceDir: string,
-  watchFile?: string,
+  watchFile?: Array<string>,
   watchIgnored?: Array<string>,
   startUrl?: Array<string>,
   target?: Array<string>,
@@ -117,6 +117,11 @@ export default async function run(
     log.info('Disabled auto-reloading because it\'s not possible with ' +
              '--pre-install');
     noReload = true;
+  }
+
+  if (watchFile != null && (!Array.isArray(watchFile) ||
+      !watchFile.every((el) => typeof el === 'string'))) {
+    throw new UsageError('Unexpected watchFile type');
   }
 
   // Create an alias for --pref since it has been transformed into an
--- a/src/extension-runners/index.js
+++ b/src/extension-runners/index.js
@@ -239,7 +239,7 @@ export class MultiExtensionRunner {
 export type WatcherCreatorParams = {|
   reloadExtension: (string) => void,
   sourceDir: string,
-  watchFile?: string,
+  watchFile?: Array<string>,
   watchIgnored?: Array<string>,
   artifactsDir: string,
   onSourceChange?: OnSourceChangeFn,
@@ -276,7 +276,7 @@ export function defaultWatcherCreator(
 export type ReloadStrategyParams = {|
   extensionRunner: IExtensionRunner,
   sourceDir: string,
-  watchFile?: string,
+  watchFile?: Array<string>,
   watchIgnored?: Array<string>,
   artifactsDir: string,
   ignoreFiles?: Array<string>,
--- a/src/program.js
+++ b/src/program.js
@@ -619,7 +619,7 @@ Example: $0 --help run.
                   ' file changes. This is useful if you use a custom' +
                   ' build process for your extension',
         demandOption: false,
-        type: 'string',
+        type: 'array',
       },
       'watch-ignored': {
         describe: 'Paths and globs patterns that should not be ' +
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -161,12 +161,22 @@ describe('run', () => {
     }, {firefoxApp, firefoxClient});
   });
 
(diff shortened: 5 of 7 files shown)

### CODE REVIEW

File: src/watcher.js [hunk 1]
```diff
@@ -57,18 +57,22 @@ export default function onSourceChange(
...
 
   const watchedDirs = [];
   const watchedFiles = [];
 
   if (watchFile) {
-    if (fs.existsSync(watchFile) && !fs.lstatSync(watchFile).isFile()) {
-      throw new UsageError('Invalid --watch-file value: ' +
-        `"${watchFile}" is not a file.`);
-    }
+    for (const path of watchFile) {
```
rpl, line 68: @Dexterp37 would you mind to rename `const path` into `file` or `filePath` (or something similar and still short enough)? 

I think that it is pretty common in nodejs apps and scripts to use `path` as the name for the result of importing the 'path' nodejs module, and so I would prefer to avoid using it for the string of the file path.

rpl (changes requested):
@Dexterp37 my apologies, I really wanted to get to this sooner.

Anyway, this PR looks pretty great to me, I do have only one small request for changes (the small nit described below), would you mind to also remove the draft state from it after your next update on it?

I'm going to also give it a try locally, just in case I may notice something more that I missed by just looking to the diff in this PR.

(let me know if you are too busy to come back to update this PR, I'm ok to make that change and then land this for you).

rpl:
Note to self (and for @Dexterp37 because this may require some more changes):

- by changing the expected type of the watchFile option, there is a chance that we may need to also consider what will happen if a user do have a web-ext config file where that option is defined as a string
- if we will have to handle the potential issue describe above, then we may either opt to auto convert the option from string to array while loading the file or make sure we raise an error that makes it clear for the user what needs to be changed to fix it

src/watcher.js
```diff
@@ -57,18 +57,22 @@ export default function onSourceChange(
...
-    }
+    for (const filePath of watchFile) {
+      if (fs.existsSync(filePath) && !fs.lstatSync(filePath).isFile()) {
+        throw new UsageError('Invalid --watch-file value: ' +
+          `"${filePath}" is not a file.`);
+      }
 
-    watchedFiles.push(watchFile);
+      watchedFiles.push(filePath);
+    }
```
rpl, line 75: While looking into some of the corner cases that may have not been covered yet, I did notice another detail that may be worth to cover explicitly:

If `web-ext` is imported as a library in a nodejs script, the caller may be passing `watchFile` as a string and no runtime check would prevent that to reach this part with a string value (assuming that the developer using `web-ext` isn't using flow and so it wouldn't be caught by the flow type checks).

In that case I think that no error would be detected because of the unexpected watchFile type, but errors will likely be raised because we would be trying to watch files named as every single letter of the watchFile string.

I think that it may be reasonable to do an explicit runtime check in `src/cmd/run.js` and throw an explicit error, which will also make us notice that there is a test case in "test.run.js" that was still passing the watchFile option as a string.

Here are the changes that I did apply locally to cover this corner case, would you mind to merge them into your PR?

```diff
diff --git a/src/cmd/run.js b/src/cmd/run.js
index 6a820dc..5334554 100644
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -119,6 +119,11 @@ export default async function run(
     noReload = true;
   }
 
+  if (watchFile != null && (!Array.isArray(watchFile) ||
+      !watchFile.every((el) => typeof el === 'string'))) {
+    throw new UsageError('Unexpected watchFile type');
+  }
+
   // Create an alias for --pref since it has been transformed into an
   // object containing one or more preferences.
   const customPrefs = pref;
diff --git a/tests/unit/test-cmd/test.run.js b/tests/unit/test-cmd/test.run.js
index f0648a1..5171025 100644
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -161,12 +161,22 @@ describe('run', () => {
     }, {firefoxApp, firefoxClient});
   });
 
+  it('throws if watchFile is not an array', async () => {
+    const cmd = prepareRun();
+    await assert.isRejected(
+      cmd.run({noReload: false, watchFile: 'invalid-value.txt' }),
+      /Unexpected watchFile type/
+    );
+  });
+
   it('can watch and reload the extension', async () => {
     const cmd = prepareRun();
     const {sourceDir, artifactsDir} = cmd.argv;
     const {reloadStrategy} = cmd.options;
 
-    const watchFile = fixturePath('minimal-web-ext', 'manifest.json');
+    const watchFile = [
+      fixturePath('minimal-web-ext', 'manifest.json'),
+    ];
 
     await cmd.run({noReload: false, watchFile });
     assert.equal(reloadStrategy.called, true);
```

rpl (changes requested):
@Dexterp37 I did a final review pass on this, and double-checked those other two corner cases I did mention in my previous comments:
-  behavior on invalid type of the `watchFile` option when it is coming from a web-ext config file
  - without any further changes, we would show to the user the following error message (which seems pretty understandable as it is, we can re-evaluate if we want to turn it into a warning and auto-convert the option value from `src/config.jg` as a followup before releasing the next major release, but in my opinion isn't a blocking issue for this PR): 
    > UsageError: The config file at /home/rpl/mozlab/projects/webext-tools/web-ext/test-config.js specified the type of "watchFile" incorrectly as "string" (expected type "array")  
-  behavior when watchFile option value is coming from the `WEB_EXT_WATCH_FILE` environment variable:
   - yargs does already convert automatically `WEB_EXT_WATCH_FILE="a-file.txt"` into `watchFile = ["a-file.txt"]`, it wouldn't support passing more than one but that is an upstream yargs issue and it wouldn't impact backward compatibility at all.

If you don't mind, I would like to cover the third corner case that is described in the inline review comment below, after that we can proceed to merge this to the main branch.

src/watcher.js
Dexterp37: Thanks! I added this code to the PR:)

rpl (approved):
Thanks a lot @Dexterp37, this looks ready and I'm going to merge it to the main branch asap.

PR #2129, opened 2021-01-19
fix: Replaced unmaintained dependencies firefox-client and node-firefox-connect

rpl:
This PR contains a new `src/firefox/rdp-client.js` module that replaces the subset of feature we do use in web-ext from the unmaintained dependencies firefox-client and node-firefox-connect (and additional  unit tests to cover this new module).

By removing the firefox-client npm package from the dependencies we do also avoid one of the deprecation warnings related to `js-select@0.6.0` being logged when the web-ext package is installed by a user:
```
npm WARN deprecated js-select@0.6.0: Package no longer supported. Contact support@npmjs.com for more info.
```

### DIFF
--- a/src/firefox/rdp-client.js
+++ b/src/firefox/rdp-client.js
@@ -1,0 +1,299 @@
+/* @flow */
+import net from 'net';
+import EventEmitter from 'events';
+import domain from 'domain';
+
+export type RDPRequest = {
+  to: string,
+  type: string,
+}
+
+export type RDPResult = {
+  from: string,
+  type: string,
+}
+
+export type Deferred = {|
+  resolve: Function,
+  reject: Function,
+|}
+
+type ParseResult = {|
+  data: Buffer,
+  rdpMessage?: Object,
+  error?: Error,
+  fatal?: boolean,
+|}
+
+export const DEFAULT_PORT = 6000;
+export const DEFAULT_HOST = '127.0.0.1';
+
+const UNSOLICITED_EVENTS = new Set([
+  'tabNavigated',
+  'styleApplied',
+  'propertyChange',
+  'networkEventUpdate',
+  'networkEvent',
+  'propertyChange',
+  'newMutations',
+  'frameUpdate',
+  'tabListChanged',
+]);
+
+// Parse RDP packets: BYTE_LENGTH + ':' + DATA.
+export function parseRDPMessage(data: Buffer): ParseResult {
+  const str = data.toString();
+  const sepIdx = str.indexOf(':');
+  if (sepIdx < 1) {
+    return {data};
+  }
+
+  const byteLen = parseInt(str.slice(0, sepIdx));
+  if (isNaN(byteLen)) {
+    const error = new Error('Error parsing RDP message length');
+    return {data, error, fatal: true};
+  }
+
+  if (data.length - (sepIdx + 1) < byteLen) {
+    // Can't parse yet, will retry once more data has been received.
+    return {data};
+  }
+
+  data = data.slice(sepIdx + 1);
+  const msg = data.slice(0, byteLen);
+  data = data.slice(byteLen);
+
+  try {
+    return {data, rdpMessage: JSON.parse(msg.toString())};
+  } catch (error) {
+    return {data, error, fatal: false};
+  }
+}
+
+export function connectToFirefox(port: number): Promise<FirefoxRDPClient> {
+  const client = new FirefoxRDPClient();
+  return client.connect(port).then(() => client);
+}
+
+export default class FirefoxRDPClient extends EventEmitter {
+  _incoming: Buffer;
+  _pending: Array<{| request: RDPRequest, deferred: Deferred |}>;
+  _active: Map<string, Deferred>;
+  _rdpConnection: net.Socket;
+  _onData: Function;
+  _onError: Function;
+  _onEnd: Function;
+  _onTimeout: Function;
+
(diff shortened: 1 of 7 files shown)


rpl:
@Rob--W I just did the split of the patches in this PR as we agreed yesterday:
- 1st patch (4aafe2a) introduce the new `rdp-client` module and the new unit tests to cover that module
- 2nd patch (cd932eb) replace in the `firefox/remote` module the old npm dependencies with the `rdp-client` module and does tweak the tests accordingly
- the last patch does remove the unused dependencies from package.json and package-lock.json files (I kept it separated to more easily drop it and update it when the package.json and lock files are changed on the main branch due to merged renovate bot PRs)

(as a side note, the coverage data for the file on the codecov side reports 100% coverage for it: https://codecov.io/gh/mozilla/web-ext/tree/8bcfb1e288cf50f9b7a51e78d715e3df476f6d46/src/firefox, but codecov is currently unable to show the coverage data on top of the content of the file, there is some additional configuration needed to fix that and I filed #2134 to take care of that).

src/firefox/rdp-client.js
```diff
@@ -0,0 +1,303 @@
...
+        conn.on('timeout', this._onTimeout);
+
+        // Resolve once the expected initial root message
+        // has been received.
+        this._expectReply('root', {resolve, reject});
+      });
+    });
+  }
+
+  disconnect(): Promise<void> {
```
Rob--W, line 127: Why is `disconnect` async when the implementation is not? There are some callers (in the implementation and in the test) that treat `disconnect` as synchronous, while others do not treat it as async. I suggest to be consistent.

At this point, I guess that it's better to make it synchronous, as there is no wait for the underlying connection to really be closed.

src/firefox/remote.js
```diff
@@ -43,133 +41,138 @@ export type FirefoxRDPResponseAny = Object;
...
-      })
-      .then((addons) => {
-        for (const addon of addons) {
-          if (addon.id === addonId) {
-            return addon;
-          }
+  async getInstalledAddon(addonId: string): Promise<FirefoxRDPAddonActor> {
+    try {
+      const response = await this.client.request('listAddons');
+      if (response.addons == null) {
```
Rob--W, line 156: Why this explicit check? This code shouldn't ever be reached, the backend should send the `noAddons` error if there are no add-ons, and otherwise the error becomes `Remote Firefox: listAddons() error: addons is not iterable`, which is sufficiently clear.

tests/unit/helpers.js
```diff
@@ -207,47 +207,11 @@ export function createFakeProcess(): any {
  * Returns a fake Firefox client as would be returned by
  * connect() of 'node-firefox-connect'
```
Rob--W, line 208: `node-firefox-connect`? comment is stale

src/firefox/rdp-client.js
rpl: Fine by me.

I don't actually recall right now if I did make it async because it was in the origin firefox-client implementation or just because I was planning to make it await until the connection was really closed, anyway I'll just make it sync for now (and we can evaluate to make it async if we decide to let it wait for the actual disconnection)

src/firefox/remote.js
rpl: When I did read this comment I was convinced that I may have added the explicit check on `response.addons == null` because of flowtype errors, but flow didn't seem to care at the moment, and so I'm removing it (along with the related unit test)

tests/unit/test-firefox/test.rdp-client.js
```diff
@@ -0,0 +1,379 @@
...
+  describe('FirefoxRDPClient', () => {
+    let fakeRDPServer;
+    let client: FirefoxRDPClient;
+
+    beforeEach(() => {
+      fakeRDPServer = createFakeRDPServer();
+    });
+
+    afterEach(async () => {
+      await client?.disconnect();
```
Rob--W, line 117: No `await`.

tests/unit/test-firefox/test.rdp-client.js
```diff
@@ -0,0 +1,379 @@
...
+      // $FlowIgnore
+      c._onData = function fakeOnData() {};
+      // $FlowIgnore
+      c._onError = function fakeOnError() {};
+      // $FlowIgnore
+      c._onEnd = function fakeOnEnd() {};
+      // $FlowIgnore
+      c._onTimeout = function fakeOnTimeout() {};
+
+      await c.disconnect();
```
Rob--W, line 364: No `await`.

tests/unit/test-firefox/test.rdp-client.js
```diff
@@ -0,0 +1,379 @@
...
+      sinon.assert.calledOnce(fakeConn.end);
+      sinon.assert.callCount(fakeConn.off, 4);
+      sinon.assert.calledWith(fakeConn.off, 'data', c._onData);
+      sinon.assert.calledWith(fakeConn.off, 'error', c._onError);
+      sinon.assert.calledWith(fakeConn.off, 'end', c._onEnd);
+      sinon.assert.calledWith(fakeConn.off, 'timeout', c._onTimeout);
+    });
+
+    it('does resolve when not connected', async () => {
```
Rob--W, line 373: This test is redundant now that `disconnect()` isn't async.

Rob--W (approved):
r+c

Rob--W (approved):

PR #2144, opened 2021-02-03
chore(npm): Required npm to be strictly < v7 in package.json (fixes #2145)

### LINKED ISSUE #2145: Restrict npm version explicitly set in the package.json as supported to prevent renovatebot from updating the lock file to version 2
Some recent renovatebot PRs (#2141, #2142, #2143) are failing on CI while running the `npm ci` step.

The reason for the failure (as described in https://github.com/mozilla/web-ext/pull/2142#issuecomment-772584368) seems to be triggered because the `lockfileVersion: 2` that renovatebot did include in those PR is currently incompatible with npm v6.

As a short term solution we agreed to try to require in the package.json an npm version strictly < than npm v7, which should hopefully be enough to force renovatebot to don't upgrade the `lockfileVersion` yet.

We should then remove that restriction from the package.json once we are ready to fully migrate to npm v7 (or if in the meantime the incompatibility of the `lockfileVersion: 2` with npm v6 is gone).

rpl:
This PR does explicitly set npm engine compatibility to be strictly < v7
This should be hopefully be enough to prevent renovatebot to try to upgrade our lockfile to `lockfileVersion: 2`, as it seems to result in a lockfile incompatible with npm v6 and as a side effect it does make all recent CI jobs related to renovatebot PR to fail on the `npm ci` step (as described in https://github.com/mozilla/web-ext/pull/2142#issuecomment-772584368).

I double-checked locally that even if the npm engine required has to be strictly < v7, `npx npm@7 ci` is still completing successfully with the `lockfileVersion: 1` (and just printing a warning to mention that the npm version being used is not officially compatible with the web-ext package).

Unfortunately we can only be 100% sure that this change is going to be enough to force renovatebot to don't upgrade the lockfile in its PRs only after we have merged this PR and then asked renovatebot to rebase its own PR on top of this change.

Fixes #2145.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   ],
   "engines": {
     "node": ">=10.0.0",
-    "npm": ">=5.6.0"
+    "npm": ">=5.6.0 <7.0.0"
   },
   "engine-strict": true,
   "bin": {


Rob--W (approved):
This fixes an actual issue at the expense of generating an innocent warning.

LGTM. Hopefully we can remove the limit in the future.

Can you update the commit message to add the justification (or a link to it), for future reference?

Issue #2130, opened 2021-01-19
Fix deprecation warnings logged while installing web-ext from npm


rpl:
When web-ext is installed from npm, the following deprecation warnings are being logged:
```
npm WARN deprecated js-select@0.6.0: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated jetpack-id@1.0.0: Jetpack extensions are no longer supported by Firefox. Use web-ext instead to build a WebExtension.
npm WARN deprecated core-js@2.6.12: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated cheerio-select-tmp@0.1.1: Use cheerio-select instead
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
``` 

None of these deprecated npm packages is a direct web-ext dependencies.

Follows a list of the packages listed in the deprecation warnings, along with notes about the dependencies through which they are inherited, actions to fix the deprecation warnings, opened pull request and/or issues (either from the web-ext repo or upstream repos).

# TODO

## request@2.88.2 and har-validator@5.1.5

Notes:
- These dependencies seems to be inherited through the `sign-addon` dependency
- `har-validator` is a `request` dependency

Actions:
- [x] file issue in the sign-addon repo

## event-to-promise@0.8.0

Notes:
- The warning message suggests to use `promise-toolbox/fromEvent` instead

Actions:
- [ ] this is a direct web-ext dependency, we should just replace its usage with the new npm package suggested

## uuid@3.4.0

Notes:
- it seems to be a transitive dependency we inherit from request through sign-addon

Action:
- [ ] double-check if fixing the warning related to request would also fix this issue as a side effect (or if there are dependencies that would still keep that version in the transitive dependencies).
 
## babel-eslint

Notes:
[...]

PR #2151, opened 2021-02-08
chore: Removed  deprecated @babel/polyfill package from web-ext npm dependencies

rpl:
This PR does remove the explicit @babel/polyfill dependency (which is how we get the core-js@2 package as a transitive dependency).

This issue is part of the fixes needed for #2130 

I was originally planning to replace it with a direct core-js@3 dependency, but at a first glance I'm not really sure that we do actually need it for the ecmascript features that we are actually using (and the one already supported by nodejs releases officially supported), and so I opted to just remove the @babel/polyfill dependency for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,137 +50,6 @@
             "@babel/highlight": "^7.12.13"
           }
         },
-        "@babel/generator": {
-          "version": "7.12.15",
-          "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.12.15.tgz",
-          "integrity": "sha512-6F2xHxBiFXWNSGb7vyCUTBF8RCLY66rS0zEPcP8t/nQyXjha5EuK4z7H5o7fWG8B4M7y6mqVWq1J+1PuwRhecQ==",
-          "dev": true,
-          "requires": {
-            "@babel/types": "^7.12.13",
-            "jsesc": "^2.5.1",
-            "source-map": "^0.5.0"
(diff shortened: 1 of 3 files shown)


rpl:
@Rob--W there isn't much to review in this PR, but I assigned a review request to you to get your perspective about what described in the issue description (in case you may think of some there are other details that we may want to double-check explicitly before merging this PR).

Rob--W (commented):
Tests and CI passes, so I don't have anything against this change in principle.

As a quick extra check, did you compare the generated build output before and after the patch? If it looks reasonable, just merge it.

rpl:
> Tests and CI passes, so I don't have anything against this change in principle.
> 
> As a quick extra check, did you compare the generated build output before and after the patch? If it looks reasonable, just merge it.

I've been finally able to get back to this and compare the bundle build on master with the one build from this branch and their md5sum (and content) seems to be exactly the same (not super surprising to be honest, given that we were not requiring the polyfill explicitly anywhere and babel transpiled js code wasn't requiring it neither):

```
md5sum with*
5888ac6fc96aa80b9f431f1f910a60b4  master--web-ext.js
5888ac6fc96aa80b9f431f1f910a60b4  pr251--web-ext.js
```

I did also rebase this PR to regenerate the package-lock.json file in this PR on top of the one currently on master.
As soon as all the CI jobs did run and complete successfully I'm gonna merge this on master.

PR #2178, opened 2021-03-03
fix: Move filename from global to build command option

### LINKED ISSUE #2172: --filename cli option is currently defined as a global option, but only used by the build command

bug

The --filename cli option is currently defined as a global option, but the build command is the only one expecting it.

It doesn't make a big difference from a CLI usage perspective, but having it defined as a global option affects its format in the config file, and that may be surprising from a user perspective (e.g. see #2171).

Move the --filename cli option definition into the build-specific options.

:warning: This would be a backward incompatible change (from a config file format perspective) and so it would be better to do it as part of a major release and/or include a backward compatible behavior (e.g. warn but still load a config file with a filename global option).

rpl:
Fixes #2172

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -484,18 +484,6 @@ Example: $0 --help run.
       default: true,
       type: 'boolean',
     },
-    'filename': {
-      alias: 'n',
-      describe: 'Name of the created extension package file.',
-      default: undefined,
-      normalize: false,
-      demandOption: false,
-      requiresArg: true,
-      type: 'string',
-      coerce: throwUsageErrorIfArray(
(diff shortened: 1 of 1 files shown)


rpl:
@Rob--W as also pointed out in the issue, this is technically a backward incompatible change from a config file perspective (if any user did figure out on their own that the filename config option had to be part of the global options and they were already using it) but given that we are going to release this in a new major version I'm not sure that is worth to apply more changes to keep compatibility with the "old config file format for this particular option". wdyt?

Rob--W (approved):
sgtm. If we're bumping the version anyway, let's keep it simple.

Issue #950, opened 2017-05-19
Allow manual reloading, when using --no-reload option


paulkamer:
### Is this a feature request or a bug?
Feature request

### What is the current behavior?

Running web-ext with `--no-reload` works as advertised, but no longer offers a way to manually force a reload (by typing "R"). 

### What is the expected or desired behavior?
Running web-ext with `--no-reload` should still offer a way to manually force a reload:

> Press R to reload (and Ctrl-C to quit)

### Version information (for bug reports)

* **Firefox version**: 53
* **Your OS and version**: Ubuntu 16.04.2 LTS (xenial)
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````
v6.3.1
3.10.8
1.9.1

kumar303:
Oh yeah, that seems useful. Thanks for filing. 

I think `--no-reload` may get simulated internally for older Firefoxes that don't support reloading -- just something to be aware of for anyone works on this issue. Maybe it won't pose a problem though.

giodamelio:
It would also be great to allow a way to externally trigger the build. Perhaps listening for a [POSIX signal](http://man7.org/linux/man-pages/man7/signal.7.html) like `USR1`?

antekone:
Any chances of implementing this feature?

Rob--W:
I didn't know that Ctrl-R is disabled with `--no-reload` (https://github.com/mozilla/web-ext/issues/1758#issuecomment-557111582 ). It seems reasonable to implement it.

PR #2182, opened 2021-03-05
docs: Added alias for watch-file option and updated option description

ankushduacodes:
Added alias for watch-file option and updated option description to reflect recent changes to the option

Continuation of #2125. Opening this PR in response to the discussion in [extension-workshop PR #950 C3](https://github.com/mozilla/extension-workshop/pull/910#issuecomment-776797076)

@rpl Are there any tests to check if the alias does work as expected?

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -617,9 +617,12 @@ Example: $0 --help run.
         type: 'boolean',
       },
       'watch-file': {
-        describe: 'Reload the extension only when the contents of this' +
-                  ' file changes. This is useful if you use a custom' +
-                  ' build process for your extension',
+        alias: ['watch-files'],
+        describe: 'Reload the extension only when the content of these' +
+                  ' files change. This is useful if you use a custom' +
+                  ' build process for your extension or if you want web-ext' +
+                  ' to explicitly watch for changes to specific files,' +
+                  ' without watching the extension directory tree\n',
         demandOption: false,
         type: 'array',
       },


ankushduacodes:
For some reason, I am not able to see open ci job that failed (sending back a blank screen), Will try again soon to see what the issue is.

src/program.js
```diff
@@ -617,9 +617,12 @@ Example: $0 --help run.
         type: 'boolean',
       },
       'watch-file': {
-        describe: 'Reload the extension only when the contents of this' +
-                  ' file changes. This is useful if you use a custom' +
-                  ' build process for your extension',
+        alias: ['watch-files'],
+        describe: 'Reload the extension only when the content of these' +
+                  ' files change. This is useful if you use a custom' +
+                  ' build process for your extension or if you want web-ext' +
```
rpl, line 623: Using a "custom build process" is the most common reason for being interested into watching a specific file (or set of files) for changes, and that didn't change with the addition of the alias.
I think that in the end we may just leave the description as is.

rpl (changes requested):
@ankushduacodes thanks a lot for creating a PR for the alias, just in time for including it into the new version.

About the tests: this kind of tests can be covered with a functional test, e.g. we may be able to cover it with just a couple of small changes to [tests/functional/test.cli.run.js](https://github.com/mozilla/web-ext/blob/master/tests/functional/test.cli.run.js).

About the CI job that did fail:
oddly I got a white page when I tried to follow the link for the failure, but I got to the failure logs by navigating to it from the circleci homepage.
Anyway, the failure is only related to the commit message not following the convention:
```
There is only one commit in this pull request, we are going to check the single commit message...
Failures during changelog linting the pull request:

⧗   input: added alias for watch-file opt and updated option description to reflect recent changes related to supporting multiple files to watch #2125
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
⚠   header must not be longer than 72 characters, current length is 139 [header-max-length]

✖   found 2 problems, 1 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

Don't panic! If the CI job is failing here, please take a look at

 - https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#writing-commit-messages
``` 

Updating the commit message to follow the convention will fix that linting failure.

btw, I think in this case the conventional prefix should be `fix: ...` or `chore: ...`.

src/program.js
ankushduacodes: Got it. Going back...

(PS: started a review by accident)

ankushduacodes:
@rpl I added a test that does two things at the same time:
1. Check for alias's working
2. Providing multiple files in one option (Although yargs treat `--watch-file file1 --watch-file file2` and `--watch-file file1 file2` the same) I added this test because I had added same kind of test for `watch-ignored` a well. 

Please let me know if any changes are required :)

rpl (approved):
lgtm

Thanks @ankushduacodes

PR #2193, opened 2021-03-11
fix: target npm versions shipping with nodejs v12+

joelpurra:
By default, `npm` ships with `node`. When targeting Node.js v12+, the targeted NPM versions should
also be upgraded.

- Node.js v12.0.0 (minimum supported by `web-ext`) ships with NPM v6.9.0.
- Node.js v14.16.0 (current LTS) ships with NPM v6.14.11.
- Node.js v15.11.0 (current version) ships with NPM v7.6.0.
- https://nodejs.org/en/download/releases/

Changes

- Updates versions mentioned in `README.md`.
- The NPM version in `engines` in `package.json` has had the lower range limit updated to match
  the NPM version shipped with Node.js v12.
- The NPM version upper range limit is due to #2146 and must be kept for now. Note that
  this yields `EBADENGINE` warnings when using NPM v7.0.0+, which can be ignored.


joelpurra:
Just found #2146 (was greyed out because it's automated by a bot), #2145, #2144. Since the versions were untouched in `README.md` I had assumed that the version range in `package.json` was an oversight. Up to maintainers to close or keep this pull request.

rpl:
> Just found #2146 (it's [dimmed](https://user-images.githubusercontent.com/1402241/65263190-44c52b00-db36-11e9-9b33-d275d3c8479d.gif) by [refined-github](https://github.com/sindresorhus/refined-github) because the PR was automated by a bot), #2145, #2144. Since the versions were untouched in `README.md` I had assumed that the version range in `package.json` was an oversight. Up to maintainers to close or keep this pull request.

Not updating the README.md is definitely an oversight on my side, on the contrary (as you already noticed) not changing the version range `package.json` is unfortunately something we had to do on purpose due to the issue described in #2146.

There is a chance that the issue we had with npm v7 may have been solved in the meantime (I'm going to give it another try asap), but we didn't want to block the release on that (also given that npm doesn't show a warning but doesn't prevent the installation from completing successfully).

On a second thought (and after looking to the changes in this PR) it would have been likely better to update the npm range to `>= 6.9.0 < 7.0.0`, and then remove the upper bound as soon as the the issue we did have in the CI job can't be triggered anymore.

I'm personally happy to merge the change to the README, and I think it would be also ok to merge an update to the npm range if we do change it to `>= 6.9.0 < 7.0.0` in this PR (and defer removing the upper bound, after I'm going to have some time to look into that issue again).

@joelpurra how that sounds to you?

rpl:
> @rpl: updated the pull request to reflect your comments.

Thank you so much! 
 
> This does not fix my initial concern (5 prominent warning lines when using `npm` v7.0.0+ shipping with current node), but I see that there are greater issues to keep in mind. `README.md` does mention that using Node.js LTS is a prerequisite for installation using `npm`.

yeah, I'm not happy neither about those warning, but I'm confident that we will be able to get rid of the upper bound on the npm versions listed as supported in the package.json soon enough and then release a fix in one of the next minor versions we are going to release.

rpl (approved):

PR #2198, opened 2021-03-12
docs: fix URL for good first bug label

ParkerM:
The contributor docs contain a link to issues with a `good first bug` label that no longer exists. This PR fixes the URL to point to the updated label `contrib: good first bug`.


Rob--W (approved):
Thanks for noticing and the patch!

PR #2222, opened 2021-04-13
fix: Sanitize manifest property values before interpolation

### LINKED ISSUE #2119: Cannot start extension on Android if name contains a slash
Bug

Create an extension and put a slash in the `name` field in manifest.json
Then run `web-ext run -t firefox-android ....` to run the extension.
What happens is that the browser starts, but `web-ext` quits, ending with:

```
Starting org.mozilla.fenix...

Waiting for org.mozilla.fenix Remote Debugging Server...
Make sure to enable "Remote Debugging via USB" from Settings -> Developer Tools if it is not yet enabled.

Building web extension from /path/to/extension/with/slash/in/name

s: Invalid filename template "{name}-{version}.zip". Filename "Name with /slash-1.zip" should not contain a path
```

Fenix starts with the extension.

* **Firefox version**: Not relevant
* **Your OS and version**: Not relevant

````
````

v15.5.0
6.14.10
5.2.0

### PR DESCRIPTION
**Pull request by Rob--W**
Fixes #2119 (which was a regression by #1900).

### DIFF
--- a/tests/fixtures/slashed-name/manifest.json
+++ b/tests/fixtures/slashed-name/manifest.json
@@ -1,0 +1,6 @@
+{
+    "name": "Name w/o slash \\o/",
+    "description": "ends with.zip",
+    "version": "1",
+    "manifest_version": 2
+}
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -132,7 +132,8 @@ function getPackageNameFromTemplate(
   const packageName = filenameTemplate.replace(
     /{([A-Za-z0-9._]+?)}/g,
     (match, manifestProperty) => {
-      return getStringPropertyValue(manifestProperty, manifestData);
+      return safeFileName(
+        getStringPropertyValue(manifestProperty, manifestData));
     }
   );
 
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -121,6 +121,22 @@ describe('build', () => {
     );
   });
 
+  it('sanitizes characters in extension name in filename from template', () => {
+    return withTempDir(
+      (tmpDir) =>
+        build({
+          sourceDir: fixturePath('slashed-name'),
+          artifactsDir: tmpDir.path(),
+          // name contains a slash. description ends with ".zip".
+          filename: 'prefix-{name}{description}',
+        })
+          .then((buildResult) => {
+            assert.match(buildResult.extensionPath,
+                         /prefix-name_w_o_slash_o_ends_with\.zip$/);
+          })
+    );
+  });
+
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: tests/fixtures/slashed-name/manifest.json [hunk 1]
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "Name w/o slash \\o/",
+    "description": "ends with.zip",
```
rpl, line 3: Nit/question: why putting `.zip` is the description? it is definitely expected that ".zip" ends up into the generated filename, but I was wondering what is the role of it in the test (because it is not immediately clear what those that test, may be good to mention why we have included it in the description as part of the inline comment at line 130 in test.build.js, which is currently mentioning that .zip is coming from the description but not why we are doing it in that test).

rpl (approved):
lgtm, thanks Rob.
I just have one question, but it should affect the fix applied in this PR and so I'm also approving this PR as is.

tests/fixtures/slashed-name/manifest.json
Rob--W: It's not significant, and unrelated to the fixed issue. I just put it there so that there is test coverage for the scenario where the extension is derived from the interpolated value. That would catch the case if someone ever tried to enforce the validation on the template (instead of the parsed result).

I ended up fixing the issue by sanitizing the interpolated values, but an alternative/additional change that I considered was to move the path check to before the interpolation (since it's logically equivalent now)

rpl: > That would catch the case if someone ever tried to enforce the validation on the template (instead of the parsed result).

I see, then I think that is something that would be good to be pointed out in that inline comment in at test.build.js line 130, because that comment does explicitly say that we expect it but not that it is intended to catch a possible regression.

Rob--W: I just found that I had non-commited changes with the following comment:
```diff
diff --git a/tests/unit/test-cmd/test.build.js b/tests/unit/test-cmd/test.build.js
index 04745db..cc12d48 100644
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -128,6 +128,9 @@ describe('build', () => {
           sourceDir: fixturePath('slashed-name'),
           artifactsDir: tmpDir.path(),
           // name contains a slash. description ends with ".zip".
+          // This allows us to verify that the requirement (i.e. file name
+          // should end with a zip or xpi extension) is enforced for the
+          // resulting filename and not just the template.
           filename: 'prefix-{name}{description}',
         })
           .then((buildResult) => {
```

This is just a test, so I don't really care that much about the inline comment, but if you consider it important I could submit a PR with the comment fix.

PR #2269, opened 2021-06-16
fix: Fixed noReload in configuration files not taking effect

wingdeans:
In the run command, noReload is defined using yargs's boolean negation
to automatically create the `--no-reload` parameter. This does not
allow setting `noReload` from a configuration file and the valid
option `reload` does not affect the behavior of `run`.

This fix defines an explicit `no-reload` option in the exact same way
as `no-input` is defined, allowing `noReload` to be used in
configuration files.


rpl:
@wingdeans codecov is currently complaining because [line 179 from src/program.js](https://github.com/mozilla/web-ext/blob/c866678c95bfb755fee190bb8dbe163d90aa6c9c/src/program.js#L179) seems to not be covered anymore after adding the new option.

On one side it sound s kind of ok to workaround the issue using the same approach we used the issue with --no-input in #2074, but while looking to the code in program.js I got the feeling that the config option is not having any effect because [web-ext is "manually" setting the negated version of the CLI option](https://github.com/mozilla/web-ext/blob/c866678c95bfb755fee190bb8dbe163d90aa6c9c/src/program.js#L178-L180) before we load the config file and so I think that after loading the config file we end up with an argv that looks like the following:
```
{
  noReload: false,
  reload: false
}
``` 

You may double-check that I'm not wrong and that is the case by temporarily adding a log to print out what are the properties of the `adjustedArv` variable right after we have loaded and applied all the config files (e.g. [at line 316 in program.js](https://github.com/mozilla/web-ext/blob/c866678c95bfb755fee190bb8dbe163d90aa6c9c/src/program.js#L316)).

And so I think that, another way to make sure a config file like the following:
```
module.exports = {
  run: {
    reload: false,
  },
};
```

does work as expected, may be to move this part:
https://github.com/mozilla/web-ext/blob/c866678c95bfb755fee190bb8dbe163d90aa6c9c/src/program.js#L172-L189

into a separate method (e.g. one named `fixNegatedCLIOptions` or something like that) and call that method (and passing to it `adjustedArgv`) right after we loaded all the config file.

With that change `noReload` should be set consistently to the value set for `run.reload` in the config file (and then we will be passing the arguments to the run command with a `noReload` property set to the expected value loaded from the config file).

In either case, it would be good to also include a test case to cover the fix, to prevent it to regress without being noticed.
This pre-existing test case from test.program.js may be a useful source of inspiration for the new test case that this fix would need:
https://github.com/mozilla/web-ext/blob/c866678c95bfb755fee190bb8dbe163d90aa6c9c/tests/unit/test.program.js#L623-L645

PR #2321, opened 2021-09-22
fix: Fixed Chrome Not Opening Extensions Pages provided in URL

kodergeek:


Rob--W:
Do you need anything else from me to proceed? I posted a very brief comment in reply to your question at https://github.com/mozilla/web-ext/issues/1979#issuecomment-924761488. It does answer your question, but I can imagine if some more detail is needed to understand my reply.

kodergeek:
Yes, I appreciate if you can point in code where I should add the call so that bg.js extension can successfully open the extensions tab as you described in your comment. I couldn't figure it out and I tried in a few different places in code to no avail.

Rob--W:
If your goal is to use the WebSocket to transmit the message, then you'll have to send a message once [the manager extension](https://github.com/mozilla/web-ext/blob/6.4.0/src/extension-runners/chromium.js#L283-L349) has connected.

Here is an example of a snippet that connects: https://github.com/mozilla/web-ext/blob/6.4.0/src/extension-runners/chromium.js#L147

You would have to send the message once (not on every connect), to prevent the extension from opening tabs again when the WebSocket connection is temporarily interrupted for some reason.

For the current requirements, a simpler alternative is to hard-code the startup URL in the source of the manager extension. After all, the URLs are known when the extension is generated, and they need to be opened only once. To implement this approach, you'd need to pass the list of URLs to https://github.com/mozilla/web-ext/blob/6.4.0/src/extension-runners/chromium.js#L154 and write it to the generated bg.js

To make sure that the tabs are opened only once, use `chrome.runtime.onInstalled` and open the tabs if `details.reason === "install"`.

In your current patch you're also filtering on `chrome://extensions` specificially. I suggest to just do this for any `chrome:` URL. E.g. `chrome://version` would have the same issue.

kodergeek:
I have now updated the PR with the second approach as suggested by you, eg hard-coding the startup URL in the source of the manager extension. 

I have also tested it with some different URL combinations and it seems to be working. Please check it out and let me know if there's anything that should be improved.

Also, there are 4 failing checks for my PR which I have no idea what they relate to. So if you think the fix is good to be merged I'd appreciate if you could advise me how to fix the failing checks :)

Rob--W:
The test failures are linting issues. Run `eslint` to see the exact errors.

This is the output of circleci:

```
/home/circleci/web-ext/src/extension-runners/chromium.js
  163:34  error  Unexpected string concatenation                      prefer-template
  163:52  error  Strings must use singlequote                         quotes
  163:55  error  Missing semicolon                                    semi
  164:36  error  Missing semicolon                                    semi
  174:1   error  This line has a length of 95. Maximum allowed is 80  max-len
  294:1   error  This line has a length of 84. Maximum allowed is 80  max-len

✖ 6 problems (6 errors, 0 warnings)
  4 errors and 0 warnings potentially fixable with the `--fix` option.
```

src/extension-runners/chromium.js
```diff
@@ -150,8 +150,28 @@ export class ChromiumExtensionRunner {
...
+    let startingUrl;
+    var specialStartingUrls = '';
+    if (this.params.startUrl) {
+      const startingUrls = Array.isArray(this.params.startUrl) ?
+        this.params.startUrl : [this.params.startUrl];
+
+      // Remove URLs starting with chrome:// from startingUrls and let bg.js open them instead
+      for (let i = 0; i < startingUrls.length; i++) {
+        if (startingUrls[i].toLowerCase().startsWith('chrome://')) {
+          specialStartingUrls += startingUrls[i] + " "
```
Rob--W, line 163: The URLs should be appended to an array, not to a string.

src/extension-runners/chromium.js
```diff
@@ -150,8 +150,28 @@ export class ChromiumExtensionRunner {
...
+          startingUrls.splice(i, 1)
```
Rob--W, line 164: By removing the element during the iteration, you'll end up skipping URLs.

For example:

```
[
  "chrome://extensions",
  "chrome://version"
]
```

At `i=0`, `"chrome://extensions"` would be appended to the list of start URLs.
But due to the `startingUrls.splice(i, 1)` call, the array will be `["chrome://version"]`.
At the next iteration, `i=1`, which would result in `"chrome://version"` to be skipped.

To fix this, you could iterate over the URLs in reverse order and prepend the URLs to the `specialStartingUrls` list. Then `.splice(i, 1)` wouldn't be an issue any more.

src/extension-runners/chromium.js
```diff
@@ -334,6 +345,19 @@ export class ChromiumExtensionRunner {
       const ws = new window.WebSocket(
         "ws://${wssInfo.address}:${wssInfo.port}");
 
+      
+      if (${specialStartingUrls.length} > 0)
+      {   
+        const chromeTabList = "${specialStartingUrls}".trim().split(" ")
```
Rob--W, line 351: Serialize the list of URLs with `${JSON.stringify(specialStartingUrls)}` instead of splitting by strings. Otherwise, if the URL contains a `"`, this could be abused to execute code in the context of the helper extension. That's problematic.

src/extension-runners/chromium.js
```diff
@@ -334,6 +345,19 @@ export class ChromiumExtensionRunner {
...
+        chrome.runtime.onInstalled.addListener(details => {
+          if (details.reason === chrome.runtime.OnInstalledReason.INSTALL ) {
+            chromeTabList.forEach(url => {
+              chrome.tabs.create({ url });
+            });
+            }
```
Rob--W, line 357: Whitespace is off, please fix this.

Rob--W (changes requested):
Please address the implementation issues below (and also check that `eslint` passes), and add a unit test that verifies the expected parameters and opened tabs in https://github.com/mozilla/web-ext/blob/master/tests/unit/test-extension-runners/test.chromium.js.

You'd have to verify that the expected args are present, and that the special URLs are in the generated helper extension.

src/extension-runners/chromium.js
kodergeek: Thanks for the feedback. I had a hard time with template literals and arrays that's why I used string with spaces, but JSON.stringify() solves both this problem and potential code injection issue you mentioned nicely!

kodergeek:
Thanks for your feedback @Rob--W. I have now resolved the issues you mentioned and have also added a unit test to both check for presence of special URLs in generated script, and also lack of them in the args list sent to chrome.

One issue I encountered during running existing unit tests was with the ordering of arguments being pushed to `chromeFlags` array. I had to move [some code up](https://github.com/kodergeek/web-ext/blob/bug-1979/src/extension-runners/chromium.js#L153-L158) to be able to check starting URLs and send them to `createReloadManagerExtension`  because `createReloadManagerExtension` is now expecting the special arguments list, and it resulted in reordering of chromeFlags array because `startingUrls` was pushed first in `chromeFlags` array, and it resulted in 2 unit tests failing. So I have rearranged the code slightly and it has caused a slight code duplication at [here](https://github.com/kodergeek/web-ext/blob/bug-1979/src/extension-runners/chromium.js#L156-L158) and [here](https://github.com/kodergeek/web-ext/blob/bug-1979/src/extension-runners/chromium.js#L230-L233). But no unit tests are failing as a result.

kodergeek:
Well yes the easiest way to remove code duplication and keeping the functionality is to make a small change to 2 unit tests by shifting around the list of arguments that are being asserted. I have updated the PR with the suggested fix. Is that something that is acceptable or I should not change the existing unit tests at all?

kodergeek:
Hi @Rob--W Just wondering if you've had time to review this?

Rob--W:
> Hi @Rob--W Just wondering if you've had time to review this?

Unfortunately I did not. I'll try to get back to you next week.

Issue #2154, opened 2021-02-10
Add an environment variable to remove the progress bar for validating & signing the add on with the web-ext sign command


hmx789:
### Is this a feature request or a bug?
This is a feature request.

### What is the current behavior?
Currently when we run web-ext sign command there is a progress bar that for validating the add on and signing the addon(added images below). Don't get me wrong, the progress bar is very cool on my local terminal but on CI this causes unnecessary logging.
![image](https://user-images.githubusercontent.com/36273963/107573521-3f490480-6bb3-11eb-8e2a-5612ce69d5d7.png)
![image](https://user-images.githubusercontent.com/36273963/107573550-47a13f80-6bb3-11eb-8f2a-3a747d3a4e88.png)
![image](https://user-images.githubusercontent.com/36273963/107573644-69022b80-6bb3-11eb-8706-93416afd4860.png)

### What is the expected or desired behavior?
I would like it if there was an environment variable that I could pass into the command to remove the progress updates. I would only like to remove the progress bar updates. I believe it is still valuable for us to provide the validation results link, the output of downloaded file and the Extension ID.

Rob--W:
Patches are welcome. To resolve this issue, the `sign-addon` dependency needs to be updated to support this (see https://github.com/mozilla/sign-addon/issues/710 for details), and then web-ext needs to account for it.

If it is possible to automatically detect the scenario (e.g. by using `isTTY` to detect interactive output), then we can automatically detect and fix it so that `web-ext` users don't need special configuration. Otherwise a new CLI option or environment variable would be acceptable too.

manthony9:
Hey @Rob--W  can I take this issue up? I've been in conversation with @willdurand and I think I have some headway. Would need a little assistance to go further.

willdurand:
> Hey @Rob--W can I take this issue up? I've been in conversation with @willdurand and I think I have some headway. Would need a little assistance to go further.

@manthony9 FWIW, we need to update `sign-addon` first before being able to move forward with _this_ issue.

manthony9:
Oh! my bad...let me work on sign-add on first..

Rob--W:
@willdurand When you're back, can you reply to https://github.com/mozilla/sign-addon/pull/775 ?

willdurand:
`sign-addon` v3.7 now provides a option to disable the pseudo progress bar: https://github.com/mozilla/sign-addon/releases/tag/3.7.0

rpl:
[...]

PR #2347, opened 2021-11-17
fix(#2154): Added condition to toggle progress bar

manthony9:
Fix#2154

Hi @Rob--W @rpl 

Some progress, I've made a simple condition block that checks for a CLI option and toggles the progressbar basis the presence of the option. Please review and share your feedback. I am not sure if i'm utilizing isTTY the correct way and I think this could be done a better way. Could you point me to the approach you have in mind or some resources that I check for the same. I read up the node documentation on TTY but could not get much context.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4674,9 +4674,9 @@
       "dev": true
     },
     "eslint-plugin-flowtype": {
-      "version": "5.9.2",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-flowtype/-/eslint-plugin-flowtype-5.9.2.tgz",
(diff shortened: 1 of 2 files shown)


rpl:
@manthony9 unfortunately this PR can't be really reviewed in its current form:
- there are too many changes unrelated (e.g. code reformatted in many of the files, I guess it is likely because of some autoformatting feature enabled in your editor, + other files you may have used while testing this our that seems to have slipped in the PR by mistake)
- what I described above is also triggering failures in all CI jobs (in particular on the linting task).

Please, update this PR by removing all unrelated files added and by reverting the code reformatting (if that keeps happening every time you save a web-ext source file, then you should look into how to disable the autoformatting behavior in your editor), please try to keep the coding style consistent with the surrounding code and run the test and lint locally using the related npm scripts part of web-ext package.json file.

Issue #2249, opened 2021-05-29
Watch for changes - stops reacting if file changes multiple times


Juraj-Masiar:
### Is this a feature request or a bug?
Bug

### What is the current behavior?
1) using any project
1) open console and navigate to the folder where web-ext is running
1) execute `touch some_file` multiple times in a short succession (every 500ms)
```touch 128.png```
1) notice how the addon got reloaded only once?

### What is the expected or desired behavior?
It would be nice if the addon got reload after the last change.
The reload throttle is fine, but if there are no more changes being reported, it should reload it again. Right?

### Version information (for bug reports)

* **Firefox version**: 89
* **Your OS and version**: Win 21H1
* Paste the output of these commands:

````
v14.16.1
6.14.11
6.1.0
````

https://user-images.githubusercontent.com/6358971/120074132-b5f30a80-c09b-11eb-9ffb-bf7c29f7882f.mp4

The reason I'm reporting this is actually different. Sometimes when I make a change, the web-ext won't reload the addon at all. It's like 50% chance it will or won't reload. As I was experimenting with it, I've noticed this issue.
After A LOT of work I've managed to hotfix it by using the `--watch-file version` and by touching a dummy `version` file when the webpack makes changes.

Rob--W:
To avoid too many repeated reloads, the reload is debounced using [debounce](https://www.npmjs.com/package/debounce), i.e. additional changes within (by default) 1 second are ignored: https://github.com/mozilla/web-ext/blob/8bd0d13f425bbaed6cfc2f8127883c5c43010f1d/src/watcher.js#L61

What you're asking for is to refresh immediately for the first change, and if anything has happened in the following second, to reload once more thereafter. When the source changes files many times, this would then always result in two reloads. This is not ideal either.

To have predictable reloading behavior when a build tool is used, touching a file after the completion of a build is recommended (in combination with `--watch-file`).

Juraj-Masiar:
Actually what I would prefer is to defer the execution by some time (~400ms) and postpone it by this time every time it's called again within the delay (so that it's called once after this time when things stops changing).

**The current behavior is not stable for me, is it for you?** Sometimes the addon reloads broken or disappears completely and then reappears after further changes. 
I use Webpack 4, maybe I need to upgrade to 5?

Anyway, the workaround with watching a file seems to be working fine so it's OK for me. I use [webpack-touch](https://www.npmjs.com/package/webpack-touch).

Rob--W:
[...]

PR #2359, opened 2021-11-25
fix: Improve debounce to catch multiple consecutive changes

fregante:
- Fixes https://github.com/mozilla/web-ext/issues/2249

The current behavior is effectively ignoring file changes, hence breaking the expectation that **"the extension will update on file changes"**

Changing it to update _after_ all the changes are done ensures that no file changes are lost, while only delaying the reload by half a second.

An alternative solution would be to use a "before and after" debouncer, so that the first change will be immediate, and any further changes will _also_ be handled, but after `1000ms` have gone by without events. The drawback of this is that some users might see 2 reloads instead of one, so IMHO it's a worse solution.

If this change is accepted I'll fix the tests.


Rob--W:
This change should be merged with a major version bump because it changes the behavior in a way that is not backwards-compatible.

Rob--W:
CI shows a test failure at the `npm run test-coverage` phase:

```
  1 failing

  1) watcher
       watches for changes in the sourceDir:
     AssertError: expected spy to be called once but was called 0 times
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:120:25)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:66:20)
      at Object.calledOnce (node_modules/sinon/lib/sinon/assert.js:95:17)
      at Context.<anonymous> (tests/unit/test.watcher.js:105:18)

```

fregante:
CI is failing due to the title, but I don’t understand what's wrong with it. Can you fix it and merge the PR?

rpl:
Merged as #2385 
Thanks @fregante 
(and my sincere apologies I couldn't get to this sooner, but I didn't forgot it was waiting for me to take care of rebase and merge it ;-)).

fregante:
Grazie! Looking forward to the next release

PR #2363, opened 2021-12-03
fix(#2154): Added cli option not recognized

### LINKED ISSUE #2154: Add an environment variable to remove the progress bar for validating & signing the add on with the web-ext sign command

manthony9:
Fixes #2154 

Hi @Rob--W @rpl 

Cleaned my file structure and formatting as per the previous comment. Now trying to add a cli option for sign in program.js under sign but it does not get recognized. Tried setting it under global options as well but that does not work either. Could you help in finding out why.


rpl:
@manthony9 there are still some cleanups necessary in this PR, in particular:
- the package-lock.json should have not been included (and now it is also conflicting with changes that are already landed on the main repo branch), the diff is due to the fact that you are using a newer npm version (but don't worry about it, npm format changes do not matter in this case, and we can remove it from the PR)
- the ` web-ext-artifacts/some-signed-file-1.2.3.xpi` file should also not be part of the PR
- there are some empty line changes that aren't needed, likely just added while you were experimenting with the change locally and then they got into the patch submitted by mistake

Once we got the basic PR clean, there are a couple more details that I see were part of the initial guidance provided by Rob (e.g. automatically disable the progress bar automatically also if web-ext is being executed without a TTY), and after that we can double-check if there is some more test coverage that we may want to add along with the final set of changes.

rpl:
Closing as inactive.

In comment https://github.com/mozilla/web-ext/pull/2363#issuecomment-1014758172 there are more details about what needed to be changed in this PR to clean it up.

Feel free to ping me if you have updated it and I'll take a look and reopen this.

PR #2367, opened 2022-01-06
fix: Replaced event-to-promise with promise-toolbox/fromEvent

equal-l2:
ref: #2130


rpl:
@equal-l2 thanks a lot for looking into replacing the event-to-promise deprecated dependency.

Before we can merge this we need to solve the conflict on the package.json (I bet there is just some dependency that I have updated as part of the 6.7.0 release that was part of the context of the package.json change in this PR)

If you have time to rebase this PR, I can wait for that and then approve and merge this PR, but if you don't have time for that I'm also happy to do it for you (and keep the git metadata from this PR while doing so).

equal-l2:
@rpl Thank you for the information, I have rebased the commits.

PR #2397, opened 2022-05-21
fix(deps): Update dependency adbkit to v3.2.3

rpl:

PR #2399, opened 2022-05-23
ci: remove configure_global_npm task

willdurand:
This does not seem required anymore and that actually breaks recent Circle CI builds


willdurand:
ha well, epic fail

willdurand:
I suppose we should do what @rpl did in https://github.com/mozilla/web-ext/pull/2397/commits/ca7587dbf8bc53fe5d9ab0ff66603a53df049444 then...

willdurand:
closing because it is now fixed in https://github.com/mozilla/web-ext/pull/2397

PR #2405, opened 2022-05-31
feat: convert web-ext package into a nodejs native ES module 

### LINKED ISSUE #1306: Use dynamic imports "await import" instead of require
In #1302, we started to lazily load some modules for performance.

At first `await import` was used, but because the test coverage tool cannot instrument these, we switched to `require` instead (https://github.com/mozilla/web-ext/pull/1302#issuecomment-380779622).

Ideally we should not need CommonJS's `require`, and only the (async) import.
This issue tracks the migration from `require` to `await import`.

### LINKED ISSUE #2311: Prepare web-ext npm package to switch to ES modules
Same as mozilla/addons-linter#3842 for the addons-linter npm package:

- npm dependencies used by web-ext itself are being gradually migrating to ES modules, and we should prepare web-ext npm package to also switch to ES modules. 
- Node.js versions >= than `12.20` / `14.14` / `16.0` seems to not require a command line flag to enable native loading of the ES modules.

rpl:
Fixes #2311
Fixes #1306

This pull requests goal is to convert web-ext package into a nodejs native ES module.

- Changes to npm scripts to work as ESM modules (and replaced webpack with babel-cli as build step)

- Add small custom node module loader (tests/babel-loader.js) to transpile on the fly (used to make mocha able to load tests modules using flowtype syntaxes)

- Re-export custom node module loader exports provided by testdouble / quibble to be able to mock ES modules in some tests (used as part of workaround for sinon.stub limitations on stubbing ES modules)

- Added babel-plugin-transform-inline-environment-variables to interpolate environment variable during babel transpiling (replaces webpack interpolated WEBEXT_BUILD_ENV global)

- converted build scripts and fake firefox and amo scripts to ES modules

- fixed sign-addon import when it is loaded using the nodejs native ES module loader (caught by functional tests)

### package size comparison

The new package size is 165 kB (unpacked 699 kB) vs old package size 103 kB (unpacked 442 kB).  

###  TODO list
- [x] fix unexpected failures while running unit test on the windows CI worker
- [x] consider changing unit tests to import modules from "src/" instead of importing the transpiled ones from "lib/"  (and make sure they are being transpile on the fly by tests/babel-loader.js)
- [x] consider changing the way we are re-exporting the adb helper module (e.g. change it to export the adb transpiled module directly from the package.json)
- [x] fix tests related to importing web-ext as a library
- [x] investigate how the tests importing web-ext as a library are able to pass as is (because it seems suspicious they pass without any change)
- [x] compare new npm package size (with babel per-file transpile) with previous one (with webpack bundled)

### other notes

NOTE: the following jscodeshift transform script has been created and used to aid the conversion to ES modules that loads successfully while using the nodejs native ES module loader:
- https://gist.github.com/rpl/7b50f9d2a347c068d1b1185a0d173b09


.babelrc
```diff
@@ -1,13 +1,21 @@
 {
   "presets": [
-    ["@babel/env", {
-      "targets": {"node": "10"}
-    }],
+    [
+      "@babel/env", {
+        "targets": {"node": "14"},
```
willdurand, line 5: I am wondering if this is needed given we should use the right node version in the CI job that does the build + publish.

I am actually wondering if that wouldn't actually have the opposite effect and hide potential bugs on Node > 14 (because that will force babel to stick to Node 14 features).

.babelrc
```diff
@@ -1,13 +1,21 @@
...
+        "modules": false
```
willdurand, line 6: Why?

.eslintrc
```diff
@@ -26,10 +26,10 @@
   "settings": {
     "ft-flow": {
       "onlyFilesWithFlowAnnotation": true
-    },
+    }
   },
   "env": {
-    "node": true,
+    "node": false,
```
willdurand, line 32: Why?

bin/web-ext.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import webExt from '../lib/main.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
willdurand, line 9: Any reason for naming these variables like the (old) built-in node variables?

package.json
```diff
@@ -3,24 +3,25 @@
   "version": "6.8.0",
   "description": "A command line tool to help build, run, and test web extensions",
   "main": "dist/web-ext.js",
+  "type": "module",
```
willdurand, line 6: 🔥

tests/babel-loader.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * NOTE: This ESM module is being used as a node loader (https://nodejs.org/api/esm.html#esm_experimental_loaders)
+ * while running mocha tests, e.g.:
+ *
+ *     npx mocha -n "loader=./tests/babel-loader.js" -r tests/setup.js tests/unit/test.config.js
+ *
+ * it is responsable to transpile on the fly the tests module using babel (otherwise the flowtype syntaxes
```
willdurand, line 7:
```suggestion
 * it is responsible to transpile on the fly the tests module using babel (otherwise the flowtype syntaxes
```

tests/babel-loader.js
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * - https://github.com/giltayar/babel-register-esm
+ *
+ * COMPATIBILITY NOTES:
+ *
+ * nodejs module loader API is experimental and so different nodejs versions expects
+ * a different set of hooks:
+ *
+ * - in nodejs < 16: `getSource` and `transformSource`
+ * - in nodejs >= 16: `resolve` and `load`
```
willdurand, line 20: 😞

tests/unit/test-extension-runners/test.firefox-android.js
```diff
@@ -984,37 +984,37 @@ describe('util/extension-runners/firefox-android', () => {
         {
           params: {profilePath: '/fake/dir'},
           expectedMessage: (
-            /Android target does not support --profile-path/
+            (/Android target does not support --profile-path/)
```
willdurand, line 987: Why do we need two sets of parens?

.eslintrc
rpl: it was temporary measure to ask eslint to please shut up until I got a good picture of what I needed to change :-)
I can revert that to `true` along with removing those last __filename and __dirname from bin/web-ext.js.

bin/web-ext.js
rpl: good catch, I did that at the start as a quick fix to get enough of web-ext to run `web-ext run` in the initial set of exploratory changes, but I was planning to remove them along with that "node: false" set in .eslintrc.

tests/babel-loader.js
rpl: hehe, I told you that I got "multiple source of fun" in the process of adapting web-ext to type module :rofl:

tests/unit/test-extension-runners/test.firefox-android.js
rpl: really no reasons.

I suspect this may have been a jscodeshift doing, maybe a side-effect of turning the transformed AST back into js code (e.g. in a previous version of the transformer I was missing a `quote: 'single'` option and all strings got a double quote in the transformed files).

I'm pretty sure that I removed another similar one that I noticed but in the big diff I missed this one.

.babelrc
```diff
@@ -1,13 +1,22 @@
...
+      "@babel/env", {
+        "targets": {"node": "14"},
+        // Leave import/export statements unchanged in the babel transpiling output.
+        "modules": false
+      }
+    ],
     "@babel/flow"
   ],
   "plugins": [
     "@babel/plugin-proposal-class-properties",
```
willdurand, line 13: I see this is included in ES2022 and we're using `preset-env` so could this plugin be removed?

.eslintrc
```diff
@@ -39,7 +39,10 @@
   "globals": {
     "exports": false,
     "module": false,
-    "require": false
+    "require": false,
+    "process": true,
```
willdurand, line 43: shouldn't this one be already part of the `node` env? (and maybe some of the new additions as well)

tests/unit/helpers.js
```diff
@@ -316,3 +320,37 @@ export function createFakeStdin(): tty.ReadStream {
...
+  defaultExport,
+}: {
+  moduleURL: string,
+  importerModuleURL: ?string,
+  namedExports?: ?Object,
+  defaultExport?: any,
+}): void {
+  // Compute the full URL to the module to mock, otherwise
+  // quibble will compute the wrong module URL when running
+  // on windows (which would be looking as "C:\\C:\\Users|...").
```
willdurand, line 344:
```suggestion
  // on windows (which would be looking as "C:\\C:\\Users\...").
```

package.json
```diff
@@ -116,8 +119,7 @@
     "prettyjson": "1.2.1",
     "shelljs": "0.8.5",
     "sinon": "12.0.1",
-    "webpack": "5.24.2",
```
willdurand, line 119: So we don't need webpack anymore? Which means no minification and no bundling anymore?

rpl: exactly, the modules in src/ are transpiled into the modules in lib/, no bundling just native nodejs module loading.

.babelrc
rpl: yep, it looks like we can remove both this one and the one that follows it `@babel/plugin-transform-runtime`, this last one doesn't actually seem to make any actual difference in the resulting transpiled modules, I double-checked that locally by looking to the diffs).

Removing `@babel/plugin-proposal-class-properties` instead gets rid of a imported module to polyfill that using defineProperty:

```
--- lib.with-plugin-class-properties/util/temp-dir.js        2022-06-01 12:15:39.783001797 +0200
+++ lib/util/temp-dir.js    2022-06-01 12:14:38.777841288 +0200
@@ -1,4 +1,3 @@
-import _defineProperty from "@babel/runtime/helpers/defineProperty";
 import { promisify } from 'util';
 import tmp from 'tmp';
 import { createLogger } from './logger.js';
@@ -46,10 +45,6 @@

 export class TempDir {
   constructor() {
-    _defineProperty(this, "_path", void 0);
-
-    _defineProperty(this, "_removeTempDir", void 0);
-
     this._path = undefined;
     this._removeTempDir = undefined;
   }
```

The unit tests did pass just fine without those _defineProperty calls on both node 14 and node 16, and the unit tests are using the transpiled modules and so polyfilling that was not really necessary anymore.

Based on the compatibility table on MDN, that JS syntax should be supported in nodejs starting from nodejs 12.

I'll update this PR accordingly.

src/cmd/run.js
```diff
@@ -1,26 +1,26 @@
...
-} from '../util/desktop-notifier';
-import * as defaultFirefoxApp from '../firefox';
+} from '../util/desktop-notifier.js';
+import * as defaultFirefoxApp from '../firefox/index.js';
 import {
   connectWithMaxRetries as defaultFirefoxClient,
-} from '../firefox/remote';
-import {createLogger} from '../util/logger';
-import defaultGetValidatedManifest from '../util/manifest';
-import {UsageError} from '../../src/errors';
```
rpl, line 14: @willdurand I looked into making sure the tests are importing the module they cover from src/ instead of lib/. 

One reason for doing that was that the module in src/ have the flowtype information and I can remove a couple of suppress comments that I had to add in the previous version to workaround flow complaining about missing type info.

The other reason is that I was a bit concerned that unit tests may run on an obsolete version of the transpiled files while developing a change locally, which isn't great from a developer workflow perspective.

While doing that, I caught a mistake that was already here:
this import was wrongly importing from a relative path that goes out of the src modules root dir and then points back to a src module, which was then turned into importing it from lib/ because it was rewritten by the jscodeshift transformer (which was expecting '/src/' to be part of a module url only for the tests).

That wasn't caught by the test suites because the module was there in lib and so everything kind of worked just fine, but I would say "just because of pure luck".

I fixed that in the last update to this PR, but I'm thinking that it would be nice if an eslint rule could error if any import is doing that by mistake in a new PR.

rpl:
@willdurand This PR does now cover everything that we discussed earlier today, in particular:

- I looked into the reasons why the tests covering importing web-ext as a library were not failing and fixed the tests to make sure they cover the package being tested and not the last version released on npm (which I'm pretty sure it was the reason why those two tests were not yet failing)

- added explicit coverage for the new behavior expected when trying to use web-ext as a library from a CommonJS module (asserting that it throws the error that nodejs throws internally in that scenario, and that the commonjs module can still use dynamic import to use web-ext internally)

- changed util.logger and util.adb into additional exports declared in the package.json, covered that expected behavior in the functional tests and ripped off the workaround we were previously using to export these utilities from the webpack bundle (and related unit tests)

- removed all inline TODOs related to #1306 (given that it is also fixed in this PR).

Then I also compared the package size as agreed, it is slightly bigger but imho not that much to be concerned about the difference:
> new package size 165 kB (unpacked 699 kB) vs old package size 103 kB (unpacked 442 kB).

This is now ready for another review pass from your perspective.

index.js
```diff
@@ -0,0 +1,11 @@
+// This module provides a nicer ESM module wrapper around the module generated by webpack.
```
willdurand, line 1:
```suggestion
// This file is loaded when developers import 'web-ext' in their own code.
```

index.js
```diff
@@ -0,0 +1,11 @@
...
+
+// NOTE: disabled eslint rules:
+// - import/no-unresolved: the lib/main.js file has likely not build yet when this runs in the CI jobs
+//   and this is also covered by automated tests (which would catch an unresolved import issue here).
```
willdurand, line 5:
```suggestion
// - import/no-unresolved: in the CI jobs, the `lib/main.js` file has likely not been built yet. It's fine to
//   disable this rule because we have automated tests (which would catch an unresolved import issue
//   here).
```

willdurand (approved):
r+wc, thanks

package.json
```diff
@@ -2,25 +2,27 @@
...
+    ".": "./index.js",
+    "./util/adb": "./lib/util/adb.js",
+    "./util/logger": "./lib/util/logger.js"
   },
   "files": [
     "CODE_OF_CONDUCT.md",
     "README.md",
-    "dist/*.js",
+    "lib/**",
     "src/**",
```
willdurand, line 16: Do we still need the sources in the npm packages?

rpl: at the moment the only thing that the src/ modules provides that the lib/ one don't is the flow typing information, I think we may actually remove them, but I was considering looking if it is easy enough to properly export the flowtype data along with removing the src/ files from the package (and so to defer that to a separate PR), does is sound ok to you too?

willdurand: sounds good yeah, also TIL that `README.md` is added by default so we could remove it from this list of `files` (and maybe we don't need `CODE_OF_CONDUCT.md` in the final package.

README.md
```diff
@@ -112,13 +112,15 @@ need to relink it.
 
 ## Using web-ext in NodeJS code
 
-Aside from [using web-ext on the command line][web-ext-user-docs], you may wish to execute `web-ext` in NodeJS code. There is limited support for this. Here are some examples.
+Aside from [using web-ext on the command line][web-ext-user-docs], you may wish to execute `web-ext` in NodeJS code.
+
+`web-ext` npm package exports nodejs native ES modules, and so from a NodeJS CommonJS module they can only be imported using dynamic import.
+
```
willdurand, line 118:
```suggestion
As of version `7.0.0`, the `web-ext` npm package exports NodeJS native ES modules only. If you are using CommonJS, you will have to use [dynamic imports][].

```

README.md
```diff
@@ -112,13 +112,15 @@ need to relink it.
...
+There is limited support for this. Here are some examples.
```
willdurand, line 119:
```suggestion
### Examples
```

README.md
willdurand: Unfortunately, I can make a suggestion at the bottom of this file so you'll have to include:

```
[dynamic imports]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports
```

rpl:
@willdurand Thanks a lot for all the support and help reviewing this PR :heart: 
I'll be merging it in a minute.

bin/web-ext.js
```diff
@@ -0,0 +1,13 @@
...
+
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import webExt from '../lib/main.js';
+
+const absolutePackageDir = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
+  '..'
+);
```
fregante, line 11: I think that that the new way would be to use URL instead of path, something like `fileURLToPath(new URL('..', import.meta.url))`, and you'd have the full path.

fregante (commented):
Great to see this change!

PR #2433, opened 2022-06-13
fix: rewrite scripts/develop to use ESM

willdurand:


scripts/develop.js
```diff
@@ -50,11 +50,11 @@ async function runTasks(changes) {
     return;
   }
 
-  console.log('\nBuilding web-ext webpack bundle');
-  const webpackSuccess = await webpack().then(() => true, () => false);
+  console.log('\nBuilding web-ext');
+  const babelSuccess = await babel().then(() => true, () => false);
```
rpl, line 54: babel function isn't async and so this line should just be:
```
const babelSuccess = babel();
```

rpl (changes requested):
@willdurand thanks a lot for taking care of this script, I don't use it in my usual workflow and so I missed to convert it. Just one small fix and then this is good to go.

rpl (approved):

PR #2436, opened 2022-06-13
feat: introduce a --firefox-preview flag for 'web-ext run'

willdurand:
This flag allows to enable the [Firefox Dev Preview for Manifest V3][get-involved] (at the moment, given that was my main use case). I've been light on testing but it covers enough. I don't think we need to be super fancy right now (personally).

[get-involved]: https://blog.mozilla.org/addons/2022/06/08/manifest-v3-firefox-developer-preview-how-to-get-involved/

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -43,6 +43,7 @@ export type CmdRunParams = {|
   startUrl?: Array<string>,
   target?: Array<string>,
   args?: Array<string>,
+  firefoxPreview: Array<string>,
 
   // Android CLI options.
   adbBin?: string,
@@ -89,6 +90,7 @@ export default async function run(
     startUrl,
     target,
     args,
+    firefoxPreview = [],
     // Android CLI options.
     adbBin,
     adbHost,
@@ -126,7 +128,12 @@ export default async function run(
 
   // Create an alias for --pref since it has been transformed into an
   // object containing one or more preferences.
-  const customPrefs = pref;
+  const customPrefs: FirefoxPreferences = {...pref};
+  if (firefoxPreview.includes('mv3')) {
+    log.info('Configuring Firefox preferences for Manifest V3');
+    customPrefs['extensions.manifestV3.enabled'] = true;
+  }
+
   const manifestData = await getValidatedManifest(sourceDir);
 
   const profileDir = firefoxProfile || chromiumProfile;
--- a/src/program.js
+++ b/src/program.js
@@ -691,6 +691,12 @@ Example: $0 --help run.
         demandOption: false,
         type: 'array',
       },
+      'firefox-preview': {
+        describe: 'Turn on developer preview features in Firefox',
+        demandOption: false,
+        default: ['mv3'],
+        type: 'array',
+      },
       // Firefox for Android CLI options.
       'adb-bin': {
         describe: 'Specify a custom path to the adb binary',
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -454,4 +454,22 @@ describe('run', () => {
          );
        });
   });
+
+  describe('firefox-preview', () => {
+    it('supports the MV3 preview', async () => {
+      const cmd = await prepareRun();
+      const runOptions = {
+        firefoxPreview: ['mv3'],
+      };
+
+      await cmd.run(runOptions);
+
+      sinon.assert.calledOnce(desktopRunnerStub);
+      const runnerParams = desktopRunnerStub.firstCall.args[0];
+
+      assert.deepEqual(runnerParams.customPrefs, {
+        'extensions.manifestV3.enabled': true,
+      });
+    });
+  });
 });
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -912,6 +912,16 @@ describe('program.main', () => {
     }
   });
 
+  it('sets the default firefox preview to "mv3"', async () => {
+    const fakeCommands = fake(commands, {
+      run: () => Promise.resolve(),
+    });
+
+    await execProgram(['run', '--firefox-preview'], {commands: fakeCommands});
+
+    const {firefoxPreview} = fakeCommands.run.firstCall.args[0];
+    assert.deepEqual(firefoxPreview, ['mv3']);
+  });
 });
 
 describe('program.defaultVersionGetter', () => {


rpl (approved):
Thanks Will, as I anticipated you I think this is a actually a great idea and the PR looks great as is.

We should create a followup issue in the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop/issues) repo to add a section for this option in the web-ext command reference here https://extensionworkshop.com/documentation/develop/web-ext-command-reference/

And I think that as part of that it may be worth also to consider calling this option out as part of documentation page providing guidance for transitioning or developing mv3 extensions in Firefox.
wdyt?

In the meantime I'll file the issue and I'll mention in its description that we coordinate merging the changes to the web-ext references with the releasing on npm of a new web-ext version that provides this new cli option.

Rob--W:
Let's call it `--enable-mv3` or something from which it's obvious that mv3 is enabled with it?

Side note: Firefox Preview is the name of the initial test version of the GeckoView-based Firefox for Android.

rpl:
> Let's call it `--enable-mv3` or something from which it's obvious that mv3 is enabled with it?
> 
> Side note: Firefox Preview is the name of the initial test version of the GeckoView-based Firefox for Android.

There were a couple of reasons why we ended up settling on --firefox-preview:
- this is Firefox only, it doesn't have any effect on the Chromium extension runner
- if we call it `--something-mv3` (I think I have initially said to William that I was likely going to prefer something like `--firefox-mv3-devprefix` or something like that), then we would also become an obsolete cli option soon enough and we would need to remove it in a new major release, and so we were settling on `--firefox-preview=mv3` (with mv3 as the current default) to add an option that may actually make sense to keep for longer (e.g. use it to also enable other "developer preview" features, like event pages in MV2 and some future one we may need to intrduce)

I'm not sure worried that the option may sound like a temporary code name assigned to old GeckView-based Firefox for Android versions, but given that we explicitly used the term "Developer Preview" to refer to to the experimental mv3 support locked behind a pref maybe we could tweak it a little bit into: `--firefox-devpreview`

@willdurand how `--firefox-devpreview` would sound to you?

willdurand:
> And I think that as part of that it may be worth also to consider calling this option out as part of documentation page providing guidance for transitioning or developing mv3 extensions in Firefox.
> wdyt?

Yeah, I agree

rpl:
> > @willdurand how `--firefox-devpreview` would sound to you?
> 
> too long :) `--dev-preview` would work for me, though

@willdurand how about `--firefox-devpreview` as the option name but with also a shorter alias like `--devpreview` or `--dev-preview`? :grin:

Rob--W:
In Chrome, MV3 is not behind a flag, so something like --enable-mv3 doing nothing would not be an issue.

Since conciseness is apparently desired, we could even go with
--mv3

This also leaves the option open to offer more specific mv3 flags in the future, e.g. --mv3=force (which could treat mv2 as mv3, to quickly allow devs to check whether the extension could work if they only change mv2 to mv3).

willdurand:
> @willdurand how about `--firefox-devpreview` as the option name but with also a shorter alias like `--devpreview` or `--dev-preview`? 😁

I don't think an alias would work because if we refer to the alias instead of the full flag name, let's just not use an alias.

I think this discussion means `--firefox-preview` is the best choice so far given we want to (1) have a Firefox specific flag, (2) not too long and (3) that we can reuse for something else in the future.

(FTR the initial flag name I had was `--preview` because of our "dev preview")

> Since conciseness is apparently desired, we could even go with --mv3
> 
> This also leaves the option open to offer more specific mv3 flags in the future, e.g. --mv3=force (which could treat mv2 as mv3, to quickly allow devs to check whether the extension could work if they only change mv2 to mv3).

I would be OK with `--mv3` (short, self-explanatory, little to no side effects on non-Firefox browsers) but that goes against (3) above.

rpl:
> > @willdurand how about `--firefox-devpreview` as the option name but with also a shorter alias like `--devpreview` or `--dev-preview`? grin
> 
> I don't think an alias would work because if we refer to the alias instead of the full flag name, let's just not use an alias.
> 
> I think this discussion means `--firefox-preview` is the best choice so far given we want to (1) have a Firefox specific flag, (2) not too long and (3) that we can reuse for something else in the future.
> 
> (FTR the initial flag name I had was `--preview` because of our "dev preview")
> 
> > Since conciseness is apparently desired, we could even go with --mv3
> > This also leaves the option open to offer more specific mv3 flags in the future, e.g. --mv3=force (which could treat mv2 as mv3, to quickly allow devs to check whether the extension could work if they only change mv2 to mv3).
> 
> I would be OK with `--mv3` (short, self-explanatory, little to no side effects on non-Firefox browsers) but that goes against (3) above.

would `--devpreview` (and `["firefox-mv3"]` as the current default value) be a choice that matches all 3 points and be short and self-explanatory enough from all our 3 perspective?

In my opinion, this is shorter in its most common use cases, and it would still make it pretty clear if a `devpreview` feature being requested applies only to a specific browser.

This would also allow to extend the CLI option to more quickly enable experimental features in Chrome based browsers as well (just as an example, if the promises API in chrome was locked behind a CLI flag in Chrome then we could support a `'chrome-promised-api'` devpreview value that would pass the chrome flag automatically, as this PR does for the Firefox about:config prefs to enable manifest_version 3 support).

As a side note, `web-ext run` does actually accept more than one browser target at the time, e.g. the following command would run a manifest version 3 extension on both Chrome and Firefox:

```
$ web-ext run --devpreview -s /path/to/test-mv3-extension/ -t firefox-desktop -t chromium
```

and so being able to extend a `--devpreview` cli option to the non-Firefox extension runners seems something that could turn out to be useful, and so a cli option that we could keep for longer than just the Firefox MV3 DevPreview phase. 

@willdurand @robwu how that sounds to you?

willdurand:
> @willdurand @Rob--W how that sounds to you?

Can we just stick to what we have in the current patch?

willdurand:
(rebase)

willdurand:
As per our Slack discussion, I am landing this patch now.

willdurand:
> We should create a followup issue in the [mozilla/extension-workshop](https://github.com/mozilla/extension-workshop/issues) repo to add a section for this option in the web-ext command reference here [extensionworkshop.com/documentation/develop/web-ext-command-reference](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/)

I submitted a PR: https://github.com/mozilla/extension-workshop/pull/1339

PR #2454, opened 2022-07-08
fix: firefoxPreview option should be set to mv3 only when explicitly requested

rpl:
The --firefox-preview web-ext option we introduced in v7.1.0 should not default to mv3 if it wasn't explicitly passed as an option (either on the command line or through the web-ext config file).

Unfortunately setting a default value through the yargs `default` property will always set `argv.firefoxPreview` to `['mv3']` even when it is not explicitly passed, `demandOptions` set to `false` makes the cli option not mandatory but it doesn't ignore the default value.

This PR adds an additional test case to make sure that `argv.firefoxPreview` is `undefined` when not explicitly passed as an option, and a small tweak to program.js that make sure `argv.firefoxPreview` defaults to `['mv3']` if it was set to an empty array (and so when `--firefox-preview` was explicitly passed but without any value specified explicitly).

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -221,6 +221,10 @@ export class Program {
       throw new UsageError('Not enough arguments following: start-url');
     }
 
+    if (Array.isArray(argv.firefoxPreview) && !argv.firefoxPreview.length) {
+      argv.firefoxPreview = ['mv3'];
+    }
+
     return argv;
   }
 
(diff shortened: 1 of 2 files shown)


rpl:
@willdurand my apologies, I should have caught this during the review.

willdurand (approved):

willdurand:
oops

PR #2477, opened 2022-08-09
feat: allow run command to use flatpak

### LINKED ISSUE #2465: Run with flatpak version
Unable to `web-ext` run with `flatpak` version of firefox on `fedora 36`.

```bash
❯ web-ext run -f "flatpak run org.mozilla.firefox"
Running web extension from /mnt/Projects/Products/joinblob/extension

Error: not found: flatpak run org.mozilla.firefox
    at F (/home/aniket/.config/yarn/global/node_modules/fx-runner/node_modules/which/which.js:63:19)
    at E (/home/aniket/.config/yarn/global/node_modules/fx-runner/node_modules/which/which.js:72:29)
    at /home/aniket/.config/yarn/global/node_modules/fx-runner/node_modules/which/which.js:81:16
    at /home/aniket/.config/yarn/global/node_modules/fx-runner/node_modules/isexe/index.js:44:5
    at /home/aniket/.config/yarn/global/node_modules/fx-runner/node_modules/isexe/access.js:8:5
    at FSReqCallback.oncomplete (node:fs:201:23)
```

willdurand:
Fixes #2465

---

~~This patch isn't enough because the [fx-runner](https://github.com/mozilla/node-fx-runner/) does weird things with the command parameters. With some small changes,~~ things work fine, though:

Screenshot 2022-08-09 at 10 08 57

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -21,6 +21,7 @@ import type {
   FirefoxPreferences,
 } from './preferences';
 import type {ExtensionManifest} from '../util/manifest.js';
+import type {Extension} from '../extension-runners/base.js';
 
 
 const log = createLogger(import.meta.url);
@@ -50,6 +51,7 @@ export type FirefoxRunnerParams = {|
   'foreground'?: boolean,
   'listen': number,
   'binary-args'?: Array<string> | string,
+  'binary-args-first'?: boolean,
   'env'?: {
     // This match the flowtype signature for process.env (and prevent flow
     // from complaining about differences between their type signature)
@@ -87,6 +89,7 @@ export type FirefoxRunOptions = {
   firefoxBinary?: string,
   binaryArgs?: Array<string>,
   args?: Array<any>,
+  extensions: Array<Extension>,
 };
 
 /*
@@ -97,7 +100,9 @@ export async function run(
   {
     fxRunner = defaultFxRunner,
     findRemotePort = defaultRemotePortFinder,
-    firefoxBinary, binaryArgs,
+    firefoxBinary,
+    binaryArgs,
+    extensions,
   }: FirefoxRunOptions = {}
 ): Promise<FirefoxInfo> {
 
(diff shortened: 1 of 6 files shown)


src/firefox/index.js
```diff
@@ -97,14 +99,34 @@ export async function run(
...
+    const flatpakAppId = firefoxBinary.substring(8);
+    log.debug(`Configuring Firefox with flatpak: appId=${flatpakAppId}`);
+
+    // This should be resolved by the fx-runner.
+    firefoxBinary = 'flatpak';
+    binaryArgs = [
+      'run',
+      `--filesystem=${profile.path()}`,
+      ...extensions.map(({ sourceDir }) => `--filesystem=${sourceDir}`),
+      // We need to share the network namespace because we want to connect to
```
Rob--W, line 122: If for devtools only, Firefox supports UNIX domain sockets with the `--start-debugger-server` option: https://searchfox.org/mozilla-central/rev/c0bed29d643393af6ebe77aa31455f283f169202/devtools/startup/DevToolsStartup.jsm#955

Could that be used to avoid the need for sharing the same network namespace?

src/firefox/index.js
```diff
@@ -97,14 +99,34 @@ export async function run(
...
 
+  if (firefoxBinary && firefoxBinary.startsWith('flatpak:')) {
+    const flatpakAppId = firefoxBinary.substring(8);
+    log.debug(`Configuring Firefox with flatpak: appId=${flatpakAppId}`);
+
+    // This should be resolved by the fx-runner.
+    firefoxBinary = 'flatpak';
+    binaryArgs = [
+      'run',
+      `--filesystem=${profile.path()}`,
```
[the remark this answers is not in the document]
willdurand: I was more thinking about adding `:ro` to the extension folders because Firefox could write to the profile, no?

src/firefox/index.js
willdurand: Maybe but our code does not handle that so we can maybe support UNIX sockets in a follow-up.

willdurand:
> This patch isn't enough because the [fx-runner](https://github.com/mozilla/node-fx-runner/) does weird things with the command parameters.

There is the patch for fx-runner: https://github.com/mozilla/node-fx-runner/pull/96

rpl (approved):
@willdurand this looks pretty good! thank you so much for digging into this :heart: 

I'm also fine with updating fx-runner dependency as part of this PR, the lock file change is also pretty minimal. Renovatebot will detect that we already updated fx-runner and will close the other PR on its own.

rpl:
@willdurand I just filed https://github.com/mozilla/extension-workshop/issues/1375 to track the related updates to the web-ext commands reference available at https://extensionworkshop.com/documentation/develop/web-ext-command-reference/

Issue #759, opened 2017-01-18
`web-ext run` should allow for starting with debugger enabled.


sblask:
Feature request: it would be great if `web-ext run` would have a `--debugging-enabled` flag so I don't have to go to about:debugging and click on `debug` for my add-on on every start.

kumar303:
Yes, this would be nice, especially since auto-reloading exits the debugger (at least it used to but I can no longer find the bug).

@rpl would we need a Firefox patch for this? Maybe to use a remote actor to turn this on? If so, perhaps we should file the bugs ASAP.

rpl:
@sblask :+1:

@kumar303 the WebExtensions Addon Debugger should survive auto-reloading (the old one didn't, but when I worked on the new one I took care of this scenario)

Nevertheless I totally agree that it would be helpful if we could "auto-open" the tool on the installed addon without the need to start it manually, I'm pretty sure that we need a patch for it, as an example we could define an additional `{openAddonDebugger: true}` parameter on the [addons RDP actor `installTemporaryAddon` method](https://dxr.mozilla.org/mozilla-central/source/devtools/server/actors/addons.js#19), which will open the addon debugger as soon as the addon has been installed.

kumar303:
@rpl yeah, I like `installTemporaryAddon(path, {openAddonDebugger: true})` -- can you file a bug so it can get into triage?

rpl:
@kumar303 sure thing :+1:

rpl:
@kumar303 Filed on bugzilla as [Bug 1332299 - Optionally open the Addon Debugger automatically after an addon is installed using the RDP method addonsActor.installTemporaryAddons](https://bugzilla.mozilla.org/1332299).

yangwen2:
bump on this. Is web-ext still a focus given Mozilla's recent strategy pivot?

rpl:
> bump on this. Is web-ext still a focus given Mozilla's recent strategy pivot?

This tool is actively maintained :slightly_smiling_face: 
The priority of this particular issue is still ‘enhancement’ and not currently being worked on.

A reasonable path forward for this issue would be interest from a contributor to work on it
(we didn't mark it as a good first bug because implementing it will likely require a more experienced contributor).

curtisbelt:
> Feature request: it would be great if web-ext run would have a --debugging-enabled flag so I don't have to go to about:debugging and click on debug for my add-on on every start.

Here's a solution that is working very well for me.

[...]

PR #2488, opened 2022-08-26
feat: Introduce webext run --devtools to open DevTools for the installed add-on right away.

ochameau:
This depends on bug [1787409](https://bugzilla.mozilla.org/show_bug.cgi?id=1787409) and aims at addressing issue #759.

### DIFF
--- a/src/firefox/index.js
+++ b/src/firefox/index.js
@@ -90,6 +90,7 @@ export type FirefoxRunOptions = {
   binaryArgs?: Array<string>,
   args?: Array<any>,
   extensions: Array<Extension>,
+  devtools: boolean,
 };
 
 /*
@@ -103,6 +104,7 @@ export async function run(
     firefoxBinary,
     binaryArgs,
     extensions,
+    devtools,
   }: FirefoxRunOptions = {}
 ): Promise<FirefoxInfo> {
 
@@ -164,9 +166,13 @@ export async function run(
     throw error;
   });
 
-  log.info(
-    'Use --verbose or open Tools > Web Developer > Browser Console ' +
-    'to see logging');
+  if (!devtools) {
+    log.info('Use --verbose or --devtools to see logging');
+  }
+  if (devtools) {
+    log.info('More info about WebExtensions debugging:');
+    log.info('https://extensionworkshop.com/documentation/develop/debugging/');
+  }
 
   firefox.stderr.on('data', (data) => {
     log.debug(`Firefox stderr: ${data.toString().trim()}`);
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -28,6 +28,7 @@ const log = createLogger(import.meta.url);
 export type CmdRunParams = {|
   artifactsDir: string,
   browserConsole: boolean,
+  devtools: boolean,
   pref?: FirefoxPreferences,
   firefox: string,
   firefoxProfile?: string,
@@ -75,6 +76,7 @@ export default async function run(
   {
     artifactsDir,
     browserConsole = false,
+    devtools = false,
     pref,
     firefox,
     firefoxProfile,
@@ -174,6 +176,7 @@ export default async function run(
       profilePath: firefoxProfile,
       customPrefs,
       browserConsole,
+      devtools,
       preInstall,
 
       // Firefox runner injected dependencies.
--- a/src/extension-runners/firefox-desktop.js
+++ b/src/extension-runners/firefox-desktop.js
@@ -243,6 +243,7 @@ export class FirefoxDesktopExtensionRunner {
       firefoxBinary,
       binaryArgs,
       extensions,
+      devtools,
     });
 
     this.runningInfo.firefox.on('close', () => {
--- a/src/extension-runners/firefox-desktop.js
+++ b/src/extension-runners/firefox-desktop.js
@@ -33,6 +33,7 @@ import type {FirefoxInfo} from '../firefox/index'; // eslint-disable-line import
 type FirefoxDesktopSpecificRunnerParams = {|
   customPrefs?: FirefoxPreferences,
   browserConsole: boolean,
+  devtools: boolean,
   firefoxBinary: string,
   preInstall: boolean,
 
@@ -212,6 +213,7 @@ export class FirefoxDesktopExtensionRunner {
   async startFirefoxInstance() {
     const {
       browserConsole,
+      devtools,
       extensions,
       firefoxBinary,
       preInstall,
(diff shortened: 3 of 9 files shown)


rpl (changes requested):
@ochameau thank you so much for looking into the improvements to the webextensions debugging, I just tried this patch locally combined with (most of) the changes from [Bug 1787409](https://bugzilla.mozilla.org/show_bug.cgi?id=1787409) (and the other patch in that stack) and it definitely provides a nicer developer experience.

It seems also (at least if I'm not mistaken) that on older Firefox versions where the changes from Bug 1787409 will not be available the additional `openDevTools` property in the installTemporaryAddon RDP request is just ignored, and so it looks that we will not need to do any runtime detection and fallback, which is also great.

Follows a couple of small tweaks to let the PR to run all tests in the CI job (fixing linting and flow error, and an unrelated test failure due to the change to the method signature that the sinon stub wasn't expecting).

Would you mind to also remove the package-lock.json change? that is unrelated (and pretty big because running `npm install` with an older npm version rewrite the entire lock file to lockfileVersion 1).

You should be able to avoid the automatic rewite of the lock file by running `npm ci` instead of `npm install`.

ochameau:
> It seems also (at least if I'm not mistaken) that on older Firefox versions where the changes from Bug 1787409 will not be available the additional `openDevTools` property in the installTemporaryAddon RDP request is just ignored, and so it looks that we will not need to do any runtime detection and fallback, which is also great.

Yes I confirm, no backward compat is needed, unless we want to warn when using --devtools on an unsupported version.

Thanks for all the suggestion to fix all ci issues \o/

ochameau:
I only started asking for review on https://bugzilla.mozilla.org/show_bug.cgi?id=1787409
So no rush in reviewing/landing this.

ochameau:
I'm also trying to document this new command line argument [over there](https://github.com/mozilla/extension-workshop/pull/1398).

ochameau:
I pushed a change to fix lint issues, but jobs are still failing whereas `yarn test` works locally.

rpl:
> I pushed a change to fix lint issues, but jobs are still failing whereas `yarn test` works locally.

@ochameau sorry for not having got back to this sooner, I noticed it and I was meant to comment but didn't got back to this quickly enough (or at least not as quickly as you ;-))

Those failures are coming from the function, which are the ones that can be also run locally using [`npm run test-functional`](https://app.circleci.com/pipelines/github/mozilla/web-ext/1643/workflows/5cb26fc3-d663-4662-b515-25fbc12e7687/jobs/6654?invite=true#step-113-2), the reason for the failure (and related fix) is actually trivial:
- the `web-ext run` functional test is currently using a nodejs script that mocks Firefox and it doesn't currently expect the new option and so it fails, the reason is not immediately visible in the failure message but it is part of the failure logs emitted by the `tests/functional/fake-firefox-binary.js` script [here where it says `Fake Firefox received an unexpected message`](https://app.circleci.com/pipelines/github/mozilla/web-ext/1643/workflows/5cb26fc3-d663-4662-b515-25fbc12e7687/jobs/6654?invite=true#step-113-253)

The following one line patch should fix that remaining CI job failure:
```
diff --git a/tests/functional/fake-firefox-binary.js b/tests/functional/fake-firefox-binary.js
index 2f86902..00ffb3f 100755
--- a/tests/functional/fake-firefox-binary.js
+++ b/tests/functional/fake-firefox-binary.js
@@ -9,6 +9,7 @@ const REQUEST_INSTALL_ADDON = {
   to: 'fakeAddonsActor',
   type: 'installTemporaryAddon',
   addonPath: process.env.addonPath,
+  openDevTools: false, // Introduced in Firefox 106 (Bug 1787409 / Bug 1789245)
 };
 const REPLY_INSTALL_ADDON = {
   from: 'fakeAddonsActor',
```

ochameau:
Thanks for the investigate. I hope jobs will be green now.

src/firefox/index.js
```diff
@@ -164,9 +168,13 @@ export async function run(
...
 
-  log.info(
-    'Use --verbose or open Tools > Web Developer > Browser Console ' +
-    'to see logging');
+  if (!verbose && !devtools) {
+    log.info('Use --verbose or --devtools to see logging');
+  }
+  if (devtools) {
+    log.info('More info about WebExtension debugging:');
+    log.info('https://extensionworkshop.com/documentation/develop/debugging/');
```
rpl, line 176: @ochameau I'm thinking that we may as well always log this couple of messages, instead of condition them on the verbose and devtools flags.

wdyt?

As a side-effect, that may also make codecov to don't complain about these logs emitted on `devtools === true` not being covered in tests (which are not a big deal on their own, given that its just two informative logs I wouldn't definitely block on that codecov coverage failure).

rpl (commented):
@ochameau looks pretty great, I have just one more thoughts described in the inline review comment below (I'm basically wondering if we really need to condition those logs, if we think it is not necessary we can revert a couple of changes related to propagating verbose and avoid the codecov failure as a side effect).

src/firefox/index.js
ochameau: TBH, I'm not working enough with this tooling to have a strong opinion.

So I would prefer to defer to you/your team judgement.
I'm fine removing, or adding proper coverage.

rpl: ok, let's remove the `verbose` flag propagation and keep that log unconditioned, while condition the new one to the `devtools` flag.

Let's also change `WebExtension` to `WebExtensions` in the new logged message and let's add something like this to `test.firefox.js` to explicitly cover it and make codecov happy again:
```
import {
  consoleStream, // instance is imported to inspect logged messages
} from '../../../src/util/logger.js';

...

    it('logs link to debugging docs', async () => {
      const runner = createFakeFxRunner();
      consoleStream.startCapturing();

      const expectedMessage = 'More info about WebExtensions debugging:';
      const expectedURLToDocs =
        'https://extensionworkshop.com/documentation/develop/debugging/';
      await runFirefox({fxRunner: runner, devtools: false});
      assert.notOk(consoleStream.capturedMessages.find(
        (msg) => msg.includes(expectedMessage)
      ));
      assert.notOk(consoleStream.capturedMessages.find(
        (msg) => msg.includes(expectedURLToDocs)
      ));

      await runFirefox({fxRunner: runner, devtools: true});
      const foundMessage = consoleStream.capturedMessages.find(
        (msg) => msg.includes(expectedMessage)
      );
      const foundDocURL = consoleStream.capturedMessages.find(
        (msg) => msg.includes(expectedURLToDocs)
      );

      // Expect the logs to be found.
      assert.ok(foundMessage);
      assert.ok(foundDocURL);

      // Expected to be emitted as info level logs.
      assert.ok(foundMessage?.includes('[info]'));
      assert.ok(foundDocURL?.includes('[info]'));
    });
```

rpl (changes requested):

ochameau:
Thanks a lot for the test snippet!

willdurand:
I am going to land this PR to avoid future conflicts.

PR #2489, opened 2022-08-26
Allow JSON metadata to be submitted to addon submission API

### LINKED ISSUE #2487: Support passing JSON/form data to the addon submission API
follow-on from #2451 
AMOs addon submission api supports setting a number of add-on properties on add-on and version creation - and strictly requires some of these properties be provided while creating the first listed version of an add-on - so web-ext needs to support providing these parameters from the user and passing them to the API.

eviljeff:
fixes #2487


src/program.js
```diff
@@ -569,29 +569,31 @@ Example: $0 --help run.
...
-          describe: 'The channel for which to sign the addon. Either ' +
-          '\'listed\' or \'unlisted\'',
+          describe: 'The channel for which to sign the addon.',
+          type: 'string',
+          choices: ['listed', 'unlisted'],
+        },
+        'metadata': {
+          describe: 'A JSON object containing metadata to be passed to the ' +
+          'API.  Only used with `use-submission-api`',
           type: 'string',
```
rpl, line 597: `metadata` feels a bit generic to me, even in the context of the sign command, I'm wondering if `listing-metadata` or `store-metadata` may be more immediately clear about what metadata that refers to.

rpl, line 597: Question: would make sense to let the user pass a path to json data stored in a file instead of having to pass the json string as the option value?

I'm guessing that it would be easier for a developer to store and edit the json metadata from a file than doing it on the command line, wdyt?

If we will agree on doing that, then I think that `submit-addon.js` may still be getting the metadata json already parsed, as in this version, and instead `cmd/sign.js` will be getting the file path and turning that into the parsed metadata object that `submit-addon.js` will receive.

src/program.js
```diff
@@ -569,29 +569,31 @@ Example: $0 --help run.
           describe:
             'Use a proxy to access the signing API. ' +
             'Example: https://yourproxy:6000 ',
-          demandOption: false,
```
rpl, line 572: are these removed `demandOption: false` related to this set of changes?

src/program.js
```diff
@@ -569,29 +569,31 @@ Example: $0 --help run.
...
         },
         'channel': {
-          describe: 'The channel for which to sign the addon. Either ' +
-          '\'listed\' or \'unlisted\'',
+          describe: 'The channel for which to sign the addon.',
+          type: 'string',
+          choices: ['listed', 'unlisted'],
+        },
+        'metadata': {
+          describe: 'A JSON object containing metadata to be passed to the ' +
```
rpl, line 595: I think that for this option it may be useful if we could include in the doc string the url of the doc page where we describe what is the format for this metadata.

rpl (commented):
Follows an initial set of feedback comments (mainly thoughts and questions in this pass).

src/program.js
eviljeff: We don't refer to AMO as a `store` anywhere, so not `store-metadata`.  I guess `listing-metadata` is almost accurate - though there are a few properties you can set/update that aren't listing related (`approval_notes` and `source` - though we don't support uploads for `source`, nor editing the version to add it)

eviljeff: that could be easier to manage for a developer.  I was wondering if there was some existing file we can leverage for this rather than yet-another-random-file in the directory.  `package.json`?

rpl: The only that comes to my mind is the web-ext config file, a `package.json` is likely common enough but I wouldn't assume that to be always the case (technically an extension may be as well written withou a package.json, and web-ext doesn't require one to exist).

The web-ext config file is strictly paired with the config option defined on the web-ext side, and so what it is going to accept depends from the format of the cli option (and also the option name and what command it belongs to).

How common would be for a developer to pass the metadata in all submission?

if we are convinced that would be common enough, then we may opt to provide two alternative cli options:

- `--listing-metadata` which would expect the JSON stringified data, and that could be used in a config file to avoid having to refer to a separate json file for the listing metadata

- `--listing-metadata-file` which would expected to be used when the metadata is in a separate json file, which would be easier to use when the developer want to manually pass the listing metadata only for particular sign requests (e.g. when first submitting a new addon, and when submitting an update that would require changes, e.g. if they are changing license in that version may be one possible reason I can think of)

It may be also worth for web-ext to be raising a UserError if the user passed both options (which may also happen if one is in the config file or passed through environment variables), or we may ignore the file if the `--listing-metadata` is passed and print a warning.

@eviljeff how that sounds to you?

rpl: An alternative may be `addon-metadata` but I'm not sure if that may be clear enough or likely be confused for something related to the manifest file (like the browser specific settings), if we believe that `listing-metadata` is accurate enough (even if technically a couple of those are not part of what a listing is on the AMO side), I'd be ok with going for `listing-metadata`.

src/program.js
rpl: k, then let's omit it from the new one and add back the ones that are technically the unrelated change for this PR, we can consider doing that if that is the yargs default now, but I'd prefer to be doing that in a separate PR.

src/program.js
eviljeff: > How common would be for a developer to pass the metadata in all submission?

Currently the same metadata _could_ be passed to subsequent versions for the same add-on, but it'd be redundant at best (i.e. you don't need set the categories each time) and possibly overwrite anything you'd changed via devhub in the meantime.  If we expand `web-ext` to support icon arbitrary metadata updates then it'd conflict with those requests so we'd only be able to use it for `sign`.

The web-ext config file felt like the wrong place to me primarily because the one in the home directory would be used for every add-on submission while you can use `web-ext` to develop any number of different add-ons. But I suppose we can document that while the config _can_ be added to the config file in the home directory, we would recommend it be in the local folder `web-ext-config.js` file so per project.

According to https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/#uautomatic-discovery-of-configuration-files it could be in the `package.json` file too.  The annoying part here would be that it'd be JSON data, encoded in a single string, stored in a JSON file 😬.

eviljeff: So, the no-extra-code option is we rely on the existing config setup (reading from either `.web-ext-config.js` or `web-ext-config.js` or `package.json`).
Though it's sub-optimal because a) it'd be used for all `sign` requests even though only the first request could need it - unless we can reliably know what is the first request so not have to send it for all; and b) it'd be JSON encoded as a string in a JSON (or Javascript) file.

rpl: @eviljeff I mentioned this to @willdurand (while we have been discussed about also other things that I wanted to gather his perspective for once he was back) and we both agreed that your proposed name `amo-metadata` sounds even better (and it is also shorter, which is also nice) and so let's go for `amo-metadata` as you also suggested.

----

As a side note (I'm mentioning here mainly because it is also related to the submit-addon module and about using the same "option name convention" being chosen here):

@willdurand and I also discussed about that remaining `api-host` vs `api-url-prefix` detail form the initial pull request for the submit-addon.js module, and we agreed on some changes, and one was to rename the new option `amo-base-url`, which would also make the two new options to nicely share `amo-` in their option name.
But let's discuss between the three of us (you, @willdurand and I) in more detail about changes related to the `api-host` option name tomorrow.

rpl: @eviljeff I read this exchange once more and at this point I think I'd vote to keep it simple and only add a `--amo-metadata-file` option for now.

I think that should be enough to cover how the users will be using it (which I'd expect to be either through command line option, or the environment variables that yargs translates automatically into cli options, added only for the requests that are creating a new addon or specifically meant to update the metadata).

We can always consider adding another way of specifying the metadata as a potential separate low priority followup, something worth looking into if issues or enhancement requests filed from extension developers using the new options will convince us that aren't covered in a good enough way by the initial `--amo-metadata-file` we will be supporting.

how that sounds?

rpl (approved):
@eviljeff lgtm thanks!

eviljeff:
Note: whichever of https://github.com/mozilla/web-ext/pull/2511 and this pr you merge second will likely need a rebase for conflicts.

PR #2498, opened 2022-08-31
make api auth more pluggable in submit-addon.js::Client

### LINKED ISSUE #2493: Implement support for AMO session cookie auth
I'd like to be able to use `web-ext` as a node dependency in an website that could re-use AMO session cookie auth that's authenticated via Firefox Accounts (i.e. a website on or under addons.mozilla.org), rather than the JWT token auth we provide for 3rd party API use.  All we need for this is a slightly different auth header and the ability to pass the cookie value rather than `apiKey` and `apiSecret`.

eviljeff:
closes #2493 - doesn't add session token auth (as the issue is currently describing) but instead abstracts the API auth into a separate class, so a consumer of this package could provide their own easily.

### DIFF
--- a/tests/unit/test-util/test.submit-addon.js
+++ b/tests/unit/test-util/test.submit-addon.js
@@ -8,7 +8,8 @@ import { afterEach, before, beforeEach, describe, it } from 'mocha';
 import * as sinon from 'sinon';
 import { File, FormData, Response } from 'node-fetch';
 
-import Client, { signAddon } from '../../../src/util/submit-addon.js';
+import Client, { JwtApiAuth, signAddon }
+  from '../../../src/util/submit-addon.js';
 import { withTempDir } from '../../../src/util/temp-dir.js';
 
 class JSONResponse extends Response {
@@ -80,6 +81,7 @@ describe('util.submit-addon', () => {
       const apiHost = 'https://foo.host';
       const downloadDir = '/foo';
       const clientSpy = sinon.spy(Client);
+      const apiAuthSpy = sinon.spy(JwtApiAuth);
 
       await signAddon({
         ...signAddonDefaults,
@@ -88,13 +90,20 @@ describe('util.submit-addon', () => {
         apiHost,
         downloadDir,
         SubmitClient: clientSpy,
+        ApiAuthClass: apiAuthSpy,
       });
 
+      sinon.assert.calledOnce(apiAuthSpy);
+      assert.deepEqual(
+        apiAuthSpy.firstCall.args[0], {
+          apiKey,
+          apiSecret,
+        }
+      );
       sinon.assert.calledOnce(clientSpy);
       assert.deepEqual(
         clientSpy.firstCall.args[0], {
-          apiKey,
-          apiSecret,
+          apiAuth: {},
           apiHost,
           validationCheckTimeout: signAddonDefaults.timeout,
           approvalCheckTimeout: signAddonDefaults.timeout,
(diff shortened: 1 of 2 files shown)


tests/unit/test-util/test.submit-addon.js
```diff
@@ -80,6 +81,7 @@ describe('util.submit-addon', () => {
       const apiHost = 'https://foo.host';
       const downloadDir = '/foo';
       const clientSpy = sinon.spy(Client);
+      const apiAuthSpy = sinon.spy(JwtApiAuth);
```
rpl, line 84: While we are here, let's also assert explicitly that the expected authorization header ends up into the actual request, e.g. something like the following (diff on top of this version of the pull request) may be enough for getting some explicit covered about that:

```diff
diff --git a/tests/unit/test-util/test.submit-addon.js b/tests/unit/test-util/test.submit-addon.js
index fa17b03..5e99197 100644
--- a/tests/unit/test-util/test.submit-addon.js
+++ b/tests/unit/test-util/test.submit-addon.js
@@ -199,19 +199,30 @@ describe('util.submit-addon', () => {
       status: 'unreviewed',
     };
 
+    const getAuthHeaderSpy = sinon.spy(apiAuth, 'getAuthHeader');
+
+    afterEach(() => {
+      getAuthHeaderSpy.resetHistory();
+    });
+
     describe('doUploadSubmit', () => {
       it('submits the xpi', async () => {
         const client = new Client(clientDefaults);
         sinon.stub(client, 'fileFromSync')
           .returns(new File([], 'foo.xpi'));
+
+        const nodeFetchStub = sinon.stub(client, 'nodeFetch');
         mockNodeFetch(
-          sinon.stub(client, 'nodeFetch'),
+          nodeFetchStub,
           `${apiHostPath}/addons/upload/`,
           'POST',
           [
             {
               body: sampleUploadDetail,
               status: 200,
+              headers: {
+                Authorization: 'test123',
+              },
             },
           ]
         );
@@ -224,6 +235,20 @@ describe('util.submit-addon', () => {
         const returnUuid = await client.doUploadSubmit(xpiPath, channel);
         assert.equal(sampleUploadDetail.uuid, returnUuid);
         sinon.assert.calledWith(waitStub, sampleUploadDetail.uuid);
+
+        // Verify the jwt Authorization header are included in the response.
+        sinon.assert.calledOnce(getAuthHeaderSpy);
+        const authHeaderValue = await getAuthHeaderSpy.getCall(0).returnValue;
+        assert.match(authHeaderValue, /^JWT .*/);
+        sinon.assert.calledOnceWithMatch(
+          nodeFetchStub,
+          sinon.match.instanceOf(URL),
+          sinon.match({
+            headers: {
+              Authorization: authHeaderValue,
+            },
+          })
+        );
       });
     });
```

rpl (changes requested):
@eviljeff lgtm, besides allowing to open up the use case you have been looking into, I like it because it feels like a nice small cleanup of the internal implementation.

Follows just a small change requested on the test side to add some minimal explicit coverage on the authorization header itself (mainly the amount that would make sure one test fails if the header isn't included or it is empty).

The other comment is just a small nit (really not a bit deal, kind of optional in the end).

tests/unit/test-util/test.submit-addon.js
eviljeff: re:
```diff
+              headers: {
+                Authorization: 'test123',
+              },
```
AMO don't send the Authorization header in responses.

rpl: eh, yeah sorry about that, that was a leftover from an initial attempt 
(I was looking into how to pass to the helper the request headers to expect and initially I looked into passing it through the array of the responses for the mocked calls, then I thought that was not looking great and opted for the approach in the rest of the diff).

Thanks for spotting that left over and omit it from the changes applied to this PR.

rpl (approved):
lgtm, thanks!

PR #2506, opened 2022-09-09
chore: Add Prettier

willdurand:
We shouldn't have to think or argue about codestyle so let's use Prettier like in most other projects we maintain.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,6 +195,7 @@ jobs:
           condition:
             equal: [*nodejs_current, << parameters.nodejs >>]
           steps:
+            - run: npm run prettier-ci
             - run:
                 name: run linting check and unit tests with coverage
                 command: npm run test-coverage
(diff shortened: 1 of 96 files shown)


.circleci/config.yml
```diff
@@ -195,6 +195,7 @@ jobs:
           condition:
             equal: [*nodejs_current, << parameters.nodejs >>]
           steps:
+            - run: npm run prettier-ci
```
willdurand, line 198: This works as shown here 🤣 https://app.circleci.com/pipelines/github/mozilla/web-ext/1653/workflows/f4c65822-c0c2-4d21-8a50-9a7600c3648c/jobs/6694

willdurand:
I just rebased this PR.

willdurand:
rebased!

rpl (approved):
lgtm thanks Will, I'm going to merge this in a minute.

I think it may be worth adding a mention to the prettier npm script to run before submitting a PR in CONTRIBUTING.md, but let's not do that in this PR, which has already updated the CONTRIBUTING .md file due to prettier reformatting it.

As a side note: I updated the PR title from feat to chore, given that this is an internal rework of the js code and doesn't need to be called out as a feature in the next release changelog.

PR #2507, opened 2022-09-09
Rename apiHost to amoBaseUrl

### LINKED ISSUE #2504: Replace submission-api --api-host cli option with --amo-base-url
As a followup to the original discussion in #2458 about the --api-host option.

@willdurand @Rob--W and I have discussed about it one more time during today's issue triage and the following is a summary of the changes we agreed on:

- rename the option from --api-host to --amo-base-url
- the default value for --amo-base-url will be set to "https://addons.mozilla.org/api/v5"

This option is meant to mainly be used by QA to run QA verifications on the staging or dev AMO servers, and it is reasonable to most extension developers will not need to use it or to customize the url, and so we agreed that we may not need to apply any special check or restrictions to the value got from it (well besides being a valid url that doesn't throw when passed to `new URL(...)`).

eviljeff:
fixes #2504 + makes the base url constractor arg a `URL` in `Client` to be consistent with internal use in the class.

### DIFF
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -54,7 +54,7 @@ export class JwtApiAuth {
 
 type ClientConstructorParams = {|
   apiAuth: ApiAuth,
-  apiHost: string,
+  baseUrl: URL,
   validationCheckInterval?: number,
   validationCheckTimeout?: number,
   approvalCheckInterval?: number,
@@ -73,7 +73,7 @@ export default class Client {
 
   constructor({
     apiAuth,
-    apiHost,
+    baseUrl,
     validationCheckInterval = 1000,
     validationCheckTimeout = 300000, // 5 minutes.
     approvalCheckInterval = 1000,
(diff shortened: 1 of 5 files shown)


rpl (approved):
lgtm 

Thanks @eviljeff!

src/util/submit-addon.js
```diff
@@ -354,9 +354,16 @@ export async function signAddon({
     throw new Error(`error with ${xpiPath}: ${statError}`);
   }
 
+  let baseUrl;
+  try {
+    baseUrl = new URL(amoBaseUrl);
+  } catch (err) {
+    throw new Error(`Invalid amoBaseUrl: ${amoBaseUrl}`);
```
willdurand, line 361:
```suggestion
    throw new Error(`Invalid AMO API base URL: ${amoBaseUrl}`);
```

willdurand, line 361:
```suggestion
    throw new Error(`Invalid AMO API base URL: ${amoBaseUrl}`);
```

willdurand: Also, I don't see a test case for this.

willdurand (approved):
r+wc

src/util/submit-addon.js
eviljeff: There wasn't one before 😉 but 👍

tests/unit/test-util/test.submit-addon.js
```diff
@@ -150,6 +150,15 @@ describe('util.submit-addon', () => {
         'Error: ENOENT: no such file or directory'
       );
     });
+
+    it('throws error if xpiPath is invalid', async () => {
```
willdurand, line 154:
```suggestion
    it('throws error if amoBaseUrl is invalid', async () => {
```

Issue #2490, opened 2022-08-26
Make changes to handle the addon id existing from the previous submission


eviljeff:
Background:
AMO's submission API has three endpoints for creating a version from an xpi `Upload`:
1. POST to `/addons/addon/` to create a new add-on listing with the new version
2. POST to `/addons/addon/<amo-id|addon-id|slug>/versions/` to create a new version under an existing add-on listing. There isn't support for this endpoint in `submit-addon.js` code currently.
3. PUT to `/addons/addon/<addon-id>`/ to either create a new add-on listing with the new version if the `addon-id` doesn't exist already; or to update an existing add-on listing and create a new version under that listing if the `addon-id` does already exist.  This endpoint also enforces that the addon-id is defined in the manifest (so there is only one possible place to define the `addon-id`).

Current web-ext behaviour is to use endpoint 3. when an `addon-id` is defined in the manifest, provided via command-line arg, or exists in `.web-extension-id`; and to use endpoint 1. otherwise.  After a successful add-on submission (including approval and download of the xpi) the `addon-id` from the API response is used to write the `.web-extension-id` file.

However, this behaviour is now flawed with the submission api if the developer didn't define the `addon-id` in the manifest - after a new add-on is created and the `addon-id` written to `.web-extension-id`, the next attempt to use `web-ext sign` will raise an error because there is an `addon-id` provided but it's not in the manifest file.

There are two options to fix this, as I see them:
- add the `addon-id` to the manifest if not already there (in the `build` function? in `getValidatedManifest`?) if it's either provided via command line arg or in `.web-extension-id`.  Requiring the `addon-id` in the manifest is a planned change anyway so `web-ext` will likely need to add support at some point regardless.
- implement support for endpoint 2. and use that endpoint when there is `.web-extension-id` file instead.  An `addon-id` in the manifest would continue to use endpoint 3.  Note this still won't cover the scenario when a developer only provides the `addon-id` via the command line.

eviljeff:
@rpl and @willdurand for opinions.

eviljeff:
> * add the `addon-id` to the manifest if not already there (in the `build` function? in `getValidatedManifest`?) if it's either provided via command line arg or in `.web-extension-id`.  Requiring the `addon-id` in the manifest is a planned change anyway so `web-ext` will likely need to add support at some point regardless.

fwiw, I see this as the "correct" fix.

willdurand:
[...]

PR #2511, opened 2022-09-12
write id to file upon successful submission

### LINKED ISSUE #2491: save the addon id earlier, after a successfull submission?
Right now the `addon-id` is saved in `.web-extension.id` _after_ the xpi has been downloaded.  If the approval check times out, or the user cancels the wait before the timeout, the `addon-id` isn't saved. If the user then decides to submit a new version they'll be creating a new add-on, possibly without realising(!)

This isn't an unlikely scenario - in the best case there are a few minutes of waiting after a successful submission, after it's been auto-signed by the cron job, and although the user may wait they may also conclude something is amiss before the approval is complete.  But in many cases (i.e. the first listed version for an add-on) the signing is purposefully held back for up to 24 hours to allow it to be content reviewed; and in some cases the add-on will be chosen for manual review and signing could be delayed further depending on when it's submitted, review team capacity, etc.

With the submission api we know the `addon-id` immediately after a successfully request because it's returned in the response, so we could write the id immediately once we know an add-on will be created.

eviljeff:
fixes #2491 - and could address #2490 (at least in part) by clearly stating the addon id after it's available from AMO.

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -194,7 +194,7 @@ export default function sign(
           }
           result = { id: newId, downloadedFiles };
           // All information about the downloaded files would have already been
-          // logged by signAddon(). submitAddon() calls saveIdToSourceDir itself.
+          // logged by signAddon(). submitAddon() calls saveIdToFile itself.
           await saveIdToFile(savedIdPath, newId);
           log.info(`Extension ID: ${newId}`);
           log.info('SUCCESS');
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -10,7 +10,10 @@ import {prepareArtifactsDir} from '../util/artifacts.js';
 import {createLogger} from '../util/logger.js';
 import getValidatedManifest, {getManifestId} from '../util/manifest.js';
 import type {ExtensionManifest} from '../util/manifest.js';
-import { signAddon as defaultSubmitAddonSigner } from '../util/submit-addon.js';
+import {
+  signAddon as defaultSubmitAddonSigner,
+  saveIdToFile,
+} from '../util/submit-addon.js';
 import type { SignResult } from '../util/submit-addon.js';
 import {withTempDir} from '../util/temp-dir.js';
 
@@ -81,6 +84,7 @@ export default function sign(
       await prepareArtifactsDir(artifactsDir);
 
       let manifestData;
+      const savedIdPath = path.join(sourceDir, extensionIdFile);
 
       if (preValidatedManifest) {
         manifestData = preValidatedManifest;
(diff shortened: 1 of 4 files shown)


src/cmd/sign.js
```diff
@@ -155,8 +155,13 @@ export default function sign(
...
       try {
         if (useSubmissionApi) {
-          // $FlowIgnore: we verify 'channel' is set above
-          result = await submitAddon({...signSubmitArgs, amoBaseUrl, channel});
+          result = await submitAddon({
+            ...signSubmitArgs,
+            amoBaseUrl,
+            // $FlowIgnore: we verify 'channel' is set above
+            channel,
+            saveIdFunc: (genId: string) => saveIdToSourceDir(sourceDir, genId),
```
willdurand, line 163: Is this _only_ needed because `saveIdToSourceDir()` is declared in this file?

Instead of passing a function as a parameter, I would rather call `saveIdToSourceDir()` in the `submitAddon()` module directly personally.

src/util/submit-addon.js
```diff
@@ -293,6 +294,11 @@ export default class Client {
       [versionObject]: {id: newVersionId},
     } = await this.doNewAddonSubmit(uploadUuid, metaDataJson);
 
+    await saveIdFunc(addonId);
+    log.info(`Generated extension ID: ${addonId}.`);
+    log.info('You must add the following to your manifest:');
+    log.info(`"browser_specific_settings": {"gecko": "${addonId}"}`);
```
willdurand, line 300: If we move `saveIdToSourceDir()` somewhere else (could be in this file even), then those logs could also be moved to `saveIdToSourceDir()`. If there is a chance that this message does not apply, we could update the content a bit (e.g. from "You must add [...]" to "Make sure you have [...] in your manifest:").

src/cmd/sign.js
eviljeff: It also means we don't need to pass down the source dir. Having the function as the parameter seemed to be a cleaner change than passing down the source dir just for that, and moving the `saveIdToSourceDir` function elsewhere. But I'm also not clear where you draw the line between functionality that's supposed to be in `sign.js` v.s. `submit-addon.js`

src/util/submit-addon.js
eviljeff: The `id` is never needed for the `signAddon` route, while AMO doesn't allow MV3 submissions. (If we're specifically adding support for MV3 I'd prefer that in a separate PR).

We only call the function for `id`-less submissions for `submit-addon`, so it's always needed.

willdurand (approved):

eviljeff:
had to squash and rebase to address the conflicts from #2489

src/cmd/sign.js
```diff
@@ -185,28 +193,26 @@ export default function sign(
             throw new Error('The extension could not be signed');
           }
           result = { id: newId, downloadedFiles };
+          // All information about the downloaded files would have already been
+          // logged by signAddon(). submitAddon() calls saveIdToSourceDir itself.
```
rpl, line 197: nit, `saveIdToSourceDir` => `saveIdToFile` in this inline comment?

rpl (approved):
lgtm, just a small nit (also not a super big deal, it is just a function name mentioned in an inline comment that may need to be updated and so I'm also approving this version in the meantime).

PR #2528, opened 2022-10-18
add apiProxy support for submission API

### LINKED ISSUE #2472: support proxy with addon submission api
follow-up for #2451 - it's not going to support proxies at first, but to be feature comparable with sign-addon that's used for the signing api, it should support proxies too.

eviljeff:
fixes #2472


eviljeff:
(I can't verify this is working with local testing, so hold off any reviews until it's clear it's functional or a fix comes along)

willdurand (approved):
Thanks.

eviljeff:
> aww conflict :(

it'll be that small change where all the typing was removed 😬

PR #2540, opened 2022-10-27
feat: send UserAgentHeader with signing requests

### LINKED ISSUE #2538: Send web-ext version in `User-Agent` header when calling the AMO API
See: https://github.com/mozilla/addons-server/issues/18803#issuecomment-1034141273

eviljeff:
fixes #2538

### DIFF
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -66,7 +66,7 @@ type ClientConstructorParams = {|
   approvalCheckInterval?: number,
   approvalCheckTimeout?: number,
   downloadDir?: string,
-  userAgentString?: string,
+  userAgentString: string,
 |};
 
 export default class Client {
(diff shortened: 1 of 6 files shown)


src/util/submit-addon.js
```diff
@@ -66,6 +66,7 @@ type ClientConstructorParams = {|
   approvalCheckInterval?: number,
   approvalCheckTimeout?: number,
   downloadDir?: string,
+  userAgentString?: string,
```
willdurand, line 69: I think we should make this argument mandatory, not optional and without a default value.

eviljeff: it has a default value (`'webext'`) defined in the constructor.  How do you define defaults in the `type` declaration?

willdurand: I think you don't need `?` since it will always have a value but what I mean is making it required and without default value.

eviljeff: it's an internal function where we always provide a value so it doesn't make much difference (the default was primarily to make changing the tests less annoying) but sure, can do.

willdurand (approved):
That looks reasonable, thanks.

rpl (approved):

eviljeff:
Thanks both. Please merge when it's convenient/appropriate.

Issue #2549, opened 2022-11-10
Enable MV3 by default


willdurand:
When AMO starts accepting MV3 extensions, we should update web-ext to enable the MV3 features by default. Currently, these features are behind a CLI flag: `--firefox-preview=mv3`.

PR #2557, opened 2022-11-18
feat: enable MV3 by default in `web-ext lint` command

willdurand:
Refs https://github.com/mozilla/web-ext/issues/2549

---

I'll update `web-ext run` in a different patch.


rpl (approved):
@willdurand lgtm thanks, let's ignore codecov, the only line we are not covering is the newly added warning and that isn't that big of a deal.

PR #2565, opened 2022-11-27
feat: add chromium debugger port option

### LINKED ISSUE #2284: Debugging over a remote port occasionally fails
A bug report.

1. I run the following command:
    ```shell
    web-ext run --no-config-discovery --start-url https://www.youtube.com/watch?v=dQw4w9WgXcQ --arg="--remote-debugging-port=9229" --source-dir dist --verbose --target chromium
    ```
1. Google Chrome Canary opens up.
1. After a moment, `web-ext` throws an error and exits (Chrome stays open):
    ```
    [program.js][error]
    Error: connect ECONNREFUSED 127.0.0.1:3570
        at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1148:16)

    [program.js][error] Error code: ECONNREFUSED

    [program.js][debug] Command executed: run
    ERROR  Command failed with exit code 1
    ```
Being able to launch Chrome such that it's listening to debugging, without `web-ext` crashing.

* **Google Chrome version**: Canary 94.0.4589.2
* **Your OS and version**: Windows 10 x64 21H1
    ```powershell
    node -v; npm -v; pnpm -v; web-ext --version
    ```
    ```
    v14.17.3
    7.15.0
    6.10.3
    6.2.0
    ````
    ↑ I added PNPM because that's the package manager I'm using.

### Note
`web-ext` doesn't *always* crash. It's more like a 50:50.

fuji44:
Fixes #2284

This PR adds an option to specify the debugger port number for chromium-based browsers.
The motivation is to make debugger attachments declarative and manageable.

Note: It seems that the unit test is failing in an area unrelated to my modification.

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -28,6 +28,7 @@ type ChromiumSpecificRunnerParams = {|
   chromiumBinary?: string,
   chromiumProfile?: string,
   chromiumLaunch?: typeof defaultChromiumLaunch,
+  chromiumDebuggerPort?: number,
 |};
 
 export type ChromiumExtensionRunnerParams = {|
@@ -232,6 +233,7 @@ export class ChromiumExtensionRunner {
       userDataDir,
       // Ignore default flags to keep the extension enabled.
       ignoreDefaultFlags: true,
+      port: this.params.chromiumDebuggerPort,
     });
 
     this.chromiumInstance.process.once('close', () => {
--- a/src/program.js
+++ b/src/program.js
@@ -674,6 +674,11 @@ Example: $0 --help run.
         demandOption: false,
         type: 'string',
       },
+      'chromium-debugger-port': {
+        describe: 'Listen port number of the Chromium debugger',
+        demandOption: false,
+        type: 'number',
+      },
       'profile-create-if-missing': {
         describe: 'Create the profile directory if it does not already exist',
         demandOption: false,
--- a/src/program.js
+++ b/src/program.js
@@ -675,7 +675,9 @@ Example: $0 --help run.
         type: 'string',
       },
       'chromium-debugger-port': {
-        describe: 'Listen port number of the Chromium debugger',
+        describe:
+          'An unused port number, for use by the Chromium debugger. ' +
+          'If not specified, a random port will be used.',
         demandOption: false,
         type: 'number',
       },
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -329,6 +329,26 @@ describe('run', () => {
     );
   });
 
+  it('provides a chromiumDebuggerPort option to the Chromium runner', async () => {
+    const fakeChromiumDebuggerPort = 59222;
+    const cmd = await prepareRun();
+    await cmd.run({
+      target: ['chromium'],
+      chromiumDebuggerPort: fakeChromiumDebuggerPort,
+    });
+
+    sinon.assert.calledWithMatch(chromiumRunnerStub, {
+      chromiumDebuggerPort: sinon.match.number,
+    });
+
+    const { chromiumDebuggerPort } = chromiumRunnerStub.firstCall.args[0];
+    assert.equal(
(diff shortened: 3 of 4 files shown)


saadbahir:
@fuji44 do you know how we could merge this one?

saadbahir:
@Rob--W can we merge this branch?

fuji44:
@saadbahir 
I don't have permission to merge. waiting for review

src/extension-runners/chromium.js
```diff
@@ -232,6 +233,7 @@ export class ChromiumExtensionRunner {
       userDataDir,
       // Ignore default flags to keep the extension enabled.
       ignoreDefaultFlags: true,
+      port: this.params.chromiumDebuggerPort,
```
Rob--W, line 236: Passing this flag without validation is dangerous, because it may break the functionality of `web-ext`.

`web-ext` has to launch a new Chrome browser instance, because the extension and auxiliary extension are installed via a command-line flag, `--load-extension` to be specific.

When `chrome-launcher` receives a given port, it tries to connect to that port, and if it exists, it will not start a new Chrome browser, but connect to an existing one instead: https://github.com/GoogleChrome/chrome-launcher/blob/v0.15.1/src/chrome-launcher.ts#L248-L264

IF we want to support taking a fixed flag as input, then we should validate that the port is unused, and throw an explicit error if it is used.

src/program.js
```diff
@@ -674,6 +674,11 @@ Example: $0 --help run.
         demandOption: false,
         type: 'string',
       },
+      'chromium-debugger-port': {
+        describe: 'Listen port number of the Chromium debugger',
```
Rob--W, line 678:
```suggestion
        describe:
          'An unused port number, for use by the Chromium debugger. ' +
          'If not specified, a random port will be used.',
```

Suggested edit to clarify the meaning, constraints and default behavior.

tests/unit/test-cmd/test.run.js
```diff
@@ -329,6 +329,26 @@ describe('run', () => {
     );
   });
 
+  it('provides a chromiumDebuggerPort option to the Chromium runner', async () => {
```
Rob--W, line 332: This test is not enough for test coverage. It merely verifies that the CLI parameter is passed to the extension runner.

You need to add more tests, to verify that the port is received correctly in https://github.com/mozilla/web-ext/blob/005bed86dff796cca48281b1928f7109b0972c80/tests/unit/test-extension-runners/test.chromium.js . And the check requested in the other comment (i.e. validity of port) should be tested too.

Rob--W (changes requested):

src/extension-runners/chromium.js
fuji44: I agree with the basic idea that not validating external parameters is dangerous, but I don't think it applies in this case. web-ext is just passing a port number, and it is essentially the browser's responsibility to validate the port number. Well, chrome doesn't seem to check it.

Even if you implement the verification code, wouldn't it be enough to verify that the specified number is within the TCP port range?

Verifying whether a port is in use may be useful, but it is a "nice to have" feature.
Considering that the user of web-ext is a developer, it is common sense not to use duplicate ports, and even if duplicate ports are used by mistake, they will immediately consider the possibility.

So the bottom line is that for most web developers, the ability to verify that a port is in use is an unnecessary feature. This is undesirable because it adds complexity to the code and reduces maintainability.

Rob--W: I explained why the port has to be unused in my previous comment.

Summarized, web-ext would not work and be unreliable. The `web-ext` program expects the chrome-launcher library to launch the browser with the extension loaded, but if the port is in use, it won't launch Chrome, and therefore the core feature of `web-ext run` will be stuck waiting for the browser+extension to start. Explicitly failing with a useful message is more desirable than getting stuck without further information.

fuji44: When I get time to work on it, I will check the stacking behavior.

fuji44:
It is not good to leave it there forever, so I am going to close it and work on it again when I have more free time.

PR #2618, opened 2023-01-19
Use Node 18 in Circle CI

### LINKED ISSUE #2617: Use Node 18 in Circle CI
We should use Node 18 instead of Node 17 here: https://github.com/mozilla/web-ext/blob/6356fa9b4cbfa7b2b87e8a03d3e2e3b42988b531/.circleci/config.yml#L30

willdurand:
Fixes #2617


willdurand:
Looks like it's working :)

rpl (approved):

PR #2621, opened 2023-01-20
add trailing slash to amo-base-url & enforce within submit-addon Client

### LINKED ISSUE #2579: --use-submission-api results in HTML response with "Oops! We can’t find that page"

Bug.

So far I've been successfully using `sign` and publishing my addon without `--use-submission-api`.
I've decided to give this flag a go:

```
'npm', 'run', 'web-ext', '--', 'sign', 
'--use-submission-api', 
'--channel', 'unlisted',
'--source-dir', '.../code/grasp/extension/dist/firefox', 
'--artifacts-dir', '.../grasp/extension/dist/artifacts/firefox', 
'--api-key', 'KEY', '--api-secret', 'SECRET'
```

My extension's manifest version is 2 (not sure whether this matters). I also had to add extension ID to the manifest, otherwise web-ext complained.
```
    "gecko": {
      "id": "{37e42980-a7c9-473c-96d5-13f18e0efc74}"
    }
  },
```

However this resulted in a cryptic error. I nailed it down to this line
https://github.com/mozilla/web-ext/blob/master/src/util/submit-addon.js#L246
Somehow instead of JSON, I got a html page with the following content
```
Oops! We can’t find that page

If you’ve followed a link from another site for an extension or theme, that item is no longer available. This could be because:

    The developer removed it. Developers commonly do this because they no longer support the extension or theme, or have replaced it.
    Mozilla removed it. This can happen when issues are found during the review of the extension or theme, or the extension or theme has been abusing the terms and conditions for addons.mozilla.org. The developer has the opportunity to resolve the issues and make the add-on available again.
```

I added a bit more debug logging to web-ext and this is the fetch call that fails:

```
URL URL {
  href: 'https://addons.mozilla.org/api/addons/upload/',
  origin: 'https://addons.mozilla.org',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'addons.mozilla.org',
  hostname: 'addons.mozilla.org',
  port: '',
  pathname: '/api/addons/upload/',
  search: '',
  searchParams: URLSearchParams {},
  hash: '',
  cannotBeBase: false,
  special: true,
  [Symbol(context)]: URLContext {
    flags: 400,
    scheme: 'https:',
    username: '',
    password: '',
    host: 'addons.mozilla.org',
    port: null,
    path: [ 'api', 'addons', 'upload', '', [length]: 4 ],
    query: null,
    fragment: null
  }
}
METHOD 'POST'
BODY [
  [ 'channel', 'unlisted', [length]: 2 ],
  [
    'upload',
    File {
      [name]: [Getter],
      [lastModified]: [Getter],
      [Symbol(Symbol.toStringTag)]: [Getter],
      size: [Getter],
      type: [Getter]
    },
    [length]: 2
  ],
  [length]: 2
]
HEADERS {
  Authorization: 'JWT <REMOVED>',
  Accept: 'application/json'
}
```

Anything I'm doing wrong?

Extension is uploaded to addons.mozilla.org

### Version information
```
$ node --version && npm --version && npm run web-ext -- --version
v18.12.1
8.19.2
7.4.0
```

eviljeff:
fixes #2579

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -568,7 +568,7 @@ Example: $0 --help run.
         'amo-base-url': {
           describe:
             'Signing API URL prefix - only used with `use-submission-api`',
-          default: 'https://addons.mozilla.org/api/v5',
+          default: 'https://addons.mozilla.org/api/v5/',
           demandOption: true,
           type: 'string',
         },
--- a/src/program.js
+++ b/src/program.js
@@ -48,6 +48,8 @@ type ExecuteOptions = {
   globalEnv?: string | void,
 };
 
+export const AMO_BASE_URL = 'https://addons.mozilla.org/api/v5/';
+
 /*
  * The command line program.
  */
@@ -568,7 +570,7 @@ Example: $0 --help run.
         'amo-base-url': {
           describe:
             'Signing API URL prefix - only used with `use-submission-api`',
-          default: 'https://addons.mozilla.org/api/v5/',
+          default: AMO_BASE_URL,
           demandOption: true,
           type: 'string',
         },
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -90,6 +90,9 @@ export default class Client {
     userAgentString,
   }: ClientConstructorParams) {
     this.apiAuth = apiAuth;
+    if (!baseUrl.href.endsWith('/')) {
+      baseUrl = new URL(`${baseUrl.href}/`);
+    }
     this.apiUrl = new URL('addons/', baseUrl);
     this.validationCheckInterval = validationCheckInterval;
     this.validationCheckTimeout = validationCheckTimeout;
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -90,8 +90,9 @@ export default class Client {
     userAgentString,
   }: ClientConstructorParams) {
     this.apiAuth = apiAuth;
-    if (!baseUrl.href.endsWith('/')) {
-      baseUrl = new URL(`${baseUrl.href}/`);
+    if (!baseUrl.pathname.endsWith('/')) {
+      baseUrl = new URL(baseUrl.href);
+      baseUrl.pathname += '/';
     }
     this.apiUrl = new URL('addons/', baseUrl);
     this.validationCheckInterval = validationCheckInterval;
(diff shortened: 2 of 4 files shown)


src/util/submit-addon.js
```diff
@@ -90,6 +90,9 @@ export default class Client {
     userAgentString,
   }: ClientConstructorParams) {
     this.apiAuth = apiAuth;
+    if (!baseUrl.href.endsWith('/')) {
```
Rob--W, line 93: If the URL somehow contains a `?` or `#` but not a `/`, then this would not work as desired.

This would fix the issue:

```js
if (!baseUrl.pathname.endsWith('/')) {`
  baseUrl = new URL(baseUrl);
  baseUrl.pathname += "/";
}
```

P.S. The query and ref will be stripped by `new URL('addons/', baseUrl)` anyway, so an alternative is to assert that `baseUrl.search` and `baseUrl.hash` are both empty.

src/program.js
```diff
@@ -568,7 +568,7 @@ Example: $0 --help run.
         'amo-base-url': {
           describe:
             'Signing API URL prefix - only used with `use-submission-api`',
-          default: 'https://addons.mozilla.org/api/v5',
+          default: 'https://addons.mozilla.org/api/v5/',
```
willdurand, line 571: Did we have coverage for this URL/value? I see we're updating tests below but I don't get why nothing failed before. Maybe the default value in tests should be coming from this file instead of having yet another value that might or might not resemble this one here.

eviljeff: Nothing failed because the the default value was defined 2 files (and numerous calls/processing) further up from where it was used.

Relying on locally defined defaults in unit tests rather than importing other files seemed to be the normal style of writing tests but it could be changed I guess.

src/util/submit-addon.js
eviljeff: The fix suggestion sounds good.  I'm not sure it's worth the extra effort to assert on `baseUrl.search` and `baseUrl.hash` though - the code is too specialised to the AMO API here, so trying to support another API that has a different url style is out of scope. (My vote is just to ignore it and let them be stripped)

eviljeff: https://github.com/mozilla/web-ext/pull/2621/commits/757f47a646311ab50966e64e49035dbd884734f7 is your fix(ish)

src/program.js
eviljeff: https://github.com/mozilla/web-ext/pull/2621/commits/f8cfb9f8fda5b9940c097c1e64d107202ee53af9

willdurand (approved):

willdurand:
@Rob--W if that looks good to you, please merge :) (so that I'll know that you saw both this message and the updated patch)

Issue #2206, opened 2021-03-18
Wrong activity being started for Fenix


palant:
### Is this a feature request or a bug?

Bug

### What is the current behavior?

The following command fails:

    web-ext run --target=firefox-android --android-device=id

Error message:

> Error: Activity class {org.mozilla.fenix.debug/org.mozilla.fenix.App} does not exist.

This works correctly with an explicit `--firefox-apk-component '.HomeActivity'` parameter.

### What is the expected or desired behavior?

This should launch Fenix using the correct activity class.

### Version information (for bug reports)

* **Firefox version**: 88.0 (built from source)
* **Your OS and version**: Android 11
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
````
> v15.11.0
> 7.6.0
> 6.0.0

palant:
Actually, an explicit `--firefox-apk-component org.mozilla.fenix.debug.App` works as well. This only fails when Fenix is compiled from source, not with the official Nightly.

Rob--W:
@jonalmeida The current logic for determining the "APK component" (activity identifier) is based on the discussion starting from your comment at https://github.com/mozilla/web-ext/pull/1871#issuecomment-605400070 :

> What it's sending: `org.mozilla.reference.browser.debug/org.mozilla.reference.browser.debug.BrowserActivity`
> What it should be sending: `org.mozilla.reference.browser.debug/org.mozilla.reference.browser.BrowserActivity`

The logic was implemented at https://github.com/mozilla/web-ext/pull/1941

In this bug report, it is apparent that `org.mozilla.fenix.debug.App` should be used instead of `org.mozilla.fenix.App`, which is in contradiction with the previously described logic.

I guess that we can avoid prepending a fixed package identifier when `.App` is used, i.e. changing the condition from `if (apkComponent.startsWith('.')) {` to `if (apkComponent !== ".App" && apkComponent.startsWith('.')) {` at https://github.com/mozilla/web-ext/blame/ac207c6fdca5c806949adcaa7c54105ce278f606/src/util/adb.js#L328 

@jonalmeida What do you suggest to resolve this bug? Does my suggestion make sense?

jonalmeida:
Maybe there is some inconsistency within the two mobile apps, there were some variant changes that could have affected this?

Juraj-Masiar:
Just to confirm, **using `web-ext` with Nightly on Android is currently not possible**, correct?

I have these 3 commands for Nightly / Beta / Release:
[...]

PR #2682, opened 2023-03-20
Add support for Fenix debug builds

willdurand:
This PR adds support for fenix debug builds (with a new package
identifier). This is probably more useful to Mozilla engineers but we
never know :)


rpl (approved):
lgtm, thanks Will!

src/util/adb.js
```diff
@@ -336,20 +336,23 @@ export default class ADBUtils {
...
       }
     }
 
-    // if `apkComponent` starts with a '.', then adb will expand
-    // the following to: `${apk}/${apk}.${apkComponent}`
-    const component = `${apk}/${apkComponent}`;
+    // If `apkComponent` starts with a '.', then adb will expand the following
+    // to: `${apk}/${apk}.${apkComponent}`
+    let component = `${apk}`;
+    if (apkComponent) {
```
Rob--W, line 353: What's the reason for this change?

Is it #2206 ?

What is the test case that failed before and works with this part of the change?

Rob--W (commented):
While this patch specifically introduces support for the Debug build of Fenix, it is not a generic solution because it required the explicit addition of `".debug"` to the hard-coded list of app identifiers. E.g. it doesn't work for debug builds of other apps.

The ideal solution would be to get it working without adding `org.mozilla.fenix.debug` to the list.

src/util/adb.js
willdurand: That was just flow being annoying.

willdurand:
> The ideal solution would be to get it working without adding org.mozilla.fenix.debug to the list.

IMO the ideal solution would be to remove most of this overly complex logic and simply add a table with the supported android apps/variants and their activity names.

PR #2684, opened 2023-03-20
chore: fix audit-deps.js script to handle references to vulnerabilities

willdurand:


rpl (approved):
good catch!

lgtm, ty!

PR #2712, opened 2023-04-06
don't re-upload an upload if the xpi hasn't changed

### LINKED ISSUE #2492: Allow a new add-on or a new version submission to continue with an existing upload
With the submission api the validation is effectively split up into two stages - upload validation, including the linter, scanners, etc, and then the addon/version validation, which is mainly (but not exclusively) about requiring certain metadata for listed add-ons.  So it's more than possible to submit an xpi upload which passes validation and returns an upload uuid, which is then rejected with a 400 response because the additional metadata submitted is either incorrect or deficient (one example of many possible: the category name is misspelt).  Currently to fix this you execute the command again - with the same content, which is then built into an xpi; then submitted to AMO; validated, scanned, etc; and a new upload uuid is returned; which you have to wait for - when the POST/PUT to `/addons/addon/` could occur with the existing upload instantaneously.

Ideally the default would be to resume the submission with the same upload if the add-on contents haven't changed.  (Details of how this would work technically tbd)

eviljeff:
fixes #2492 - a hash of the crc32 values from the zip seemed a pretty easy way of checking if the built zip was different than the previously built zip but open to alternatives.


Rob--W:
Is there a reason to thoroughly check the CRC checksums instead of just the version number?

eviljeff:
> Is there a reason to thoroughly check the CRC checksums instead of just the version number?

yes.  The version number may not have changed but the other contents could be different.

Rob--W:
> > Is there a reason to thoroughly check the CRC checksums instead of just the version number?
> 
> yes. The version number may not have changed but the other contents could be different.

Does web-ext create reproducible builds? If yes, would it be acceptable to hash the zip file instead of its individual entries? That would simplify the implementation and avoid 2 additional dependencies.

eviljeff:
> > > Is there a reason to thoroughly check the CRC checksums instead of just the version number?
> > 
> > yes. The version number may not have changed but the other contents could be different.
> 
> Does web-ext create reproducible builds? If yes, would it be acceptable to hash the zip file instead of its individual entries? That would simplify the implementation and avoid 2 additional dependencies.

The datetime stamps of the files in the zip file would be different so the zip file itself would have a different hash each time.

> 2 additional dependencies.

It's only 1 extra really - jszip is a dependency of `zip-dir`

Rob--W:
> > Does web-ext create reproducible builds? If yes, would it be acceptable to hash the zip file instead of its individual entries? That would simplify the implementation and avoid 2 additional dependencies.
> 
> The datetime stamps of the files in the zip file would be different so the zip file itself would have a different hash each time.

That wouldn't be a problem if the zip process is reproducible, i.e. creating the same zip file for the same file input.
Or do you explicitly want to account for a flow where one uses another build process that touches the timestamps of the files?

eviljeff:
> > > Does web-ext create reproducible builds? If yes, would it be acceptable to hash the zip file instead of its individual entries? That would simplify the implementation and avoid 2 additional dependencies.
> > 
> > 
> > The datetime stamps of the files in the zip file would be different so the zip file itself would have a different hash each time.
> 
> That wouldn't be a problem if the zip process is reproducible, i.e. creating the same zip file for the same file input. 

The zip process isn't strictly reproducible - if the hash of the zip file is calculated it is different for each build, even when the hash of the files themselves would stay the same. My assumption is it's the timestamp of the manifest file which is generated each time, though it could be some other metadata that `zip-dir` (actually `jszip` under the hood) is adding.

eviljeff:
> We already have a (transitive) dependency to read XPIs, see: https://github.com/mozilla/addons-scanner-utils/blob/master/src/io/xpi.ts, which uses https://github.com/thejoshwolfe/yauzl under the hood. Could we reuse this library instead?

I guess?  It's swapping one transitive dependency (jszip) for another (addons-scanner-utils) so I don't see there is much difference.  (the extra handling added in xpi.ts isn't really needed as we're reading a zip file with jszip that was previously written with jszip - zip-dir that we use in build.js is just a wrapper around jszip)

willdurand:
> zip-dir that we use in build.js is just a wrapper around jszip

ah, right. I missed that part. I guess that's fine as it is now, then

tests/unit/test-util/test.submit-addon.js
```diff
@@ -257,6 +265,133 @@ describe('util.submit-addon', () => {
...
+        sinon.assert.calledWith(doUploadStub, xpiPath, newChannel);
+        sinon.assert.calledWith(saveUploadUuidStub, uuidFilePath, {
+          uploadUuid: newUuid,
+          channel: newChannel,
+          xpiCrcHash,
+        });
+      });
+    });
+
+    describe('hashXpiCrcs', () => {
```
Rob--W, line 370: Need more test cases, where we don't necessarily care about the output (i.e. CRC32), but that the hashes are different. At least the following (but depending on the implementation of hashXpiCrcs you may need more):
- zip with directory != zip without that directory.
- zip with directory != zip with empty file.
- zip with file A == zip with file a (same content, file name differs)
- zip with file a b == zip with b a (file order differs).

Rob--W (changes requested):
The hashing mechanism has multiple flaws, see the comments there. The other feedback is minor.

Questions:
- The method uses one file to track previous uploads. It explicitly supports multiple channels, but if one uploads to one channel and then to another, and then back to the first one, then another file is uploaded. Is this something you'd be concerned about?
- TOCTOU race: It is possible for the UUID to be out-of-sync with the file, when the xpi file changes between starting the check and uploading. Is this a concern?
- What are the supported uses of this submit-addon functionality? The implementation relies on the assumption that only the most recent upload is relevant, which seems plausible, but I don't have enough information to verify whether the assumption holds. For example, if the method can be used in parallel by as library to submit different XPIs, then the "don't reupload" logic wouldn't work at all (unless keyed by something more specific, such as the addon ID and/or XPI path).

eviljeff:
> Questions:
> 
>     * The method uses one file to track previous uploads. It explicitly supports multiple channels, but if one uploads to one channel and then to another, and then back to the first one, then another file is uploaded. Is this something you'd be concerned about?

It'd be an extra unnecessary upload but I'm not concerned about that edge case.

>     * TOCTOU race: It is possible for the UUID to be out-of-sync with the file, when the xpi file changes between starting the check and uploading. Is this a concern?

It's a small window so afaik it'd only be a concern if web-ext was attempted to be used in parallel - which seems like would have other problems like the generated add-on ID file being incorrect/overwritten.  Suggestions welcome as for how to workaround it though.
 
>     * What are the supported uses of this submit-addon functionality? The implementation relies on the assumption that only the most recent upload is relevant, which seems plausible, but I don't have enough information to verify whether the assumption holds. For example, if the method can be used in parallel by as library to submit different XPIs, then the "don't reupload" logic wouldn't work at all (unless keyed by something more specific, such as the addon ID and/or XPI path).

I think if someone was trying to use `submit-addon.js` as a library they'd be expected to rely on calling the `Client` methods directly rather than the helper `signAddon` function.  Then `const uploadUuid = await client.getPreviousUuidOrUploadXpi` could be swapped out for `uploadUuid = await client.doUploadSubmit`

tests/unit/test-util/test.submit-addon.js
eviljeff: > zip with file a b == zip with b a (file order differs).

@Rob--W this case should return a _different_ hash, or the same?  (AMO rewrites the zip file server-side so the file order shouldn't be relied upon by the extension.)

Rob--W: Same hash.

src/util/submit-addon.js
```diff
@@ -308,15 +315,60 @@ export default class Client {
...
 
-  async postNewAddon(
+  async hashXpiCrcs(
+    filePath: string,
+    asyncFsReadFile: typeof defaultAsyncFsReadFile = defaultAsyncFsReadFile
+  ): Promise<string> {
+    const zip = await JSZip.loadAsync(asyncFsReadFile(filePath));
+    const hash = createHash('sha256');
+    zip.forEach((relativePath, entry) =>
+      hash.update(entry._data.crc32.toString())
```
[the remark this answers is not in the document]
eviljeff: I believe I've addressed the concerns in https://github.com/mozilla/web-ext/pull/2712/commits/17a52764435fb0078d45b742e71042f05a697ad0

src/util/submit-addon.js
```diff
@@ -308,15 +313,80 @@ export default class Client {
...
+    filePath: string,
+    asyncFsReadFile: typeof defaultAsyncFsReadFile = defaultAsyncFsReadFile
+  ): Promise<string> {
+    const zip = await JSZip.loadAsync(
+      asyncFsReadFile(filePath, { createFolders: true })
+    );
+    const hash = createHash('sha256');
+    const entries = [];
+    zip.forEach((relativePath, entry) => {
+      let path = relativePath;
```
Rob--W, line 337: Not sure if / is included or stripped, but if unsure, you can do this:

```suggestion
      let path = relativePath.replace(/\/+$/, "");
```

The source code suggests that this is needed for consistency: https://github.com/Stuk/jszip/blob/2ceb998e29d4171b4f3f2ecab1a2195c696543c0/lib/zipEntry.js#L143-L157

src/util/submit-addon.js
```diff
@@ -308,15 +313,80 @@ export default class Client {
...
+      if (entry.dir) {
+        path += '/';
+      }
+      entries.push({ path, crc32: entry._data?.crc32 || 0 });
```
Rob--W, line 341: Would be nice to add a comment explaining why this is correct, even if it is just a link to my previous comment.

Rob--W (approved):
LGTM!

src/util/submit-addon.js
eviljeff: added

PR #2734, opened 2023-04-25
fix: Android 13 startActivity with correct intent action and category

zmckenney:
Due to Android 13 changes described [here](https://developer.android.com/guide/components/intents-filters#match-intent-filter) which require activities to now match action and categories this PR updates the `.startActivity` to do this. 

Locally building and testing shows this to now work in Fenix as expected with the command:
`web-ext run -t firefox-android --adb-device DEVICE_ID_HERE --firefox-apk org.mozilla.fenix.debug --firefox-apk-component App`

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -355,12 +355,26 @@ export default class ADBUtils {
     }
 
     await wrapADBCall(async () => {
-      await adbClient.getDevice(deviceId).startActivity({
-        wait: true,
-        action: 'android.activity.MAIN',
-        component,
-        extras,
-      });
+      try {
+        // TODO: once Fenix (release) uses Android 13, we can get rid of this
+        // call and only use the second call in the `catch` block.
+        await adbClient.getDevice(deviceId).startActivity({
+          wait: true,
+          action: 'android.activity.MAIN',
+          component,
+          extras,
+        });
+      } catch {
+        // Android 13+ requires a different action/category but we still need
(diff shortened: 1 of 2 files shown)


willdurand (approved):
I updated a test case and added support for Android < 13 builds because this change is in nightly. I verified that web-ext works with: Fenix Release, Fenix Nightly and a Fenix local/debug build.

willdurand:
@Rob--W since I updated this PR, could you please take a look too? I'd like to get that fix backported to `7.6` soon since it is a problem for the Firefox Translations team and likely other folks (including Zac's team).

Rob--W (approved):
Confirmed that the adjustment matches the intent-filter from the relevant manifests:

- Fenix - https://github.com/mozilla-mobile/firefox-android/blob/487bc890e4fd221ea90209034c522d54ab237284/fenix/app/src/main/AndroidManifest.xml#L66-L74
- GVE - https://searchfox.org/mozilla-central/rev/1f4f99a8f331cce8467a50742178b6d46914ab89/mobile/android/geckoview_example/src/main/AndroidManifest.xml#24-35 (different name)
- Focus - https://github.com/mozilla-mobile/firefox-android/blob/481768ce88bec0e5a2ae998d0eb7029f4463670a/focus-android/app/src/main/AndroidManifest.xml#L89-L102 (even though it doesn't support extensions)

Issue #2483, opened 2022-08-16
use `pipeline` directly, without `promisify` once we support node16 minimum


eviljeff:
The node stream api doesn't have promise support until [node 15](https://nodejs.org/api/stream.html#streams-promises-api), so it needs to be wrapped in `promisify` under node 14.  Once we up our minimum node to 16 we can stop doing that.

eviljeff:
I tried to fix this - it seemed like an easy fix now we're only >=16.0 node - but our flow type definitions in 0.182.0 are lacking.  The stream promises api was only added to flow in 0.204.0 so we're blocked on merging https://github.com/mozilla/web-ext/pull/2759

willdurand:
> I tried to fix this - it seemed like an easy fix now we're only >=16.0 node - but our flow type definitions in 0.182.0 are lacking. The stream promises api was only added to flow in 0.204.0 so we're blocked on merging #2759

@eviljeff we removed flow 😅

PR #2765, opened 2023-05-25
chore: Remove flow types annotations and dependencies

rpl:
This PR applies a number of changes to remove flow types annotations, related npm dependencies, use of flow-related npm dependencies in npm scripts and circle ci jobs, mentions to flow in the docs and cleanups of all the (now unnecessary) flow suppress comments.

Technically this PR will also be superseding #2759 and unblock #2483.


willdurand (approved):
Thanks!

More red than green award, congratulations!

PR #2774, opened 2023-06-03
fix: Chrome not opening special chrome URLs or extension specific pages

dleetr:
This continues the PR from https://github.com/mozilla/web-ext/pull/2321 and resolves the original issues in https://github.com/mozilla/web-ext/issues/1979

Additionally, I have expanded the special handler cases to handle urls in the format `chrome-extension://`.

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -129,17 +129,18 @@ export class ChromiumExtensionRunner {
 
     const chromeFlags = [...DEFAULT_CHROME_FLAGS];
     let startingUrl;
-    var specialStartingUrls = [];
+    let specialStartingUrls = [];
     if (this.params.startUrl) {
+      const specialUrlFormats = ['chrome://', 'chrome-extension://'];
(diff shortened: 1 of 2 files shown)


dleetr:
@Rob--W or @willdurand would either of you have some time to review this?

nishu-murmu:
@Rob--W can anyone review and merge this PR?
We extensively require this feature.

Rob--W:
I've added it to my to do list. Thanks for the ping.

Rob--W:
@dleetr The latest version of this patch is very sparse. The original PR relied on the extension helper to open tabs. That reload helper was removed in favor of using the Chrome DevTools Protocol (CDP) as part of #3388.

The PR in its current form is not doing anything any more.

If you are interested in exploring the introduction of the feature with the CDP, see the CDP documentation at https://chromedevtools.github.io/devtools-protocol/

The PR that introduced CDP support is #3434. I suggest looking at individual commits instead of the whole diff if you want a better understanding of how the current code morphed from the original version.

PR #2790, opened 2023-06-16
allow approval timeout to be customized & skipped

### LINKED ISSUE #2704: Allow to skip "wait for approval" when submitting a new version
When we sign a new version of an extension, `web-ext` waits for approval in order to download the signed XPI (it waits for some time, then it gives up). For a end user, it is probably a sensible behavior but `web-ext` is also used in CI.

In this case, it'd be interesting to expose a new CLI flag to disable this waiting loop and only store the metadata we get from the submission call itself. Ideally, the metadata should be enough to retrieve the signed XPI at a later point in time. Maybe we could even print a link that will _eventually_ point to an actual XPI (once this is signed).

eviljeff:
fixes #2704


willdurand (approved):
Sorry for the delay, looks fine.

PR #2808, opened 2023-07-06
chore: fix audit-deps npm script on deeply nested inherited advisories

rpl:
While reviewing the CI jobs failure hit by most of the dependabot PR opened over the last week we noticed that it was hitting the `audit-deps` step of the CI job but the output of our `audit-deps` npm script (which parses the json output of the `npm audit` command) wasn't always providing the details about the security advisory associated to the reported npm dependency, e.g. like here for nyc: https://app.circleci.com/pipelines/github/mozilla/web-ext/2388/workflows/f22e2ff0-ed9c-4373-88d5-bccb0bf6869c/jobs/9630?invite=true#step-116-141

Looked into the reasons behind that, I noticed that the `audit-deps` script was currently assuming that the via attribute in each of the reported npm packages details was going to be either an object with the details for the security advisory or a string that is meant to be the package name from which that package is inheriting the security advisory, but it wasn't accounting for more deeply nested cases

This PR does apply a small change to recursively resolve the elements of the `via` properties, which as an example turns the output from the linked PR:
```
nyc (isDirect: true, severity: moderate, fixAvailable: true):
  undefined
    undefined undefined
    undefined
  undefined
    undefined undefined
    undefined
  ...
```

into the one we expected:
```
nyc (isDirect: true, severity: moderate, fixAvailable: true):
  https://github.com/advisories/GHSA-c2qf-rxjj-qqgw
    semver <7.5.2
    semver vulnerable to Regular Expression Denial of Service
...
```

### DIFF
--- a/scripts/audit-deps.js
+++ b/scripts/audit-deps.js
@@ -83,10 +83,30 @@ if (auditReport) {
       // packages in the audit json report. We need to normalize the data so
       // that we always deal with a list of objects.
       item.via = item.via.reduce((acc, via) => {
-        if (typeof via === 'object') {
-          acc.push(via);
+        const addAdvisoryDetails = (entries, newEntry) => {
+          if (entries.some((entry) => entry.url === newEntry.url)) {
+            // The advisory url is already listed, no need to add a new entry.
+            return;
+          }
+          entries.push(newEntry);
+        };
+
+        if (typeof via === 'string') {
+          // Resolve the actual security advisory details recursively.
+          const recursivelyResolveVia = (currVia) => {
+            const resolvedVia = auditReport.vulnerabilities[currVia].via;
+            for (const viaEntry of resolvedVia) {
+              if (typeof viaEntry === 'string') {
+                recursivelyResolveVia(viaEntry);
+              } else {
+                addAdvisoryDetails(acc, viaEntry);
+              }
+            }
+          };
+
+          recursivelyResolveVia(via);
         } else {
-          acc.push(...auditReport.vulnerabilities[via].via);
+          addAdvisoryDetails(acc, via);
         }
 
         return acc;


rpl:
After fixing the issue in our `audit-deps` script, it becomes clear that there are two advisory that are triggering all the audit failures, in particular the semver one is the one that is hit by most of them as a transitive dependency (and based on a quick look not all the dependencies have been updated to fix their semver npm dependency):

```
diff --git a/.nsprc b/.nsprc
index 70c3aa7..8641594 100644
--- a/.nsprc
+++ b/.nsprc
@@ -1,6 +1,12 @@
 {
   "exceptions": [
     // See: https://github.com/mozilla/web-ext/issues/2678
-    "https://github.com/advisories/GHSA-p8p7-x288-28g6"
+    "https://github.com/advisories/GHSA-p8p7-x288-28g6",
+
+    // semver advisory
+    "https://github.com/advisories/GHSA-c2qf-rxjj-qqgw",
+
+    // word-wrap advisory
+    "https://github.com/advisories/GHSA-j8xg-fqg3-53r7"
   ]
 }
```

@willdurand Before considering adding these two to our nsprc and file bugs to remove the exceptions, I think that we may want to consider merging the dependabot PRs that are hitting only the audit-deps failure and then double-check how many other dependency are still make us hit the npm audit-deps because of this two advisories (and if any of them is actually part of runtime dependency which would also be hit by users of the web-ext package).

wdyt?

willdurand:
> @willdurand Before considering adding these two to our nsprc and file bugs to remove the exceptions, I think that we may want to consider merging the dependabot PRs that are hitting only the audit-deps failure and then double-check how many other dependency are still make us hit the npm audit-deps because of this two advisories (and if any of them is actually part of runtime dependency which would also be hit by users of the web-ext package).

+1

willdurand (approved):

PR #2843, opened 2023-08-14
feat: use `.web-ext-config.js` for automatic discovery

### LINKED ISSUE #2812: Feat: support `.web-ext-config.js` in the current directory.
### Feat 

>### Automatic discovery of configuration files
>
>web-ext will load existing configuration files in the following order:
>- A config file named `.web-ext-config.js` in your home directory
>- ......
>- A config file named `web-ext-config.js` in the current directory.

about Automatic discovery of configuration files, I hope that support `.web-ext-config.js` in the current directory, instead of  `web-ext-config.js`.

i dont undestand why the name of configuration files is different home dir and current dir. I don't think it's necessary

uiolee:
Fix #2812.

use `./.web-ext-config.js` for automatic discovery, ~instead of `./web-ext-config.js`~


src/config.js
```diff
@@ -145,12 +145,12 @@ export function loadJSConfigFile(filePath) {
...
 
 export async function discoverConfigFiles({ getHomeDir = os.homedir } = {}) {
-  const magicConfigName = 'web-ext-config.js';
+  const magicConfigName = '.web-ext-config.js';
 
   // Config files will be loaded in this order.
   const possibleConfigs = [
     // Look for a magic hidden config (preceded by dot) in home dir.
-    path.join(getHomeDir(), `.${magicConfigName}`),
+    path.join(getHomeDir(), magicConfigName),
```
rpl, line 153: **requested changes**: Personally I'd prefer if we would add `.web-ext-config.js` as an additional config file we look for as part of the discovery, in addition to `web-ext-config.js`, and it seems that was also what @Rob--W was proposing in https://github.com/mozilla/web-ext/issues/2812#issuecomment-1643919085

Would you mind to tweak this PR to do that? (looking for `.web-ext-config.js` in the home and in the project dir right after the one named `web-ext-config.js`.

-----

**thought:** I'm wondering if we should also add a warning message when loading for the files named `web-ext-config.js` (the ones without the leading dot) so that we can consider to deprecate them in the future, but I would keep that out of the PR for now, we can eventually consider to emit warnings in a separate followup PR.

rpl (changes requested):

src/config.js
uiolee: Do you mean to use `web-ext-config.js` and `.web-ext-config.js`? I will tweak it.

src/config.js
```diff
@@ -155,12 +155,20 @@ export async function discoverConfigFiles({ getHomeDir = os.homedir } = {}) {
...
   const configs = await Promise.all(
     possibleConfigs.map(async (fileName) => {
       const resolvedFileName = path.resolve(fileName);
       if (await fileExists(resolvedFileName)) {
+        if (path.basename(magicConfigName) === magicConfigName) {
+          log.warn(
+            `Config file "${magicConfigName}" is deprecated. ` +
+              `Please use ".${magicConfigName}"`,
+          );
+        }
```
rpl, line 171: Sorry for not having made clear that adding a warning was a thought I had but not yet something I had touched base with the rest of the team to confirm if we really wanted that.

We discussed about this PR in the triage we did today and we agreed that we can keep both filenames for the project dir in the longer term and so that we don't need nor want to warn here.

Apologies again for not having made clear that the warning was not yet an actual "request for a change" yet in my previous review comment.

tests/unit/test.config.js
```diff
@@ -1029,6 +1029,23 @@ describe('config', () => {
       });
     });
 
+    it('finds a config without dot in your working directory', () => {
+      return withTempDir(async (tmpDir) => {
+        const lastDir = process.cwd();
+        process.chdir(tmpDir.path());
+        try {
+          const undotedConfig = path.resolve(
```
rpl, line 1037: nit, I guess this is a typo (which is not really a super big deal but thought to mention it anyway) `undotedConfig` => `undottedConfig`

tests/unit/test.program.js
```diff
@@ -651,7 +651,7 @@ describe('program.main', () => {
       return configObject;
     });
 
-    await execProgram(['lint', '--config', 'path/to/web-ext-config.js'], {
+    await execProgram(['lint', '--config', 'path/to/.web-ext-config.js'], {
```
rpl, line 654: I don't think this needs to be changed, a user of web-ext can still decide to use a config file without a dot.

rpl (commented):
@uiolee Apologies for the long time we needed to get back to this PR, I just took a look and it looks pretty r+ with just the warning removed and one or two small tweaks on the tests side (the  one I actually care more about on the test side is the review comment related to leave the custom config file passes using --config to not be dotted as it was originally in that test case, but just add to that test the needed changes to confirm we also looked for the dotted file name in the project dir).

tests/unit/test.config.js
uiolee: thanks

src/config.js
uiolee: never mind. just reset it

tests/unit/test.program.js
uiolee: Sure. User can use any config file. It is just a example

rpl (approved):
@uiolee thanks a lot for contributing this enhancement, this version looks good to me and so I'm approving and then merge it.

PR #2857, opened 2023-09-01
fix: Cannot find Firefox package on device with work profile

lm1:
On some devices with work profile enabled, `adb shell pm list packages` will fail with permission error.
Avoid reaching work profile and scan packages for the primary user instead.

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -86,13 +86,31 @@ export default class ADBUtils {
     return devices.map((dev) => dev.id);
   }
 
-  async discoverInstalledFirefoxAPKs(deviceId, firefoxApk) {
-    log.debug(`Listing installed Firefox APKs on ${deviceId}`);
+  async getCurrentUser(deviceId) {
+    log.debug(`Retrieving  current user on ${deviceId}`);
 
+    const currentUser = await this.runShellCommand(deviceId, [
+      'am',
+      'get-current-user',
+    ]);
+
+    const userId = parseInt(currentUser.trim());
+    if (isNaN(userId)) {
+      throw new WebExtError(`Unable to retrieve current user on ${deviceId}`);
+    }
+    return userId;
+  }
+
+  async discoverInstalledFirefoxAPKs(deviceId, firefoxApk) {
+    const userId = await this.getCurrentUser(deviceId);
+
+    log.debug(`Listing installed Firefox APKs on ${deviceId}`);
     const pmList = await this.runShellCommand(deviceId, [
       'pm',
       'list',
       'packages',
+      '--user',
+      `${userId}`,
     ]);
 
     return pmList
--- a/tests/unit/test-util/test.adb.js
+++ b/tests/unit/test-util/test.adb.js
@@ -224,11 +224,51 @@ describe('utils/adb', () => {
       });
 
       sinon.assert.calledOnce(adb.fakeADBDevice.shell);
+    });
+
+    it('retrieves current user', async () => {
+      const adb = getFakeADBKit({
+        adbkitUtil: {
+          readAll: sinon.spy(() => {
+            return Promise.resolve(Buffer.from('123\n'));
+          }),
+        },
+      });
+
+      const adbUtils = new ADBUtils({ adb });
+      const promise = adbUtils.getCurrentUser();
+
+      sinon.assert.calledOnce(adb.fakeADBDevice.shell);
       sinon.assert.calledWith(adb.fakeADBDevice.shell, [
-        'pm',
-        'list',
-        'packages',
+        'am',
+        'get-current-user',
       ]);
+      const result = await assert.isFulfilled(promise);
+      assert.equal(result, 123);
+    });
+
+    it('rejects invalid get-current-user output', async () => {
+      const adb = getFakeADBKit({
+        adbkitUtil: {
+          readAll: sinon.spy(() => {
+            return Promise.resolve(Buffer.from('No user'));
+          }),
+        },
+      });
+
+      const adbUtils = new ADBUtils({ adb });
+      const promise = adbUtils.getCurrentUser();
+
+      sinon.assert.calledOnce(adb.fakeADBDevice.shell);
+      sinon.assert.calledWith(adb.fakeADBDevice.shell, [
+        'am',
+        'get-current-user',
+      ]);
+      await assert.isRejected(
+        promise,
+        WebExtError,
+        /Unable to retrieve current user/,
+      );
     });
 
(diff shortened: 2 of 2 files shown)


src/util/adb.js
```diff
@@ -93,6 +93,8 @@ export default class ADBUtils {
       'pm',
       'list',
       'packages',
+      '--user',
+      '0',
```
Rob--W, line 97: While this may work for your specific setup, I think that a more general fix is to use `adb shell am get-current-user` to get the actual user that's currently running on the phone, and use that instead of hardcoding 0. Can you give that a try?

lm1: Unfortunately you are right that would only work "on most devices" per [Android docs](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/Users.md).
I can give it a try, though not immediately due to some other hustle.
That seem to be an ADB bug to some extent. Disabled user does not seem to be the best default on their end.

Rob--W: @lm1 Have you had a chance to revisit this PR and test it?

lm1: Updated and tested.

lm1:
Codecov complains about the new exception line testing the result of (`am get-current-user`), but that's extremally unlikely condition the as command result is tested earlier, so perhaps does not require an additional test.

src/util/adb.js
```diff
@@ -86,13 +86,31 @@ export default class ADBUtils {
     return devices.map((dev) => dev.id);
   }
 
+  async getCurrentUser(deviceId) {
+    log.debug(`Retreiving current user on ${deviceId}`);
```
Rob--W, line 90:
```suggestion
    log.debug(`Retrieving current user on ${deviceId}`);
```

src/util/adb.js
```diff
@@ -86,13 +86,31 @@ export default class ADBUtils {
...
+
+    const currentUser = await this.runShellCommand(deviceId, [
+      'am',
+      'get-current-user',
+    ]);
+
+    const userId = parseInt(currentUser.trim());
+    if (isNaN(userId)) {
+      throw new WebExtError(`Unable to retreive current user on ${deviceId}`);
```
Rob--W, line 99: > Codecov complains about the new exception line testing the result of (`am get-current-user`), but that's extremally unlikely condition as the command result is tested earlier, so perhaps does not require an additional test.

codecov is correct. There is no test case that confirms what happens when `adb shell am get-current-user` returns an unexpected value.

tests/unit/test-util/test.adb.js
```diff
@@ -223,12 +223,28 @@ describe('utils/adb', () => {
         testFn: (adbUtils) => adbUtils.discoverInstalledFirefoxAPKs('device1'),
       });
 
+      sinon.assert.calledOnce(adb.fakeADBDevice.shell);
```
Rob--W, line 226: Restore the test coverage that verified the received parameters (`calledWith`).

tests/unit/test-util/test.adb.js
```diff
@@ -242,11 +258,25 @@ describe('utils/adb', () => {
...
+        ADBUtils.prototype,
+        'getCurrentUser',
+      );
+      stubGetCurrentUser.resolves(0);
+
       const adbUtils = new ADBUtils({ adb });
 
       const promise = adbUtils.discoverInstalledFirefoxAPKs('device1');
       const packages = await assert.isFulfilled(promise);
       sinon.assert.calledOnce(adb.fakeADBDevice.shell);
```
Rob--W, line 272: Could you also verify that `stubGetCurrentUser` was calledOnce?

tests/unit/test-util/test.adb.js
```diff
@@ -223,12 +223,28 @@ describe('utils/adb', () => {
         testFn: (adbUtils) => adbUtils.discoverInstalledFirefoxAPKs('device1'),
       });
 
+      sinon.assert.calledOnce(adb.fakeADBDevice.shell);
+    });
+
+    it('retrieves current user', async () => {
```
Rob--W, line 229: Please add a test case with a userId other than 0, just for completeness. For example 1234.

src/util/adb.js
lmielick: Yes, deemed that testcase irrelevant given it would either fail or return a user ID.

tests/unit/test-util/test.adb.js
lmielick: This is intentionally removed, assuming this test case purpose is to check recovery from the adb failure.
It should not depend on the command really. The command is now tested in the positive case below.

tests/unit/test-util/test.adb.js
lmielick: I propose updating this testcase to test non-zero ID, since zero ID is already covered by 'resolves the array of the installed firefox APKs' test.

src/util/adb.js
lmielick: Fixed

src/util/adb.js
lmielick: Fixed

tests/unit/test-util/test.adb.js
lm1: Done

tests/unit/test-util/test.adb.js
lm1: This is a negative test and it's named 'rejects an UsageError on adb binary not found'.
With the change the first shell command is no longer `pm list packages` but `am get-current-user`.
I'm quite convinced the command is not relevant in this test as the test purpose is clearly to test a recovery from failure on accessing the adb shell. Both commands are now being verified in other tests. Might be a matter of taste but verifying the same behavior in multiple tests may cause unnecessary coupling and make the code harder ro change.
Not sure would one want to change this specific test and other tests when another command is issued as first one in this flow.

tests/unit/test-util/test.adb.js
lm1: I agree with the explanation, the `resolves the array of the installed firefox APKs' does not test the `am get-current-user` result parsing.
Still we have the result parsing covered by positive and negative test already. While '0' might have not been the most representative value (updated), I'm not sure I see a reason in replicating the test to cover '0' vs '123', or any other numeric ID for that matter. That code has 100% coverage already.

lm1:
> I started the review before the patch update; I'll drop the comments that I had below and will revisit this PR later for another review round.

Was in a bit of a haste due to a trip last week, but hope the PR is good to merge as of the last code push. Responded to the remaining comments.

Rob--W (approved):
Thanks for the patch and tests!

Issue #2497, opened 2022-08-30
support uploading file content to addons submission api


eviljeff:
https://addons-server.readthedocs.io/en/latest/topics/api/addons.html supports some fields which are set or updated by uploading a file, as multipart-formdata.  The addon `icon`, version `source`, and preview images are all file uploads - they are all also sent as subsequent `PATCH` requests* because JSON and form data encoding can't be combined.

* a general update/edit function for addons/versions (i.e. without uploading a new xpi) would also be useful, but is a slightly different ask, so would be a separate issue - but would share a lot of code.

PR #2859, opened 2023-09-04
implement source file upload

eviljeff:
part of #2497 - supports uploading sources for versions

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -601,6 +601,11 @@ Example: $0 --help run.
             'Only used with `use-submission-api`',
           type: 'string',
         },
+        'version-source': {
+          describe:
+            'Path to a zip file containing human readable source code for a version. ' +
+            'Only used with `use-submission-api`',
+        },
       },
     )
     .command('run', 'Run the extension', commands.run, {
--- a/src/program.js
+++ b/src/program.js
@@ -601,10 +601,13 @@ Example: $0 --help run.
             'Only used with `use-submission-api`',
           type: 'string',
         },
-        'version-source': {
+        'upload-source-code': {
           describe:
-            'Path to a zip file containing human readable source code for a version. ' +
-            'Only used with `use-submission-api`',
+            'Path to an archive file containing human readable source code of this submission, ' +
+            'if the code in --source-dir has been processed to make it unreadable. ' +
+            'See https://extensionworkshop.com/documentation/publish/source-code-submission/ for ' +
+            'details. Only used with `use-submission-api`',
+          type: 'string',
         },
       },
     )
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -206,7 +206,11 @@ export default class Client {
     }
   }
 
-  async doAfterSubmit(addonId, newVersionId, editUrl) {
+  async doAfterSubmit(addonId, newVersionId, editUrl, patchData) {
+    if (patchData && patchData.version) {
+      log.info(`Submitting ${Object.keys(patchData.version)} to version`);
+      await this.doFormDataPatch(patchData.version, addonId, newVersionId);
+    }
     if (this.approvalCheckTimeout > 0) {
       const fileUrl = new URL(
         await this.waitForApproval(addonId, newVersionId),
@@ -379,17 +383,12 @@ export default class Client {
       version: { id: newVersionId, edit_url: editUrl },
     } = await this.doNewAddonSubmit(uploadUuid, metaDataJson);
 
-    if (patchData && patchData.version) {
-      log.info('Submitting source zip');
-      await this.doFormDataPatch(patchData.version, addonId, newVersionId);
-    }
-
     await saveIdToFileFunc(savedIdPath, addonId);
     log.info(`Generated extension ID: ${addonId}.`);
     log.info('You must add the following to your manifest:');
     log.info(`"browser_specific_settings": {"gecko": {"id": "${addonId}"}}`);
 
-    return this.doAfterSubmit(addonId, newVersionId, editUrl);
+    return this.doAfterSubmit(addonId, newVersionId, editUrl, patchData);
   }
 
   async putVersion(uploadUuid, addonId, metaDataJson, patchData) {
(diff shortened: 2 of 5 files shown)


src/program.js
```diff
@@ -601,6 +601,11 @@ Example: $0 --help run.
             'Only used with `use-submission-api`',
           type: 'string',
         },
+        'version-source': {
+          describe:
+            'Path to a zip file containing human readable source code for a version. ' +
+            'Only used with `use-submission-api`',
```
Rob--W, line 607: type:string is missing here.

(also demandOption, like the other params)

src/program.js
```diff
@@ -601,6 +601,11 @@ Example: $0 --help run.
             'Only used with `use-submission-api`',
           type: 'string',
         },
+        'version-source': {
+          describe:
+            'Path to a zip file containing human readable source code for a version. ' +
```
Rob--W, line 606:
```suggestion
            'Path to a zip file containing human readable source code of this submission. ' +
```

"a version" is vague, how about this? Other suggestions are welcome.

Ideally the documentation should be clear in expressing the requirement that the submitted add-on can be built from the zip, e.g. by linking to the extension workshop article that documents the requirements.

src/program.js
```diff
@@ -601,6 +601,11 @@ Example: $0 --help run.
             'Only used with `use-submission-api`',
           type: 'string',
         },
+        'version-source': {
```
Rob--W, line 604: Is --version-source the best public-facing option that we can think of?

Here are some other random options that I can think of that are more descriptive:

--submission-source=
--source-zip=
--private-source=

src/util/submit-addon.js
```diff
@@ -237,7 +258,7 @@ export default class Client {
   }
 
   async fetch(url, method = 'GET', body) {
-    log.info(`Fetching URL: ${url.href}`);
+    log.info(`${method}ing URL: ${url.href}`);
```
Rob--W, line 261: Nit: "GETing URL"?

```suggestion
    log.info(`Fetching URL: ${method} ${url.href}`);
```

src/util/submit-addon.js
```diff
@@ -423,14 +455,23 @@ export async function signAddon({
     channel,
     savedUploadUuidPath,
   );
+  // if we have a source file we need to upload we patch after the create
+  const versionPatchData = versionSource
```
Rob--W, line 459: Shouldn't we validate the source before attempting submission?

- That it looks like a zip file (opposed to non-existing file, a non-file like a directory, or a non-zip like a tar file)
- That it hasn't been submitted before, to ensure that the dev doesn't forget to update the zip file.

+ unit tests for these cases

Rob--W (changes requested):

src/program.js
eviljeff: well, it's the most accurate... it's the source for the version that's being created (that uploads to a field called `source` on the `/version/` endpoint) but if you think there's naming that would be more intuitive for web-ext users, sure.  Which one do you want?

src/util/submit-addon.js
eviljeff: It was a deliberate  - I quite liked `GETing URL: https://...` and `PATCHing URL: https://...` - the verbs read well as, well, verbs in a sentence.  ("CLIMBing the mountain: Everest" vs "Doing the mountain: CLIMB Everest")

But I'm not committed to it if you think keeping "Fetching" in the text is better.

src/util/submit-addon.js
eviljeff: >  Shouldn't we validate the source before attempting submission?

Nothing else is really validated before submission - otherwise we'd be duplicating the validation that takes place in the API and opening ourselves up to two diverging standards.

> That it looks like a zip file (opposed to non-existing file, a non-file like a directory, or a non-zip like a tar file)

I don't think a non-existing file, or a directory would successfully be read by `fileFromSync`, so it'd just fail with an error at that point.  I've just looked and AMO actually accepts tar files too for sources, which kind of reinforces why we shouldn't try to do too much in `web-ext` locally because what is accepted could change at any time.  (And I should adjust the wording of the help text)

> That it hasn't been submitted before, to ensure that the dev doesn't forget to update the zip file.

I'm not sure this is practical - there are cases where the xpi could have changed but the source is still the same (e.g. bad build) - but if you want to log a follow-up issue we could spec out how it would work.

src/program.js
eviljeff: throwing in some more options:
--readable-source
--source-archive
--review-sources

src/program.js
eviljeff: `demandOption` is inconsistent here and some of the other options don't specify either - it defaults to `false` so maybe we should just omit for non-required cases?

Rob--W: Let's not worry about demandOption for now then.

src/program.js
Rob--W: EW is here: https://extensionworkshop.com/documentation/publish/source-code-submission/

src/program.js
Rob--W: --upload-source-code= ? That is the label of the upload form at https://extensionworkshop.com/documentation/publish/source-code-submission/ and it's obvious that the source code is going to be uploaded somewhere.

src/program.js
eviljeff: reworded

src/util/submit-addon.js
Rob--W: > >  Shouldn't we validate the source before attempting submission?
> > That it looks like a zip file (opposed to non-existing file, a non-file like a directory, or a non-zip like a tar file)
> 
> I don't think a non-existing file, or a directory would successfully be read by `fileFromSync`, so it'd just fail with an error at that point.

My request is more about having a good developer experience than anything else. If the fileFromSync tool already emits useful error messages when a directory or non-existing file is passed, then the implementation doesn't need more. But in any case I'd like to see test coverage to ensure sensible long-term behavior.

> > That it hasn't been submitted before, to ensure that the dev doesn't forget to update the zip file.
> 
> I'm not sure this is practical - there are cases where the xpi could have changed but the source is still the same (e.g. bad build) - but if you want to log a follow-up issue we could spec out how it would work.

AMO requires unique version numbers. I would consider it quite unlikely for a source to be completely unchanged. I would consider it much more likely for someone to pass a path to a zip file to make the tool happy, and then forget to update the zip file later. Of course, when a reviewer catches this, they would reject the add-on on review, but IMO it would be worth detecting that mistake in advance.

Detection of source checks can be done similarly to how xpi duplicate checks are done now (i.e. comparing with the hash of the last successful submission). If you are concerned about blocking submission of the same source, another parameter to override the check can be added.

I'm willing to move this to a follow-up.

eviljeff: https://github.com/mozilla/web-ext/issues/2892 for dealing with dup'd source files

eviljeff: https://github.com/mozilla/web-ext/pull/2859/commits/26efa493111b7ce80e6b794ffc4704e9c9898ac0 adds tests and specific checks.

Rob--W (approved):

fregante:
Yesss looking forward to seeing this published

PR #2868, opened 2023-09-05
upload files icon

eviljeff:
further part of #2497, depends on https://github.com/mozilla/web-ext/pull/2859 - implements add-on icon upload.


src/util/submit-addon.js
```diff
@@ -207,10 +208,19 @@ export default class Client {
...
-      await this.doFormDataPatch(patchData.version, addonId, newVersionId);
+      promises.push(
+        this.doFormDataPatch(patchData.version, addonId, newVersionId),
+      );
     }
+    if (patchData && patchData.addon) {
+      log.info(`Submitting ${Object.keys(patchData.addon)} to addon`);
+      promises.push(this.doFormDataPatch(patchData.addon, addonId));
+    }
+    await Promise.all(promises);
```
Rob--W, line 222: The version-specific (source code) data and addon-global (icon) data are independent. A failure in any of these can result in an early termination of the logic. Is that desirable?

Instead of termination as soon as any failure occurs, it may make sense to wrap individual promises in always-fulfilling promises and doing error handling there.

Alternatively, if you'd like to report errors late, use `Promise.allSettled`, and then check whether there are any errors (and print useful information if needed).

Rob--W (commented):
I posted a question below. More broadly I'm also wondering what the desired behavior is when a post-submission failure occurs. I guess that the current failure mode is to still create the XPI on AMO, but to not have it available locally.

I can imagine conflicting requirements, depending on the most desired scenario:

- Someone wants to publish an update and doesn't care about source code submission or icon/metadata
  - In this case, it would be desirable to upload the xpi and ignore any icon/source code errors.
  - The downside to ignoring errors is that if used in CI, that people won't know that the icon/source code submission has failed. So perhaps it's better to not ignore the error completely, but at least wait until the next appropriate checkpoint before exiting with an error.
- Someone wants to publish an update and cares about the add-on listing & branding to be in sync with the add-on package.
  - In this case, it would be desirable to update everything in one transaction, and not commit the changes until every part succeeded.

src/util/submit-addon.js
eviljeff: I think, broadly, proceeding with a warning is probably the way forward - because at that point they can't fix the problem without a whole new version being uploaded (with a version number bump). 

The downside is there isn't a function in `web-ext` to make arbitrary API calls without uploading a versions. So the "fix" would be for them to either make a manual API request (difficult, without the tooling to make the jwt for the Auth); or use devhub (a different UX to command line). And yeah, they might miss the message in the other logging if there isn't a hard error thrown.

Given this is a flaw with the source code upload too, we should probably address that as a separate issue first (blocking V8?)

PR #2912, opened 2023-10-15
feat: Add `--chromium-pref` flag

### LINKED ISSUE #2899: Pass in custom chromium prefs (enable developer mode by default)

Feature request.

Currently, web-ext uses `chrome-launcher@0.15.1`, but v1.1.0 was just released, with a fix to make the `prefs` option work. This enables us to set custom preferences for the profile, including `extensions.ui.developer_mode`, which would enable developer mode by default when using web-ext.

```ts
import { launch } from 'chrome-launcher';

launch({
  prefs: {
    extensions: {
      ui: {
        developer_mode: true,
      },
    },
  },
});
```

Just like the `--pref` for firefox, it would be awesome if we could add `--chromium-pref` to the CLI and `chromiumPrefs` to the JS API. Or at least enable `extensions.ui.developer_mode` by default. I always have to toggle the setting when developing an extension before debugging the background in chrome.

This might also be a solution for #2874.

I'd be happy to submit a PR, but I can't get tests to pass after upgrading `chrome-launcher`, there's a error with the watcher that I can't figure out :/

- **Firefox version**: deved 119.0b2
- **Your OS and version**: MacOS Ventura 13.4.1c
- Paste the output of these commands:

```
v18.16.1
9.5.1 
7.8.0
```

aklinker1:
## Overview

This closes #2899.

Add a new parameter, `--chromium-pref`, which lets users configure [some profile settings](https://chromium.googlesource.com/chromium/src/+/main/chrome/common/pref_names.h) when running the `web-ext run` command. For default preferences, I set `extensions.ui.developer_mode=true` so you don't have to toggle the developer mode switch every time you run an extension on chromium.

https://github.com/mozilla/web-ext/blob/9d0512a1a5c3d9100764ff4bfec037c23e44ed15/src/extension-runners/chromium.js#L30-L32

## Manual Testing

Here's the commands I ran to do manual testing:

```sh
# Default, dev mode should be enabled at chrome://extensions
node .\bin\web-ext.js run -s .\tests\fixtures\minimal-web-ext\ -t chromium
```

```sh
# Disable dev mode, change download dir (chrome://settings/downloads)
node .\bin\web-ext.js run -s .\tests\fixtures\minimal-web-ext\ -t chromium --chromium-pref extensions.ui.developer_mode=false --chromium-pref download.default_directory="C:\Users\Aaron\Downloads\test"
```


tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -66,6 +66,9 @@ describe('util/extension-runners/chromium', async () => {
       '--password-store=basic',
       '--use-mock-keychain',
       '--force-fieldtrials=*BackgroundTracing/default/',
+      '--disable-hang-monitor',
+      '--disable-prompt-on-repost',
+      '--disable-domain-reliability',
```
aklinker1, line 71: These changes were required after upgrading chrome-launcher, similar to https://github.com/mozilla/web-ext/pull/2825#discussion_r1268584304

aklinker1:
Rebased and fixed conflicts

VersoriumX (approved):

VersoriumX:
No conflicts present with Branch

VersoriumX (approved):

VersoriumX (approved):
add

VersoriumX (approved):

VersoriumX (approved):

pftom:
@aklinker1 why not merge this pr? I still encounter the content script source map issues

aklinker1:
I can't merge this if it hasn't been approved by a maintainer

VersoriumX (approved):
Thanks

VersoriumX (approved):

aklinker1:
@VersoriumX I appreciate your approvals, but since you're not a maintainer, you approving this more than once doesn't really help

aklinker1:
@Rob--W no worries! Thanks for the update

Rob--W (changes requested):
This PR contains unrelated code (maybe partially due to a merge conflict).

Since you need to fix up the PR, could you split this PR in two parts:
- `chore: Update chrome-launcher` (this can be merged quickly)
- `feat: Add --chromium-pref flag` (can be reviewed and merged separately).

If it makes it easier, you could create a new branch branched off the `main` branch that updates chrome-launcher (and create a PR for that), and then rebase the current branch on top of that. Once the chrome-launcher PR is merged into main, this PR can be rebased on top of the main branch.

aklinker1:
@Rob--W Thanks for the review! I'll update this branch and make the suggested changes once `chrome-launcher` is updated. I created a PR: https://github.com/mozilla/web-ext/pull/3200

src/extension-runners/chromium.js
```diff
@@ -414,4 +420,18 @@ export class ChromiumExtensionRunner {
       }
     }
   }
+
+  /**
+   * Returns a deep preferences object based on a set of flat preferences, like
+   * "extensions.ui.developer_mode".
+   */
+  getPrefs() {
+    return Object.entries({
```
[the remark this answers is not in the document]
aklinker1: I'm going to base the function off this implementation. It's very compact and doesn't require using maps: https://youmightnotneed.com/lodash/#set

Will add some basic tests too.

Why do you want to avoid overwriting existing values? None of the default prefs I set in this PR are necessary for running the extension, they're just nice defaults that someone could turn off if they wanted. I also can't think of a use-case in the future where that would be useful.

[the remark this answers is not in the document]
Rob--W: > I'm going to base the function off this implementation. It's very compact and doesn't require using maps: https://youmightnotneed.com/lodash/#set

This snippet you've selected and lodash (`_.set`) are both vulnerable to a Prototype pollution vulnerability. The `set-value` library that you used before fixed that in response to a report (https://github.com/advisories/GHSA-4g88-fppr-53pp). Despite the fix, it is still possible to overwrite properties on inherit built-in prototypes. For example:

```js
var setValue = require('set-value');
setValue({}, 'hasOwnProperty.call', 1);
// Expected: true. Actual: TypeError: Object.prototype.hasOwnProperty.call is not a function
console.log(Object.prototype.hasOwnProperty.call(Math, "min"));
```

> Why do you want to avoid overwriting existing values?

Are you referring to "TODO: Consider log.warn() if submap.has(lastProp) before overwriting."? The purpose of that is to let the user know that a built-in value has been overwritten. Not that big of a deal. I'm also fine with omitting it. Overwriting the value when there is a `Map` at that point may be an error though, consider the following:
```
download.download_path=/path/to/somewhere
download=1
```

It is not meaningful to set "download" to 1, and a warning could be nice. Not required though.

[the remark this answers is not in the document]
aklinker1: I see now. I misunderstood lots of things about the original comment lol. But now that I've implemented and tested your original suggestion, we're on the same page.

Only comment is about throwing the error. See these test cases:

https://github.com/aklinker1/web-ext/blob/eefc4adeea0b3930adcdf69b7b92a751e81eddc7/tests/unit/test-util/test.expand-prefs.js#L39-L61

We throw an error on the first, but not the second. I think we should make both behave the same way, either throw an error or allow the later value to override the earlier. Thoughts?

src/cmd/run.js
```diff
@@ -80,6 +81,11 @@ export default async function run(
   // Create an alias for --pref since it has been transformed into an
   // object containing one or more preferences.
   const customPrefs = { ...pref };
+
+  // Create an alias for --chromium-pref since it has been transformed into an
+  // object containing one or more preferences.
+  const customChromiumPrefs = { ...chromiumPref };
```
Rob--W, line 87: I see that you basically copied the previous few lines. The logic there did not completely make sense, I've sent a PR to fix that up: https://github.com/mozilla/web-ext/pull/3218

Let's simplify to the following:

```suggestion
  const customChromiumPrefs = chromiumPref;
```

tests/unit/test-util/test.expand-prefs.js
```diff
@@ -0,0 +1,68 @@
...
+      'hasOwnProperty.call': call,
+    };
+    const expected = {
+      hasOwnProperty: {
+        call,
+      },
+    };
+    const actual = expandPrefs(input);
+
+    assert.notEqual(Object.prototype.hasOwnProperty.call, call);
```
Rob--W, line 41: Instead of notEqual, save the original value before modifying the test, then do a strict comparison. A unit test that tests for equality is stronger than one that tests for inequality.

tests/unit/test-util/test.expand-prefs.js
```diff
@@ -0,0 +1,68 @@
...
+          f: 'f',
+        },
+      },
+    };
+    const actual = expandPrefs(input);
+
+    assert.deepEqual(actual, expected);
+  });
+
+  it("doesn't pollute the object prototype", () => {
```
Rob--W, line 29: The test case here is a specific one that failed the original library that you proposed. Let's also prepend a test for a more obvious test case:

```js
const input = {
  '__proto__.x': 1
};
// Using JSON.parse because an object literal with __proto__
// would be mistaken for an object with a different prototype.
const expected = JSON.parse('{"__proto__":{"x":1}}');
```

Besides the obvious comparison like above, you should also add a test `Object.getPrototypeOf(actual)` and confirm that it is `Object.prototype`. 

And let's also have the same test, but with the following:

```js
const input = JSON.parse('{"__proto__":[]}');
const expected = JSON.parse('{"__proto__":[]}');
```

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -619,6 +619,56 @@ describe('util/extension-runners/chromium', async () => {
...
+
+    await runnerInstance.exit();
+  });
+
+  it('does pass custom prefs to chrome', async () => {
+    const { params } = prepareExtensionRunnerParams({
+      params: {
+        customChromiumPrefs: {
+          'download.default_directory': '/some/directory',
+          'extensions.ui.developer_mode': false,
```
Rob--W, line 647: Let's split the test in two: that a custom pref can be passed, and that a default pref can be overridden.

We want to make sure that custom prefs are merged with the default prefs, and not changed. The current implementation is correct, but the test as constructed would still pass if the implementation failed to merge, which is a sign of an incomplete test.

src/util/chromium-preferences.js
```diff
@@ -0,0 +1,33 @@
+import { UsageError } from '../errors.js';
+
+export function coerceCLICustomChromiumPreference(cliPrefs) {
+  const customPrefs = {};
```
Rob--W, line 4:
```suggestion
  const customPrefs = new Map();
```

Suggesting a few changes to avoid prototype pollution.

src/util/chromium-preferences.js
```diff
@@ -0,0 +1,33 @@
...
+      throw new UsageError(`Invalid custom preference name: ${key}`);
+    }
+
+    if (value === `${parseInt(value)}`) {
+      value = parseInt(value, 10);
+    } else if (value === 'true' || value === 'false') {
+      value = value === 'true';
+    }
+
+    customPrefs[`${key}`] = value;
```
Rob--W, line 29:
```suggestion
    customPrefs.set(`${key}`, value);
```

src/util/chromium-preferences.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  return customPrefs;
```
Rob--W, line 32:
```suggestion
  return Object.fromEntries(customPrefs);
```

src/util/chromium-preferences.js
```diff
@@ -0,0 +1,33 @@
...
+      throw new UsageError(
+        `Incomplete custom preference: "${pref}". ` +
+          'Syntax expected: "prefname=prefvalue".',
+      );
+    }
+
+    const key = prefsAry[0];
+    let value = prefsAry.slice(1).join('=');
+
+    if (/[^\w_.]/.test(key)) {
```
Rob--W, line 19: Is there any point in this check or can it be dropped?

`\w` means `[A-Za-z0-9_]`, so what your code here does is rejecting anything other than alphanumeric characters and period.

It is stricter than the actual values permitted by Chromium's Preferences keys, which is quite permissive actually. For example it can contain URLs (including `:` characters which are rejected by your current code), e.g. as seen at:
- example of file: https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/data/diagnostics/user/Default/Preferences;l=58;drc=f0f6afc99205acab3f069d915cdad756a52d8809
- example in code comment: https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_server_properties_manager.cc;l=245-256;drc=4787fce6c51383f5631643ac3d14cc512d656de6

VersoriumX (approved):

VersoriumX (approved):

VersoriumX (approved):

VersoriumX:
@VersoriumX squssh and merge

aklinker1:
Woops, this completely slipped my mind... I'm gonna close this. I haven't had time to make the next set of changes, and I feel bad leaving this open for... over a year, yikes.

Also, google keeps changing where settings are stored, so things like the "developer mode" toggle break multiple times a year :/

PatrykKuniczak:
It'll be merge someday or maybe it's merged already on other PR? @Rob--W 
It could be a `game changer` :)

Issue #2917, opened 2023-10-17
Support develoment environment using Node.js 20


yan12125:
### Is this a feature request or a bug?

A feature request

### What is the current behavior?

Similar to #2564, I was testing if web-ext works with the latest Node.js or not, and noticed many tests failed

<details>
<summary>test failures</summary>

```
  1) webExt
       exposes commands
         lazily loads cmd/run:
     TypeError: The "path" argument must be of type string. Received undefined
      at new NodeError (node:internal/errors:406:5)
      at validateString (node:internal/validators:162:11)
      at Object.join (node:path:1175:7)
      at getValidatedManifest (file:///home/yen/tmp/web-ext/src/util/manifest.js:579:55)
      at run (file:///home/yen/tmp/web-ext/src/cmd/run.js:1594:56)
      at run (file:///home/yen/tmp/web-ext/src/cmd/index.js:297:10)
      at Context.<anonymous> (file:///home/yen/tmp/web-ext/tests/unit/test.web-ext.js:38:24)

  2) webExt
       exposes commands
         lazily loads cmd/lint:
     TypeError: The "paths[0]" argument must be of type string. Received undefined
      at new NodeError (node:internal/errors:406:5)
      at validateString (node:internal/validators:162:11)
      at Object.resolve (node:path:1101:7)
      at new FileFilter (file:///home/yen/tmp/web-ext/src/util/file-filter.js:925:22)
      at createFileFilter (file:///home/yen/tmp/web-ext/src/util/file-filter.js:1021:10)
      at lint (file:///home/yen/tmp/web-ext/src/cmd/lint.js:405:48)
      at lint (file:///home/yen/tmp/web-ext/src/cmd/index.js:289:10)
      at Context.<anonymous> (file:///home/yen/tmp/web-ext/tests/unit/test.web-ext.js:38:24)

  3) webExt
       exposes commands
         lazily loads cmd/build:
     TypeError: The "paths[0]" argument must be of type string. Received undefined
      at new NodeError (node:internal/errors:406:5)
      at validateString (node:internal/validators:162:11)
      at Object.resolve (node:path:1101:7)
      at new FileFilter (file:///home/yen/tmp/web-ext/src/util/file-filter.js:925:22)
      at createFileFilter (file:///home/yen/tmp/web-ext/src/util/file-filter.js:1021:10)
      at build (file:///home/yen/tmp/web-ext/src/cmd/build.js:2401:45)
[...]

Issue #2564, opened 2022-11-26
Support develoment environment using Node.js 19


yan12125:
### Is this a feature request or a bug?

A feature request

### What is the current behavior?

I was testing if web-ext works with the latest Node.js or not, and noticed some tests failed
```
$ CI_SKIP_FLOWCHECK=y npm test
(...some other logs...)
  1) build
       checks locale file for malformed json:
     AssertionError: expected 'Error parsing messages.json file at /…' to match /Unexpected string in JSON at position 14/
      at file:///build/web-ext/src/web-ext-7.3.1-build/tests/unit/test-cmd/test.build.js:187:16
  2) util/manifest
       getValidatedManifest
         reports an error for invalid manifest JSON:
     AssertionError: expected 'Error parsing manifest.json file at /…' to include 'Unexpected token "," (0x2C) in JSON a…'
      at file:///build/web-ext/src/web-ext-7.3.1-build/tests/unit/test-util/test.manifest.js:36:16
      at file:///build/web-ext/src/web-ext-7.3.1-build/src/errors.js:1047:14
```

### What is the expected or desired behavior?

web-ext works fine on Node.js 19.x and all tests pass.

### Version information (for bug reports)

- **Firefox version**: irrelevant, installation issue
- **Your OS and version**: Arch Linux latest
- Paste the output of these commands:

```
$ node --version && npm --version && web-ext --version
v19.1.0
8.19.2
7.3.1
```
Logs are from web-ext 7.3.1. The issue still persists with web-ext 7.4.0.

yan12125:
Looks like those failures are caused by changed error messages from `JSON.parse()`. Tests can pass with the following changes:

```Diff
diff --git a/tests/unit/test-cmd/test.build.js b/tests/unit/test-cmd/test.build.js
index 9a63278..97e1295 100644
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -269,7 +269,7 @@ describe('build', () => {
           assert.instanceOf(error, UsageError);
           assert.match(
             error.message,
-            /Unexpected string in JSON at position 14/
+            /Expected ':' after property name in JSON at position 14/
           );
           assert.match(error.message, /^Error parsing messages\.json/);
           assert.include(error.message, messageFileName);
diff --git a/tests/unit/test-util/test.manifest.js b/tests/unit/test-util/test.manifest.js
index 7e9e17a..5fc665e 100644
[...]

PR #2951, opened 2023-11-13
ci: bump next/experimental Node to 20/21

Rob--W:
I put the latest Node.js versions as listed at https://nodejs.org in CI.
I did not bump the min version yet, to avoid backwards-incompatible breakages. We could probably do that as part of our 8.x release.

I expect CI to be green (in particular, that #2564 and #2917 are not causing failures).


Rob--W:
Well fun, CI is red. I don't have time to investigate right now.

rpl:
closing as superseded by #3014

PR #2964, opened 2023-11-22
Switch to Node 18 by default

willdurand:
Node 18 is the current LTS.


Rob--W:
Wouldn't raising the supported Node version require a major version bump?

I only bumped the maximum versions for that reason, in PR #2951.

willdurand:
> Wouldn't raising the supported Node version require a major version bump?

Yes, but we haven't release v8 yet.

rpl:
closing as superseded by #3014

PR #3018, opened 2024-01-16
fix: Remove READ_EXTERNAL_STORAGE permission requirement

### LINKED ISSUE #2312: Remove android.permission.READ_EXTERNAL_STORAGE permission requirement
Follow-up to https://github.com/mozilla/web-ext/pull/2306/files#r698444621

web-ext doesn't need to ask for the READ_EXTERNAL_STORAGE permission on Android, because the profile/add-on is now stored in `/data/local/tmp` instead of `/sdcard`: https://github.com/mozilla/web-ext/blame/c884696fd3ea496c67959337f98d4e39bdb08871/src/util/adb.js#L14 (#2200).

Let's remove the code that does it to simplify the implementation.

Rob--W:
The READ_EXTERNAL_STORAGE permission is not needed, and actually unavailable on some apps such as RefBrow. This patch removes the check and all supporting implementation.

Fixes #2312


rpl:
I also gave a try to this branch with nodejs 20 and Firefox for Android release, and no surprises with that combination neither, all worked as I was already expecting it to.

PR #3029, opened 2024-01-22
feat: Add support for config files in ES module format

### LINKED ISSUE #2604: web-ext-config ES Module Support

Feature

If you've migrated your browser extension to no longer use CommonJS modules then Node will no longer parse the `web-ext-config.js` file without additional changes.

This is the correct behavior as per the `web-ext` docs:

>The file should be a CommonJS module [as understood by NodeJS](https://nodejs.org/docs/latest/api/modules.html#modules_modules) and must export each configuration value.

and

>Automatic discovery of configuration files
>
>web-ext will load existing configuration files in the following order:
>
> * A config file named `.web-ext-config.js` in your home directory, for example: - On Windows: `C:\Users\<username>\.web-ext-config.js` - On macOS: `/Users/<username>/.web-ext-config.js` - On Linux: `/home/<username>/.web-ext-config.js`
> * A config property named `"webExt"` included in a `package.json` file in the current directory.
> * A config file named `web-ext-config.js` in the current directory.

https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/#setting-option-defaults-in-a-configuration-file

### Example

In `package.json` you have set:

```json
{
  "type": "module",
}
```

and you then run any `web-ext` command you get an error from Node.

```$shell
$ npx web-ext lint
Applying config files: ./package.json, ./web-ext-config.js

UsageError: Cannot read config file: /REDACTED/web-ext-config.js
Error: require() of ES Module /REDACTED/web-ext-config.js from /REDACTED/node_modules/import-fresh/index.js not supported.
web-ext-config.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename web-ext-config.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /REDACTED/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

Quickest win would be to add `./web-ext-config.cjs` to the pool of default files checked.

Less quick would be to allow ES Modules `export {}` in `web-ext-config.{,m}js`.

The workaround for now is to add `--config web-ext-config.cjs` as a suffix to `web-ext` commands.

It seems setting `config` in the `webExt` block in `package.json` _doesn't_ work:

```json
{
  "webExt": {
    "config": "web-ext-config.cjs"
  }
}
```

```
node --version && npm --version && npx web-ext --version
v16.15.1
8.11.0
7.4.0
```

rpl:
This PR includes a small set of changes to:
- add support for loading config files in the ESM module format
- add guidance help message to UserError error messages raised when hitting parsing error because the config file is not in the nodejs instance expects it to be (e.g. either a .js config file in CommonJS format from a type module package, or a .js config file in ES module format from a type commonjs package)
- proposed addition of an additional command which does just print all config data loaded from the discovered files

Fixes #2604 

TODO:
- [x] agree with the team if we would prefer the behavior on .js config files to be just throwing and force the user to conver them to either .cjs or .mjs config files (instead of allow them to load successfully whenever possible)
- [x] agree with the team if we feel that introducing an additional command which just dump the config data as discovered and loaded (mainly to make it easier for extensions developers to inspect the resulting data got from loading and merging all discovered config files, e.g. to debug and/or report to us issue with config file loading)
  - [x] either cover the new command with a few new tests or remove the command prototype all together (based on which is the agreement we reached on the parent todo item)

### DIFF
--- a/src/cmd/dump-config.js
+++ b/src/cmd/dump-config.js
@@ -3,5 +3,5 @@ import { createLogger } from '../util/logger.js';
 const log = createLogger(import.meta.url);
 
 export default async function config(configData) {
-  log.info(`${JSON.stringify(configData, null, 2)}`);
+  log.info(JSON.stringify(configData, null, 2));
 }
--- a/src/cmd/dump-config.js
+++ b/src/cmd/dump-config.js
@@ -3,5 +3,5 @@ import { createLogger } from '../util/logger.js';
 const log = createLogger(import.meta.url);
 
 export default async function config(configData) {
-  log.info(`Config data:\n\n${JSON.stringify(configData, null, 2)}`);
+  log.info(`${JSON.stringify(configData, null, 2)}`);
 }
--- a/src/config.js
+++ b/src/config.js
@@ -140,7 +140,7 @@ export async function loadJSConfigFile(filePath) {
     if (resolvedFilePath.endsWith('.mjs')) {
       const nonce = `${Date.now()}-${Math.random()}`;
       const { default: configDefault, ...esmConfigMod } = await import(
-        `${resolvedFilePath}?nonce=${nonce}`
+        `file://${resolvedFilePath}?nonce=${nonce}`
       );
       // ESM modules may expose both a default and named exports and so
       // we merge the named exports of top of what may have been set in
--- a/src/config.js
+++ b/src/config.js
@@ -11,6 +11,20 @@ import { UsageError, WebExtError } from './errors.js';
 
 const log = createLogger(import.meta.url);
 
+export const HELP_ERR_REQUIRE_ESM = [
+  'This config file belongs to a package.json file with "type" set to "module".',
+  'Rename the config file to end in .cjs or rewrite it to an ES module.',
+].join(' ');
+
+export const HELP_ERR_IMPORTEXPORT_CJS = [
+  'This config file is defined as an ESM module, but it belongs to either a project dir',
+  'with a package.json file with "type" set to "commonjs" or one without any package.json file.',
+  'Rename the config file to end in ".mjs" to fix the config loading error.',
+].join(' ');
+
+const ERR_IMPORT_FROM_CJS = 'Cannot use import statement outside a module';
+const ERR_EXPORT_FROM_CJS = "Unexpected token 'export'";
+
 export function applyConfigToArgv({
   argv,
   argvFromCLI,
@@ -115,7 +129,7 @@ export function applyConfigToArgv({
   return newArgv;
 }
 
-export function loadJSConfigFile(filePath) {
+export async function loadJSConfigFile(filePath) {
   const resolvedFilePath = path.resolve(filePath);
   log.debug(
     `Loading JS config file: "${filePath}" ` +
@@ -123,12 +137,31 @@ export function loadJSConfigFile(filePath) {
   );
   let configObject;
   try {
-    configObject = importFresh(resolvedFilePath);
+    if (resolvedFilePath.endsWith('.mjs')) {
+      const nonce = `${Date.now()}-${Math.random()}`;
+      const { default: configDefault, ...esmConfigMod } = await import(
+        `${resolvedFilePath}?nonce=${nonce}`
+      );
+      // ESM modules may expose both a default and named exports and so
+      // we merge the named exports of top of what may have been set in
+      // the default export.
+      configObject = { ...configDefault, ...esmConfigMod };
+    } else {
+      configObject = importFresh(resolvedFilePath);
+    }
   } catch (error) {
     log.debug('Handling error:', error);
+    let errorMessage = error.message;
+    if (error.code === 'ERR_REQUIRE_ESM') {
+      errorMessage = HELP_ERR_REQUIRE_ESM;
+    } else if (
+      [ERR_IMPORT_FROM_CJS, ERR_EXPORT_FROM_CJS].includes(error.message)
+    ) {
+      errorMessage = HELP_ERR_IMPORTEXPORT_CJS;
+    }
     throw new UsageError(
       `Cannot read config file: ${resolvedFilePath}\n` +
-        `Error: ${error.message}`,
+        `Error: ${errorMessage}`,
     );
   }
   if (filePath.endsWith('package.json')) {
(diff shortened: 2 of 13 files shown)


rpl:
@willdurand this PR includes an initial draft of a proposed set of changes to handle the issue tracked by #2604 

As briefly mentioned in the task list at the end of https://github.com/mozilla/web-ext/pull/3029#issue-2094469693

- the current version of this PR still allows and loads config data from .js config files, if we are able to load them successfully, and show an help message to guide the developers with config files that would not parse for mismatching between the format of the config file (ESM or CommonJS) and the one that nodejs expects. Alternatively we could consider to still discover .js config file but forcefully fail to force their conversion into explicit .mjs or .cjs config files. Let me know your thoughts around this two possible behaviors (or if you had some other behavior that may be worth considering)

- the new command is not stricly necessary but I was considering the idea while drafting this PR and so I thought to include it in a separate commit to bring it up and touch base about it (and so that I can just remove the commit if we agree that we are not going to add it).

rpl:
@willdurand this is ready for another pass from your perspective, this updates includes the changes we agreed on yesterday and few more small things:
- added a warning on loading from .js config file
- renamed and tweaked the dump command to emit only valid json to stdout as we discussed yesterday and covered it with a basic test case
- removed the import-fresh dependency while I was here.

src/config.js
```diff
@@ -1,16 +1,38 @@
...
+
+export const HELP_ERR_MODULE_FROM_ESM = [
+  'This config file belongs to a package.json file with "type" set to "module".',
+  'Rename the config file to end in .cjs or rename it to .mjs and rewrite it to an ES module.',
+].join(' ');
+
+export const HELP_ERR_IMPORTEXPORT_CJS = [
+  'This config file is defined as an ESM module, but it belongs to either a project dir',
+  'with a package.json file with "type" set to "commonjs" or one without any package.json file.',
+  'Rename the config file to end in ".mjs" to fix the config loading error.',
```
willdurand, line 29:
```suggestion
  'Change the file extension to ".mjs" to fix the config loading error.',
```

src/config.js
```diff
@@ -1,16 +1,38 @@
...
 
 import fileExists from './util/file-exists.js';
 import { createLogger } from './util/logger.js';
 import { UsageError, WebExtError } from './errors.js';
 
 const log = createLogger(import.meta.url);
 
+export const WARN_LEGACY_JS_EXT = [
+  'should be renamed to .cjs or .mjs to ensure its format is not ambiguous.',
+  'Config files with .js extensions are deprecated and will not be loaded anymore',
```
willdurand, line 17:
```suggestion
  'Config files with the ".js" file extension are deprecated and will not be loaded anymore',
```

src/config.js
```diff
@@ -1,16 +1,38 @@
...
+import parseJSON from 'parse-json';
 
 import fileExists from './util/file-exists.js';
 import { createLogger } from './util/logger.js';
 import { UsageError, WebExtError } from './errors.js';
 
 const log = createLogger(import.meta.url);
 
+export const WARN_LEGACY_JS_EXT = [
+  'should be renamed to .cjs or .mjs to ensure its format is not ambiguous.',
```
willdurand, line 16:
```suggestion
  'should be renamed to use the ".cjs" or ".mjs" file extension to ensure its format is not ambiguous.',
```

willdurand, line 16: Maybe add a comment to indicate that this message is added to another string when it is printed. Other strings below are full sentences but this one is partial.

src/config.js
```diff
@@ -115,20 +137,49 @@ export function applyConfigToArgv({
...
+      configModule = parseJSON(
+        await fs.readFile(resolvedFilePath, { encoding: 'utf-8' }),
+      );
+    } else {
+      configModule = await import(`file://${resolvedFilePath}?nonce=${nonce}`);
+    }
+
+    if (configModule.default) {
+      const { default: configDefault, ...esmConfigMod } = configModule;
+      // ESM modules may expose both a default and named exports and so
```
willdurand, line 163:
```suggestion
      // ES modules may expose both a default and named exports so
```

src/config.js
```diff
@@ -115,20 +137,49 @@ export function applyConfigToArgv({
...
+      // we merge the named exports of top of what may have been set in
```
willdurand, line 164:
```suggestion
      // we merge the named exports on top of what may have been set in
```

src/program.js
```diff
@@ -521,6 +521,12 @@ Example: $0 --help run.
         },
       },
     )
+    .command(
+      'dump-config',
+      'Run config discovery and dump the resulting config data',
```
willdurand, line 526:
```suggestion
      'Run config discovery and dump the resulting config data as JSON',
```

(or something like that)

src/util/logger.js
```diff
@@ -31,6 +31,8 @@ export class ConsoleStream {
       const msg = this.format(packet);
       if (this.isCapturing) {
         this.capturedMessages.push(msg);
+      } else if (packet.level > bunyan.INFO) {
+        localProcess.stderr.write(msg);
```
willdurand, line 35: ❤️

src/program.js
```diff
@@ -521,6 +521,12 @@ Example: $0 --help run.
...
+      commands.dumpConfig,
+      {},
+    )
```
willdurand, line 529: I filed an issue to document this command eventually: https://github.com/mozilla/extension-workshop/issues/1784

willdurand (approved):
r+wc, thanks

src/cmd/dump-config.js
```diff
@@ -0,0 +1,7 @@
+import { createLogger } from '../util/logger.js';
+
+const log = createLogger(import.meta.url);
+
+export default async function config(configData) {
+  log.info(`${JSON.stringify(configData, null, 2)}`);
```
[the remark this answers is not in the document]
rpl: yep, that is definitely a left over :facepalm:

src/program.js
rpl: Thank you for taking care of filing that followup as in issue in the extension-workshop! :heart:

Issue #3116, opened 2024-04-24
web-ext has no way to specify a user to use on Android, causing it to fail.


Flyingbacen:
### Is this a feature request or a bug?
bug
### What is the current behavior?
I'm attempting to install a plugin to my android device to test it, but I'm getting the error `No Firefox packages were found on the selected Android device`. Running verbose, I found it running `pm list packages`, so I ran it myself, and got the next error `Shell does not have permission to access user 10` with no extra arguments. I have to manually specify in the command to use user 0 to find firefox, but I am unable to do that as far as I'm aware

### What is the expected or desired behavior?
have a fallback or ask the user to enter the user id
### Version information (for bug reports)
android nightly - 127.0a1
OS - Ubuntu 20.04
Phone: Samsung A13
node: 20.12.2
npm: 10.5.0
web-ext: 7.11.0

Rob--W:
Could you share your verbose log?

We added auto-detection of `--user` in https://github.com/mozilla/web-ext/pull/2857. We need more information before we can improve this.

kizaski:
I'm experiencing the same issue. Here are my verbose logs 
````
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\program.js][info] Version: 7.11.0
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\program.js][debug] Discovering config files. Set --no-config-discovery to disable
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\config.js][debug] Discovered config "C:\Users\kiril\.web-ext-config.js" does not exist or is not readable
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\config.js][debug] Discovered config "C:\dev\github\borderify\package.json" does not exist or is not 
readable
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\config.js][debug] Discovered config "C:\dev\github\borderify\web-ext-config.js" does not exist or is not readable
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\cmd\run.js][info] Running web extension from C:\dev\github\borderify
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\util\manifest.js][debug] Validating manifest at C:\dev\github\borderify\manifest.json
[C:\ProgramData\nvm\v20.11.1\node_modules\web-ext\lib\extension-runners\firefox-android.js][debug] Listing android devices
[...]

Issue #2678, opened 2023-03-20
`request` library (via `sign-addon`) has a security issue


willdurand:
```
> web-ext@7.5.0 audit-deps /home/circleci/web-ext
> node ./scripts/audit-deps

== audit-deps: blocking security issues

request (https://github.com/advisories/GHSA-p8p7-x288-28g6):
  2.88.2, paths: sign-addon>request
```

Araxeus:
Shame that https://github.com/mozilla/web-ext/pull/2688 wasn't included in the last hotfix

@willdurand any ETA for when will a new version be released?

hymccord:
@Araxeus That wouldn't help anyhow. `sign-addon` 6.0.0 still depends on `request` 2.88.2

willdurand:
See also: https://github.com/mozilla/web-ext/pull/2822#issuecomment-1640054212

aspiers:
It's not just `request` which is the issue here; [`sign-addon`](https://github.com/mozilla/sign-addon) is [no longer maintained](https://github.com/mozilla/sign-addon/releases/tag/6.4.0).

willdurand:
Yeah, sign-addon is no longer maintained because we're about to release web-ext v8.

willdurand:
There is a new one:

```
jose (https://github.com/advisories/GHSA-hhhv-q57g-882q):
  4.13.1, paths: jose
```

Rob--W:
Is https://github.com/mozilla/web-ext/issues/3106 a duplicate of this?

willdurand:
> Is #3106 a duplicate of this?

I think so, yeah

GabenGar:
There is still a warning in v8.0.0, this time asking to downgrade to `5.3.0`:
```
express  <4.19.2
Severity: moderate
Express.js Open Redirect in malformed URLs - https://github.com/advisories/GHSA-rv95-896h-c2vc
fix available via `npm audit fix --force`
Will install web-ext@5.3.0, which is a breaking change
node_modules/express
  addons-scanner-utils  *
  Depends on vulnerable versions of express
  node_modules/addons-linter/node_modules/addons-scanner-utils
    addons-linter  >=2.11.0
    Depends on vulnerable versions of addons-scanner-utils
    node_modules/addons-linter
      web-ext  >=5.4.0
      Depends on vulnerable versions of addons-linter
      node_modules/web-ext

```

willdurand:
[...]

PR #3146, opened 2024-05-27
fix: Cannot find Firefox package on device with work profile (#2857)

willdurand:
Backport to v7, see https://github.com/mozilla/web-ext/issues/3116#issuecomment-2133844621

### DIFF
--- a/src/util/adb.js
+++ b/src/util/adb.js
@@ -104,16 +104,34 @@ export default class ADBUtils {
     return devices.map((dev) => dev.id);
   }
 
+  async getCurrentUser(deviceId: string): Promise<number> {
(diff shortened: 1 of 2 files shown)


willdurand:
There is a failure caused by `npm audit`, which is not unknown: https://github.com/mozilla/web-ext/issues/2678

Rob--W (approved):
Approved because of trivial backport.

The PR also includes an advisory ignore. Was that intentional? Would you consider moving the ignore to a separate pr+commit to untangle it from this unrelated PR?

willdurand:
> The PR also includes an advisory ignore. Was that intentional? Would you consider moving the ignore to a separate pr+commit to untangle it from this unrelated PR?

Sure, done.

PR #3177, opened 2024-06-21
fix: typo and missed async keyword

Decodetalkers:
I am reading the repo to learn something, but typos-lsp told me that there are some typos. So I think maybe I can make a pr.. and lsp said that there are some `async` keywords are missing.


Rob--W (approved):

Rob--W:
`async` is not strictly necessary if a Promise is already returned, but adding it does not hurt either.

PR #3188, opened 2024-07-08
feat: allow port option when run with target chromium

dever23b:
This suggestion allows us to create a dependable debug workflow to attach an IDE to the browser that gets spawned with the `run` command.

By default, the underlying `chrome-launcher` flow will choose a random debugging port when launching the browser.  This is not ideal for those of us hoping to attach  our IDE/other debugging tool to the browser.  Ideally, we launch the browser with a predictable port that we can configure into our debugging flow.  The underlying `chrome-launcher` actually already supports this; however, `web-ext` does not currently provide any way to pass the required option through.  This PR bridges that gap.

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -202,6 +202,12 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...startingUrls);
     }
 
+    let port;
+    if(this.params.chromiumPort && !isNaN(this.params.chromiumPort)) {
+      port = this.params.chromiumPort;
+      log.debug(`(port: ${port})`);
+    }
+
     this.chromiumInstance = await this.chromiumLaunch({
       enableExtensions: true,
       chromePath: chromiumBinary,
@@ -210,6 +216,7 @@ export class ChromiumExtensionRunner {
       userDataDir,
       // Ignore default flags to keep the extension enabled.
       ignoreDefaultFlags: true,
+      port,
     });
 
     this.chromiumInstance.process.once('close', () => {
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -235,7 +235,7 @@ export class ChromiumExtensionRunner {
 
     log.debug('(config: %O)', chromiumConfig);
     this.chromiumInstance = await this.chromiumLaunch(chromiumConfig);
-
+    log.debug('(process: %O)', this.chromiumInstance);
     this.chromiumInstance.process.once('close', () => {
       this.chromiumInstance = null;
 
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -18,7 +18,7 @@ import { TempDir } from '../util/temp-dir.js';
 import isDirectory from '../util/is-directory.js';
 import fileExists from '../util/file-exists.js';
 import { validatePort } from '../util/verify-chromium-port.js';
-import { PortInUseError, PortVerificationFailedError } from '../errors.js';
+import { PortInUseError, PortInvalidError } from '../errors.js';
 
 const log = createLogger(import.meta.url);
 
@@ -156,17 +156,11 @@ export class ChromiumExtensionRunner {
     const { chromiumPort } = this.params;
     if (chromiumPort) {
       log.debug('(port: %d)', chromiumPort);
-      const isPortUsable = await validatePort(
-        chromiumPort,
-        chromiumBinary,
-        chromeFlags,
-      ).catch(async (error) => {
-        // Validation wasn't able to complete successfully
-        throw new PortVerificationFailedError(error.message);
+      const isPortUsable = await validatePort(chromiumPort).catch((error) => {
+        throw new PortInvalidError(error.message);
       });
 
       if (!isPortUsable) {
-        // Validation completed, and the port definitely isn't usable
         throw new PortInUseError();
       }
     }
@@ -222,7 +216,7 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...startingUrls);
     }
 
-    const chromiumConfig = {
+    this.chromiumInstance = await this.chromiumLaunch({
       enableExtensions: true,
       chromePath: chromiumBinary,
       chromeFlags,
@@ -231,11 +225,7 @@ export class ChromiumExtensionRunner {
       // Ignore default flags to keep the extension enabled.
       ignoreDefaultFlags: true,
       port: chromiumPort,
-    };
-
-    log.debug('(config: %O)', chromiumConfig);
-    this.chromiumInstance = await this.chromiumLaunch(chromiumConfig);
-    log.debug('(process: %O)', this.chromiumInstance);
+    });
     this.chromiumInstance.process.once('close', () => {
       this.chromiumInstance = null;
 
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -17,6 +17,8 @@ import { createLogger } from '../util/logger.js';
 import { TempDir } from '../util/temp-dir.js';
 import isDirectory from '../util/is-directory.js';
 import fileExists from '../util/file-exists.js';
+import { validatePort } from '../util/verify-chromium-port.js';
+import { PortInUseError, PortVerificationFailedError } from '../errors.js';
 
 const log = createLogger(import.meta.url);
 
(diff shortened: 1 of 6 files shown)


Rob--W (changes requested):
Someone else attempted the same functionality before. The same review comments that I put there also apply here, please take a look at https://github.com/mozilla/web-ext/pull/2565

dever23b:
Oh, I apologize-- I got ahead of myself and failed to check that other PR.  I see your concern regarding a potential to attach to an instance that doesn't have the extension loaded.  Good thinking!  

I can try to work on a mitigation for that.  Since I haven't gotten started yet, I don't know if what I'm asking for might be more effort than it's worth to write, but are you hard set on aborting if the port is in use?  I'm wondering if it would be permissible to attach to an existing instance on that port, if I could find a way to verify that the existing instance was launched with the same `--load-extension` parameter.

dever23b:
Alright, sorry for the abundance of commits.  I didn't realize that this PR would cause everything to automatically sync from my fork.  

I believe the current state meets your stated concerns, @Rob--W.  Would you mind giving this another review?

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
+import { createServer } from 'node:http';
```
Rob--W, line 1:
```suggestion
import { createServer } from 'node:net';
```

We don't really need http capabilities, `node:net` is enough and we don't need `node:http`.

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {number} port Port to test
+ * */
+async function isPortAvailable(port) {
+  return new Promise((resolve) => {
+    const server = createServer();
+    server.once('listening', () => {
+      server.close(() => resolve(true));
+    });
+    server.once('error', () => resolve(false));
+    server.listen(port);
```
Rob--W, line 16:
```suggestion
    server.listen(port, '127.0.0.1');
```

By default, `listen()` listens on `::` and/or `0.0.0.0`, which would enable external clients on the network to connect to the server. This may also trigger a firewall notification. This all is undesirable.

To avoid these issues, explicitly pass in the localhost IP, i.e. `127.0.0.1`.

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+import { UnusablePortError } from '../errors.js';
+
+/**
+ * Determine if a port is available
+ * @param {number} port Port to test
+ * */
+async function isPortAvailable(port) {
+  return new Promise((resolve) => {
+    const server = createServer();
+    server.once('listening', () => {
```
Rob--W, line 12: For simplicity, you could also consider passing the callback as a parameter to the `listen` function, which is equivalent to manually registering a `listening` listener beforehand.

src/extension-runners/chromium.js
```diff
@@ -151,6 +153,18 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...this.params.args);
     }
 
+    const { chromiumPort } = this.params;
+    if (chromiumPort) {
```
Rob--W, line 157: Are you intentionally excluding `--chromium-port=0` here? The validation helpers that you have added are not 100% correct when `0` is passed.

The expected behavior when `0` is passed is to select a random port (which chrome-launcher does indeed).

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+      server.close(() => resolve(true));
+    });
+    server.once('error', () => resolve(false));
```
Rob--W, line 15: This assumes that the only error is that the port is in use (in which case `e.code === 'EADDRINUSE'`). This is not correct. E.g. ports 1-1023 require privileges, and when not granted (which is the most common case), EACCES is raised.

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+ * @returns {boolean}
+ */
+function isPortValid(port) {
+  if (!port) {
+    return false;
+  }
+  if (Number.isNaN(port) || !Number.isInteger(port)) {
+    throw new UnusablePortError(`Port provided is not an integer (${port})`);
+  }
+  if (port < 0 || port > 65535) {
```
Rob--W, line 32:
```suggestion
  if (port <= 0 || port > 65535) {
```
`0` should be excluded here. It is not reachable and if it were, it would result in a random port being chosen.

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Validate that requested port is a valid port
+ * @param {number} port Debugging port
+ * @returns {boolean}
+ */
+function isPortValid(port) {
+  if (!port) {
+    return false;
+  }
```
Rob--W, line 28:
```suggestion
```

The caller is expected to pass a port, this branch is unreachable. For simplicity let's just drop it (an error would be raised if `port` is somehow omitted).

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return true;
+}
+
+/**
+ * Validate user-supplied port to ensure it is suitable for use
+ * @param {number} port User-supplied port request
+ * @param {string} chromiumBinary Chromium binary being requested for launch
+ * @param {string[]} chromeFlags Array of flags requested for launch
```
Rob--W, line 43:
```suggestion
```

These parameters don't exist.

src/util/verify-chromium-port.js
```diff
@@ -0,0 +1,52 @@
...
+    });
+    server.once('error', () => resolve(false));
+    server.listen(port);
+  });
+}
+
+/**
+ * Validate that requested port is a valid port
+ * @param {number} port Debugging port
+ * @returns {boolean}
```
Rob--W, line 23: This `isPortValid` helper` and the exported `isPortValid` helper document a boolean return value, but throw errors when the port is invalid. This is not documented, and strange.

To avoid the strange "true / false / Error` tristate, you could consider dropping the boolean return values and just have error vs non-Error.

```js
// Raises an error if the port is invalid for any reason
await ensureValidPort(port);
```

Another option is to have a boolean return value (port available or not) and throw one error (`PortNotAvailableError`), with the diagnostic information exported via the `log` object ([example of createLogger](https://github.com/mozilla/web-ext/blob/1c015188bc215cdbd327bb5d2213727f178d6231/src/extension-runners/firefox-android.js#L17-L20)).

tests/unit/helpers.js
```diff
@@ -193,6 +193,15 @@ export class StubChildProcess extends EventEmitter {
   stderr = new EventEmitter();
   stdout = new EventEmitter();
   kill = sinon.spy(() => {});
+
+  constructor(params = {}) {
```
Rob--W, line 197: Drop this all - it is not needed by `StubChildProcess` and should not be added.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -615,6 +615,36 @@ describe('util/extension-runners/chromium', async () => {
       }),
   );
 
+  it('does pass port to chrome', async () => {
+    const port = 9222;
```
Rob--W, line 619: This unit test will fail if 9222 is used at any point. You should choose a free port here.

Note that we already have a utility to find a free port, at https://github.com/mozilla/web-ext/blob/1c015188bc215cdbd327bb5d2213727f178d6231/src/firefox/remote.js#L233

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -23,7 +23,7 @@ import isDirectory from '../../../src/util/is-directory.js';
 
 function prepareExtensionRunnerParams({ params } = {}) {
   const fakeChromeInstance = {
-    process: new StubChildProcess(),
+    process: new StubChildProcess(params),
```
Rob--W, line 26: If you want to more realistically mimic the result of chromium-launch, then you can assign the result to `fakeChromeInstance` instead of forwarding it through `StubChildProcess`.

Here is the type from chrome-launcher: https://github.com/GoogleChrome/chrome-launcher/blob/v1.1.1/src/chrome-launcher.ts#L85

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -615,6 +615,36 @@ describe('util/extension-runners/chromium', async () => {
...
+    assert.isDefined(
+      runnerInstance.chromiumInstance,
+      'returned process instance',
+    );
+    assert.isDefined(
+      runnerInstance.chromiumInstance.process.port,
+      'process instance has port value',
+    );
+    assert.equal(
+      runnerInstance.chromiumInstance.process.port,
```
Rob--W, line 640: `port` is not a member of process. Fix the stub above and read it off `chromiumInstance.port`.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -615,6 +615,36 @@ describe('util/extension-runners/chromium', async () => {
...
+      port,
+    });
+    assert.isDefined(
+      runnerInstance.chromiumInstance,
+      'returned process instance',
+    );
+    assert.isDefined(
+      runnerInstance.chromiumInstance.process.port,
+      'process instance has port value',
+    );
```
Rob--W, line 638:
```suggestion
```

This assertion is not useful. It merely tests the behavior of the internal stub logic.

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -615,6 +615,36 @@ describe('util/extension-runners/chromium', async () => {
       }),
   );
 
+  it('does pass port to chrome', async () => {
```
Rob--W, line 618: This unit test only tests the good-weather behavior. You should also test the bad weather behavior, i.e. the error conditions.

Here are the relevant cases:

- port is available.
- port is clearly invalid, e.g. -1 and 99999
- edge case: port 0. Ordinarily this means to select a random port.
- port is unavailable, which the test can guarantee by reserving a port.

Rob--W (changes requested):

Rob--W:
`web-ext` has to switch to communication by pipes to communicate with Chrome to continue support for `web-ext run` (https://github.com/mozilla/web-ext/issues/3388).
That is mutually exclusive with `--remote-debugging-port`, and therefore I am closing this PR.

Issue #2456, opened 2022-07-12
Remove `bunyan`


avi12:
### Is this a feature request or a bug?
In-between

### What is the current behavior?

`web-ext` depends on
https://github.com/mozilla/web-ext/blob/0d9f5e1835100e9365bcfe2148abe679aae831ef/package.json#L61
which depends on [moment@2.19.3](https://github.com/trentm/node-bunyan/blob/0ff1ae29cc9e028c6c11cd6b60e3b90217b66a10/package.json#L39), that has a [security vulnerability](https://github.com/avi12/skillshare-player-control/security/dependabot/17) that affects my Chrome extensions

### What is the expected or desired behavior?
Please switch to a different package other than [bunyan](https://www.npmjs.com/package/bunyan), as it seems to no longer be maintained (the last update was 2 years ago)

willdurand:
That makes sense. I think we adopted [Pino](https://github.com/pinojs/pino) in our other projects. Pino used to be a drop-in replacement for bunyan a long time ago but it has evolved a lot since then. That might still be a straightforward replacement, though.

fregante:
Are you open to a PR? Dependency-wise the project is huge and quite behind:

- https://npmgraph.js.org/?q=web-ext
- https://github.com/mozilla/web-ext/pulls


I worked on this before and I can do a little more cleaning again:

- https://github.com/mozilla/addons-linter/issues/3601

willdurand:
sure thing

fregante:
Sounds good! I can send a PR after most of the dependabot PRs are merged, there are a lot of low-hanging fruits there that would cause conflict with a Bunyan replacement PR

fregante:
Good to see that those PRs were merged last week! I sent a PR for Pino:

- https://github.com/mozilla/web-ext/pull/3214

Also there the `open` PRs failed but for unrelated reasons. It just needs a rebase:

- https://github.com/mozilla/web-ext/pull/3075


You can see how big v9 was: https://packagephobia.com/result?p=open@10.0.0

PR #3214, opened 2024-08-04
feat: replace Bunyan logger with Pino

fregante:
- Part of https://github.com/mozilla/web-ext/issues/3216
- Will close https://github.com/mozilla/web-ext/issues/2456


src/util/logger.js
```diff
@@ -17,21 +14,24 @@ export class ConsoleStream {
...
   }
 
   makeVerbose() {
     this.verbose = true;
   }
 
-  write(packet, { localProcess = process } = {}) {
-    const thisLevel = this.verbose ? bunyan.TRACE : bunyan.INFO;
+  write(json, { localProcess = process } = {}) {
+    const packet = typeof json === 'string' ? JSON.parse(json) : json;
```
rpl, line 26: Thought: I was wondering about this part conditioned on `typeof json === 'string'` and why it was needed, but I see now that pino is actually going to call this always with a string as the first parameter while the only case where it is a js object is when this class is exercised by the related unit test test.logger.js. I feel like I'd prefer to adjust the test to match how this method will be called by the actual pino logger and remove this special handling from here (and rename the parameter from `json` to `jsonString` or something like that), so that the unit test doesn't do something different from what the actual pino logger would be doing.

It looks like we just need to do a couple more tweaks to the test.logger.js unit test to get rid of this:

```diff
diff --git a/tests/unit/test-util/test.logger.js b/tests/unit/test-util/test.logger.js
index edb8991..bf738e4 100644
--- a/tests/unit/test-util/test.logger.js
+++ b/tests/unit/test-util/test.logger.js
@@ -23,12 +23,12 @@ describe('logger', () => {
 
   describe('ConsoleStream', () => {
     function packet(overrides) {
-      return {
+      return JSON.stringify({
         name: 'some name',
         msg: 'some messge',
         level: logLevels.values.info,
         ...overrides,
-      };
+      });
     }
 
     function fakeProcess() {
@@ -52,13 +52,11 @@ describe('logger', () => {
     it('logs names in verbose mode', () => {
       const log = new ConsoleStream({ verbose: true });
       assert.equal(
-        log.format(
-          packet({
-            name: 'foo',
-            msg: 'some message',
-            level: logLevels.values.debug,
-          }),
-        ),
+        log.format({
+          name: 'foo',
+          msg: 'some message',
+          level: logLevels.values.debug,
+        }),
         '[foo][debug] some message\n',
       );
     });
@@ -66,7 +64,7 @@ describe('logger', () => {
     it('does not log names in non-verbose mode', () => {
       const log = new ConsoleStream({ verbose: false });
       assert.equal(
-        log.format(packet({ name: 'foo', msg: 'some message' })),
+        log.format({ name: 'foo', msg: 'some message' }),
         'some message\n',
       );
     });
```

@fregante how that sounds? any reasons for preferring to keep it as in this version of this patch that I may have missed to notice?

rpl (changes requested):
@fregante this PR looks pretty good already (thanks for taking care of replacing buyan with pino while keeping the change 100% contained inside the logger module and its unit test!!!), there is only a small change I'd like us to consider before proceeding to merge this PR (getting rid of the check from the ConsoleStream write method to account for both js object vs json string passed as its first parameter).

src/util/logger.js
fregante: > any reasons

No, it was just a bad fix on my part. You're right it should accept just a string like it will in production.

I merged your patch and it appears to work

rpl (approved):
@fregante ty for quickly re-updating this PR! this one is now also ready to go :partying_face: I'm signing it off (and merging it righ after).

fregante:
The logs have been handled 🤝

image

PR #3215, opened 2024-08-04
Drop `mz` dependency

fregante:
https://github.com/normalize/mz

I think it offers nothing over `fs/promises`, which is node v14+

### DIFF
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -1,6 +1,6 @@
 import path from 'path';
+import fs from 'fs/promises';
 
-import { fs } from 'mz';
 import { it, describe } from 'mocha';
 import { assert } from 'chai';
 import * as sinon from 'sinon';
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -1,6 +1,6 @@
 import path from 'path';
+import fs from 'fs/promises';
 
-import { fs } from 'mz';
 import { afterEach, beforeEach, describe, it } from 'mocha';
 import { assert } from 'chai';
 import * as sinon from 'sinon';
--- a/tests/unit/test.program.js
+++ b/tests/unit/test.program.js
@@ -1,8 +1,8 @@
 import path from 'path';
+import fs from 'fs/promises';
 
 import { describe, it } from 'mocha';
 import git from 'git-rev-sync';
-import { fs } from 'mz';
 import * as sinon from 'sinon';
 import { assert } from 'chai';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "jszip": "3.10.1",
         "mkdirp": "3.0.1",
         "multimatch": "6.0.0",
-        "mz": "2.7.0",
         "node-fetch": "3.3.2",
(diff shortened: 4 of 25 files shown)


tests/unit/test-cmd/test.run.js
```diff
@@ -1,6 +1,7 @@
 import path from 'path';
+import nodeFs from 'fs';
+import fs from 'fs/promises';
```
fregante, line 3: Sometimes both APIs are used. For consistency, I let `fs` stay the promised API and used `nodeFs` for the other APIs. This name was actually used elsewhere too.

tests/unit/test-cmd/test.build.js
```diff
@@ -1,6 +1,7 @@
 import path from 'path';
+import fs from 'fs/promises';
+import { writeFileSync } from 'fs';
```
fregante, line 3: In some cases I used the single export instead. This pattern was also already used in parts of the repo. However this is not always possible due to stubbing.

fregante:
Due to the number of touched files, it would be good to merge this sooner rather than later

tests/unit/test.program.js
```diff
@@ -898,17 +898,19 @@ describe('program.defaultVersionGetter', () => {
...
-      const commit = `${git.branch(projectRoot)}-${git.long(projectRoot)}`;
-      const testBuildEnv = { globalEnv: 'development' };
-      assert.equal(
-        await defaultVersionGetter(projectRoot, testBuildEnv),
-        commit,
-      );
-    });
+      },
+    );
   });
```
rpl, line 914: Nit: not really mandatory, but given that we need to touch this test case a bit then we may as well rewrite into an async function, and make it slightly more readable:

```js
  it('returns git commit information in development', async function () {
    try {
      await fs.access(path.join(projectRoot, '.git'));
    } catch {
      this.skip();
      return;
    }
    const commit = `${git.branch(projectRoot)}-${git.long(projectRoot)}`;
    const testBuildEnv = { globalEnv: 'development' };
    assert.equal(
      await defaultVersionGetter(projectRoot, testBuildEnv),
      commit,
    );
  });
```

but I'd also like to merge this sooner rather than later, and in practice the test would be basically the same, and so we can eventually do that later after this is merged.

tests/unit/test-cmd/test.build.js
rpl: Makes sense (and thanks a lot for mentioning the motivations behind these choices explicitly in this comment, that really helped to speed up the review and sign off :heart:)

rpl (approved):
@fregante thank you so much for these dependencies cleanup PRs, I just looked to this one and looks great. I'm signing it of as is and I'm going to be merging it right after.

tests/unit/test.program.js
fregante: Definitely, in my PRs here I just limited my changes to the bare minimum to avoid unnecessary bikeshedding. 

I think both this repo and addons-linter would benefit from a strict lint config and a whole bunch of rules (like `eslint-plugin-unicorn`) with autofixes. 

Your suggestion is https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md

Issue #3216, opened 2024-08-04
Dependency cleanup


fregante:
I'll use this issue to collect my cleanup PRs related to web-ext:

- https://github.com/mozilla/web-ext/pull/3214
- https://github.com/mozilla/web-ext/pull/3215
- https://github.com/mozilla/node-fx-runner/pull/138
- https://github.com/mozilla/node-fx-runner/pull/136
- https://github.com/mozilla/node-fx-runner/pull/137

PR #3217, opened 2024-08-04
chore(deps): drop `node-fetch` dependency

fregante:
- Part of https://github.com/mozilla/web-ext/issues/3216

The `fetch` globals are available unflagged in v18+


tests/unit/test-util/test.submit-addon.js
```diff
@@ -8,8 +8,8 @@ import { assert, expect } from 'chai';
 import JSZip from 'jszip';
 import { afterEach, before, beforeEach, describe, it } from 'mocha';
 import * as sinon from 'sinon';
-// eslint-disable-next-line no-shadow
-import { File, FormData, Response } from 'node-fetch';
+// eslint-disable-next-line no-shadow -- Testing sub-100 Responses, impossible with native
+import { File, Response } from 'node-fetch';
```
fregante, line 12: The tests would fail when the mock `Response` was instantiated with <100 responses instead of throwing in the client. So I left node-fetch as a dev dependency.

src/util/submit-addon.js
```diff
@@ -1,9 +1,8 @@
 import { createHash } from 'crypto';
-import { createWriteStream, promises as fsPromises } from 'fs';
+import { createWriteStream, promises as fsPromises, readFileSync } from 'fs';
 import { promises as streamPromises } from 'stream';
 
 // eslint-disable-next-line no-shadow
```
rpl, line 5: Nit, this eslint-disable-next-line should also be removed

rpl (changes requested):
@fregante on this PR there are a couple of additional changes to consider:

- one is an actual issue (due to the fact that the AMO server API validating the file extension on the form-data upload filename)
- the other one is an idea to get rid of the remaining use of node-fetch on the test side.

Let me know wdyt.

src/util/submit-addon.js
```diff
@@ -89,7 +88,9 @@ export default class Client {
   }
 
   fileFromSync(path) {
-    return fileFromSync(path);
+    // create a blob from a file path
+    const file = readFileSync(path);
+    return new Blob([file]);
   }
```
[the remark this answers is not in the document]
fregante: That worked on first paste! 🙌 Which is great because I was ready to throw in the towel 😅 

[](https://www.youtube.com/watch?v=g-nmx5iw9oY)

tests/unit/test-util/test.submit-addon.js
fregante: _me when I paste code from rpl-GPT and everything is still green_

mother-of-god-omg

rpl (approved):
:wave: bye bye node-fetch, yours @rpl and @fregante :rofl: 

:heart: lovely to see this one also ready to go!!! 

I'm signing it off and merging it

PR #3260, opened 2024-09-27
README - add a note about extensions now requiring Firefox 58+

wesinator:
this PR updates the webextension readme to note that new published web extensions now require Firefox 58 and later.
https://github.com/mozilla/addons-server/pull/22408

Feel free to suggest different ways communicating this info if it doesn't make sense to be as a `Note:`

[NO_TRAIN]::


Rob--W:
web-ext itself loads extensions directly from the local directory, without requiring signing. The comment on Firefox version is therefore redundant.

Issue #3093, opened 2024-03-31
Unable to Test Spotify with Chromium, Missing Components


flynncao:
### Is this a feature request or a bug?

Bug.

### What is the current behavior?

* Description:
Encountering difficulties while attempting to test Spotify with Chromium using the `web-ext run serve -t chromium` via http://open.spotify.com/. 
![image](https://github.com/mozilla/web-ext/assets/29702341/57763064-c7b8-4be7-b15d-79394b465cc0)

   This issue stems from missing components, particularly the Widevine plugin, necessary for Spotify functionality. I inspected via 
   `chrome://components`, no components are installed when using this command. 

* Solutions Attempted:
Attempted to specify Chrome version using the `--chrome-binary command`: Unfortunately, this didn't resolve the issue, as the version of Chrome launched under the `-t` command remains the same.
Tried specifying Chrome profile using the `--chromium-profile` command: Despite loading explicit profiles, it had no effect on the components Chrome loaded. The components list remains empty.

* References:
https://support.spotify.com/my-en/article/web-player-help/?utm_source=open&utm_campaign=drm_error&utm_medium=web

Seeking assistance in finding a solution to this issue. Any help would be greatly appreciated!

Web Extension: [slyrics-1.5.33.zip](https://github.com/mozilla/web-ext/files/14814745/slyrics-1.5.33.zip)

### What is the expected or desired behavior?

Allow me to test web extensions on spotify.com without encountering previously displayed promotional text within the Chrome browser environment. 

### Version information (for bug reports)
- **OS**: Windows10 22H2
- **Chrome version**: 123.0.6312.86
- **Your OS and version**: Windows 10 22H2
- Paste the output of these commands:
```
node --version && npm --version && web-ext --version
```
- node version: v16.20.0 
- npm version:  8.19.4
- web-ext version: 7.11.0

Rob--W:
Widevine is a CDM (content decryption module) used to implement DRM.

The underlying chrome-launcher dependency starts Chrome with several flags, listed here:

https://github.com/GoogleChrome/chrome-launcher/blob/v0.15.1/src/flags.ts

When `web-ext run` launches Chrome, Chrome starts with these flags minus those listed in `EXCLUDED_CHROME_FLAGS` at https://github.com/mozilla/web-ext/blob/7.11.0/src/extension-runners/chromium.js#L41.

[...]

PR #3298, opened 2024-11-13
feat: remove --disable-component-update from default args

andrea11:
Similarly to puppeteer/puppeteer#13010, 
Quoting @mathiasbynens:

> Summarizing the “why” here:  
The --disable-component-update command-line flag does two things: it disables component updates (desired) but it also prevents any bundled components from being initialised (undesired).

Disable this flag grant more flexibility, and it can always be disabled via the `web-ext run --arg option`

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -19,7 +19,11 @@ import fileExists from '../util/file-exists.js';
 
 const log = createLogger(import.meta.url);
 
-const EXCLUDED_CHROME_FLAGS = ['--disable-extensions', '--mute-audio'];
+const EXCLUDED_CHROME_FLAGS = [
+  '--disable-extensions',
+  '--mute-audio',
+  '--disable-component-update',
+];
 
 export const DEFAULT_CHROME_FLAGS = ChromeLauncher.defaultFlags().filter(
   (flag) => !EXCLUDED_CHROME_FLAGS.includes(flag),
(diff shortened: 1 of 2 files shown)


andrea11:
Can someone review this PR? (@willdurand maybe?)

Rob--W (approved):
I'd prefer to not auto-download a bunch of extra stuff, but since it is currently possible to add it again, but not to opt out, I reluctantly approve this change.

When merged, this will fix https://github.com/mozilla/web-ext/issues/3093

willdurand:
> I'd prefer to not auto-download a bunch of extra stuff, but since it is currently possible to add it again, but not to opt out, I reluctantly approve this change.

I am not clear on what that means, @Rob--W.

Rob--W:
> > I'd prefer to not auto-download a bunch of extra stuff, but since it is currently possible to add it again, but not to opt out, I reluctantly approve this change.
> 
> I am not clear on what that means, @Rob--W.

I am approving this change, even if it means that Chrome will auto-download some additional components after startup. This will utilize extra bandwidth and CPU when launching (new) browser profiles instances. In exchange for that downside, the upside is that the launched Chrome browser will more closely match production. One example of that is that websites requiring DRM will work.

If devs want to revert to the behavior from before, they can manually add the command line flag to `web-ext run`.

TL;DR: LGTM, let's merge.

Issue #3309, opened 2024-11-22
UsageError message could be more helpful: "Cannot use... auto-generated extension ID"


mcomella:
### Is this a feature request or a bug?
Bug?

### What is the current behavior?
I'm not sure what the exact extension state is that triggers this. I haven't used this tool in a long time and when I updated it and ran `web-ext sign`, I received the following error message:

> UsageError: Cannot use previously auto-generated extension ID {redacted} - This extension ID must be specified in the manifest.json file.

### What is the expected or desired behavior?
The supplied error message could be more helpful and tell you specifically which key to add, e.g.:

    UsageError: Cannot use previously auto-generated extension ID {redacted} - This extension ID must be specified in the manifest.json file:
    
    // manifest.json
    {
        "browser_specific_settings": {
            "gecko": {
                "id": "{redacted}"
            }
        },

        ...
    }

As a user, I spent about twenty minutes looking for a solution. I started in the [the manifest.json docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json) (which doesn't have the answer) and eventually found a hint to the answer by searching in the `web-ext` source code.

The code that triggers this error message is here: https://github.com/mozilla/web-ext/blob/cde10a871e60f1fede2a79539563b13122f299bf/src/cmd/sign.js#L67-L70

### Version information (for bug reports)

- **Firefox version**: 131.0b5
- **Your OS and version**: macOS 14.6.1
- Paste the output of these commands:
```
$ node --version && npm --version && npx web-ext --version
v22.9.0
10.8.3
8.3.0
```

mcomella:
I submitted a pull request: https://github.com/mozilla/web-ext/pull/3310

PR #3310, opened 2024-11-22
fix: add suggestion to 'cannot use... autogenerated' msg (#3309)

mcomella:
The new output looks like:
```
UsageError: Cannot use previously auto-generated extension ID {redacted} - This extension ID must be specified in the manifest.json file. For example:

    // manifest.json
    {
            "gecko": {
                "id": "{redacted}"
            }
        },

    ...
    }

```


Rob--W (changes requested):
Could you drop the example and link to the documentation instead? https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings

The documentation already has an example at the top.

PR #3326, opened 2024-12-19
chore: Ensure submit-addon exported methods fallbacks to 'web-ext-lib' userAgentString when not specified

rpl:
As a followup to #3250, this PR sets 'web-ext-lib' as a fallback when userAgentString property has not been explicitly set.

### DIFF
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -91,7 +91,7 @@ export default class Client {
     approvalCheckInterval = 1000,
     approvalCheckTimeout = 900000, // 15 minutes.
     downloadDir = process.cwd(),
-    userAgentString,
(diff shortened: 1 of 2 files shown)


src/util/submit-addon.js
```diff
@@ -91,7 +91,7 @@ export default class Client {
     approvalCheckInterval = 1000,
     approvalCheckTimeout = 900000, // 15 minutes.
     downloadDir = process.cwd(),
-    userAgentString,
+    userAgentString = 'web-ext-lib',
```
Rob--W, line 94: Do we not want to include the version number?

E.g. `web-ext-lib/<whatever is in package.json>`

PR #3329, opened 2025-01-05
Work around vercel/ncc import issues

kewisch:
Here is a workaround so web-ext compiles with vercel's ncc. It doesn't like dynamic imports, but will happily ignore them if indirect. Any chance you'd accept this patch?

See https://github.com/vercel/ncc/issues/935 for details.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -157,7 +157,9 @@ export async function loadJSConfigFile(filePath) {
         await fs.readFile(resolvedFilePath, { encoding: 'utf-8' }),
       );
     } else {
-      configModule = await import(`file://${resolvedFilePath}?nonce=${nonce}`);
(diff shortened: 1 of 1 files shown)


Rob--W (changes requested):
This is a hack to work around ncc. We're not in favor of such a hack.

The purpose of the code here is to evaluate a config file as JS code and use its result. The use of `nonce` shows that `import()` is not exactly the perfect API here, because of its module caching behavior.

We are willing to consider non-hacky alternatives, such as logic that uses the `vm` module to execute code (https://nodejs.org/api/vm.html).

kewisch:
I think `vm` might not work without `--experimental-vm-modules` for ESM config files, but I'll see if there is another way to solve this. Let me know if you have ideas!

Issue #3388, opened 2025-04-03
Replace deprecated --load-extension with Extensions.loadUnpacked CDP to continue "web-ext run" support in Chrome


Rob--W:
Chrome is [looking into removing the `--load-extension` flag](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/aEHdhDZ-V0E/m/UWP4-k32AgAJ). `web-ext run` currently [relies on `--load-extension`](https://github.com/mozilla/web-ext/blob/a19675df339c13deab4a6e100c10c44ea4043161/src/extension-runners/chromium.js#L151) to load/run an extension and to load [a helper extension](https://github.com/mozilla/web-ext/blob/a19675df339c13deab4a6e100c10c44ea4043161/src/extension-runners/chromium.js#L268-L334) (whose purpose is to reload the extension on demand). When `--load-extension` is removed, `web-ext run` will be unable to load extensions in Chrome (and Chromium based browsers that follow the same implementation).

A supported alternative to `--load-extension` is the `Extensions.loadUnpacked` Chrome DevTools Protocol (CDP) command: 
https://chromedevtools.github.io/devtools-protocol/tot/Extensions/#method-loadUnpacked

To use this, Chrome's DevTools server has to be set up with `--remote-debugging-pipe` instead of `--remote-debugging-port`. But the `chrome-launcher` library that `web-ext` relies on right now [hardcodes a dependency on `--remote-debugging-pipe`](https://github.com/GoogleChrome/chrome-launcher/blob/324780dbb3e260a645dcf66be26fe628dc3f5097/src/chrome-launcher.ts).

For reference, Puppeteer supports pipes (and port):
- https://github.com/puppeteer/puppeteer/blob/5b777616198ae31625b9b2c1eba82c0968b82340/packages/puppeteer-core/src/node/ChromeLauncher.ts#L86-L100
- https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/node/BrowserLauncher.ts#L112-L140
- https://github.com/puppeteer/puppeteer/blob/5b777616198ae31625b9b2c1eba82c0968b82340/packages/puppeteer-core/src/node/PipeTransport.ts

Ideally, `chrome-launcher` would also accept a mechanism to switch to a pipe-based communication.

[...]

PR #3389, opened 2025-04-06
fix: Update Reload Manager Extension for Chromium to MV3

Hibi-10000:
Related to #3388

Upgrade `web-ext Reload Manager Extension` to Manifest Version 3 to prevent it from being disabled by Chromium-based browsers.

image

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -282,12 +282,12 @@ export class ChromiumExtensionRunner {
     await fs.writeFile(
       path.join(extPath, 'manifest.json'),
       JSON.stringify({
-        manifest_version: 2,
+        manifest_version: 3,
         name: 'web-ext Reload Manager Extension',
         version: '1.0',
         permissions: ['management', 'tabs'],
         background: {
-          scripts: ['bg.js'],
+          service_worker: 'bg.js',
         },
       }),
     );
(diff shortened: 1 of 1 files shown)


Rob--W:
Thanks for the PR. I am not merging this yet because there is other necessary work required to maintain long term support for Chromium browsers (#3388).

I had already noticed this issue before, but haven't migrated yet. It is not as trivial as bumping the manifest version + background property: In MV2, the background page is persistent, while in MV3 the background service worker may suspend on inactivity, which would break the WebSocket connection.

Hibi-10000:
For your information, it seemed to work fine, at least when I tested it by running it for a short time.
(In fact, I confirmed that it is able to receive information via WebSocket.)
Of course, I suppose it would need to be tested with it running for a longer period of time.

unmade-spangle:
> Thanks for the PR. I am not merging this yet because there is other necessary work required to maintain long term support for Chromium browsers (#3388).
> 
> I had already noticed this issue before, but haven't migrated yet. It is not as trivial as bumping the manifest version + background property: In MV2, the background page is persistent, while in MV3 the background service worker may suspend on inactivity, which would break the WebSocket connection.

Thanks. I believe you can incorporate a self-ping to keep it alive, @plasmohq/persistent/background has this implemented.

PR #3395, opened 2025-04-16
fix: restore stdin raw mode to off after ctrl-c quit

qupig:
When using `web-ext` via [API](https://github.com/mozilla/web-ext#using-web-ext-in-nodejs-code), after quit via `Ctrl-C`, the parent process can no longer respond to any keyboard input.

This PR fixes the issue by correctly returning `stdin` to the parent process.

This fix is ​​only tested and confirmed in the firefox-desktop version.
The firefox-android process may have similar issue, but I cannot test it at the moment.

### DIFF
--- a/src/extension-runners/index.js
+++ b/src/extension-runners/index.js
@@ -255,7 +255,9 @@ export function defaultReloadStrategy(
   extensionRunner.registerCleanup(() => {
     watcher.close();
     if (allowInput) {
-      if (isTTY(stdin)) setRawMode(stdin, false);
+      if (isTTY(stdin)) {
+        setRawMode(stdin, false);
+      }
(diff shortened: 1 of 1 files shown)


rpl:
@qupig would be possible from the nodejs app using web-ext as a dependency to do `process.stdin.destroy()` call on its side or the API as exposed by web-ext not allowing that? (e.g. missing a callback or some other hook point to do that kind of cleanup)?

Rob--W:
If you use it as a library, would the noInput option better match your needs? Then web-ext won't capture the stdin.

rpl (approved):
Thank you for further investigating the issue :heart: 

This change looks reasonable given that it is balancing the call to setRawMode true.

PR #3434, opened 2025-05-28
fix: support Chrome 137+ in web-ext run -t chromium (+functional tests)

### LINKED ISSUE #3388: Replace deprecated --load-extension with Extensions.loadUnpacked CDP to continue "web-ext run" support in Chrome

Rob--W:
Fixes #3388.

This PR gets `web-ext run -t chromium` working again in Chrome 137+.

It uses the `Extensions.loadUnpacked` CDP command, with a fallback to `--load-extension` for versions before Chrome 126 (all based on feature detection; there are no explicit version checks in the implementation, to allow it to work with any Chromium-based browser).

Since the implementation now uses CDP, it also replaces the "web-ext Reload Manager Extension" helper extension with a CDP-based mechanism to reload the extension.

All of this is covered by extensive functional tests, with a fake chrome binary that accurately simulates the relevant behaviors. I have verified the behaviors  Chrome 69, 70, 75, 77, 88, 100, 122, 125, 126, 134, 139. Relevant findings of version-specific behavior is documented in https://github.com/mozilla/web-ext/issues/3388#issuecomment-2906982117. The tests with real Chrome are disabled by default to not slow down CI, but if interested one can run them by setting the `TEST_WEBEXT_USE_REAL_CHROME=1` environment variable. More tips on debugging are documented in `tests/functional/test.cli.run-target-chromium.js`.

### DIFF
--- a/tests/functional/common.js
+++ b/tests/functional/common.js
@@ -131,10 +131,11 @@ export function execWebExt(argv, spawnOptions) {
 export function monitorOutput(spawnedProcess) {
   const callbacks = new Set();
   let outputData = '';
+  let errorData = '';
   function checkCallbacks() {
     for (const callback of callbacks) {
       const { outputTestFunc, resolve } = callback;
-      if (outputTestFunc(outputData)) {
+      if (outputTestFunc(outputData, errorData)) {
         callbacks.delete(callback);
         resolve();
       }
@@ -142,6 +143,10 @@ export function monitorOutput(spawnedProcess) {
   }
   spawnedProcess.stdout.on('data', (data) => {
     outputData += data;
+    checkCallbacks();
+  });
+  spawnedProcess.stderr.on('data', (data) => {
+    errorData += data;
     checkCallbacks();
   });
 
--- a/tests/functional/fake-chrome-binary.js
+++ b/tests/functional/fake-chrome-binary.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { createReadStream, createWriteStream } from 'node:fs';
+import { createReadStream, createWriteStream, existsSync } from 'node:fs';
 import { readFile } from 'node:fs/promises';
 import http from 'node:http';
 import net from 'node:net';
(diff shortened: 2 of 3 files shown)


Rob--W:
This PR is part of the 8.8.0 release (https://github.com/mozilla/web-ext/releases/tag/8.8.0).

It is possible for websites to treat the launched Chrome differently when they check `navigator.webdriver` as explained in https://github.com/mozilla/web-ext/issues/3388#issuecomment-2906988005

A work-around to that is to use `web-ext run -t chromium --args=--disable-blink-features=AutomationControlled`

PR #3450, opened 2025-06-19
chore: fix npm audit for brace-expansion

willdurand:


willdurand:
> for example, to make `github-pr-title-lint` check happier in CI?

ah right, I don't like this linting thing...

PR #3553, opened 2025-11-13
fix: Ensured `navigator.webdriver != true` in launched Chromium

### LINKED ISSUE #3511: Chromium launched via web-ext run is marked unsafe by Google, works fine when launched directly
```sh
web-ext run --target chromium \
  --chromium-binary "/usr/local/bin/chromium" \
  --args="--user-data-dir=/Users/xxx/Library/Application\ Support/Chromium" \
  -s /Users/xxx/.output/chrome-mv3-dev
```

web-ext version: 8.9.0
OS: macOS 26.0 Beta (25A5351b)

When I launch Chromium using the above command, Google treats the browser as "unsafe."
This causes a lot of trouble, not just account login failures:

Google Search always requires CAPTCHA verification.

YouTube videos stop after a few seconds, and captions cannot be loaded.

Since my extension depends on YouTube services, this issue makes development very difficult.

Strangely, I didn’t have this problem before. It only started in the last couple of days, even though I haven’t upgraded Chromium or web-ext. Maybe Google recently changed its detection logic?

When I launch Chromium manually from the terminal with:

sh
Copy code
/usr/local/bin/chromium
everything works fine — login, search, YouTube playback.

So my final question is:
👉 What is the actual difference between launching Chromium directly (/usr/local/bin/chromium) and launching it via web-ext run --target chromium --chromium-binary "/usr/local/bin/chromium"?

rnwst:
The `--remote-debugging-pipe` flag causes `navigator.webdriver == true` in the launched Chromium instance (provided there is no existing Chromium instance). Flags need to be added to prevent this, otherwise many websites with bot detection features are broken.

Fixes #3511

### DIFF
--- a/src/extension-runners/chromium.js
+++ b/src/extension-runners/chromium.js
@@ -256,6 +256,21 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...this.params.args);
     }
 
+    // `--remote-debugging-pipe` results in `navigator.webdriver == true` in
+    // the launched Chromium instance (unless a Chromium session exists
+    // already, in which case the existing session is used, which typically
+    // has `navigator.webdriver == false`). This breaks websites with bot
+    // detection features. The following flag prevents this.
+    // Only pass this flag if `--enable-blink-features=AutomationControlled`
+    // is not passed.
+    if (
+      !chromeFlags.some((flag) =>
+        /^--enable-blink-features=(.*,)?AutomationControlled(,|$)/.test(flag),
+      )
+    ) {
+      chromeFlags.push('--disable-blink-features=AutomationControlled');
+    }
+
     // eslint-disable-next-line prefer-const
     let { userDataDir, profileDirName } =
       await ChromiumExtensionRunner.getProfilePaths(
--- a/tests/unit/test-extension-runners/test.chromium.js
+++ b/tests/unit/test-extension-runners/test.chromium.js
@@ -117,6 +117,7 @@ describe('util/extension-runners/chromium', async () => {
         ...DEFAULT_CHROME_FLAGS,
         '--remote-debugging-pipe',
         '--enable-unsafe-extension-debugging',
+        '--disable-blink-features=AutomationControlled',
       ],
       startingUrl: undefined,
     });
@@ -227,6 +228,7 @@ describe('util/extension-runners/chromium', async () => {
         ...DEFAULT_CHROME_FLAGS,
         '--remote-debugging-pipe',
         '--enable-unsafe-extension-debugging',
+        '--disable-blink-features=AutomationControlled',
       ],
       startingUrl: undefined,
     });
@@ -250,6 +252,7 @@ describe('util/extension-runners/chromium', async () => {
         ...DEFAULT_CHROME_FLAGS,
         '--remote-debugging-pipe',
         '--enable-unsafe-extension-debugging',
+        '--disable-blink-features=AutomationControlled',
         'url2',
         'url3',
       ],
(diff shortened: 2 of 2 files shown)


rnwst:
@Rob--W -- thanks for the review and the link to the issue, I hadn't realised that this is a known issue. Without `--test-type`, the message shown by Chromium is rather annoying. I also discovered in my testing that the `--remote-debugging-port` option does not exhibit the above issue, so a potential solution could be to switch to that instead. Before I start looking into that, are there any reasons you can think of why this solution wouldn't work or wouldn't be desired?

Rob--W:
> @Rob--W -- thanks for the review and the link to the issue, I hadn't realised that this is a known issue. Without `--test-type`, the message shown by Chromium is rather annoying.

Even without `--test-type` set by default, you can start `web-ext run` with `--args=--test-type` (or its shorthand) if you'd like.

> I also discovered in my testing that the `--remote-debugging-port` option does not exhibit the above issue, so a potential solution could be to switch to that instead. Before I start looking into that, are there any reasons you can think of why this solution wouldn't work or wouldn't be desired?

The `--remote-debugging-port` option does not allow loading of extensions, see #3388.

For this and other alternatives that I considered, see https://github.com/mozilla/web-ext/issues/3388#issuecomment-2906988005

rnwst:
@Rob--W -- I removed the `--test-type` flag and added logic to avoid passing `--disable-blink-features=AutomationControlled` when --enable-blink-features is set with `AutomationControlled`. I think the right place to mention `--test-type` would be in the [documentation](https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/), though I couldn't find the source for that.

Not sure why the dependency audit is failing, I didn't change any dependencies.

src/extension-runners/chromium.js
```diff
@@ -256,6 +256,23 @@ export class ChromiumExtensionRunner {
...
+    // already, in which case the existing session is used, which typically
+    // has `navigator.webdriver == undefined`). This breaks websites with bot
+    // detection features. The following flag prevents this.
+    // Only pass this flag if `--enable-blink-features=AutomationControlled`
+    // is not passed.
+    if (
+      !chromeFlags.some(
+        (flag) =>
+          flag.startsWith('--enable-blink-features') &&
+          flag.match(/AutomationControlled/),
```
Rob--W: Generally the condition on these lines is broader than needed, but given the length and content of the strings, it is unlikely to match something unrelated, so I am okay with that.

If we wanted to make it very precise, we could change the whole check to `/^--enable-blink-features=(.*,)AutomationControlled(,|)/.test(flag)`.

If we want to avoid unnecessary use of rege, this line here can change to `flag.includes("AutomationControlled"),`.

Rob--W (approved):
LGTM, thanks!

src/extension-runners/chromium.js
```diff
@@ -256,6 +256,23 @@ export class ChromiumExtensionRunner {
       chromeFlags.push(...this.params.args);
     }
 
+    // `--remote-debugging-pipe` results in `navigator.webdriver == true` in
+    // the launched Chromium instance (unless a Chromium session exists
+    // already, in which case the existing session is used, which typically
+    // has `navigator.webdriver == undefined`). This breaks websites with bot
```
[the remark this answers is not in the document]
rnwst: You are right!

rnwst:
@Rob--W -- Thanks for the suggestions, I've committed those now.

src/extension-runners/chromium.js
```diff
@@ -256,6 +256,21 @@ export class ChromiumExtensionRunner {
...
+    // `--remote-debugging-pipe` results in `navigator.webdriver == true` in
+    // the launched Chromium instance (unless a Chromium session exists
+    // already, in which case the existing session is used, which typically
+    // has `navigator.webdriver == false`). This breaks websites with bot
+    // detection features. The following flag prevents this.
+    // Only pass this flag if `--enable-blink-features=AutomationControlled`
+    // is not passed.
+    if (
+      !chromeFlags.some((flag) =>
+        /^--enable-blink-features=(.*,)AutomationControlled(,|)/.test(flag),
```
Rob--W:
```suggestion
        /^--enable-blink-features=(.*,)?AutomationControlled(,|$)/.test(flag),
```

I realize that this regex is a copy paste of what I sketched before, but it has two mistakes that need to be corrected to have the intended outcome:

- add `?` to make the leading comma optional, test case `--enable-blink-features=AutomationControlled`
- add `$` after the `|` to make sure that we require an anchor. Otherwise something like `--enable-blink-features=AutomationControlledButNotReally` would also be matched.

Rob--W: To verify the fix automatically (instead of relying on code review), the unit test could be expanded to test it.

One way to do it is to copy paste the test case twice more, but I'd prefer to not have that level of repetition and reuse the test logic in a new function instead.

The test framework (mocha) has an example showing how this can be done: https://mochajs.org/#dynamically-generating-tests

Rob--W (approved):
Marking as approved to clearly signal that the current shape of the patch looks good to me. In a comment I suggested a way to improve test coverage further, but it is not required for merging.

I'll let another team member do the actual merge and release since I am travelling.

PR #3579, opened 2025-12-31
chore(readme): fix formatting of package.json example

y-koj:
Before this commit, the example of package.json in README.md didn't follow the JSON grammar as the key-value pair was not surrounded by `{}`, and there was a trailing comma after the last key-value pair.
This commit fixes the README.md to include a valid package.json example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,10 @@ the source code for your extension.
 `package.json`
 
 ```json
-"scripts": {
-  "start:firefox": "web-ext run --source-dir ./extension-dist/",
+{
+  "scripts": {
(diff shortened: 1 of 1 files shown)


rpl:
We briefly talked about this pull request in our triage today, we agreed that we prefer that snippet to not look like it is the complete json file but just the relevant fragment. That property is expected to be part of a package.json file, the change in this PR makes that a valid json but as a package.json it would be incomplete, and potentially misleading.

y-koj:
I understand. Thank you for the review!

Issue #3460, opened 2025-07-01
Feature: GitHub Action `Mozilla-Actions/web-ext@v8`


reneleonhardt:
Would it be easier for developers if a GitHub Action like `Mozilla-Actions/web-ext@v8` would be available instead of rebuilding it in their CI workflows over and over again? 😄

Rob--W:
What functionality are you looking for? What are the alternatives that you are using right now?

Rob--W:
We do not have time to maintain a Github action, but here is an existing Github action from a third party that may offer what you need: https://github.com/kewisch/action-web-ext

PR #3583, opened 2026-01-04
docs: document web-ext usage in CI environments

inverse:
Saw this closed issue (#3460) and thought it might be worthwhile highlighting existing solutions to make it easier for folks to discover what's available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,12 @@ signAddon({
 });
 ```
 
+## Using web-ext in CI envionments.
+
+### GitHub actions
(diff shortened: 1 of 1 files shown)


rpl:
We have triage this pull request in our triage meeting today and we agreed to wontfix it with the following motivation:

as mentioned in the closed issue (#3430) we are not maintaining that custom github action and so we prefer to not link it in the web-ext README, because it would make it look like it is officially supported while it is not.

PR #3586, opened 2026-01-09
feat: Added --firefox-debugger-socket option

yureitzk:
This allows connecting to Firefox on Android devices without access to `/proc/net/unix`, bypassing automatic socket discovery.

### DIFF
--- a/src/cmd/run.js
+++ b/src/cmd/run.js
@@ -49,6 +49,7 @@ export default async function run(
     adbRemoveOldArtifacts,
     firefoxApk,
     firefoxApkComponent,
+    firefoxDebuggerSocket,
     // Chromium CLI options.
     chromiumBinary,
     chromiumProfile,
@@ -150,6 +151,7 @@ export default async function run(
       preInstall,
       firefoxApk,
       firefoxApkComponent,
+      firefoxDebuggerSocket,
       adbDevice,
       adbHost,
       adbPort,
--- a/src/extension-runners/firefox-android.js
+++ b/src/extension-runners/firefox-android.js
@@ -52,6 +52,7 @@ export class FirefoxAndroidExtensionRunner {
   exiting;
   selectedAdbDevice;
   selectedFirefoxApk;
+  firefoxDebuggerSocket;
   selectedArtifactsDir;
   selectedRDPSocketFile;
   selectedTCPPort;
@@ -458,7 +459,7 @@ export class FirefoxAndroidExtensionRunner {
       adbUtils,
       selectedAdbDevice,
       selectedFirefoxApk,
-      params: { adbDiscoveryTimeout },
+      params: { adbDiscoveryTimeout, firefoxDebuggerSocket },
     } = this;
 
     const stdin = this.params.stdin || process.stdin;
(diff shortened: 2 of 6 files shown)


Rob--W:
Could you elaborate when this would be needed, and how we can test it?

The current logic matches Firefox's logic that is used at `about:debugging` on desktop. There, we also parse `/proc/net/unix` without fallback, see this Firefox source code: https://searchfox.org/firefox-main/rev/28398d11d39984cc5db08020671e7dd0d1408ec0/devtools/client/shared/remote-debugging/adb/adb-device.js#29-40

yureitzk:
Yes. The problem is that new version of Android [don't have an access](https://developer.android.com/about/versions/10/privacy/changes#proc-net-filesystem) to `/proc/net/unix` on non-root devices, and the current code fails to extract Firefox's socket. This option allows user to skip the file check and manually specify the socket.

The socket usually looks like this `@org.mozilla.fenix/firefox-debugger-socket` . It's pretty much the same string you parse from `/proc/net/unix`. To test the connection in the code I make a temporary `FirefoxRDPClient` and make a test connection on it. Not sure if that's the best to test the connection though, but without it, the program will fail with a cryptic error for an end-user

Rob--W:
You linked to a restriction for apps in Android 10. I have tested various Android versions on different Pixel devices without issues.

Which phone are you using?
What OS is it running?
What is the output of:
`adb shell cat /proc/net/unix`
`adb shell ls -la /proc/net/unix`

yureitzk:
I think I figured it out. Some of my phones (running Android 10+) have access to `/proc/net/unix`, and dome don't. The ones that can't access it are running custom ROMs that set the file's permissions to `440`, so you can't really open the file via `adb`, you need a root shell

Rob--W:
If you have a root shell you can fix up your permissions yourself. For example, here is a (random) other project where a similar access issue was encountered, and a user shared explicit steps to fix it up temporarily: https://github.com/mpotthoff/vscode-android-webview-debug/issues/4#issuecomment-1208160672

yureitzk:
For sure, you can change the permissions, but doing this just to test a web extension is not really handy, especially if you rely on `440` mask and you need to change it each time you want to test a web extension on your phone

Rob--W:
`/proc/net/unix` 440 is not a standard configuration, and prevents `web-ext` from discovering debugging targets.

Allowing users to specify a custom debugging socket is a footgun, because it would result in an attempt to connect to the wrong app.

We'd be willing to consider a safe patch that just falls back to trying to detect the presence of the abstract socket by its well-known name if `/proc/net/unix` is unreachable. When `web-ext` runs a Firefox app, it already knows which appid it wants to launch, so you could use that to construct the socket path.

To enable the use of `about:debugging` in Firefox, a similar patch would be needed for Firefox though. Otherwise you can start an extension with `web-ext`, but not be able to debug it via `about:debugging`.

yureitzk:
But shouldn't the option be used by people who know what they're doing? I highly doubt that people who don't know what this option does will attempt to connect to the wrong app.

Do you mean that you want this option to be a flag that accepts no arguments, and just constructs the path automatically, based on the provided apk?

I simply forward the socket to a host machine using the command below, then added this network location in `about:debugging`, it makes the debugging possible.

```sh
adb forward tcp:6000 localabstract:org.mozilla.fenix/firefox-debugger-socket
```

PR #3664, opened 2026-03-20
feat: Add `--chromium-pref` to CLI options

PatrykKuniczak:
Basing on #2912, I've taken couple commits and a bunch of code.

I've added request changes of @Rob--W review.

I've converted `customPrefs` to `Map()` because of https://github.com/mozilla/web-ext/pull/2912/changes#r1704363575

Let's guide me, how to merge this PR successfully, because a bunch of community of https://github.com/wxt-dev/wxt/issues/137 waiting for this, and i want to do it and use in our codebase 😸 

I hope this is understable and nothing'll block me to do this 😆


PatrykKuniczak:
Wait a second `coerceCLICustomChromiumPreference` is the same as `coerceCLICustomPreference`, but there's ` const customPrefs = {};` 

I'll remove `coerceCLICustomChromiumPreference` because that's redundant, but @Rob--W tell me, if you want `Map()` like it's now, and i'll convert `coerceCLICustomPreference` to use `Map()` and remove regexp.

I'm waiting to your response and then i'll quickly adjust this PR 😄

Rob--W (commented):
This PR makes more changes than strictly necessary to implement the functionality, such as moving several functions around without obvious reason.

Could you undo these unnecessary changes to make it easier to review the code? It has been a long while since I have reviewed the original unmerged PR so I basically have to review from scratch because the PR has escaped my human context window in the meantime.

PatrykKuniczak:
@Rob--W Which changes are you want me to undo?

Rob--W:
> @Rob--W Which changes are you want me to undo?

Take a look at the diff of your PR: https://github.com/mozilla/web-ext/pull/3664/files
There are a few places where code moved around or modified with no apparent connection to the feature this patch is intended to provide.

PatrykKuniczak:
@Rob--W DONE

Pls review it.

I'm able to convert everything from `object` to `Map`, if you'll want that, but in other PR.

`preferences.js`
```ts
  if (/[^\w{@}.-]/.test(key)) {
    }
```
This can be removed as you've suggested [here](https://github.com/mozilla/web-ext/pull/2912/changes#r1704377300), but maybe not on this PR, because it's not related.

But your words:
`including ":" characters which are rejected by your current code`
It will be also rejected here, maybe omit this regexp for chrome?

Let's give me move advice i want to do it, not walking in the dark

PatrykKuniczak:
@Rob--W Any moves forward?

Rob--W:
Thanks for fixing the import order. I'll take a look next week.

PatrykKuniczak:
@Rob--W Fun fact, i've fixed it then, but i've forgot to push it, and today i found it by mistake and pushed it XD

PatrykKuniczak:
@Rob--W This week? :)

PatrykKuniczak:
@willdurand Maybe you're able to check it?

Rob--W:
I was away, am back now. I'll take a look later this week, ideally before Thursday.

PatrykKuniczak:
@Rob--W Thanks for response, i'll be able to make any adjustments on the end of this week, if it'll be necessary 😸

Rob--W (approved):
Looks good. Thanks for your contribution!

The PR currently has a CI failure, due to audit failures that are likely unrelated. Could you rebase the patch (or merge the main branch) to see if the tests pass?

PatrykKuniczak:
@Rob--W WOOOW, looks good :)

I'm thrilled to see it on next release.
Thanks for support

src/extension-runners/chromium.js
Rob--W, line 336:
```suggestion
      ignoreDefaultFlags: true,
      prefs: this.getPrefs(),
```

This crucial bit of implementation was missing in your patch. It was present in the original (#2912) but dropped for some unknown reason.

I can confirm that the functionality behaves as expected with this fix, however the unit tests don't care either way.

Rob--W (changes requested):
I tried out your patch, expecting Developer Mode to be turned on by default. However, that did not work. It turns out that you omitted a `prefs: this.getPrefs(),` call. I have suggested an implementation change that fixes the issue.

However, the unit tests pass with and without the change, signaling the lack of necessary test coverage. Could you add the necessary test coverage?

- verify that the `developer_mode` pref defaults to true. See the `installs and runs the extension` test for inspiration on how you can verify the launch options. The `calledWithMatch` accepts objects when the specified properties overlap, so you don't have to include all other properties.
- `merges default and custom preferences from an object` should also check that the passed prefs are forwarded to the launcher.

src/extension-runners/chromium.js
PatrykKuniczak: @Rob--W Probably i've missed it, thanks for second look

PatrykKuniczak:
> I tried out your patch, expecting Developer Mode to be turned on by default. However, that did not work. It turns out that you omitted a `prefs: this.getPrefs(),` call. I have suggested an implementation change that fixes the issue.
> 
> However, the unit tests pass with and without the change, signaling the lack of necessary test coverage. Could you add the necessary test coverage?
> 
> * verify that the `developer_mode` pref defaults to true. See the `installs and runs the extension` test for inspiration on how you can verify the launch options. The `calledWithMatch` accepts objects when the specified properties overlap, so you don't have to include all other properties.
> * `merges default and custom preferences from an object` should also check that the passed prefs are forwarded to the launcher.

@Rob--W Yeah no problem, done :)

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -70,6 +70,28 @@ function prepareExtensionRunnerParams(
 }
 
 describe('util/extension-runners/chromium', async () => {
+  describe('developer_mode pref', () => {
+    it('defaults extensions.ui.developer_mode to true', async () => {
+      const { params } = prepareExtensionRunnerParams();
+      const runnerInstance = new ChromiumExtensionRunner(params);
+      const prefs = runnerInstance.getPrefs();
+      assert.propertyVal(prefs.extensions.ui, 'developer_mode', true);
```
Rob--W: This unit test would not catch the issue I raised most recently. The purpose of the test should be to verify that the launcher is called with the right parameters.

Something like this:

```suggestion
      await runnerInstance.run();
      sinon.assert.calledWithMatch(params.chromiumLaunch, {
        prefs: {
          ui: {
            developer_mode: true,
          },
        },
      });
      await runnerInstance.exit();
```

tests/unit/test-extension-runners/test.chromium.js
```diff
@@ -70,6 +70,28 @@ function prepareExtensionRunnerParams(
...
+    });
+
+    it('allows overriding extensions.ui.developer_mode to false', async () => {
```
Rob--W: We don't need this test, it is already covered by the `merges default and custom preferences from an object` test below.

Let's drop this one, and also drop the new `describe(...)`, and move the new `defaults extensions.ui.developer_mode to true` test after the flags test:
https://github.com/mozilla/web-ext/blob/1650b3b4fa13bc9f3627e276b281052b6facb958/tests/unit/test-extension-runners/test.chromium.js#L104

tests/unit/test-extension-runners/test.chromium.js
PatrykKuniczak: That's a little wrong, you've missed `extensions`, but i fixed this :)

tests/unit/test-extension-runners/test.chromium.js
PatrykKuniczak: @Rob--W You're right 😆 

I hope, now it's done :)

Rob--W (approved):
Should be good now, thanks!

There is still a CI error, but that is due to an unrelated audit failure.

PatrykKuniczak:
@Rob--W WOW :)

PR #3688, opened 2026-04-23
chore: Remove unnecessary `// eslint-disable` comment for `import/no-unresolved`

PatrykKuniczak:
Zrzut ekranu 2026-04-23 160820

If you'll want that change, i'll remove `// NOTE ...` also

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,6 @@
 //   disable this rule because we have automated tests (which would catch an unresolved import issue
 //   here).
 //
-// eslint-disable-next-line import/no-unresolved
 import webext from './lib/main.js';
 
 export default webext;


PatrykKuniczak:
@Rob--W There's another problem with tests across PRs

Rob--W:
This is merely a warning, but if not built (e.g. on a fresh checkout) this would be a linter warning.

On the other hand, there are no CI failures without this linter suppression.

@rpl thoughts on this one?

PatrykKuniczak:
@Rob--W Yeah indeed, there's no CI error, that's why i haven't closed it, after i opened it😆

PatrykKuniczak:
Any decision?

PR #3689, opened 2026-04-23
fix: Removed node-notifier unmaintained dependency

rpl:
node-notifier doesn't seem to be actively maintained, last update pushed on npm is from 4 years ago.

Given that we agreed to remove it from the web-ext dependencies, the use case in develop mode isn't really something we depend on and the use related to `web-ext run` is more of a nice to have (the errors are still going to be reported in the console as it already was).

This pull request is removing the dependencies and adjusting the implementation and test coverage accordingly.


rpl:
@willdurand @Rob--W CircleCI has a warning saying that github webhooks are currently experiencing an outage, let's wait a bit to see if the CI jobs gets run automatically after that github webhooks issue is resolved, and to eventually look into forcing the CI job to run if that isn't the case, before proceeding to review and eventually merge this PR.

rpl:
I have rebased this patch to get past a CI job failure on the audit-deps step and then fixed a prettier linting error.

This is now ready for a review pass.

tests/unit/test-extension-runners/test.extension-runners.js
```diff
@@ -133,44 +133,47 @@ describe('util/extension-runners', () => {
...
-        sinon.match({
-          title: sinon.match(/web-ext run: extension reload error/),
-          message: sinon.match(/on "fakeExtensionRunner" - reload error 1/),
-        }),
+
+      const { capturedMessages } = consoleStream;
+
+      assert.ok(
+        capturedMessages.some(
+          (msg) => msg.match('[error]') && msg.match('reload error 1'),
```
Rob--W:
```suggestion
          (msg) => msg.includes('[error]') && msg.includes('reload error 1'),
```

string.match turns the parameter in a regexp. `/[error]/` would match any letter, not the literal "[error]".

Similarly below (2x 2x)

Rob--W (approved):
r+wc

PR #3692, opened 2026-04-23
chore(audit): add npm audit exception for uuid transitive dependency

rpl:
This patch is adding uuid advisory to the nsprc exception list as a short term workaround to fix CI jobs failures due to it.

After #3689 is signed off and merged then `uuid` would not be used by any of the dependencies included in the released package and without the node-notifier unmaintained dependency it should also be easier to remove this npm audit exception.


rpl:
@willdurand I'm going to merge this right now, so that I can rebase a few pull requests that have hit a CI job failure due on the audit-deps due to this.

As mentioned in the description once we have merged #3689 it should also be easier to remove this exception.

PR #3695, opened 2026-04-25
test: avoid intermittent failure of test.watcher.js

Rob--W:
The "changes if the watched file is touched" test failed frequently when I ran it, due to the file change not having been detected. This is because the `whenFilesChanged` promise has not resolved when the hardcode test timeout of 500 ms was reached.

Bumping that 500 to 1000 solves the issue.

The "does not change if watched file is not touched" test is directly impacted by this timeout change, because it waits for that time before checking the result. To avoid this, schedule the setup of that test earlier, so that it runs in parallel with the other test.

Before (2 seconds and intermittent failures):

```
    ✔ watches for changes in the sourceDir (1065ms)
    --watch-file option is passed in
      ✔ changes if the watched file is touched (514ms)
      ✔ does not change if watched file is not touched (514ms)
```

After (2 seconds and no intermittent failure):

```
  watcher
    ✔ watches for changes in the sourceDir (1078ms)
    --watch-file option is passed in
      ✔ does not change if watched file is not touched (setup)
      ✔ changes if the watched file is touched (519ms)
      ✔ does not change if watched file is not touched (await) (498ms)
```

(without the parallelization trick, it would have taken 2.5 seconds)

### DIFF
--- a/tests/unit/test.watcher.js
+++ b/tests/unit/test.watcher.js
@@ -60,24 +60,17 @@ describe('watcher', () => {
         tmpDirPath: tmpDir.path(),
       };
 
-      return Promise.race([
-        whenFilesChanged.then(() => {
-          watcher.close();
-          // This delay seems to avoid stat errors from the watcher
-          // which can happen when the temp dir is deleted (presumably
-          // before watcher.close() has removed all listeners).
-          return new Promise((resolve) => {
-            setTimeout(resolve, 2, assertParams);
-          });
-        }),
+      await Promise.race([
+        whenFilesChanged,
         // Time out if no files are changed
-        new Promise((resolve) =>
-          setTimeout(() => {
-            watcher.close();
-            resolve(assertParams);
-          }, 500),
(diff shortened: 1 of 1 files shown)


tests/unit/test.watcher.js
```diff
@@ -60,24 +60,17 @@ describe('watcher', () => {
...
+      await Promise.race([
+        whenFilesChanged,
         // Time out if no files are changed
-        new Promise((resolve) =>
-          setTimeout(() => {
-            watcher.close();
-            resolve(assertParams);
-          }, 500),
-        ),
+        new Promise((resolve) => setTimeout(resolve, 1000)),
```
Rob--W, line 66: This 500 to 1000 change was the only necessary change of this whole code block here. However, I moved the old code around a little bit to increase the readability (making it obvious that `whenFilesChanged` is racing against a timer and that everything else is identical).

rpl (approved):

PR #3696, opened 2026-04-25
fix: Avoid autoreloading by Chrome-generated files

### LINKED ISSUE #3468: A reload now triggers infinite reloading in Chrome
```
$ ./node_modules/.bin/web-ext --version
8.8.0
```

```
$ google-chrome --version
Google Chrome 138.0.7204.100 
```

```
$ ./node_modules/.bin/web-ext run --target chromium -s src/
Running web extension from /home/alexei/Projects/privacybadger/src
Press R to reload (and Ctrl-C to quit)
```

I press R.

The extension reloads forever.

Rob--W:
Fixes #3468

Must be merged after #3697 to avoid test failures.

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -25,10 +25,18 @@ export default function onSourceChange({
   const ignored =
     watchIgnored && process.platform === 'win32'
       ? watchIgnored.map((it) => it.replace(/\\/g, '/'))
-      : watchIgnored;
+      : (watchIgnored || []).slice(0);
+
+  // Chrome may write to the loaded extension directory:
+  // https://github.com/mozilla/web-ext/issues/3468
+  // Ignore these to avoid permanent auto-reload.
+  // Although these files are only expected at the top level, we ignore them
+  // globally in case a subdirectory is also loaded as a Chrome extension.
+  ignored.push('**/_metadata');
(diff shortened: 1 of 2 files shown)


Rob--W:
CI was repeatedly failing due to `onChange` called for the `parent` directory in the test:

```
  1) watcher
       watcher ignores _metadata
         ignores change to non-toplevel _metadata directory (await):
     expected spy to not have been called but was called once
```

I identified the root cause, to be a watchpack bug, and created a PR to fix that: #3697.

CI is currently red because #3697 is not included in the PR.

Here is a fully green CI job for #3697 + this PR: https://app.circleci.com/pipelines/github/mozilla/web-ext/4894/workflows/871b8d37-b0ce-475a-9094-182c9c889cb0

Rob--W:
While this PR works, I wonder whether I should have added the ignore patterns to `FileFilter` instead: https://github.com/mozilla/web-ext/blob/3b2a51875c20166ed2c1449daa2685fed52f98c0/src/util/file-filter.js#L32-L39

That is passed to the watcher at: https://github.com/mozilla/web-ext/blob/3b2a51875c20166ed2c1449daa2685fed52f98c0/src/extension-runners/index.js#L199-L206

PR #3746, opened 2026-06-23
Log the response data when a PATCH request fails

willdurand:
This PR prints the response json in the error when PATCH requests are not OK.

### DIFF
--- a/src/util/submit-addon.js
+++ b/src/util/submit-addon.js
@@ -244,8 +244,20 @@ export default class Client {
       }
 
       const response = await this.fetch(patchUrl, 'PATCH', formData);
+      if (response.status < 200 || response.status >= 500) {
+        throw new Error(
+          `Patch request failed: ${response.statusText || response.status}.`,
+        );
+      }
+
       if (!response.ok) {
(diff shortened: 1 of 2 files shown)


src/util/submit-addon.js
```diff
@@ -245,7 +245,13 @@ export default class Client {
 
       const response = await this.fetch(patchUrl, 'PATCH', formData);
       if (!response.ok) {
-        throw new Error(`response status was ${response.status}`);
+        const responseData = await response.json();
```
diox, line 254: I think you need something to guard against 50x errors, which would not be JSON. Elsewhere, before checking for `response.ok` we do:
```javascript
if (response.status < 200 || response.status >= 500) {
    throw new Error(
        `${errorMsg}: ${response.statusText || response.status}.`,
    );
}
```

diox (changes requested):

src/util/submit-addon.js
willdurand: ah yes, sorry about that. I missed that part when looking at what was done in `fetchJson()`.

diox (approved):

PR #3747, opened 2026-06-23
chore: run npm audit fix, and clean up the .nsprc entries

willdurand:
I would expect this PR to keep failing, but that's OK. The intent is to
update the `.npsrc` file to get rid of outdated entries.


rpl (approved):
lgtm, the followup pull request #3748 will fix the CI job failure this one is currently hitting.

PR #3762, opened 2026-06-30
Upgrade Babel to v8

willdurand:
Closes https://github.com/mozilla/web-ext/pull/3755
Closes https://github.com/mozilla/web-ext/pull/3757
Closes https://github.com/mozilla/web-ext/pull/3758
Closes https://github.com/mozilla/web-ext/pull/3759
Closes https://github.com/mozilla/web-ext/pull/3760


scripts/babel-plugin-inline-environment-variables.cjs
```diff
@@ -0,0 +1,45 @@
...
+
+  function getComputedKey(path) {
+    const { node } = path;
+    if (node.computed) {
+      return node.property;
+    }
+    if (t.isIdentifier(node.property)) {
+      return t.stringLiteral(node.property.name);
+    }
+    return node.property;
```
rpl, line 18: nit/thought: could we use a single `return node.property` if we change the `if (t.isIdentifier(...)) {...}` to `if (!node.computed && t.isIdentifier(...)) {...}`?

If I'm not mistaken, in the current version of the `getComputedKey` helper function:
- `if (node.computed)` is meant to cover the `process.env[...]` cases
- `if (t.isIdentifier(...)` is meant to cover the `process.env.SOMETHING` cases
- and the `return node.property` any other case not explicitly handled

And so I'd guess that merging the alternative version with only one if block should be equivalent in practive, but maybe we prefer to keep it as is and add a couple of inline comments to explicitly mention to our future selves what cases each part of the helper function is meant to be handling?

scripts/babel-plugin-inline-environment-variables.cjs
```diff
@@ -0,0 +1,45 @@
...
+    name: 'inline-environment-variables',
+    visitor: {
+      MemberExpression(path, { opts: { include, exclude } = {} }) {
+        if (!path.get('object').matchesPattern('process.env')) {
+          return;
+        }
+
+        const key = getComputedKey(path);
+
+        if (
```
rpl, line 34: nit/thought: if may be worth mentioning briefly here what this if block is meant to be doing, just to make it easier to recall how this plugin works, e.g.

```
// Ignore dinamic variables (e.g. `process.env[someVar]`) and assignment expressions (`process.env.FOO = ...`) and account for `include`/`exclude` plugin options (if any was specified)
```

rpl (approved):
@willdurand looks good to me, the plugin is small enough in the end, it feels ok to roll it into web-ext repo itself, the following are just small nits/thoughts (not really mandatory but I thought it may be reasonable to leave us some hints about how the plugin works in case we will need to rework it on top of a future babel major release, I doubt we will ever look at it before something like that would eventually force us to :-P)

PR #3765, opened 2026-07-02
Drop commitlint in this project

willdurand:
I've been fighting with this linter for a long time. The reality is that
this is the only project that enforces rules on commit titles. That's
the main reason why I am proposing to drop commitlint in this PR.


rpl (approved):
lgtm

PR #3781, opened 2026-07-24
Fix configured run target precedence

### LINKED ISSUE #3314: `run.target: ["chromium"]` ignored in config

bug

A plain `$ web-ext run` will read/validate `webExt.run.target` from my package.json, but then it ignores it. 

This verbose log line explains why:

```
[web-ext/lib/config.js][debug] Favoring CLI: target=firefox-desktop over configuration: target=chromium
```

The CLI parser is providing its own defaults before merging with the config file.

The config file should be respected by applying the defaults **after** reading the config file.

- **Your OS and version**: macOS 15.1.1

```
v23.1.0
10.9.0
8.3.0
```

giri256:
Fixes #3314

## Summary

- Remove the yargs default from the array-valued `run.target` option so it is not mistaken for an explicit CLI value during config merging.
- Keep Firefox Desktop as the effective default in the run command when no target is provided.
- Add regression coverage for config precedence, explicit CLI precedence, and the undefined-target fallback.

## Validation

- `npx prettier --check src/program.js tests/unit/test.program.js tests/unit/test-cmd/test.run.js`
- `npx eslint src/program.js tests/unit/test.program.js tests/unit/test-cmd/test.run.js`
- `npx mocha tests/unit/test.program.js tests/unit/test-cmd/test.run.js` (83 passing)

### DIFF
--- a/src/program.js
+++ b/src/program.js
@@ -583,7 +583,6 @@ Example: $0 --help run.
         describe:
           'The extensions runners to enable. Specify this option ' +
           'multiple times to run against multiple targets.',
-        default: 'firefox-desktop',
         demandOption: false,
         type: 'array',
         choices: ['firefox-desktop', 'firefox-android', 'chromium'],
--- a/tests/unit/test-cmd/test.run.js
+++ b/tests/unit/test-cmd/test.run.js
@@ -271,6 +271,13 @@ describe('run', () => {
     sinon.assert.calledOnce(desktopRunnerStub);
   });
 
+  it('creates a Firefox Desktop runner if target is undefined', async () => {
+    const cmd = await prepareRun();
+    await cmd.run({ target: undefined });
+    sinon.assert.notCalled(androidRunnerStub);
+    sinon.assert.calledOnce(desktopRunnerStub);
+  });
+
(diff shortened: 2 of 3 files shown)


giri256:
CI audit note: the only failing context is `build-nodejs-current`, at its final `npm run audit-deps` step. Prettier, unit/coverage tests, both functional-test modes, Node next/experimental, and Windows all pass.

The audit is reporting newly blocking advisories in existing dependencies (`adm-zip`, `brace-expansion`, `fast-uri`, and `shell-quote`); this PR does not modify dependency files. The same current-job failure pattern is also present on the repository's open Dependabot PRs #3780 and #3779, so this is not introduced by this change.

src/program.js
```diff
@@ -583,7 +583,6 @@ Example: $0 --help run.
         describe:
           'The extensions runners to enable. Specify this option ' +
           'multiple times to run against multiple targets.',
-        default: 'firefox-desktop',
```
Rob--W: Could you update the `describe` to mention that `firefox-desktop` is the default option, e.g. between parentheses?

To see the effect of the configuration, run the command with `--help`.

PR #3782, opened 2026-07-24
Expose disabled linter rules

### LINKED ISSUE #3333: cannot make linter ignore specific rules
Even though [addons-linter supports disabling specific rules](https://github.com/search?q=repo%3Amozilla%2Faddons-linter%20disableLinterRules&type=code), it seems that there is no way to use this feature when invoking `web-ext lint`.  Based on a very quick glance at the code, maybe fixing this is as simple as adding `disableLinterRules` here?

https://github.com/mozilla/web-ext/blob/61464c83b46fb2f2fa7185538c1b816c2a75d681/src/cmd/lint.js#L32-L48

The only workaround seems to be to ignore whole files, but this is undesirable for the same reason it's undesirable in other linters like eslint: it runs the risk of missing other rule violations in the ignored file.

giri256:
Fixes #3333

## Summary

- Add `--disable-linter-rules` as a comma-separated string option for `web-ext lint`.
- Support the corresponding `lint.disableLinterRules` config value.
- Forward the value unchanged to addons-linter, which performs rule-name trimming and splitting internally.
- Preserve current behavior when the option is omitted.

## Validation

- `npx prettier --check src/program.js src/cmd/lint.js tests/unit/test.program.js tests/unit/test-cmd/test.lint.js`
- `npx eslint src/program.js src/cmd/lint.js tests/unit/test.program.js tests/unit/test-cmd/test.lint.js`
- `npx mocha tests/unit/test.program.js tests/unit/test-cmd/test.lint.js` (69 passing)
- `npm run build`
- `node bin/web-ext.js lint --source-dir tests/fixtures/minimal-web-ext --disable-linter-rules no-implied-eval --output json`

### DIFF
--- a/src/cmd/lint.js
+++ b/src/cmd/lint.js
@@ -11,6 +11,7 @@ export default function lint(
   {
     artifactsDir,
     boring,
+    disableLinterRules,
     enterprise,
     ignoreFiles,
     metadata,
@@ -39,6 +40,7 @@ export default function lint(
     metadata,
     output,
     boring,
+    disableLinterRules,
     selfHosted,
     enterprise,
     shouldScanFile: (fileName) => fileFilter.wantFile(fileName),
--- a/src/program.js
+++ b/src/program.js
@@ -792,6 +792,11 @@ Example: $0 --help run.
         type: 'boolean',
         default: false,
       },
+      'disable-linter-rules': {
+        describe: 'Disable a comma-separated list of linter rules',
+        type: 'string',
+        requiresArg: true,
+      },
       pretty: {
         describe: 'Prettify JSON output',
         type: 'boolean',
--- a/tests/unit/test-cmd/test.lint.js
+++ b/tests/unit/test-cmd/test.lint.js
@@ -93,6 +93,29 @@ describe('lint', () => {
     });
   });
 
+  it('passes disableLinterRules to the linter', () => {
+    const { lint, createLinter } = setUp();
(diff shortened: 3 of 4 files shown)


giri256:
CI audit note: the only failing context is `build-nodejs-current`, at its final `npm run audit-deps` step. Prettier, unit/coverage tests, both functional-test modes, Node next/experimental, and Windows all pass.

The audit is reporting newly blocking advisories in existing dependencies (`adm-zip`, `brace-expansion`, `fast-uri`, and `shell-quote`); this PR does not modify dependency files. The same current-job failure pattern is also present on the repository's open Dependabot PRs #3780 and #3779, so this is not introduced by this change.

src/cmd/lint.js
```diff
@@ -39,6 +40,7 @@ export default function lint(
     metadata,
     output,
     boring,
+    disableLinterRules,
```
Rob--W: A potential concern with supporting linter rule suppression is that someone configures the suppression, forgets about it, and then encounters differences between what AMO reports and what they encounter.

To mitigate that concern, let's log a warning below when the option is set.

Rob--W (commented):
Thanks for the PR! It's nice that the chosen flag name matches the flag we already use in the `addons-linter`, added back in https://github.com/mozilla/addons-linter/pull/3525

Just one request - could you log a message when the option is set, to make the `web-ext` user aware of the usage of the flag?

PR #3808, opened 2026-08-15
Fix web-ext sign falsely failing on AMO 502/503/504 timeouts during source upload (Fixes #3807)

### LINKED ISSUE #3807: `web-ext sign` falsely returns a 5xx error
Hi, I've been encountering a strange error several times when running `web-ext sign` to submit a build to AMO for signing:

```
Waiting for validation...
Submitting source to version
Upload of source failed: Error: response status was 503.

WebExtError: Uploading source failed
at file:///dev_build-release_misc_web-ext-v8/lib/cmd/sign.js:101:13
at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
at async Program.execute (file:///dev_build-release_misc_web-ext-v8/lib/program.js:270:7)
at async file:///dev_build-release_misc_web-ext-v8/bin/web-ext.js:13:1
```

It returns a 503 error code, but AMO _does_ successfully sign the build, which is very misleading and causing CI builds to fail. I see that another issue addressing this (#3484), which states that this error is resolved, yet I'm still running into this error.

itsnjstyle27:
This PR addresses an issue where web-ext sign fails and crashes CI pipelines by throwing a WebExtError: Uploading source failed when the AMO server returns a 502, 503, or 504 status code during the source code upload phase.

Because AMO's backend often continues to successfully process and attach these larger source code bundles in the background despite timing out the client's HTTP request, hard-failing in the CLI is misleading and unnecessarily breaks the build.

Changes Made
src/util/submit-addon.js: Updated doFormDataPatch to intercept 502, 503, and 504 status codes. Instead of throwing an error, it now logs a warning indicating the Gateway Timeout / Service Unavailable response from AMO and safely proceeds with the assumption that the source upload succeeded in the background. This ensures that the remainder of the signing workflow can continue uninterrupted.
scripts/lib/mocha.js (Internal / Test): Wrapped the constructed binPath in quotes when executing on Windows to fix a local testing bug where spaces in the user's root directory path caused mocha test executions to fail to spawn.
Testing
Verified all unit tests (npm run test) continue to pass successfully.
Confirmed that eslint style formatting is maintained.

Hope this helps :)

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -4,6 +4,11 @@
     "https://github.com/advisories/GHSA-73rr-hh4g-fpgx",
     // serialize-javascript  <=7.0.4
     "https://github.com/advisories/GHSA-5c6j-r48x-rmvq",
-    "https://github.com/advisories/GHSA-qj8w-gfj5-8c6v"
+    "https://github.com/advisories/GHSA-qj8w-gfj5-8c6v",
+    // image-size DoS (via addons-linter)
+    "https://github.com/advisories/GHSA-w3rx-r6r6-pgpr",
+    "https://github.com/advisories/GHSA-5p2g-fcmc-qvqq",
+    // js-yaml Quadratic CPU consumption (via babel-plugin-istanbul)
+    "https://github.com/advisories/GHSA-5p4m-2wfm-xmqj"
   ]
 }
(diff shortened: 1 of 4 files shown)


rpl:
We discussed about this proposed change and we agreed that assuming that when AMO returned a 5XX error then the submission have been able to pass through consistently would be misleading.

There have been a short window during which AMO had intermittent issues around a month ago and in that case it may have been true but web-ext can't assume that to be the case in general.

An alternative approach would be to handle it on the CI job side if the developer wants to relax how strict the behavior of the CI job is.

Issue #3806, opened 2026-08-13
npm audit failures: image-size and js-yaml


Rob--W:
Latest landing commits are having permanent CI failures due to npm audit failures.
E.g. https://app.circleci.com/pipelines/github/mozilla/web-ext/5155/workflows/b3cbae27-58b5-492c-8470-b6c1cdd305f5/jobs/20759

- the `image-size` dependency is referenced via addons-linter.
- the `js-yaml` dependency is referenced via`@eslint/eslintrc`, which is a direct dependency of addons-linter as well, but also others.

```
== audit-deps: blocking security issues

addons-linter (isDirect: true, severity: high, fixAvailable: true):
  https://github.com/advisories/GHSA-w3rx-r6r6-pgpr
    image-size <=2.0.2
    image-size: ICNS parser allows denial of service through an infinite loop
  https://github.com/advisories/GHSA-5p2g-fcmc-qvqq
    image-size <=2.0.2
    image-size: JXL and HEIF parsers allow denial of service through infinite loops
image-size (isDirect: false, severity: high, fixAvailable: true):
  https://github.com/advisories/GHSA-w3rx-r6r6-pgpr
    image-size <=2.0.2
    image-size: ICNS parser allows denial of service through an infinite loop
  https://github.com/advisories/GHSA-5p2g-fcmc-qvqq
    image-size <=2.0.2
    image-size: JXL and HEIF parsers allow denial of service through infinite loops
js-yaml (isDirect: false, severity: high, fixAvailable: true):
  https://github.com/advisories/GHSA-5p4m-2wfm-xmqj
    js-yaml >=3.0.0 <3.15.1
    JS-YAML: Quadratic CPU consumption in !!omap resolution (3.x and 4.x) — CVE-2026-59870 fix not backported
```

carlosjeurissen:
There is no image-size above 2.0.2. See also https://github.com/github/advisory-database/issues/9028

Rob--W:
The lack of update is likely because the author of image-size is tired of LLM-generated advisories: https://codeberg.org/image-size/image-size#tired-maintainers-doesn-t-want-your-slop

If maintenance for the package is ceased, we should drop it eventually.

Rob--W:
Resolved by:
- https://github.com/mozilla/addons-linter/pull/6191
- https://github.com/mozilla/web-ext/pull/3850
- https://github.com/mozilla/web-ext/pull/3851
- https://github.com/mozilla/web-ext/pull/3855

PR #3828, opened 2026-09-09
run npm audit fix

willdurand:
refs https://github.com/mozilla/web-ext/issues/3806


willdurand:
Not enough, but that can't hurt... r=me

PR #3829, opened 2026-09-09
Remove npm audit exceptions

willdurand:
These exceptions are no longer necessary.


rpl (approved):
Thanks @willdurand! It looks like the CI jobs failed, but it failed because of new npm audit reported advisories that were not listed even before this change, and so expected in this pull request. Signing it off as is so that we can proceed with removing the old npm audit exceptions.

PR #3832, opened 2026-09-11
fix: Treated localized extension names as literal filename text

Hanabi9248:
A localized extension named `Price $& Tracker` builds as `price_msg_name_tracker-1.0.zip`: the replacement string expands `$&` back to `{name}`, which is then resolved against the original manifest. Likewise, `Edition {version}` substitutes the manifest version, and `Tools/Kit` fails filename validation.

Sanitize the localized name before inserting it into the filename template, using a replacement callback. This gives localized names the same filename treatment as ordinary manifest names and keeps their contents from being interpreted as template syntax.

All three regressions exercise actual package builds and fail before the change. On Windows with Node 24.19.0, the build and full test suite pass (585 passing, 3 pending), as do Prettier and ESLint for both changed files.

### DIFF
--- a/src/cmd/build.js
+++ b/src/cmd/build.js
@@ -152,7 +152,9 @@ export async function defaultPackageCreator(
       manifestData,
     });
     // allow for a localized `{name}`, without mutating `manifestData`
-    filenameTemplate = filenameTemplate.replace(/{name}/g, extensionName);
+    filenameTemplate = filenameTemplate.replace(/{name}/g, () =>
+      safeFileName(extensionName),
+    );
   }
 
   const packageName = safeFileName(
--- a/tests/unit/test-cmd/test.build.js
+++ b/tests/unit/test-cmd/test.build.js
@@ -122,6 +122,39 @@ describe('build', () => {
     );
   });
 
+  for (const [name, expectedName] of [
+    ['Price $& Tracker', 'price_tracker'],
+    ['Edition {version}', 'edition_version_'],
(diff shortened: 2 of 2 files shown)


src/cmd/build.js
```diff
@@ -152,7 +152,9 @@ export async function defaultPackageCreator(
       manifestData,
     });
     // allow for a localized `{name}`, without mutating `manifestData`
-    filenameTemplate = filenameTemplate.replace(/{name}/g, extensionName);
+    filenameTemplate = filenameTemplate.replace(/{name}/g, () =>
+      safeFileName(extensionName),
```
Rob--W: `safeFileName` call here seems not strictly needed because we already call `safeFileName` on the transformation below. Is there any reason you're adidng `safeFileName` instead of just returning `extensionName`?

Hanabi9248: The outer call runs after `getPackageNameFromTemplate`, which both expands placeholders and rejects paths. Returning the raw localized name lets that intermediate step interpret it first.

For example, `Edition {version}` becomes `edition_1.0-1.0.zip`, and `Tools/Kit` throws the path-validation error before the outer `safeFileName` can run. Sanitizing before insertion gives `edition_version_-1.0.zip` and `tools_kit-1.0.zip`. Those are the second and third regression cases in this PR.

The callback alone fixes the replacement-string `$&` case; this earlier sanitization handles the later template parsing and validation.

PR #3833, opened 2026-09-12
fix(sign): Defaulted amoBaseUrl when signing from the Node API

### LINKED ISSUE #3164: "AMO API base URL" must be specified even though docs say it has a default

After upgrading from v7 to v8, the `.sign()` command no longer works for me due to this error:

```js
Invalid AMO API base URL: undefined
```

My code is:

```js
import webExt from 'web-ext';

await webExt.cmd.sign({
      apiKey: '<hidden>',
      apiSecret: '<hidden>',
      artifactsDir: '<hidden>',
      channel: 'unlisted',
      sourceDir: '<hidden>',
})
```

I can overcome this error by adding:

`amoBaseUrl: 'https://addons.mozilla.org/api/v5/'`

But the docs say I shouldn't need to do that as this is already the default.

If no `amoBaseUrl` is specified, error is thrown.

If no `amoBaseUrl` is specified, the default is used.

- **Firefox version**: N/A
- **Your OS and version**: MacOS 14.5 (23F79)
- Paste the output of these commands:

```

v20.14.0
10.7.0
8.0.0
```

MaxFreedomPollard:
Fixes #3164.

Calling `webExt.cmd.sign({apiKey, apiSecret, artifactsDir, channel, sourceDir})` from Node fails with `Invalid AMO API base URL: undefined`. The default only exists as a yargs default in the CLI option table (src/program.js:527), so it is applied on the command line and nowhere else. src/cmd/sign.js:23 destructures `amoBaseUrl` with no default, and `signAddon` rejects a missing value at src/util/submit-addon.js:500-503. The reporter hit this moving from 7.x to 8.x and worked around it by hardcoding `amoBaseUrl: 'https://addons.mozilla.org/api/v5/'`, which pins them to an API version web-ext is meant to manage for them.

This moves the constant into a new dependency-free module, src/util/constants.js. src/program.js imports it and re-exports it, so `import { AMO_BASE_URL } from './program.js'` keeps working for existing callers and tests, and the CLI option table is untouched. src/cmd/sign.js imports it from the new module rather than from src/program.js, so a Node API caller does not pull yargs and the rest of the CLI in behind it. No command module imports src/program.js today, and src/cmd/index.js loads commands lazily for the same start-up reason (#1302).

One scope note: `signAddon()` in src/util/submit-addon.js still takes `amoBaseUrl` explicitly. It is public API, since package.json exports `./util/submit-addon`, so this is a real choice rather than an oversight. I kept the change to the entry point the issue reports, but signAddon() can get the same default here if you would rather have it in one place.

The new test is "defaults the AMO base URL when it is not passed in" in tests/unit/test-cmd/test.sign.js. `getStubs()` always supplies `amoBaseUrl`, so the test copies the stub config, deletes that key, and asserts that `submitAddon` was called with the default. Reverting the one line in src/cmd/sign.js makes it fail with `amoBaseUrl: undefined`. The README example for `webExt.cmd.sign()` also drops the hardcoded `amoBaseUrl`; the `signAddon()` example below it keeps it, because that entry point still needs it.

Unit tests, eslint and prettier pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,6 @@ webExt.cmd.sign({
   userAgentString: 'YOUR-CUSTOM-USERAGENT',
   apiKey,
   apiSecret,
-  amoBaseUrl: 'https://addons.mozilla.org/api/v5/',
   sourceDir: ...,
   channel: 'unlisted',
   ...
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -3,6 +3,7 @@ import path from 'path';
 import defaultBuilder from './build.js';
 import { isErrorWithCode, UsageError, WebExtError } from '../errors.js';
 import { prepareArtifactsDir } from '../util/artifacts.js';
+import { AMO_BASE_URL } from '../util/constants.js';
 import { createLogger } from '../util/logger.js';
 import getValidatedManifest, { getManifestId } from '../util/manifest.js';
 import {
@@ -20,7 +21,7 @@ export const uploadUuidFile = '.amo-upload-uuid';
 
 export default function sign(
   {
-    amoBaseUrl,
+    amoBaseUrl = AMO_BASE_URL,
     apiKey,
     apiProxy,
     apiSecret,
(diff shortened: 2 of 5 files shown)


Rob--W (approved):
Looks good; we should wait with merging this until right before we publish a new version, because otherwise people trying out the code from README.md will find it not working.

PR #3834, opened 2026-09-12
fix(sign): Excluded the artifacts directory from the signed package

### LINKED ISSUE #1640: sign command should ignore web-ext-artifacts folder
Bug

The 'sign' command includes the web-ext-artifacts folder when packaging the extension.

All web-ext commands should ignore the output folder used by web-ext. The 'build' command already does, which leads to someone testing with a correctly packaged extension and then publishing a bloated extension using the 'sign' command

* **Firefox version**: 67.0.4
* **Your OS and version**: Windows 10 1903

````
node --version; npm --version; web-ext --version 
v12.3.1
6.9.0
3.1.0
````

### Additional details
It seems the 'sign' command filters out zip and xpi files. So in theory that would exclude whatever is in the artifacts folder. However, I build my extension for Chrome and Opera so there are nex and crx files in the folder, as well as my extension unpacked for testing in Chrome and Opera.

Here's a capture of the output from the sign command:
[sign.log](https://github.com/mozilla/web-ext/files/3336485/sign.log)

MaxFreedomPollard:
Fixes #1640.

`web-ext build` leaves the artifacts directory out of the package it creates, but `web-ext sign` puts it in. Anything under `web-ext-artifacts/` that is not a `.zip` or `.xpi` ends up inside the archive that is uploaded to AMO: `.crx` and `.nex` files from Chrome or Opera builds, unpacked copies of the extension, screenshots. Previously built `.zip` and `.xpi` packages are already dropped by the base ignore patterns, so those are not the problem, but everything else is shipped and signed.

The cause is the temporary build directory. `sign()` builds into a temp dir (`artifactsDir: tmpDir.path()`, src/cmd/sign.js:59 on master), and `build()` creates its default file filter from whatever artifacts directory it was given (src/cmd/build.js:210-214). `FileFilter` only adds ignore rules for the artifacts directory when that directory is inside the source directory (src/util/file-filter.js:53). A temp dir is never a subpath of the source dir, so the rule is never added and the real `web-ext-artifacts/` is packaged.

The fix is to build the filter where the real artifacts directory is known. Sign now calls `createFileFilter({ sourceDir, ignoreFiles, artifactsDir })` itself and passes the result to `build()` as its `fileFilter` option, which is a default parameter, so no signature change is needed. That is the same call-site pattern `lint.js` and the extension runners already use. Those `ignoreFiles` go into the filter sign constructs, so `--ignore-files` behaves exactly as before, and the `isSubPath` guard added in #790 is untouched, so behaviour is also unchanged when `--artifacts-dir` points outside the source directory. Only the extension package is affected. The source bundle for `--upload-source-code` is a user supplied path.

Two tests were added to `tests/unit/test-cmd/test.sign.js`. The first runs the real sign command over a source directory that has a `web-ext-artifacts/previous-build.crx` in it, with only the submission call stubbed, and asserts the zip contains `manifest.json` and not the artifacts file. The zip is read inside that stub because sign removes its temporary directory as soon as it returns. The second injects a spy `createFileFilter` and asserts it is called with the real artifacts directory and that the resulting filter reaches `build()`.

The one existing test that runs the real build signs a source directory with no artifacts directory inside it, so nothing leaked; every other sign test stubs build. Both new tests fail before the change, with `AssertionError: expected [ 'web-ext-artifacts/', …(3) ] to not include 'web-ext-artifacts/previous-build.crx'`. With the change, `tests/unit/test-cmd/test.sign.js` gives 16 passing, `test.build.js` 28 passing and `tests/unit/test-util/test.file-filter.js` 18 passing.

### DIFF
--- a/src/cmd/sign.js
+++ b/src/cmd/sign.js
@@ -3,6 +3,7 @@ import path from 'path';
 import defaultBuilder from './build.js';
 import { isErrorWithCode, UsageError, WebExtError } from '../errors.js';
 import { prepareArtifactsDir } from '../util/artifacts.js';
+import { createFileFilter as defaultFileFilterCreator } from '../util/file-filter.js';
 import { createLogger } from '../util/logger.js';
 import getValidatedManifest, { getManifestId } from '../util/manifest.js';
 import {
@@ -36,6 +37,7 @@ export default function sign(
   },
   {
     build = defaultBuilder,
+    createFileFilter = defaultFileFilterCreator,
     preValidatedManifest,
     submitAddon = defaultSubmitAddonSigner,
     asyncFsReadFile = defaultAsyncFsReadFile,
@@ -54,10 +56,18 @@ export default function sign(
       manifestData = await getValidatedManifest(sourceDir);
     }
 
+    // The package is built into a temporary directory, so the filter has to
+    // be built here, where the real artifacts directory is known.
+    const fileFilter = createFileFilter({
+      sourceDir,
+      ignoreFiles,
+      artifactsDir,
+    });
+
     const [buildResult, idFromSourceDir] = await Promise.all([
       build(
         { sourceDir, ignoreFiles, artifactsDir: tmpDir.path() },
-        { manifestData, showReadyMessage: false },
+        { manifestData, showReadyMessage: false, fileFilter },
       ),
       getIdFromFile(savedIdPath),
     ]);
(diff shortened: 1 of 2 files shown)


Rob--W (changes requested):
This patch seems to fix the issue at the wrong layer. The build command should exclude the artifacts directory; sign should not exclude files that the build includes, because otherwise someone testing the output of build may get a result that differs from what users would encounter after signing the package.

PR #3835, opened 2026-09-12
fix: Fixed files not being ignored when sourceDir contains glob characters

MaxFreedomPollard:
If the directory an extension lives in has a glob character in its name, web-ext ignores nothing. A build from a folder like "/src/my ext [beta]" packs .git, node_modules, every dotfile and every previously built zip straight into the artifact, and any pattern the user put in ignoreFiles is silently dead too.

Every ignore pattern is made absolute by prefixing the resolved sourceDir, and minimatch then reads that prefix back as pattern syntax instead of as a real path. "[beta]" is parsed as a one-character class that matches one of b, e, t or a, so the pattern no longer matches the directory it came from and nothing is ignored. The extglob openers "@(", "+(", "*(" and "?(" fail the same way, and a "!(" name matches the wrong sibling directories. Bare parentheses were never a problem, since minimatch already treats a "(" with no extglob character in front of it as a literal.

The fix escapes the glob characters in the sourceDir prefix before the rest of the pattern is appended, so the directory is matched literally while the user's own pattern keeps its glob meaning. Each character is escaped as a one-character class, "[" becoming "[[]", rather than with a backslash, because on Windows minimatch rewrites the path separator to "/" inside patterns and a backslash escape would not survive that. The artifacts directory is a real path and never a pattern, so it is escaped whole.

Two cases stay unfixed. A curly brace in a directory name still breaks matching, because brace expansion runs before character classes are parsed and there is no class that can hide a brace from it. A literal backslash in a directory name on POSIX is also still read as an escape.

Two unit tests cover this: one builds a filter on "/src/my ext [beta]" and on "/src/build @(v2)" and checks that the base patterns and a user ignoreFiles entry with a negation all still behave, and one checks an artifacts directory whose own name contains a bracket class. Both fail on master. eslint and prettier --check pass on the two changed files.

### DIFF
--- a/src/util/file-filter.js
+++ b/src/util/file-filter.js
@@ -6,6 +6,16 @@ import { createLogger } from './logger.js';
 
 const log = createLogger(import.meta.url);
 
+// Glob characters that minimatch would otherwise read as pattern syntax when
+// they are part of a real directory name, such as an extension living in
+// "/home/me/my ext [beta]". They are escaped as one-character classes and not
+// with a backslash, because minimatch rewrites path.sep to "/" in patterns on
+// Windows, which would destroy a backslash escape. Two characters stay
+// unhandled: a curly brace, because brace expansion runs before character
+// classes are parsed, and a literal backslash in a directory name on POSIX,
+// which minimatch reads as an escape.
+const escapeGlobChars = (filePath) => filePath.replace(/[*?[\]()]/g, '[$&]');
+
 // check if target is a sub directory of src
 export const isSubPath = (src, target) => {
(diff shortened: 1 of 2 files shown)


Rob--W (approved):
This looks reasonable. There is a theoretical risk of breaking backwards compatibility, if a `web-ext` user relied on expanding source dir or artifacts dirs by including `*` for example. However, that has not been an intentionally supported feature, and under normal use the shell should expand the globs. Whereas on the other hand, expanding globs prevents the use of file/directory names that contain globs.

